PR #26, opened 2020-09-23
Add codecov commit status checks to monitor coverage

### LINKED ISSUE #6: Automated code coverage reports for each PR
Adding coverage reports is easy while a codebase is new and there isn't a lot of code to cover. It will also make it easier for code reviewers to ask authors to include tests if they haven't.

byCedric:
Fixes #6

This adds codecov to monitor the code coverage, using a least intrusive workflow possible. I've tested this and some other approaches in [use-expo](https://github.com/byCedric/use-expo/pull/423).

### Current configuration

- CI only collects and reports coverage when `matrix.coverage == true` (currently only for Node 14)
- Codecov PR comments are disabled
- Codecov reports "default" or monorepo-wide coverage numbers
- Codecov reports per-package coverage numbers
- All Codecov reports are informational only, and will never fail

### Random gotchas

- **Codecov badges are separated by flags only**, but that requires separate per-project reports. This isn't possible when running Jest from root. Instead of implementing a complex CI, [like expo-cli](https://github.com/expo/expo-cli/blob/master/.github/workflows/test.yml#L30-L81), I replaced `flags` with `paths`. This gives us per project coverage numbers on PRs, but doesn't allow us to create badges per project.

- **Inline commit status coverage numbers doesn't seem to work at the moment**. It looks like the github statuses should have the coverage numbers inline (like the screenshot below, from the docs). But, in my repo I saw other numbers. ~~If this happens in this repo too, I'll create a support ticket.~~ [Created a support ticket here](https://community.codecov.io/t/codecov-commit-statuses-shows-codecov-report-instead-of-actual-coverage/2060)

Expected | Use Expo
--- | ---
docs | Screenshot 2020-09-23 at 14 39 54


fson (approved):
Sorry for forgetting about this PR 😅 Looks good to me, we should probably land it. Is there some reason the coverage doesn't show up for this pull request yet?

ide:
Perhaps the coverage did not show up on this PR because we're calling into Codecov using the GH Action instead of the Codecov bot. But we are collecting coverage data now, and could render this with badges on the README, for example: https://codecov.io/gh/expo/eas-cli

Screen Shot 2020-11-10 at 10 41 23 AM

PR #34, opened 2020-10-15
Add Bash installer

fson:
Background: https://www.notion.so/expo/Release-strategies-40bd1a2d58794696a2ecec7ea3f7d36f

To test you can run (no dependencies, doesn't require Node.js): 
```sh
curl https://raw.githubusercontent.com/expo/eas-cli/%40fson/installer/bin/install.sh | bash
```
Once we've set up a subdomain and hosting for the script the final command could be something like:
```sh
curl https://install.expo.io/eas-cli | bash
```

### Test plan
- try different platforms
- try `/usr/local/bin` not writable
- try without `curl` but with `wget` installed

### How it works
- detects the platform using `uname`
- download the latest CLI release (pre-releases are ignored) from `https://github.com/expo/eas-cli/releases/latest/download/eas-$PLATFORM-$ARCH.tar.gz` (this does a couple of HTTP redirects and finally points to a URL in GitHub's S3 bucket)
- exctracts the tarball to a temp folder (we use a temp folder so that we never need sudo at this point)
- installs the CLI:
  * checks if the folders (`/usrl/local/lib` and `/usr/local/bin` by default, override with`EAS_PREFIX`) are writable by the user running the script and that the previous `eas` executable can be moved out of the way. If not, it will use `sudo`.
  * moves the `eas` directory from temp folder to `/usr/local/lib`
  * creates a symlink in `/usr/local/bin/eas`
- tests the bundled `node` (looking at other CLIs using oclif, there can be problems on e.g. Alpine, but I haven't verified this)
- tests the CLI with `eas --version`

### Next steps:
- hosting for `install.sh`
- add release script to CI
- make sure autoupdates work


.github/workflows/install.yml
```diff
@@ -0,0 +1,18 @@
+name: Run installer tests
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
ide, line 10: In a separate PR It'd be good to test this on macOS too, but I think that's a bit less important since the team is regularly testing macOS. For Windows with WSL2 we probably want to make sure the `/usr/local/{bin,lib}` detection works too.

packages/eas-cli/README.md
```diff
@@ -54,7 +52,7 @@ USAGE
   $ eas build
 ```
 
-_See code: [src/commands/build/index.ts](https://github.com/expo/eas-cli/blob/v0.0.0/src/commands/build/index.ts)_
+_See code: [build/commands/build/index.ts](https://github.com/expo/eas-cli/blob/v0.0.0/build/commands/build/index.ts)_
```
ide, line 55: Why are the .ts files under `build`?

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+# This script downloads the latest release version of EAS CLI from GitHub and
+# installs it in /usr/local/bin
+
+{
+  # All code is inside a block to ensure the script executes only when
+  # downloaded completely.
+
+  set -e
+
+  prefix="${EAS_PREFIX:-/usr/local}"
```
wkozyra95, line 12: I'm not sure about macOS, but on Linux scripts that are installed this way always default to somewhere in home dir. It looks very suspicious when a script like that requires `sudo`.

I'm not sure If we want to keep the same behavior between platform, but I think that for Linux it would be better to 
- copy all code to `~/.eas` 
- make sure that line `export PATH=$HOME/.eas/bin:$PATH` exists in `.bashrc` and `~/.zshrc`

or instead of adding it PATH we could run some other script (e.g. the same as nvm)
```
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
```

ide: We wanted to avoid automatically modifying .bashrc/.zshrc because it's fragile to update and people may use multiple shells. On macOS, `/usr/local/{lib,bin}` are set up by Homebrew with the primary user's own permissions, so they should be writeable most of the time.

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+      sudo ln -fs "$prefix/lib/eas/bin/eas" "$prefix/bin/eas"
+    else
+      
+      mv "$temp_dir/eas" "$prefix/lib/eas"
+      ln -fs "$prefix/lib/eas/bin/eas" "$prefix/bin/eas"
+    fi
+  }
+
+  if [[ ! ":$PATH:" == *":$prefix/bin:"* ]]; then
+    abort "Your path is missing $prefix/bin, you need to add this to use this installer."
```
ide, line 66:
```suggestion
    abort "Your path is missing $prefix/bin, you need to add this to \$PATH use this installer."
```

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+  fi
+  
+  hardware="$(uname -m)"
+  case "$hardware" in
+    x86_64)
+      arch="x64"
+      ;;
+    arm*)
+      arch="arm"
+      ;;
```
ide, line 76: We may want to check for arm64 specifically since we will need arm64 builds.

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+      abort "Unsupported CPU architecture $hardware."
+      ;;
+  esac
+
+  case "$(uname -s)" in
+    Darwin*)
+      platform="darwin"
+      if [[ "$arch" != "x64" ]]; then
+        abort "Unsupported CPU architecture $hardware on macOS."
+      fi
```
ide, line 87: We will need arm64 on macOS soon

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+      ;;
+    Linux*)
+      platform="linux"
+      ;;
+    *)
+      abort "This installer is only supported on macOS, Linux and Windows Subsystem for Linux."
```
ide, line 93:
```suggestion
      abort "This installer is supported only on macOS, Linux, and Windows Subsystem for Linux."
```

bin/install.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
+
+# This script downloads the latest release version of EAS CLI from GitHub and
+# installs it in /usr/local/bin
+
+{
+  # All code is inside a block to ensure the script executes only when
+  # downloaded completely.
+
+  set -e
```
ide, line 10:
```suggestion
  set -euo pipefail
```

https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
(leave off `x`)

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+      needs_sudo=true
+    fi
+    # Run commands with sudo if necessary.
+    if [ -n "$needs_sudo" ]
+    then
+      echo "Installing requires superuser access."
+      echo "The sudo command will prompt for your password."
+      sudo rm -rf "$prefix/lib/eas"
+      sudo mv "$temp_dir/eas" "$prefix/lib/eas"
+      sudo ln -fs "$prefix/lib/eas/bin/eas" "$prefix/bin/eas"
```
ide, line 57: Should we chown lib/eas and bin/eas so they are owned by the current user (not superuser?)

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+  install
+
+  # Test the bundled node binary and if it doesn't work remove it (fall back to
+  # node installed in path).
+  "$prefix/lib/eas/bin/node" -v > /dev/null || {
+    rm "$prefix/lib/eas/bin/node"
+  }
+
+  # Test the CLI
+  executable=$(command -v eas)
```
ide, line 122: We might want to put this in an `if` check and print a nice message if it fails for an unexpected reason

ide (changes requested):

bin/install.sh
wkozyra95: > fragile to update 

We need only to make sure that one specific line is present and if not add that to the end of the file

>  people may use multiple shells

If `rc` file exists we could add there automatically(.barc and .zshrc is probably enough) and for other shells, we could print the message "Make sure to add this line to you rc file.."

I think that requiring `sudo` is acceptable only with a package supported by distro(`deb`, `rpm`, ...) where it's clear how it works and how to uninstall it or update. If I saw install script doing sth like this I wouldn't use it if had a choice, I don't think that this will be a deal-breaker for most of the users, but I think it's a pretty significant red flag.

bin/install.sh
```diff
@@ -0,0 +1,129 @@
...
+    fi
+    if ! rm -rf "$prefix/lib/eas"
+    then
+      needs_sudo=true
+    fi
+    # Run commands with sudo if necessary.
+    if [ -n "$needs_sudo" ]
+    then
+      echo "Installing requires superuser access."
+      echo "The sudo command will prompt for your password."
```
wkozyra95, line 54: If the user used command with sudo before running install script it will continue without asking user.

bin/install.sh
wkozyra95: tar is preserving ownership base on UID/GID (in this case it's 1000). So if you run this script now lib/eas lib eas is owned by main user (in most cases), `bin/eas` is owned by root. 

Nothing in system directories that affects all users in the system should be ever owned by user, the only exception I have seen was users that exist for specific services( I think jenkins is doing sth like that).

> Should we chown lib/eas and bin/eas so they are owned by the current user (not superuser?)

Regardless if you agree with main previous point explicit `chown` is necessary to either `root` or current user, but I strongly believe that it should be owned by root if it's in `/usr`

ide: With Homebrew the system files are owned by one user and it generally works well because most people have their own computer.

ide:
If sudo on Linux is an issue, we could prioritize macOS and WSL2 first, where most Expo developers are. The major use case for Linux which we need to support early on is in CI containers (including Alpine I think, since alpine-node is popular).

packages/eas-cli/README.md
```diff
@@ -69,7 +67,29 @@ OPTIONS
...
+```
+USAGE
+  $ eas credentials
+```
+
+_See code: [build/commands/credentials.ts](https://github.com/expo/eas-cli/blob/v0.0.0/build/commands/credentials.ts)_
+
+## `eas device:create`
+
+register new Apple Devices to use for internal distribution
```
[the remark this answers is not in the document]
fson: These are generated from the descriptions of commands, so I'm leaving changing this outside this PR.
https://github.com/expo/eas-cli/blob/9b807bb2da5263df7cc663e60d7f64d728bcfe87/packages/eas-cli/src/commands/device/create.ts#L9

bin/install.sh
fson: Oxford comma ftw

packages/eas-cli/README.md
fson: These lines were automatically added, by the `oclif-dev readme` command, not sure why it included the paths from build output. Re-running it seems to have removed them.

.github/workflows/install.yml
fson: I've added macOS to this test now. I can do WSL in a separate PR once I figure out how to invoke it from a GH action :)

bin/install.sh
fson: I've had problems with tools that use this profile modification approach in the past. E.g. we only add it to `.zshrc` or `.zprofile` because the user is using Zsh and/or because those files exist. Now if the use (even temporarily) switches to Bash (sometimes it's a script written by someone else, possibly running inside a build phase of Xcode, in a VS Code plugin, possibilities are endless..), the user will just get an error like `bash: eas: command not found` and it's not obvious at all why this is happening or what they should do to fix it.

There are many files too: `.profile`, `.bashrc`, `.bash_profile`, `.zshrc` and `.zprofile`... just for Bash and Zsh. Which one would we pick? Search in all of them in case the user previously added the snippet manually?

Additionally, I wouldn't copy nvm too closely: not sure how it does it but its startup script adds a visible delay to opening terminal Windows which is why I switched away from nvm a long time ago.

These are tradeoffs, but since I was aiming to create a single-command, reliable way to install the CLI, having to sometimes type your password seems less bad than having the script randomly fail or require manual steps. The current approach also works well in CI systems where the code is already running as root.

wkozyra95: I agree that interacting with `rc`/`profile` file is not ideal, but (in case of linux) this is a best way if we want to have both simple install script and seamless autoupdate.

- proper solution would be to have distro specific packages, but it's probably not something we have time right now.
- current solution meets both conditions, but you are assigning ownership to non root user for files in `/usr/local` which looks very sketchy.
- If we would go with current solution, but also `chown` files as root, it would be a lot better, it's still no ideal and we lose easy auto update, because password would be required every time, but it's probably good enough. If you feel strongly against bashrc/zshrc approach I would suggest to go this way.


For all non linux cases I agree that this approach is perfectly fine, with one issue I mentioned in other comment. You are assuming that user will always have uid=1000(that is how `tar` stores ownership info), which may not be always true.

fson: I think we will be able to create at least Ubuntu/Debian packages (.deb) with relative ease a bit later, if users ask for it, because oclif dev cli has support for this out of the box (https://oclif.io/docs/releasing#ubuntudebian-packages). These won't have our automatic updates, but I think it's fair to assume that a user installing using that method wants to use the package manager of the distro to also handle updates, so that's fine.

Owning the files as root doesn't break automatic updates because oclif has built in support for them in the executable that looks for the updates in the home dir. So we could `chown` the files as root, perhaps we should do that at least on Linux. I'm still thinking whether we should do that on macOS too - always running the script with sudo would make things simpler in many ways. (If the user has permissions to `/usr/local` that adds a bit of unnecessary friction, but we can't rely on having these permissions anyway.

ide: >  it's fair to assume that a user installing using that method wants to use the package manager of the distro to also handle updates

Definitely -- it would be confusing if apt/dnf thinks the package is v1.0.0 but the automatic updater actually upgraded the package to v1.0.1.

> So we could chown the files as root, perhaps we should do that at least on Linux.

Perhaps the cleanest heuristic to is to `chown` eas-cli's files to match the user + group of the parent directory (`/usr/local/bin` and `/usr/local/lib`). On macOS with a normal Homebrew installation (I set up a new computer a few months ago), this looks like `$me:admin` for `bin` and `$me:staff` for `lib`.

nicknovitski (approved):

PR #35, opened 2020-10-19
Add tsc build to test GH action

fson:


packages/eas-cli/package.json
```diff
@@ -50,6 +50,7 @@
     "@babel/preset-typescript": "^7.10.4",
     "@expo/babel-preset-cli": "^0.2.17",
     "@expo/oclif-dev-cli": "1.23.0-expo",
+    "@types/color-string": "^1.5.0",
```
wkozyra95, line 53: This should not be necessary, configure-splash-screen is not reexporting this type anymore, but that change was not published yet. `configure-splash-screen` is now in `unlinked-packages` and I wasn't sure if I can just publish it.

For now it's probably better to just add this type. :+1:

wkozyra95 (approved):

packages/eas-cli/package.json
fson: The problem was that we depend on it through `@expo/config` which wasn't upgraded to use the newer version yet. Fixed here: https://github.com/expo/expo-cli/pull/2810

PR #38, opened 2020-10-21
[eas-cli] Move all GraphQL to one directory

barthap:
## Summary
Each commit is a different stage:
1. https://github.com/expo/eas-cli/pull/38/commits/d82d6dbbb1bce69b5b90faee8c42d3b4f85580d5 - Move almost all GQL calls to separate folder `graphql`. Partially implement all interfaces corresponding to GQL schema.
2. https://github.com/expo/eas-cli/pull/38/commits/d597a313c4eacb4976b70e88f4fa7cdb593fc69c - Add common error handling method for all GQL query calls.
3. ~~**(Proposal)** Add intelligent QueryBuilder with some TypeScript magic - it provides IntelliSense autocomplete - query result type matches.~~ Whatever

## `graphql` folder structure
- `queries` - Domain specific Query API, e.g. `BuildsQuery`
- `mutations` - Same as above, but mutations. Currently empty, we don't have any mutations yet.
- `types` - TypeScript interfaces directly corresponding to GQL schema types, e.g. `User`, `Project`, `Build`. Generated typedefs should probably land here
- `client.ts` - GraphQL client + error handling
- ~~`queryTools.ts` - Query Builder implementation and some TypeScript magic~~

<details>
<summary>

~~query builder~~

</summary>

## ~~Proposal: `QueryBuilder` usage.~~

~~When we define new query in `graphql/queries/XyzQuery.ts`, we return a new QueryBuilder instance there and provide a _execute function_. The function receives built fragments and query string and is supposed to return a promise of GQL result data.~~

```ts
export class XyzQuery {
  static byId(id: string) {
    return new QueryBuilder<Xyz>('Xyz', {
      query: ({ fields }) => gql(`{
        xyz {
          byId(id: "${id}") { ${queryString} } 
        }
      }`),
      resolve: data => data.xyz.byId
    });
  }
}
```

~~We can also define fragments:~~

```ts
export const buildArtifactsFragment = createFragment<Build, Pick<Build, 'artifacts'>>(
    'ArtifactsFragment',
    'fragment ArtifactsFragment on Build { artifacts { buildUrl } }'
);
```

~~Usage~~

```ts
const fetchedXyz = await XyzQuery.byId('someId')
  .withFields('id', 'name') // IntelliSense suggests correct field names here
  .withFragments(someFragment) // this method has an alias: withEdges()
  .executeAsync();

fetchedXyz.  // IntelliSense works here. Displays `id`, `name` and type returned by `someFragment`
```
</details>


packages/eas-cli/src/user/User.ts
```diff
@@ -1,16 +1,17 @@
...
+import { User } from '../graphql/types/User';
+import { UserAccount } from './Account';
 import { getSession, setSessionAsync } from './sessionStorage';
 
-export interface User {
-  userId: string;
-  username: string;
-  accounts: Account[];
-}
+export type CurrentUser = Pick<User, 'id' | 'username'> & {
```
wkozyra95, line 10:
```suggestion
export type User = Pick<User, 'id' | 'username'> & {
```

packages/eas-cli/src/credentials/ios/api/AppleTeam.ts
```diff
@@ -2,11 +2,11 @@ import assert from 'assert';
 import gql from 'graphql-tag';
 
 import { graphqlClient } from '../../../api';
-import { Account } from '../../../user/Account';
+import { UserAccount } from '../../../user/Account';
 
 export interface AppleTeam {
```
wkozyra95, line 7: This file should be moved to graphql directory

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const BuildFragments = {
+  Artifacts: createFragment<Build, ArtifactFragmentType>(
+    'ArtifactsFragment',
+    'fragment ArtifactFragment on Build { artifacts { buildUr } }'
+  ),
+};
+
+export class BuildQuery {
+  static byId(buildId: string) {
```
wkozyra95, line 21:
```suggestion
export namespace BuildQuery {
  export function byId(buildId: string): QueryBuilder<Build> {
```
I don't have a strong opinion on that, but if you just want force users  of this API to use certain names  and you are never instantiating this class namespaces are more appropriate

cc @dsokal

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,63 @@
...
+          byId(buildId: "${buildId}") {
+            ${fields}
+          }
+        }
+      }`),
+      resolve: data => data.builds.byId,
+    });
+  }
+
+  static allForApp(appId: string, filters?: Filters) {
```
wkozyra95, line 35: add return type

packages/eas-cli/src/graphql/queries/ProjectQuery.ts
```diff
@@ -0,0 +1,22 @@
+import gql from 'graphql-tag';
+
+import { QueryBuilder } from '../queryTools';
+import { Project } from '../types/Project';
+
+type ProjectIdType = Pick<Project, 'id'>;
+
+export class ProjectQuery {
+  static byUsernameAndSlug(username: string, slug: string) {
```
wkozyra95, line 9: return type

packages/eas-cli/src/graphql/queries/UserQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+  'UserAccountsFrag',
+  'fragment UserAccountsFrag on User { accounts { id, name } }'
+);
+
+export const UserFragments = {
+  Accounts: accountsFragment,
+};
+
+export class UserQuery {
+  static currentUser() {
```
wkozyra95, line 17: return type

wkozyra95 (commented):
I probably missed few places, so make sure that you are specifying return type everywhere

packages/eas-cli/src/user/User.ts
barthap: Why revert?
cc @dsokal

packages/eas-cli/src/graphql/queries/BuildQuery.ts
barthap: I like the idea, however we don't use namespaces anywhere else and I'm not sure if we should.

packages/eas-cli/src/credentials/ios/api/AppleTeam.ts
barthap: graphql has its own Account interface, which is a superset of this one.

packages/eas-cli/src/user/User.ts
dsokal: @wkozyra95 We decided to rename it to avoid confusion with the `User` type from graphql directory.

dsokal: Unless you have a really strong argument why we shouldn't rename this, it'd leave it as Bartek proposed.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
dsokal: Let's stay with the class + static functions approach.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
dsokal: Maybe we can add an eslint rule for that?

barthap: Can be, I intentionally didn't specify return types in some `QueryBuilder` methods, because they're too complicated and generic.

But I think here they can be safely added and they shouldn't break anything 🤔

packages/eas-cli/src/graphql/__tests__/queryTools.test.ts
```diff
@@ -0,0 +1,174 @@
+import { DocumentNode, FragmentDefinitionNode, OperationDefinitionNode } from 'graphql';
```
dsokal, line 1: Please rename this file.
The convention is to suffix tests filenames with`-test`, not `.test`.

packages/eas-cli/src/user/User.ts
wkozyra95: We will type from `user/User.ts` almost exclusively in the entire codebase and CurrentUser is not a very good name, 
In my opinion, a type should describe structure not identify exact value, in this case, a variable of that type can only have one specific value. 

Also, I don't see a problem with the confusion of those two types, in another case I would probably agree, but graphql type names will collide quite often with types we define in cli and it most cases they won't be used directly but like here via some other type with a more specific scope.

One more thing I would prefer here is to define User the way it was not with `Pick`, ts will check if types are correct either way.

packages/eas-cli/src/submissions/utils/builds.ts
```diff
@@ -1,41 +1,23 @@
...
-    throw error.graphQLErrors[0];
-  }
-
   const {
     platform: buildPlatform,
     artifacts: { buildUrl },
-  } = data.builds.byId;
+  } = await BuildQuery.byId(buildId)
+    .withFields('platform')
+    .withFragments(BuildFragments.Artifacts)
```
dsokal, line 19:
```suggestion
    . withEdges(BuildFragments.Artifacts)
```

dsokal: And maybe update other places as well.

packages/eas-cli/src/graphql/types/Actor.ts
```diff
@@ -0,0 +1,11 @@
+import { AccessToken } from './AccessToken';
+import { Account } from './Account';
+
+export interface Actor {
```
dsokal, line 4: Hmmm, maybe this could be a class instead of an interface. This way we could have both type checking and easy access to the class name (and then we wouldn't need the first argument for QueryBuilder). See how `@expo/entity` does a similar thing with entity classes.

dsokal: Feel free to change that in a follow-up PR.

dsokal (approved):
This is great. Thanks!

packages/eas-cli/src/user/User.ts
wkozyra95: discussed offline with @dsokal , let's revert back to User.

packages/eas-cli/src/graphql/__tests__/queryTools.test.ts
barthap: Ok, but I'm gonna miss that colored icon:
Screenshot 2020-10-22 at 15 43 57

Now I have to find a VS Code theme which supports this naming pattern 😅

packages/eas-cli/src/graphql/queryTools.ts
```diff
@@ -0,0 +1,192 @@
...
+    return (this as unknown) as QueryBuilder<
+      Entity,
+      T & UnionToIntersection<typeof fragments[number]['__type']>
+    >;
+  }
+
+  /**
+   * Alias for `withFragments`
+   */
+  withEdges<U extends Fragment<ArrayElement<Entity>, unknown>[]>(...edges: U) {
```
wkozyra95, line 110: missing return type

packages/eas-cli/src/graphql/client.ts
```diff
@@ -0,0 +1,60 @@
...
+        headers: {
+          'expo-session': sessionSecret,
+        },
+      };
+    }
+    return {};
+  },
+});
+
+export async function withErrorHandling<T>(promise: Promise<OperationResult<T>>): Promise<T> {
```
wkozyra95, line 44:
```suggestion
export async function withErrorHandlingAsync<T>(promise: Promise<OperationResult<T>>): Promise<T> {
```

wkozyra95 (approved):

packages/eas-cli/src/graphql/queryTools.ts
barthap: Intentionally, https://github.com/expo/eas-cli/pull/38#discussion_r510142083

fson:
I see this pull request adds a new abstraction `QueryBuilder` on top of the small `@urql/core` library that we started to use for GraphQL. Could you explain the idea behind `QueryBuilder` and why it's needed rather than using the functionality provided by urql directly?

For TypeScript types of GraphQL results, the plan was to use https://graphql-code-generator.com which supports urql with the `@graphql-codegen/typescript-urql` package. GraphQL Code Generator is also used to generate the TS types on  the`www` project, so I think it should work well for this project too. (Apologies for not having finished setting up it yet.)

dsokal:
Hey @fson,

First of all, I think we should have discussed the approach for GraphQL queries in EAS CLI earlier. Sorry that we haven't done this.
As far as I know, we considered using the graphql-code-generator. As you said, it's already used in another part of the Expo universe. However, we use it in the website's code, not in www. @barthap was researching this and decided that have a custom solution would fit our needs better. I don't remember the exact reasons but I think he can share the thought process with us and tell us about everything he found out when researching this.

Generally, the main problem with GraphQL in EAS CLI was that we already started using it quite heavily and we haven't seen any moves regarding adding standardized support for it. This is why I asked @barthap to take care of this. So far, all Graphql queries were spread across different files/directories and it was quite hard to understand/decide where to put GraphQL code. The main idea behind this PR is that we wanted to put all GraphQL code in a separate directory (`src/graphql`). The other features, like QueryBuilder, are only additions on top of the most important part (code structure).

barthap:
As Dominik said, the main thing we want to do was to move all GraphQL related logic to one place, the `graphql` folder. That's quite obvious. Then we wanted to have unified and universal/flexible API for all queries to avoid multiplicating similar methods like `getBuildArtifactsByIdAsync()`, `getBuildsPlatformAndIdForAppAsync()` etc. And TypeScript autocomplete support was a _nice to have_ feature.
I was looking for some existing solutions and found several query builders and generator libraries (including the one above), however their usage was ugly, or not that easy to use. I decided to do some experiments with TS magic and came out with this simple QueryBuilder, which should be sufficient for us in most cases.
Regarding to the generator - I'm still considering using it in the future just for generating TypeScript interfaces and putting them into `src/graphql/types` (typing them fully by hand is an error-prone nightmare). This query builder should have no problems when using generated types.

fson:
I think defining all queries in one place is a good idea and we should do that. The website has a nice example of this at https://github.com/expo/universe/tree/a4916da65c5b56e0a12a45058f518d64d35be4e5/server/website/graphql. I think we can do something very similar and have static definitions of all queries in one place.

However, we should not add our own custom query builder library on top of the GraphQL client library we use. Developing and maintaining a separate layer of abstraction like this increases the cost of maintenance for us without any obvious gains. You mention you have evaluated existing libraries, but you haven't named the libraries or mentioned the exact problems you had with them.

Anyway, I'm in favor colocating queries, so a good next step would be to update this PR to only move queries to a single folder and leave the `QueryBuilder` class out of this PR. We would keep the existing approach of using `urql` to fetch data, but now with pre-defined queries. I can also help get the typegen integrated tomorrow so there's no need to write and maintain any type defs by hand.

In the future, please discuss any major architectural changes like this with me first before investing time to implement the change.

brentvatne:
it's nice to see some movement on the graphql integration in eas-cli! we should split this into two prs:

1) uncontroversial changes: d82d6db, d597a31. we can go ahead and land these right away.
2) proposal of QueryBuilder: other commits. we should discuss this further to make sure we're all on board with maintaining our own abstraction, and if we do go that route then we have documentation in the form of the pr discussion to help future-us and other future expo developers understand our motives and decision criteria.

in (2) we could include the QueryBuilder proposal from the above PR description and add on a section that lists some of the alternatives that were considered and where they fell short of our requirements. i think with that we will all be on the same footing to proceed with deciding what the best course of action is.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+import { graphqlClient, withErrorHandlingAsync } from '../client';
+import { Build } from '../types/Build';
+
+type Filters = Partial<Pick<Build, 'platform' | 'status'>> & {
+  order?: number;
+  limit?: number;
+};
+
+type ArtifactFragmentType = Pick<Build, 'artifacts'>;
+type PlatformAndArtifactFragmentType = Pick<Build, 'platform' | 'artifacts'>;
```
dsokal, line 12: - I think the names are inaccurate (`artifact` -> `artifacts`) and the `Type` suffix seems weird.
- I don't have an idea for better names now - feel free to suggest sth.
- I'd just define the types inline if needed. You're not exporting them anyway.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export class BuildQuery {
+  static async forArtifactByIdAsync(buildId: string): Promise<PlatformAndArtifactFragmentType> {
```
dsokal, line 15:
```suggestion
  static async byIdAsync(buildId: string): Promise<PlatformAndArtifactFragmentType> {
```

dsokal: - The function name should match the query name.
- If we're going to call this query in different contexts then I'd prefer one of the following approaches:
   - pass a fragment to the function
   - pass an array of fields/edges to fetch
   - always return the same fragment (this is probably the best option as it simplifies the types)

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+            }
+          }`
+        )
+        .toPromise()
+    );
+
+    return data.builds.byId;
+  }
+
+  static async allArtifactsForAppAsync(
```
dsokal, line 37:
```suggestion
  static async allForAppAsync(
```

packages/eas-cli/src/graphql/queries/ProjectQuery.ts
```diff
@@ -0,0 +1,28 @@
+import gql from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../client';
+import { Project } from '../types/Project';
+
+type ProjectIdType = Pick<Project, 'id'>;
+
+export class ProjectQuery {
+  static async idByUsernameAndSlugAsync(username: string, slug: string): Promise<ProjectIdType> {
```
dsokal, line 9:
```suggestion
  static async byUsernameAndSlugAsync(username: string, slug: string): Promise<ProjectIdType> {
```

packages/eas-cli/src/graphql/queries/ProjectQuery.ts
```diff
@@ -0,0 +1,28 @@
+import gql from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../client';
+import { Project } from '../types/Project';
+
+type ProjectIdType = Pick<Project, 'id'>;
```
dsokal, line 6: Same here.

dsokal (changes requested):

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+          {
+            builds {
+              byId(buildId: "${buildId}") {
+                platform,
+                artifacts {
+                  buildUrl
+                }
+              }
+            }
+          }`
```
fson, line 29: Rather than string interpolation, please use [variables](http://spec.graphql.org/draft/#sec-Language.Variables) to pass parameters like this:
```suggestion
          gql`
          query ($buildId: ID!) {
            builds {
              byId(buildId: $buildId) {
                platform
                artifacts {
                  buildUrl
                }
              }
            }
          }`, { buildId }
```
This feature is built in the GraphQL language to avoid piecing queries together from strings. It also ensures the queries are maximally reusable and is less error prone when it comes to things like correctly escaping the variables.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+          gql`
+          {
+            builds {
+              allForApp(${filterData.join(', ')}) {
+                artifacts {
+                  buildUrl
+                }
+              }
+            }
+          }`
```
fson, line 68:
```suggestion
          query ($appId: String!, $limit: Int, $offset: Int, $platform: AppPlatform, $status: BuildStatus) {
            builds {
              allForApp(appId: $appId, limit: $limit, offset: $offset, platform: $platform, status: $status) {
                artifacts {
                  buildUrl
                }
              }
            }
          }, { ...filters, appId }
```
Variables should be used here as well. A nice thing about this is that it eliminates the need to generate GraphQL from string snippets.

packages/eas-cli/src/graphql/queries/ProjectQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class ProjectQuery {
+  static async idByUsernameAndSlugAsync(username: string, slug: string): Promise<ProjectIdType> {
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .query<{ project: { byUsernameAndSlug: ProjectIdType } }>(
+          gql`
+          {
+            project {
+              byUsernameAndSlug(username: "${username}", slug: "${slug}", sdkVersions: []) {
```
fson, line 16: Please use variables.

fson (changes requested):

packages/eas-cli/src/graphql/queries/ProjectQuery.ts
fson:
```suggestion
              byUsernameAndSlug(username: $username, slug: $slug) {
```
You can also omit `sdkVersions` because it's an optional parameter.

barthap: Indeed, this is a leftover after a bug in `expo/universe`, this parameter was required to avoid crashes, but now it can be omitted.

Regarding to variables, I was wondering why they weren't used before and finally I just sticked to existing string interpolation approach. But I'm glad you're in favour of using them.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
barthap: I just went with the 3rd approach, the two former needs some kind of query building and... never mind.

packages/eas-cli/src/graphql/queries/BuildQuery.ts
barthap: I've reduced those types to one, common for each query function, named: `BuildQueryResult` - IMO it's pretty accurate name. As the type is used in multiple places across the same file, I didn't inline it.

dsokal (approved):

fson (approved):
Thanks, let's merge it 👍

PR #54, opened 2020-10-29
Add release workflow

fson:


package.json
```diff
@@ -6,9 +6,11 @@
   "scripts": {
     "eas": "packages/eas-cli/bin/run",
     "lint": "eslint . --ext .ts",
+    "release": "lerna version"
```
byCedric, line 9: This is missing a trailing `,` I think?

byCedric (approved):
Looks good! Only 1 comment, that's it!

package.json
fson, line 9:
```suggestion
    "release": "lerna version",
```

PR #56, opened 2020-11-03
[eas-cli] Basic Submission command for iOS

barthap:
Core submission functionality for iOS. As for now, it only supports submitting builds for already-existing apps on App Store Connect - user has to provide Application's `Apple ID` number on his own.

> Reminder: Application's App-Store-Connect `Apple ID` number is not the same as Apple ID username! This is an unique application number in App Store Connect. It can be found there until Your App -> General -> App Information -> General Information

Basic usage:
```bash
eas build:submit --platform ios --apple-id=... --apple-app-specific-password=... --app-apple-id=...
```

This PR provides basic functionality (it is sufficient to run it e.g. on CI), but there's still much TODO:

- ~~Check for interactive/non-interactive mode and return error instead of prompts (both Android and iOS)~~ _will need to discuss it more, probably add it in subsequent PR_
- ~~Refactor code - there are some duplicates between Android and iOS which can be merged~~ -> #58 
- Improve UX
  - Custom `--help` with separated Android and iOS. Now it's a mess. - _a single-command custom help would need a deeper oclif modifications, it's worth moving to subsequent PR_
  - ~~Print summary for iOS submissions - e.g. in the table, same as for Android~~ -> #58
- Use `fastlane produce` to create app on App Store Connect if it doesn't exist. - _Will be done in subsequent PR_
  - ~~Will need modification of `travellingFastlane`~~ (https://github.com/expo/expo-cli/pull/2855 merged, published 1.15.2)
  - Automatically guess `appAppleId` based on expo-config `ios.bundleIdentifier`
  - Create application if it doesn't exist on App Store Connect (interactive mode only)


packages/eas-cli/src/submissions/archive-source/ArchiveTypeSource.ts
```diff
@@ -103,9 +107,11 @@ type ArchiveInferredType = ArchiveType | null;
 
 function inferArchiveTypeFromLocation(location: string): ArchiveInferredType {
   if (location.endsWith('.apk')) {
-    return ArchiveType.apk;
+    return AndroidArchiveType.apk;
   } else if (location.endsWith('.aab')) {
-    return ArchiveType.aab;
+    return AndroidArchiveType.aab;
+  } else if (location.endsWith('.ipa')) {
```
dsokal, line 113: I'm not sure if this logic is correct. What if someone was running `eas build:submit --platform android` and named the archive `my-android-app.ipa`? I know this is a stupid and infrequent case, but in this case, the result of this function is incorrect.

packages/eas-cli/src/submissions/ios/AppSpecificPasswordSource.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function getAppSpecificPasswordAsync(source: AppSpecificPasswordSource) {
+  if (source.sourceType === AppSpecificPasswordSourceType.userDefined) {
+    return source.appSpecificPassword;
+  } else if (source.sourceType === AppSpecificPasswordSourceType.prompt) {
+    log.addNewLineIfNone();
+    const wrap = wordwrap(process.stdout.columns || 80);
+    log(
+      wrap(
+        chalk.bold(
+          `The password is only used to upload the build to Apple TestFlight and never stored on Expo servers`
```
dsokal, line 38: I think that in the future we will actually want to store the app-specific password on our servers. Treat this comment as a heads-up :)

packages/eas-cli/src/submissions/ios/AppSpecificPasswordSource.ts
```diff
@@ -0,0 +1,54 @@
...
+        )
+      )
+    );
+
+    const { appSpecificPassword } = await promptAsync({
+      name: 'appSpecificPassword',
+      message: 'Your Apple application-specific password:',
```
dsokal, line 45:
```suggestion
      message: 'Your Apple app-specific password:',
```

Screenshot 2020-11-04 at 11 06 19

We should be consistent with Apple.

packages/eas-cli/src/submissions/ios/AppSpecificPasswordSource.ts
```diff
@@ -0,0 +1,54 @@
...
+      type: 'text',
+      validate: (val: string) =>
+        val !== '' || 'Apple application-specific password cannot be empty!',
```
dsokal, line 48:
```suggestion
        val !== '' || 'Apple app-specific password cannot be empty!',
```

packages/eas-cli/src/submissions/ios/IosSubmitCommand.ts
```diff
@@ -0,0 +1,214 @@
...
+    const { appleId } = this.ctx.commandFlags;
+    const { EXPO_APPLE_ID } = process.env;
+
+    if (appleId) {
+      return appleId;
+    } else if (EXPO_APPLE_ID) {
+      return EXPO_APPLE_ID;
+    }
+
+    return promptForAppleIdAsync();
```
dsokal, line 176:
```suggestion
    } else {
      return await promptForAppleIdAsync();
    }
```

dsokal (commented):
This is looking good so far. Just a few comments inline.

packages/eas-cli/src/submissions/archive-source/ArchiveTypeSource.ts
barthap: Right, I should make the infer method smarter.

packages/eas-cli/src/submissions/ios/AppSpecificPasswordSource.ts
barthap: +1 for this. I was consistent with Fastlane until now, but was annoyed with very long names, e.g. `FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD`

packages/eas-cli/src/submissions/ios/AppSpecificPasswordSource.ts
wkozyra95: I'm not really sure that even now we can make that claim, technically when job is sent to submission service it's inside pub sub. Technically this is already storing it.

barthap: ok so it's just safer to remove this claim for now.

packages/eas-cli/src/submissions/archive-source/ArchiveTypeSource.ts
```diff
@@ -101,11 +105,16 @@ async function handlePromptSourceAsync(
...
 type ArchiveInferredType = ArchiveType | null;
 
-function inferArchiveTypeFromLocation(location: string): ArchiveInferredType {
-  if (location.endsWith('.apk')) {
-    return ArchiveType.apk;
+function inferArchiveTypeFromLocation(
+  platform: SubmissionPlatform,
+  location: string
+): ArchiveInferredType {
+  if (platform === SubmissionPlatform.iOS) {
```
dsokal, line 112: Maybe sth like this would be cleaner:
```suggestion
  if (platform === SubmissionPlatform.iOS) {
  } else {
     if (...) {
     } else if () { 
     } 
  }
```
I know this is logically the same, but IMO this is easier to read.

packages/eas-cli/src/submissions/commons.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (chosenOptions.filter(opt => opt).length > 1) {
+    throw new Error(`Pass only one of: --url, --path, --id, --latest`);
+  }
+
+  if (url) {
+    return {
+      sourceType: ArchiveFileSourceType.url,
+      url,
+      projectId,
+      platform: SubmissionPlatform.iOS,
```
dsokal, line 32: Why `.iOS`?

packages/eas-cli/src/submissions/commons.ts
```diff
@@ -0,0 +1,69 @@
...
+      projectDir: ctx.projectDir,
+    };
+  } else if (path) {
+    return {
+      sourceType: ArchiveFileSourceType.path,
+      path,
+      projectId,
+      platform: SubmissionPlatform.iOS,
```
dsokal, line 40: The same - why? The same applies to other branches.

dsokal (approved):
Looks really good. Great job!

packages/eas-cli/src/submissions/commons.ts
barthap: Whoops, my mistake - I must have missed something when committing

wkozyra95 (approved):

brentvatne:
sorry i missed this one! if you ever need my attention on something feel free to dm me on slack :)

PR #63, opened 2020-11-11
Windows builds

fson:
This PR builds on #62 and adds Windows 32-bit and 64-bit installers, built with `oclif-dev pack:win`. We build these on Ubuntu because that's how the pack commands in `oclif-dev` are supposed to be used (they don't run on Windows).

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@eas/config",
   "description": "A library for interacting with the eas.json",
-  "version": "0.1.0-alpha.1",
+  "version": "0.1.0-alpha.2",
   "author": "Expo <support@expo.io>",
   "bugs": "https://github.com/expo/eas-cli/issues",
   "dependencies": {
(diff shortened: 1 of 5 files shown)


byCedric:
For future reference, here are some "interesting links"
- https://github.com/oclif/dev-cli/issues/104#issuecomment-489644753
- https://github.com/oclif/dev-cli/blob/master/src/commands/pack/index.ts#L20
- https://github.com/oclif/dev-cli/issues/66#issuecomment-426042760

packages/config/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@eas/config",
   "description": "A library for interacting with the eas.json",
-  "version": "0.1.0-alpha.1",
+  "version": "0.1.0-alpha.2",
```
byCedric, line 4: shouldn't this one be `alpha.3` too?

byCedric (approved):

packages/config/package.json
fson: Not sure why Lerna did it like this 🤔

PR #64, opened 2020-11-12
[eas-cli] Implement apple utils

EvanBacon:
- Implement `apple-utils` package behind `USE_APPLE_UTILS` env variable
- Reformat session cookies to match custom Fastlane YAML format
- Remove unused `ownerType` property from certificates type
- Largest change is in `packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts`

# Test Plan

- TBD

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -9,6 +9,7 @@
   "bugs": "https://github.com/expo/eas-cli/issues",
   "dependencies": {
     "@eas/config": "^0.1.0-alpha.2",
+    "@expo/apple-utils": "^0.0.0-alpha.2",
     "@expo/config": "^3.3.12",
     "@expo/eas-build-job": "0.1.1",
     "@expo/json-file": "^8.2.24",
@@ -25,6 +26,7 @@
     "env-paths": "^2.2.0",
     "form-data": "^3.0.0",
     "fs-extra": "^9.0.1",
+    "getenv": "^1.0.0",
     "got": "^11.5.2",
     "graphql-tag": "^2.11.0",
     "indent-string": "^4.0.0",
@@ -54,6 +56,7 @@
     "@types/dateformat": "^3.0.1",
     "@types/form-data": "^2.5.0",
     "@types/fs-extra": "^9.0.1",
+    "@types/getenv": "^1.0.0",
     "@types/lodash": "^4.14.161",
     "@types/node": "^12",
     "@types/node-fetch": "^2.5.7",
--- a/packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
@@ -1,9 +1,12 @@
+import { Auth, InvalidUserCredentialsError, Session, Teams } from '@expo/apple-utils';
 import chalk from 'chalk';
+import { toggleConfirmAsync } from 'utils/prompts';
 import wordwrap from 'wordwrap';
 
 import log from '../../../log';
 import { promptAsync } from '../../../prompts';
 import UserSettings from '../../../user/UserSettings';
+import { USE_APPLE_UTILS } from './experimental';
 import { runActionAsync, travelingFastlane } from './fastlane';
 import * as Keychain from './keychain';
 
@@ -15,7 +18,7 @@ export type Options = {
   teamId?: string;
 };
 
-type AppleCredentials = {
+export type AppleCredentials = {
   appleIdPassword: string;
   appleId: string;
 };
@@ -40,7 +43,46 @@ export type AuthCtx = {
   fastlaneSession: string;
 };
 
+async function authenticateWithExperimentalAsync(options: Options = {}): Promise<AuthCtx> {
+  const { appleId, appleIdPassword } = await requestAppleCredentialsAsync(options);
+  log(`Authenticating to Apple Developer Portal...`); // use log instead of spinner in case we need to prompt user for 2fa
+
+  try {
+    // TODO: The password isn't required for apple-utils. Remove the local prompt when we remove traveling Fastlane.
+    await Auth.loginAsync({ username: appleId, password: appleIdPassword });
+
+    log(chalk.green('Authenticated with Apple Developer Portal successfully!'));
+
+    // Get all of the teams
+    const teams = await Teams.getTeamsAsync();
+    const team = await chooseTeamAsync(teams, options.teamId);
+
+    // Set the selected team ID internally
+    Teams.setSelectedTeamId(team.id);
+
+    // Get the JSON cookies in the custom YAML format used by Fastlane
+    const fastlaneSession = Session.getSessionAsYAML();
+    return { appleId, appleIdPassword, team, fastlaneSession };
+  } catch (error) {
+    if (error instanceof InvalidUserCredentialsError) {
+      log.error(error.message);
+      // Remove the invalid password so it isn't automatically used...
+      await deletePasswordAsync({ appleId });
+
+      if (await toggleConfirmAsync({ message: 'Would you like to try again?' })) {
+        // Don't pass credentials back or the method will throw
+        return authenticateAsync({ teamId: options.teamId });
+      }
+    }
+    log(chalk.red('Authentication with Apple Developer Portal failed!'));
+    throw error;
+  }
+}
+
 export async function authenticateAsync(options: Options = {}): Promise<AuthCtx> {
+  if (USE_APPLE_UTILS) {
+    return await authenticateWithExperimentalAsync(options);
+  }
   const { appleId, appleIdPassword } = await requestAppleCredentialsAsync(options);
   log(`Authenticating to Apple Developer Portal...`); // use log instead of spinner in case we need to prompt user for 2fa
   try {
--- a/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
@@ -1,24 +1,82 @@
+import { BundleId, CapabilityType, CapabilityTypeOption, Session, Teams } from '@expo/apple-utils';
 import ora from 'ora';
 
-import { AuthCtx } from './authenticate';
+import { AuthCtx, authenticateAsync } from './authenticate';
+import { USE_APPLE_UTILS } from './experimental';
 import { runActionAsync, travelingFastlane } from './fastlane';
 
-export interface AppLookupParams {
+type Options = {
+  enablePushNotifications?: boolean;
+};
+
+interface BundleIdActionProps {
   accountName: string;
   projectName: string;
   bundleIdentifier: string;
 }
 
-export interface EnsureAppExistsOptions {
-  enablePushNotifications?: boolean;
+export async function ensureAuthenticatedAsync(
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'>
+): Promise<Omit<AuthCtx, 'fastlaneSession'>> {
+  if (!Session.getSessionInfo()) {
+    appleCtx = await authenticateAsync({
+      appleId: appleCtx.appleId,
+      teamId: appleCtx.team.id,
+    });
+  }
+  Teams.setSelectedTeamId(appleCtx.team.id);
+  return appleCtx;
 }
 
-export async function ensureAppExistsAsync(
-  authCtx: AuthCtx,
-  app: AppLookupParams,
-  options: EnsureAppExistsOptions = {}
-): Promise<void> {
-  const { appleId, appleIdPassword, team } = authCtx;
+async function ensureBundleIdExistsAsync(
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'>,
+  { accountName, projectName, bundleIdentifier }: BundleIdActionProps,
+  options: Options = {}
+) {
+  let spinner = ora(`Registering Bundle ID "${bundleIdentifier}"`).start();
+
+  appleCtx = await ensureAuthenticatedAsync(appleCtx);
+
+  // Get the bundle id
+  let bundleId = await BundleId.findAsync({ identifier: bundleIdentifier });
+
+  if (bundleId) {
+    spinner.succeed('Bundle ID already registered');
+  } else {
+    // If it doesn't exist, create it
+    bundleId = await BundleId.createAsync({
+      name: `@${accountName}/${projectName}`,
+      identifier: bundleIdentifier,
+    });
+    spinner.succeed(`Registered Bundle ID "${bundleIdentifier}"`);
+  }
+
+  spinner = ora(`Syncing app capabilities`).start();
+
+  // Update the capabilities
+  await bundleId.updateBundleIdCapabilityAsync({
+    capabilityType: CapabilityType.PUSH_NOTIFICATIONS,
+    option: options.enablePushNotifications ? CapabilityTypeOption.ON : CapabilityTypeOption.OFF,
+    // TODO: Add more capabilities
+  });
+  spinner.succeed(`Sync'd app capabilities`);
+}
+
+export async function ensureAppExists(
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'>,
+  { accountName, projectName, bundleIdentifier }: BundleIdActionProps,
+  options: Options = {}
+) {
+  if (USE_APPLE_UTILS) {
+    return await ensureBundleIdExistsAsync(
+      appleCtx,
+      { accountName, projectName, bundleIdentifier },
+      options
+    );
+  }
+
+  const { appleId, appleIdPassword, team } = appleCtx;
+
   const spinner = ora(`Ensuring App ID exists on Apple Developer Portal...`).start();
   try {
     const { created } = await runActionAsync(travelingFastlane.ensureAppExists, [
@@ -26,11 +84,11 @@ export async function ensureAppExistsAsync(
       appleId,
       appleIdPassword,
       team.id,
-      app.bundleIdentifier,
-      `@${app.accountName}/${app.projectName}`,
+      bundleIdentifier,
+      `@${accountName}/${projectName}`,
     ]);
     if (created) {
-      spinner.succeed(`App ID created with bundle identifier ${app.bundleIdentifier}.`);
+      spinner.succeed(`App ID created with bundle identifier ${bundleIdentifier}.`);
     } else {
       spinner.succeed('App ID found on Apple Developer Portal.');
     }
--- a/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
@@ -5,11 +5,11 @@ import { AuthCtx, authenticateAsync } from './authenticate';
 import { USE_APPLE_UTILS } from './experimental';
 import { runActionAsync, travelingFastlane } from './fastlane';
 
-type Options = {
+export interface EnsureAppExistsOptions {
   enablePushNotifications?: boolean;
-};
+}
 
-interface BundleIdActionProps {
+export interface AppLookupParams {
   accountName: string;
   projectName: string;
   bundleIdentifier: string;
@@ -30,8 +30,8 @@ export async function ensureAuthenticatedAsync(
 
 async function ensureBundleIdExistsAsync(
   appleCtx: Omit<AuthCtx, 'fastlaneSession'>,
-  { accountName, projectName, bundleIdentifier }: BundleIdActionProps,
-  options: Options = {}
+  { accountName, projectName, bundleIdentifier }: AppLookupParams,
+  options: EnsureAppExistsOptions = {}
 ) {
   let spinner = ora(`Registering Bundle ID "${bundleIdentifier}"`).start();
 
@@ -62,17 +62,13 @@ async function ensureBundleIdExistsAsync(
   spinner.succeed(`Sync'd app capabilities`);
 }
 
-export async function ensureAppExists(
+export async function ensureAppExistsAsync(
   appleCtx: Omit<AuthCtx, 'fastlaneSession'>,
-  { accountName, projectName, bundleIdentifier }: BundleIdActionProps,
-  options: Options = {}
+  app: AppLookupParams,
+  options: EnsureAppExistsOptions = {}
 ) {
   if (USE_APPLE_UTILS) {
-    return await ensureBundleIdExistsAsync(
-      appleCtx,
-      { accountName, projectName, bundleIdentifier },
-      options
-    );
+    return await ensureBundleIdExistsAsync(appleCtx, app, options);
   }
 
   const { appleId, appleIdPassword, team } = appleCtx;
@@ -84,11 +80,11 @@ export async function ensureAppExists(
       appleId,
       appleIdPassword,
       team.id,
-      bundleIdentifier,
-      `@${accountName}/${projectName}`,
+      app.bundleIdentifier,
+      `@${app.accountName}/${app.projectName}`,
     ]);
     if (created) {
-      spinner.succeed(`App ID created with bundle identifier ${bundleIdentifier}.`);
+      spinner.succeed(`App ID created with bundle identifier ${app.bundleIdentifier}.`);
     } else {
       spinner.succeed('App ID found on Apple Developer Portal.');
     }
--- a/packages/eas-cli/src/credentials/ios/appstore/experimental.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/experimental.ts
@@ -1,0 +1,26 @@
+import { BundleId, Profile } from '@expo/apple-utils';
+import { boolish } from 'getenv';
+
+export const USE_APPLE_UTILS = boolish('USE_APPLE_UTILS', false);
+
+export async function getProfilesForBundleId(bundleIdentifier: string): Promise<Profile[]> {
+  const [bundleId] = await BundleId.getAsync({
+    query: {
+      filter: {
+        identifier: bundleIdentifier,
+      },
+    },
+  });
+  if (bundleId) {
+    return bundleId.getProfilesAsync();
+  }
+  return [];
+}
+
+export async function getBundleIdForIdentifier(bundleIdentifier: string): Promise<BundleId> {
+  const bundleId = await BundleId.findAsync({ identifier: bundleIdentifier });
+  if (!bundleId) {
+    throw new Error(`Failed to find Bundle ID item with identifier "${bundleIdentifier}"`);
+  }
+  return bundleId;
+}
--- a/packages/eas-cli/src/credentials/ios/appstore/provisioningProfile.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/provisioningProfile.ts
@@ -1,3 +1,4 @@
+import { Certificate, Profile, ProfileType } from '@expo/apple-utils';
 import ora from 'ora';
 
 import { findP12CertSerialNumber } from '../utils/p12Certificate';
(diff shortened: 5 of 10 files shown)


packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -1,24 +1,78 @@
...
+): Promise<Omit<AuthCtx, 'fastlaneSession'>> {
+  if (!Session.getSessionInfo()) {
+    appleCtx = await authenticateAsync({
+      appleId: appleCtx.appleId,
+      teamId: appleCtx.team.id,
+    });
+  }
+  Teams.setSelectedTeamId(appleCtx.team.id);
+  return appleCtx;
+}
```
dsokal, line 29: Hmmm, I feel this function doesn't belong to this file.
@wkozyra95 what do you think?

packages/eas-cli/src/credentials/ios/appstore/provisioningProfile.ts
```diff
@@ -7,8 +8,68 @@ import {
...
+      query: { filter: { serialNumber: [distCertSerialNumber] } },
+    })
+  ).find(item => item.attributes.serialNumber === distCertSerialNumber);
+  if (!cert) {
+    throw new Error(
+      'No cert available to make provision profile against. Make sure you were able to make a certificate prior to this step'
+    );
+  }
+  return cert;
+}
```
dsokal, line 47: I think this should be moved to `distributionCertificate.ts`. 
Remember to update the error message afterward.

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
```diff
@@ -0,0 +1,26 @@
...
+      },
+    },
+  });
+  if (bundleId) {
+    return bundleId.getProfilesAsync();
+  }
+  return [];
+}
+
+export async function getBundleIdForIdentifier(bundleIdentifier: string): Promise<BundleId> {
```
dsokal, line 20:
```suggestion
export async function getBundleIdForIdentifierAsync(bundleIdentifier: string): Promise<BundleId> {
```

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
```diff
@@ -0,0 +1,26 @@
...
+  const bundleId = await BundleId.findAsync({ identifier: bundleIdentifier });
+  if (!bundleId) {
+    throw new Error(`Failed to find Bundle ID item with identifier "${bundleIdentifier}"`);
+  }
+  return bundleId;
+}
```
dsokal, line 26: Maybe you could move these functions to a new module, e.g. `bundleId.ts`.

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
```diff
@@ -0,0 +1,26 @@
+import { BundleId, Profile } from '@expo/apple-utils';
+import { boolish } from 'getenv';
+
+export const USE_APPLE_UTILS = boolish('USE_APPLE_UTILS', false);
```
dsokal, line 4:
```suggestion
export const USE_APPLE_UTILS = Boolean(process.env.USE_APPLE_UTILS);
```
Using `getenv` for this seems to be overkill.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfile.ts
```diff
@@ -57,6 +142,15 @@ export async function listProvisioningProfilesAsync(
 ): Promise<ProvisioningProfileStoreInfo[]> {
   const spinner = ora(`Getting Provisioning Profiles from Apple...`).start();
   try {
+    if (USE_APPLE_UTILS) {
+      let profiles = await getProfilesForBundleId(bundleIdentifier);
+      const type = ctx.team.inHouse ? ProfileType.IOS_APP_INHOUSE : ProfileType.IOS_APP_STORE;
+
+      profiles = profiles.filter(profile => profile.attributes.profileType === type);
```
dsokal, line 149:
```suggestion
      const profilesOfType (or whatever sounds better in english) = profiles.filter(profile => profile.attributes.profileType === type);
```
Please don't reassign variables. It makes code hard to read and doesn't bring any value other than reducing the number of variables (but it doesn't make sense...).

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+    if (!alreadyAddedUdids.includes(udid)) {
+      const device = await Device.createAsync({ name: 'iOS Device (added by Expo)', udid });
+      alreadyAdded.push(device);
+    }
+  }
+
+  return alreadyAdded;
+}
+
+async function findDistCert(serial_number: string, isEnterprise: boolean) {
```
dsokal, line 32:
```suggestion
async function findDistCertAsync(serialNumber: string, isEnterprise: boolean) {
```
We don't use snake_case in js.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+  const certs = await Certificate.getAsync({
+    query: {
+      filter: {
+        certificateType: CertificateType.IOS_DISTRIBUTION,
+      },
+    },
+  });
+
+  if (serial_number === '__last__') {
```
dsokal, line 41:
```suggestion
  if (serialNumber === '__last__') {
```

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+    return certs[certs.length - 1];
+  }
+
+  return certs.find(c => c.attributes.serialNumber === serial_number) ?? null;
+}
+
+async function findProfileByBundleId(bundleId: string, certSerialNumber: string) {
```
dsokal, line 48:
```suggestion
async function findProfileByBundleIdAsync(bundleId: string, certSerialNumber: string) {
```

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
dsokal, line 32: Please add the return type.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+  Teams,
+} from '@expo/apple-utils';
 import ora from 'ora';
 
 import { ProvisioningProfile } from './Credentials.types';
 import { AuthCtx } from './authenticate';
+import { USE_APPLE_UTILS, getBundleIdForIdentifier, getProfilesForBundleId } from './experimental';
 import { runActionAsync, travelingFastlane } from './fastlane';
 
+async function registerMissingDevicesAsync(udids: string[]) {
```
dsokal, line 17: Missing return type.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
dsokal, line 48: Missing return type.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+}
+
+function sortByExpiration(a: Profile, b: Profile): number {
+  return (
+    new Date(a.attributes.expirationDate).getTime() -
+    new Date(b.attributes.expirationDate).getTime()
+  );
+}
+
+async function findProfileById(profileId: string, bundleId: string) {
```
dsokal, line 99:
```suggestion
async function findProfileByIdAsync(profileId: string, bundleId: string) {
```

dsokal: Missing return type.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+  let profiles = await getProfilesForBundleId(bundleId);
+  profiles = profiles.filter(
+    profile => profile.attributes.profileType === ProfileType.IOS_APP_ADHOC
+  );
```
dsokal, line 103: Please don't reassign variables, it's an antipattern. See my other comment re this.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+  }
+
+  return certs.find(c => c.attributes.serialNumber === serial_number) ?? null;
+}
+
+async function findProfileByBundleId(bundleId: string, certSerialNumber: string) {
+  let expoProfiles = await getProfilesForBundleId(bundleId);
+  expoProfiles = expoProfiles.filter(
+    profile => profile.attributes.profileType === ProfileType.IOS_APP_INHOUSE
+  );
```
dsokal, line 52: Same here.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+
+  expoProfiles = expoProfiles.filter(profile => {
+    return (
+      profile.attributes.name.startsWith('*[expo]') &&
+      profile.attributes.profileState !== ProfileState.EXPIRED
+    );
+  });
```
dsokal, line 59: And here.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+    profile => profile.attributes.profileType === ProfileType.IOS_APP_ADHOC
+  );
+  return profiles.find(profile => profile.id === profileId) ?? null;
+}
+
+function uniqueItems<T = any>(items: T[]): T[] {
+  const set = new Set(items);
+  // @ts-ignore: downlevel iteration
+  return [...set];
+}
```
dsokal, line 111: You can use `lodash/uniq` instead - https://lodash.com/docs/4.17.15#uniq

fson: I think this is fine, `lodash.uniq` is more kilobytes.

dsokal: `lodash` is already installed so why not use it.

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
fson: `Boolean` is not the same as `getenv.boolish`:
```
> Boolean('false')
true
> Boolean('0')
true
```
With `getenv` both of these are parsed as `false` as one would expect.

dsokal: Well, if the only purpose of the env var is to use it for enabling some experimental feature then the behavior I proposed is perfectly enough.

dsokal: I mean:
- The default behavior is to run eas-cli without any env vars.
- If I wanted to test this feature I would run it like this: `USE_APPLE_UTILS=whatever eas`
- Then, if I wanted to test the original behavior I would run `eas` without any env vars set. I wouldn't even consider setting `USE_APPLE_UTILS=false`. I don't believe anyone would expect this to work.

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
wkozyra95: I agree

I'mnjot really sure what is the goal here, with partial implementation for those changes to work user would need to login twice to both apis. If that that is intedended behaviour for now i would suggest to:
- move this function to `authenticate.ts`
- move all calls to this function to AppStoreApi.ts (depending on what function is used trigger correct ensureAuth)
- Make fastlaneSession optional in AuthCtx instead of `Omit<AuthCtx, 'fastlaneSession'>`

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -1,24 +1,78 @@
...
   projectName: string;
   bundleIdentifier: string;
 }
 
-export interface EnsureAppExistsOptions {
-  enablePushNotifications?: boolean;
+export async function ensureAuthenticatedAsync(
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'>
+): Promise<Omit<AuthCtx, 'fastlaneSession'>> {
+  if (!Session.getSessionInfo()) {
```
wkozyra95, line 21: I have the same problem here as with `setSelectedTeamId`, I don't think that those values should be hidden inside the library

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,219 @@
...
+
+async function fastlaneActionAsync({
```
wkozyra95, line 113: why is this function have fastlane in name and it's not using fastlane?

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -40,7 +43,46 @@ export type AuthCtx = {
...
+    await Auth.loginAsync({ username: appleId, password: appleIdPassword });
+
+    log(chalk.green('Authenticated with Apple Developer Portal successfully!'));
+
+    // Get all of the teams
+    const teams = await Teams.getTeamsAsync();
+    const team = await chooseTeamAsync(teams, options.teamId);
+
+    // Set the selected team ID internally
+    Teams.setSelectedTeamId(team.id);
```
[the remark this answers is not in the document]
dsokal: I totally agree with Wojtek. When I was reviewing this I almost left a comment like this but I wasn't sure if anyone thinks the same.

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
fson: `lodash` is one of the largest libraries in this project:
```
4.8M	node_modules/lodash
```
I'd try to use as little of it as possible and eventually remove it (we can use individual packages like `lodash.pick` if necessary).

fson: So should we use the `lodash.uniq` package then? I'd say no, the `[...new Set(items)]` idiom works well for this purpose and this function @EvanBacon added is 9x smaller than `lodash.uniq`.

packages/eas-cli/package.json
```diff
@@ -54,16 +56,19 @@
...
     "@types/node-fetch": "^2.5.7",
     "@types/node-forge": "^0.9.5",
     "@types/progress": "^2.0.3",
     "@types/prompts": "^2.0.8",
     "@types/tar": "^4.0.3",
+    "@types/tough-cookie": "^4.0.0",
     "@types/uuid": "^8.3.0",
     "@types/wordwrap": "^1.0.0",
     "@vercel/ncc": "^0.23.0",
+    "axios": "^0.21.0",
```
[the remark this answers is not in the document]
EvanBacon: The types are exported from apple-utils but apple-utils has 0 deps so it needs to be added here for it to work 🙃 I'll investigate switching to node-fetch when I have more time.

packages/eas-cli/package.json
```diff
@@ -54,16 +56,19 @@
...
     "@types/fs-extra": "^9.0.1",
+    "@types/getenv": "^1.0.0",
     "@types/lodash": "^4.14.161",
     "@types/node": "^12",
     "@types/node-fetch": "^2.5.7",
     "@types/node-forge": "^0.9.5",
     "@types/progress": "^2.0.3",
     "@types/prompts": "^2.0.8",
     "@types/tar": "^4.0.3",
+    "@types/tough-cookie": "^4.0.0",
```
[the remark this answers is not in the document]
EvanBacon: Same reason as https://github.com/expo/eas-cli/pull/64#discussion_r523078751

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -1,9 +1,12 @@
+import { Auth, InvalidUserCredentialsError, Session, Teams } from '@expo/apple-utils';
 import chalk from 'chalk';
+import { toggleConfirmAsync } from 'utils/prompts';
```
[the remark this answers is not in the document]
EvanBacon: auto imported incorrectly, I've fixed it.

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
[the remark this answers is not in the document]
EvanBacon: We only have to manually set the team outside of the package to accommodate how Fastlane works, when we remove Fastlane this will go and team resolution will be part of the internal authentication unless specified otherwise. Similarly we resolve the password using keychain outside of apple-utils but when we remove Fastlane we'll also remove this extra step as the auth can often be resolved without retrieving the password from the keychain.

Further, the team id is passed to nearly every apple network request, Fastlane works similarly by storing the currently selected team id, but we pass the team id around because we instantiate a new Fastlane instance for seemingly every command.

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
EvanBacon: it's pretty standard for an authentication package to manage its own state. These values aren't hidden as they can be retrieved using `Session.getSessionInfo()`. Parts of the auth state contains values related to the network request like cookies, they shouldn't be set externally, and in the singular hacky case when they can be modified it's only to accommodate the ill-formatted yaml cookies that Fastlane uses.

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
EvanBacon: I think it's only temporary while we transition off of Fastlane safely, further `getenv` is a pretty good UX as @fson explained. 
```js
V='1', V='true' === true
V='0', V='false', V='null', V='undefined', V=undefined === false
```

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function getProfilesForBundleIdAsync(bundleIdentifier: string): Promise<Profile[]> {
+  const [bundleId] = await BundleId.getAsync({
+    query: {
+      filter: {
+        identifier: bundleIdentifier,
+      },
+    },
+  });
```
quinlanj, line 13: i've seen this query multiple times, it'd be nice if this was a helper so we don't have to destructure all the time (ie) `getByExplicitIdentifierAsync(explicitBundleIdentifier: string): Promise<BundleId>`

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (bundleId) {
+    return bundleId.getProfilesAsync();
+  }
+  return [];
+}
+
+export async function getBundleIdForIdentifierAsync(bundleIdentifier: string): Promise<BundleId> {
+  const bundleId = await BundleId.findAsync({ identifier: bundleIdentifier });
```
quinlanj, line 21: what's the difference between `BundleId.findAsync` and `BundleId.getAsync`?

packages/eas-cli/src/credentials/ios/appstore/provisioningProfile.ts
```diff
@@ -93,23 +176,47 @@ export async function createProvisioningProfileAsync(
...
-      teamId: ctx.team.id,
-      teamName: ctx.team.name,
-    };
+    if (USE_APPLE_UTILS) {
+      const profileType = ctx.team.inHouse
+        ? ProfileType.IOS_APP_INHOUSE
+        : ProfileType.IOS_APP_STORE;
+
+      const certificate = await Certificate.getAsync({
+        query: { filter: { serialNumber: [distCert.distCertSerialNumber] } },
```
quinlanj, line 185: replace with your `getBySerialNumberAsync` helper

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,208 @@
...
+async function registerMissingDevicesAsync(udids: string[]): Promise<Device[]> {
+  const allIosProfileDevices = await Device.getAllIOSProfileDevicesAsync();
+  const alreadyAdded = allIosProfileDevices.filter(device =>
+    udids.includes(device.attributes.udid)
+  );
+  const alreadyAddedUdids = alreadyAdded.map(i => i.attributes.udid);
+
+  for (const udid of udids) {
+    if (!alreadyAddedUdids.includes(udid)) {
+      const device = await Device.createAsync({ name: 'iOS Device (added by Expo)', udid });
```
quinlanj, line 33: batch parallelize if possible

packages/eas-cli/src/credentials/ios/appstore/provisioningProfileAdhoc.ts
```diff
@@ -1,9 +1,208 @@
...
-import { ProvisioningProfile } from './Credentials.types';
 import { AuthCtx } from './authenticate';
+import { getBundleIdForIdentifierAsync, getProfilesForBundleIdAsync } from './bundleId';
+import { ProvisioningProfile } from './Credentials.types';
+import { getDistributionCertificateAync } from './distributionCertificate';
+import { USE_APPLE_UTILS } from './experimental';
 import { runActionAsync, travelingFastlane } from './fastlane';
 
+interface ProfileResults {
+  provisioningProfileUpdateTimestamp?: number;
```
quinlanj, line 12: i originally had `provisioningProfileUpdateTimestamp` and `provisioningProfileCreateTimestamp` so i could call out to the ruby scripts and find out whether we 1.) created, 2.) updated or 3.) used an existing profile without updating. In hindsight it'd have been simpler if we just passed a `didUpdate` or `didCreate` property instead.

If you could change these to `didUpdate` and `didCreate` instead, that'd be amazing.

quinlanj (changes requested):
most of this lgtm, but let's get @dsokal  + @wkozyra95  to approve this PR before merging it. They are really familiar with this code and are actively using it for builds, so I would definitely feel more comfortable to get their seal of approval as well :)

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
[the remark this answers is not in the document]
dsokal: I don't see why we need to accommodate how Fastlane works. Using `traveling-fastlane`, we don't depend on this behavior and instead, we're passing the team id to every "action".

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
dsokal: 🤷‍♂️

packages/eas-cli/src/credentials/ios/appstore/experimental.ts
EvanBacon: getAsync accepts a complex query and returns an array of bundle ids, it also has potential to page results. findAsync is a convenient wrapper for retrieving just the item matching a bundle id.

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -40,7 +42,59 @@ export type AuthCtx = {
   fastlaneSession: string;
 };
 
+export async function ensureAuthenticatedAsync(
```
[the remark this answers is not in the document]
EvanBacon: This method comes from ensureAppAsync, PR review had me move it to this file. The method emulates this functionality https://github.com/expo/expo-cli/blob/b83c73f701fad4e168663dca8dcbea16b5d8d555/packages/traveling-fastlane/actions/ensure_app_exists.rb#L40-L41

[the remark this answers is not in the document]
EvanBacon: If the user is signed in but for some reason the incoming "auth state" has a new team ID, it would only be set outside of the sign in method.

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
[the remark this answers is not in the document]
EvanBacon: Currently with Fastlane we do the following: 
- auth context -> traveling fastlane script -> fastlane spaceship (provides to all network requests).
In apple-utils we similarly rely on the package to provide the team id to all network requests, otherwise (almost) every network request would need to accept the Team ID like `BundleId.getAsync(query)` would become `BundleId.getAsync(teamId, query)`. I say almost because the Apple API has lots of inconsistencies which our package abstracts over.

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -1,24 +1,78 @@
...
-  enablePushNotifications?: boolean;
+async function ensureBundleIdExistsAsync(
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'>,
+  { accountName, projectName, bundleIdentifier }: AppLookupParams,
+  options: EnsureAppExistsOptions = {}
+) {
+  let spinner = ora(`Registering Bundle ID "${bundleIdentifier}"`).start();
+
+  try {
+    appleCtx = await ensureAuthenticatedAsync(appleCtx);
```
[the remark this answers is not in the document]
EvanBacon: The Apple auth state is pretty complex and this `credentials/ios/appstore` logic would need to be copied over to everything that interacts with Apple through our package. Our apple package also uses its awareness of the auth state to validate when forbidden network requests are made.

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -40,7 +42,59 @@ export type AuthCtx = {
...
+  appleCtx: Omit<AuthCtx, 'fastlaneSession'> & { fastlaneSession?: string }
+): Promise<Omit<AuthCtx, 'fastlaneSession'> & { fastlaneSession?: string }> {
+  if (!Session.getSessionInfo()) {
```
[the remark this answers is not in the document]
EvanBacon: in `alpha.4` the cookies and server session are now returned and added to the AuthCtx.

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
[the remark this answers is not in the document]
wkozyra95: I'm not sure we are talking about the same thing, I'm suggesting that you move this call here
https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/credentials/ios/appstore/AppStoreApi.ts#L56

```
  public async ensureAppExistsAsync(
    app: AppLookupParams,
    options?: EnsureAppExistsOptions
  ): Promise<void> {
    let ctx 
    if (USE_APPLE_UTILS) {
      ctx = await ensureAuthenticatedAsync(this.options);
    } else {
      ctx = awiat this.ensureAuthenticatedAsync();
   }
    
    return await ensureAppExistsAsync(ctx, app, options);
  }
```

Or even better to move logic from `ensureAuthenticatedAsync` (your function) into `AppStoreApi.ensureAuthenticatedAsync` method


A lot of issues here comes from the fact that we temporarily support both implementations, so if you disagree about that lets just merge it in current form and we can refactor it after fastlane is removed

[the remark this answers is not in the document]
EvanBacon: got it, I see the method `ensureAppExistsAsync` is not being used directly like it is in expo-cli. I removed the extra auth check since I now see that `ensureAuthenticatedAsync` is run in a wrapper method. It would seem that this wrapper causes Fastlane to have an extra auth check, but I suppose that won't be an issue for much longer.

quinlanj (approved):

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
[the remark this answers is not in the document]
dsokal: We could avoid passing in the team id to all requests if we had a way to create a "client" object tied to a particular team. This is a really common pattern, e.g. see `axios.create` (https://github.com/axios/axios#axioscreateconfig)  where you can specify the base URL for all requests.

This way we wouldn't depend on a global state which I consider to be a pretty serious antipattern.

dsokal (approved):
Please consider my comment regarding setting team id. However, I don't expect this change in this PR but I'm sure we should think about this in the future.

fson (approved):

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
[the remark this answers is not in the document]
EvanBacon: We already define a `baseUrl` for each Apple API in apple-utils, originally we set the team id in the Axios object but sometimes the API requires the team id in the [parameters](https://github.com/expo/third-party/blob/627a902c24769c74eb1c3a24dc9b18b206e0c382/packages/app-store/src/portal/PortalAPI.ts#L105) and other times it requires it in the [request body](https://github.com/expo/third-party/blob/627a902c24769c74eb1c3a24dc9b18b206e0c382/packages/app-store/src/portal/PortalAPI.ts#L78) 🙃 so we had to change to a global state to accommodate the inconsistencies.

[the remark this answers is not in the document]
dsokal: The reference to `axios.create` and setting a base URL was just an example.

I think that the way of using the team id is just an implementation detail of a particular request. It doesn't automatically mean that we have to have a global state for apple id. Please correct me if I'm missing sth.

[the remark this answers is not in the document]
dsokal: @EvanBacon 
IMO it should work sth like this:
```
import AppleUtils from '@expo/apple-utils';

const clientA = AppleUtils.createClient({ ..., teamID: 'ABC123YZ', ... });
await client.Certificate.getAsync(...); // fetches cert for team ABC123YZ

const clientB = AppleUtils.createClient({ ..., teamID: 'XYZ123AB', ... });
await clientB.Certificate.getAsync(...); // fetches cert for team XYZ123AB
```
This way we don't need the global state.

PR #71, opened 2020-11-18
[eas-cli] add command release:create

jkhales:
This PR closes: https://github.com/expo/eas-cli/projects/1#card-48195186. Namely: add `eas release:create`.

Tested by creating a release both with a positional argument and interactive prompt. Confirmed in graphiQL:

Screen Shot 2020-11-17 at 5 32 53 PM

Note:
I read https://github.com/expo/eas-cli/blob/ed798057a39de7f238958048bc96fbc8c8271863/CONTRIBUTING.md but could not find the CHANGELOG.md.


packages/eas-cli/README.md
```diff
@@ -58,8 +59,6 @@ ALIASES
   $ eas login
 ```
 
-_See code: [build/commands/account/login.ts](https://github.com/expo/eas-cli/blob/v0.1.0-alpha.3/build/commands/account/login.ts)_
-
```
jkhales, line 62: These deletions related to build commands were generated when I ran `yarn release`.

They are unrelated to my PR and I assume are leftover from a PR that did not follow the `yarn release` procedure.

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+        type: 'text',
+        name: 'releaseName',
+        message: 'Please name the release:',
+      });
+      releaseName = userInput.releaseName;
+    }
+    if (!releaseName) {
+      log.error('You must specify a releaseName.');
+      return;
+    }
```
fson, line 83: If I leave the release name empty, it annoyingly exits the whole command:
Screen Shot 2020-11-18 at 15 27 19


Maybe try this instead? (It will validate that release name is not empty.)
```suggestion
    if (!releaseName) {
      ({ releaseName } = await promptAsync({
        type: 'text',
        name: 'releaseName',
        message: 'Please name the release:',
        validate: value => value ? true : 'Release name may not be empty.',
      }));
    }
```
Perhaps there are other validations we should do too? Max length? Allowed characters?

Also maybe it should show an example of a "good" release name to aid learning?

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+    try {
+      const newRelease = await createUpdateReleaseOnAppAsync({ appId: projectId, releaseName });
+
+      log.withTick(
+        `️Created a new release: ${chalk.bold(
+          newRelease.releaseName
+        )} on project with id ${chalk.bold(projectId)}.`
+      );
+    } catch (e) {
+      log.error(e);
```
fson, line 94: If there's an error, we should not exit with status 0. So it would be better to *not* catch it here, this way it propagates to the top level error handler and the process exits with an error status.

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+  ];
+
+  async run() {
+    let {
+      args: { releaseName },
+    } = this.parse(ReleaseCreate);
+
+    const projectDir = await findProjectRootAsync(process.cwd());
+    if (!projectDir) {
+      log.error("Please run this command inside your project's directory.");
```
fson, line 67:
```suggestion
      log.error("Please run this command inside a project directory.");
```

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
+import { Command, flags } from '@oclif/command';
+import chalk from 'chalk';
+import gql from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
+import log from '../../log';
+import { findProjectRootAsync } from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+import { getProjectIdAsync } from '../../submissions/commons';
```
fson, line 9: Can be done in a separate PR, but let's move this to `project/projectUtils` because it's not only for submissions and rename it to `getAppIdAsync`?

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+    graphqlClient
+      .mutation<
+        { updateRelease: { createUpdateReleaseForApp: UpdateRelease } },
+        { appId: string; releaseName: string }
+      >(
+        gql`
+          mutation CreateUpdateReleaseForApp($appId: ID!, $releaseName: String!) {
+            updateRelease {
+              createUpdateReleaseForApp(appId: $appId, releaseName: $releaseName) {
+                id
```
fson, line 33:
```suggestion
```
(Unused field.)

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+      return;
+    }
+
+    try {
+      const newRelease = await createUpdateReleaseOnAppAsync({ appId: projectId, releaseName });
+
+      log.withTick(
+        `️Created a new release: ${chalk.bold(
+          newRelease.releaseName
+        )} on project with id ${chalk.bold(projectId)}.`
```
fson, line 91: Maybe "Created a new release *name* on app @owner/app-name"? I think full name is still a more used friendly way to refer to an app.

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,97 @@
...
+      )
+      .toPromise()
+  );
+  return data.updateRelease.createUpdateReleaseForApp;
+}
+
+export default class ReleaseCreate extends Command {
+  static description = 'Create a release on the current project.';
+
+  static args = [
```
fson, line 52: Could you also implement the `--json` flag to get only JSON output?

packages/eas-cli/src/commands/release/create.ts
jkhales: >Perhaps there are other validations we should do too? Max length? Allowed characters?

I was thinking to do this on the server to avoid having multiple different verifications.

> Also maybe it should show an example of a "good" release name to aid learning?

That's a good idea... but what is a good release name? 😅 
One idea we were playing around with was suggesting the git-branch name. But this requires an extra module and also doesn't always work.

packages/eas-cli/src/commands/release/create.ts
jkhales: Is it worth printing errors with chalk.red?

packages/eas-cli/src/commands/release/create.ts
jkhales: What would the output look like in this case?

```
{
  accountName: string;
  projectName: string;
  releaseName: string;
}
```
?

fson: I think prefixing property names with the type is a code smell - if we want all these we should nest the objects.

But I think we should maybe just return the release object or some subset of its fields. Especially since this is a CRUD method for release. The only thing you actually need if you're using this for scripting is the `id` (and maybe `name` if we generate the default name somehow). So it could be something like:
```ts
{
  id: string,
  name: string,
}
```

packages/eas-cli/src/commands/release/create.ts
fson: Maybe? In any case if we want errors to be red we can add a custom handler here https://github.com/expo/eas-cli/blob/03e754e0f66bd5fa79ca7f3efa597cb5df032d63/packages/eas-cli/bin/run#L6 so we do this for all errors consistently.

packages/eas-cli/src/commands/release/create.ts
jkhales: left in as used in with the --json

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -0,0 +1,114 @@
+import { getConfig } from '@expo/config';
+import { Command, flags } from '@oclif/command';
+import chalk from 'chalk';
+import branchName from 'current-git-branch';
```
fson, line 4: The `current-git-branch` package has way too many dependencies, we should do something lighter.
```
info Direct dependencies
└─ current-git-branch@1.1.0
info All dependencies
├─ babel-plugin-add-module-exports@0.2.1
├─ cross-spawn@5.1.0
├─ current-git-branch@1.1.0
├─ get-stream@3.0.0
├─ is-git-repository@1.1.1
├─ is-stream@1.1.0
├─ isexe@2.0.0
├─ lru-cache@4.1.5
├─ npm-run-path@2.0.2
├─ p-finally@1.0.0
├─ path-is-absolute@1.0.1
├─ path-key@2.0.1
├─ pseudomap@1.0.2
├─ shebang-command@1.2.0
├─ shebang-regex@1.0.0
├─ signal-exit@3.0.3
├─ strip-eof@1.0.0
├─ which@1.3.1
└─ yallist@2.1.2
```
Maybe something like `spawnSync('git', ['rev-parse', '--abbrev-ref', 'HEAD'])` (and catch the error if it fails).

jkhales: great point.

fson: May need to use `@expo/spawn-async` (or `cross-spawn`) for Windows support, not sure

jkhales: I used `@expo/spawn-async`

jkhales: (added to the utils/git file).

fson (approved):
:shipit:

Issue #72, opened 2020-11-18
Can be done in a separate PR, but let's move this to `project/projectUtils` because it's not only for submissions and rename it to `getAppIdAsync`?


jkhales:
Can be done in a separate PR, but let's move this to `project/projectUtils` because it's not only for submissions and rename it to `getAppIdAsync`?

_Originally posted by @fson in https://github.com/expo/eas-cli/pull/71#r526096508_

PR #73, opened 2020-11-18
[eas-cli] relocate getProjectIdAsync

jkhales:
close https://github.com/expo/eas-cli/issues/72

Left name as `getProjectIdAsync` as I am under the impression that we want to move away from `app`.


fson (approved):
:shipit:

PR #77, opened 2020-11-22
[eas-cli] add release:edit

jkhales:
While the graphql allows us to edit a release's name by either its `ID`, or the combination of `projectID` + `releaseName`, this command only implements the latter: `projectId` + `releaseName`. Adding both options seemed to add more clutter than utility.

Even though there is only one `edit` option: `rename` chose to require a special flag `--rename`.  This allows extensibility in the future and will also match other edit commands with more options, like the `channel:edit` command (which will have 2 possibilities).

### DIFF
--- a/packages/eas-cli/src/commands/release/rename.ts
+++ b/packages/eas-cli/src/commands/release/rename.ts
@@ -50,7 +50,7 @@ async function renameUpdateReleaseOnAppAsync({
   return data.updateRelease.editUpdateRelease;
 }
 
-export default class ReleaseEdit extends Command {
+export default class ReleaseRename extends Command {
   static description = 'Edit a release.';
 
   static flags = {
--- a/packages/eas-cli/src/commands/release/rename.ts
+++ b/packages/eas-cli/src/commands/release/rename.ts
@@ -8,11 +8,7 @@ import log from '../../log';
 import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
 import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
-
-type UpdateRelease = {
-  id: string;
-  releaseName: string;
-};
+import { EditUpdateReleaseInput, UpdateRelease } from '../../graphql/generated';
 
 async function renameUpdateReleaseOnAppAsync({
   appId,
@@ -28,11 +24,7 @@ async function renameUpdateReleaseOnAppAsync({
       .mutation<
         { updateRelease: { editUpdateRelease: UpdateRelease } },
         {
-          input: {
-            appId: string;
-            releaseName: string;
-            newReleaseName: string;
-          };
+          input: EditUpdateReleaseInput;
         }
       >(
         gql`
--- a/packages/eas-cli/src/commands/release/rename.ts
+++ b/packages/eas-cli/src/commands/release/rename.ts
@@ -14,14 +14,14 @@ type UpdateRelease = {
   releaseName: string;
 };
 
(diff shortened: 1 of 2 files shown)


fson:
Thanks, the command looks good, although I'm not a fan of `--rename` as a flag name. You mention extensibility and `channel:edit` having more options – however it will be difficult to match the same naming scheme for other field names.

What do you think of renaming the command to `release:edit <release-name> --name <new-name>`? Then the option name can typically match the name of the field to edit.

fson:
Or alternatively it could be `release:rename --from <release-name> --to <new-name>`, where the verb "rename" is in the command name itself.

jkhales:
> Or alternatively it could be `release:rename --from <release-name> --to <new-name>`, where the verb "rename" is in the command name itself.

This makes sense! Changed it.

packages/eas-cli/src/commands/release/rename.ts
```diff
@@ -0,0 +1,140 @@
...
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation<
+        { updateRelease: { editUpdateRelease: UpdateRelease } },
+        {
+          input: {
+            appId: string;
+            releaseName: string;
+            newReleaseName: string;
+          };
```
fson, line 35:
```suggestion
          input: EditUpdateReleaseInput;
```
(Need to run `yarn generate-graphql-code` so this gets added to `generated.ts`.)

packages/eas-cli/src/commands/release/rename.ts
```diff
@@ -0,0 +1,140 @@
...
+import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
+import log from '../../log';
+import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
+import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+
+type UpdateRelease = {
+  id: string;
+  releaseName: string;
+};
```
fson, line 15:
```suggestion
import { EditUpdateReleaseInput, UpdateRelease } from '../../graphql/generated';
```

fson (approved):

packages/eas-cli/src/commands/release/rename.ts
```diff
@@ -0,0 +1,140 @@
...
+            newReleaseName: newName,
+          },
+        }
+      )
+      .toPromise()
+  );
+  return data.updateRelease.editUpdateRelease;
+}
+
+export default class ReleaseEdit extends Command {
```
fson, line 53:
```suggestion
export default class ReleaseRename extends Command {
```

jkhales: oh good catch

packages/eas-cli/src/commands/release/rename.ts
```diff
@@ -0,0 +1,132 @@
...
+  static description = 'Edit a release.';
```
fson, line 54:
```suggestion
  static description = 'Rename a release.';
```

packages/eas-cli/src/commands/release/rename.ts
```diff
@@ -0,0 +1,132 @@
...
+    json: flags.boolean({
+      description: `return a json with the edited release's ID and name.`,
+      default: false,
+    }),
+  };
+
+  async run() {
+    let {
+      flags: { json: jsonFlag, from: currentName, to: newName },
+    } = this.parse(ReleaseEdit);
```
fson, line 74:
```suggestion
    } = this.parse(ReleaseRename);
```

PR #78, opened 2020-11-24
Add tsc to CI pipeline

fson:
Turns out `tsc` was not running in the CI for all PRs at the moment.

Run `tsc` after the tests to ensure that there are no type errors.


dsokal (approved):
Just one comment: don't we want to run `tsc` before tests? Some tests can fail in case of type/syntax errors. 
It probably doesn't matter when a tests run is really short (and it is right now) but in my opinion, type-checking should be the very first step in any pipeline.

PR #80, opened 2020-11-24
Fix a typo in provisioningProfileAdhoc.ts

fson:
`travelingFastlane` doesn't have the properties `provisioningProfile` and `provisioningProfileId`.

Looks like this code was added in #59.


dsokal (approved):
Ah, sorry and thanks!

PR #82, opened 2020-11-25
[eas-cli] Support for App Store API in iOS submissions

barthap:
A step towards removing `traveling-fastlane`

- The whole process now uses the new API when `USE_APPLE_UTILS=1` env var is set
- Updated `--language` parameter to support both fastlane and new API, it is now smarter and supports both `English` and `en-US` formats interchangeably. 
- ~~Fixed `authenticateAsync()` not to prompt for team again when it's already selected.~~ Fix moved to https://github.com/expo/eas-cli/pull/85

TODO in the future:
- Add possibility to update app name on ASC (currently API doesn't allow us to do it)
- [Optional] Update bundle ID capabilities (the same way as `ensureAppExists()` does it for push notifications) based on app config

### DIFF
--- a/packages/eas-cli/src/submissions/ios/AppProduce.ts
+++ b/packages/eas-cli/src/submissions/ios/AppProduce.ts
@@ -1,8 +1,10 @@
+import { App, RequestContext, Session, User } from '@expo/apple-utils/build';
 import { getConfig } from '@expo/config';
 import wordwrap from 'wordwrap';
 
 import { getBundleIdentifier } from '../../build/ios/bundleIdentifer';
-import { authenticateAsync } from '../../credentials/ios/appstore/authenticate';
+import { authenticateAsync, getRequestContext } from '../../credentials/ios/appstore/authenticate';
+import { USE_APPLE_UTILS } from '../../credentials/ios/appstore/experimental';
 import log from '../../log';
 import { promptAsync } from '../../prompts';
 import { IosSubmissionContext } from '../types';
@@ -11,8 +13,8 @@ import { sanitizeLanguage } from './utils/language';
 
 interface ProduceOptions {
   appleId?: string;
-  appName?: string;
-  bundleIdentifier?: string;
+  appName: string;
+  bundleIdentifier: string;
   appleTeamId?: string;
   itcTeamId?: string;
   language?: string;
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/submissions/ios/AppProduce.ts
```diff
@@ -62,9 +64,86 @@ Learn more here: https://expo.fyi/bundle-identifier`
...
+    return await runProduceExperimentalAsync(options);
+  }
+
   return await runProduceAsync(options);
 }
 
+async function isProvisioningAvailableAsync(requestCtx: RequestContext): Promise<boolean> {
+  const session = await Session.getAnySessionInfo();
+  const username = session?.user.emailAddress;
+  const [user] = await User.getAsync(requestCtx, { query: { filter: { username } } });
```
EvanBacon, line 77: I worry about this since the User object has username and email, but you cannot query against `email`, and I haven't seen a case where username and email are different. Maybe make a TODO about investigating this further in the future.

packages/eas-cli/src/submissions/ios/AppProduce.ts
```diff
@@ -62,9 +64,86 @@ Learn more here: https://expo.fyi/bundle-identifier`
...
+        bundleId,
+        name: appName,
+        // TODO: Convert values like "English" to "en-US"
+        primaryLocale: language,
+        companyName,
+      });
+    } catch (error) {
+      if (error.message.match(/An App ID with Identifier '(.*)' is not available/)) {
+        log.warn(
+          `\nThe bundle identifier "${bundleId}" is not available, please change it in your app config and try again.\n`
```
EvanBacon, line 127: Should specify that the bundle id is not available to `"provider.name" (provider.providerId)` in case maybe the user chose the wrong provider.

packages/eas-cli/src/submissions/ios/AppProduce.ts
```diff
@@ -62,9 +64,86 @@ Learn more here: https://expo.fyi/bundle-identifier`
...
+        );
+        process.exit(0);
```
EvanBacon, line 129: eventually something like this is better to make testing / automation easier https://github.com/expo/expo-cli/pull/2901

packages/eas-cli/src/submissions/ios/utils/language.ts
```diff
@@ -1,49 +1,267 @@
...
+      lang.displayName === query ||
+      lang.name === query ||
+      lang.locale === query ||
+      lang.itcLocale === query
+  );
+
+  return foundLang ?? null;
+}
+
+const LANGUAGES: Language[] = [
```
fson, line 78: Maybe add a comment that tells where this list is from and how to update it?

EvanBacon (approved):

PR #85, opened 2020-11-25
prompt team once

EvanBacon:
prompt for the team using the internal apple-utils mechanism. This fixes the bug where a user is prompted for their team twice.


packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -79,16 +79,20 @@ async function authenticateWithExperimentalAsync(options: Options = {}): Promise
...
 
-    // Get all of the teams
+    // Currently, this is resolved once, inside the apple-utils package.
+    const teamId = authState.context.teamId!;
+    // Get all of the teams to resolve the rest of the user data.
+    // TODO: optimize this step.
     const teams = await Teams.getTeamsAsync();
-    const team = await chooseTeamAsync(teams, options.teamId);
+    const team = teams.find(team => team.teamId === teamId);
+    assert(team, `Your account is not associated with Apple Team with ID: ${teamId}`);
```
barthap, line 88: This message is not clear to me. Is it ever possible for this to happen?

barthap (approved):
👍

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
EvanBacon: no it isn't, I lifted it from the teams error later in the file.

PR #88, opened 2020-11-26
Improve error/warning messages

EvanBacon:
- format `Provisioning Profile (id: ${profile.provisioningProfileId}) does not exist on Apple servers` correctly
Screen Shot 2020-11-26 at 3 52 55 PM

- use a more specific config name when resolving the bundle id + fix spacing

## After

Screen Shot 2020-11-26 at 3 27 20 PM


barthap:
The app config part is interesting. Afaik we should've avoided using `app.json` etc in favour of more general term (see this comment https://github.com/expo/eas-cli/pull/65#issuecomment-731467963), but here the real filename is gathered. cc @brentvatne 

Also, there a few more places in CLI where `app config` phrase appears and they could also be revisited.

brentvatne:
@barthap - agree it's better if we actually name their exact config file, this is the way.

PR #93, opened 2020-11-27
[eas-cli] configure app identifier for iOS and Android

wkozyra95:
# Why

- getBundleIdentifier was inconsistent with the rest of configuration and validation steps when building
- We need to configure applicationId for android to support new credentials API

# How

- when reading appIdentifier(bundle identifier or application id) read from app.json and fallback to native project if possible (The same approach as (`exp.owner ?? user.username`)
- when running `build` and app identifiers in a native project do not match thro exception
- when running `configure` ask which value is correct and update either native project or app.json
  - if static config exists it will be updated, if only dynamic config is available throw an exception and ask user to update it
- for android only applicationId is updated by default
  - package name will be updated with flag `--alow-experimental` (I'm not sure about that, I'm still working on getting `AndroidConfig.Package.renamePackageOnDisk` to work)

# Teste Plan 

Run build and configure for a few different cases with matching and mismatched values for different configs.

### DIFF
--- a/packages/eas-cli/src/build/android/applicationId.ts
+++ b/packages/eas-cli/src/build/android/applicationId.ts
@@ -1,0 +1,158 @@
+import { ExpoConfig, getConfigFilePaths } from '@expo/config';
+import { AndroidConfig } from '@expo/config-plugins';
+import assert from 'assert';
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import path from 'path';
+
+import log from '../../log';
+import {
+  getAndroidApplicationIdAsync,
+  getProjectConfigDescription,
+} from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+import { gitAddAsync } from '../../utils/git';
+
+enum ApplicationIdSource {
+  AndroidProject,
+  AppJson,
+}
+
+export async function configureApplicationIdAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  allowExperimental: boolean
+): Promise<void> {
+  const configDescription = getProjectConfigDescription(projectDir);
+  const applicationIdFromConfig = AndroidConfig.Package.getPackage(exp);
+  const applicationIdFromAndroidProject = await getAndroidApplicationIdAsync(projectDir);
+
+  if (applicationIdFromAndroidProject && applicationIdFromConfig) {
+    if (applicationIdFromConfig !== applicationIdFromAndroidProject) {
+      log.newLine();
+      log.warn(
+        `We detected that your Android project is configured with a different application id than the one defined in ${configDescription}.`
+      );
+      if (!(await hasApplicationIdInStaticConfigAsync(projectDir, exp))) {
+        log(`If you choose the one defined in ${configDescription} we'll automatically configure your Android project with it.
+However, if you choose the one defined in the Android project you'll have to update ${configDescription} on your own.
+Otherwise, you'll see this prompt again in the future.`);
+      }
+      log.newLine();
+      const { applicationIdSource } = await promptAsync({
+        type: 'select',
+        name: 'applicationIdSource',
+        message: 'Which application id should we use?',
+        choices: [
+          {
+            title: `${chalk.bold(
+              applicationIdFromAndroidProject
+            )} - In build.gradle in Android project`,
+            value: ApplicationIdSource.AndroidProject,
+          },
+          {
+            title: `${chalk.bold(applicationIdFromConfig)} - In your ${configDescription}`,
+            value: ApplicationIdSource.AppJson,
+          },
+        ],
+      });
+      switch (applicationIdSource) {
+        case ApplicationIdSource.AppJson: {
+          await updateApplicationIdInBuildGradleAsync(projectDir, exp);
+          break;
+        }
+        case ApplicationIdSource.AndroidProject: {
+          await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+          break;
+        }
+      }
+    }
+  } else if (!applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    throw new Error(
+      `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
+    );
+  } else if (applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    if (getConfigFilePaths(projectDir).staticConfigPath) {
+      await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+    } else {
+      throw new Error(
+        `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
+      );
+    }
+  } else if (!applicationIdFromAndroidProject && applicationIdFromConfig) {
+    // This should never happen, adding warning just in case
+    log.warn(
+      'applicationId is not specified in your ./android/app/build.gradle file. Make sure your project is confgured correctly before building.'
+    );
+    return;
+  }
+  if (allowExperimental) {
+    // package name does not need to be updated
+    await updatePackageNameAsync(projectDir, exp);
+    await gitAddAsync(path.join(projectDir, 'android'), { intentToAdd: true });
+  }
+}
+
+async function updateAppJsonConfigAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  newApplicationId: string
+): Promise<void> {
+  const paths = getConfigFilePaths(projectDir);
+  assert(paths.staticConfigPath, "can't update dynamic configs");
+
+  const rawStaticConfig = await fs.readJSON(paths.staticConfigPath);
+  rawStaticConfig.expo = {
+    ...rawStaticConfig.expo,
+    android: { ...rawStaticConfig.expo.android, package: newApplicationId },
+  };
+  await fs.writeJson(paths.staticConfigPath, rawStaticConfig, { spaces: 2 });
+
+  exp.android = { ...exp.android, package: newApplicationId };
+}
+
+/**
+ * Check if static config exists and if expo.android.package is defined there.
+ * It will return false if value in static confgi is different than exp.android.package
+ */
+async function hasApplicationIdInStaticConfigAsync(
+  projectDir: string,
+  exp: ExpoConfig
+): Promise<boolean> {
+  if (!exp.android?.package) {
+    return false;
+  }
+  const paths = getConfigFilePaths(projectDir);
+  if (!paths.staticConfigPath) {
+    return false;
+  }
+  const rawStaticConfig = await fs.readJson(paths.staticConfigPath);
+  return rawStaticConfig?.expo?.android?.package === exp.android.package;
+}
+
+async function updateApplicationIdInBuildGradleAsync(projectDir: string, exp: ExpoConfig) {
+  const buildGradlePath = AndroidConfig.Paths.getAppBuildGradle(projectDir);
+  let buildGradleContent = await fs.readFile(buildGradlePath, 'utf8');
+  buildGradleContent = AndroidConfig.Package.setPackageInBuildGradle(exp, buildGradleContent);
+  await fs.writeFile(buildGradlePath, buildGradleContent);
+}
+
+async function updatePackageNameAsync(projectDir: string, exp: ExpoConfig): Promise<void> {
+  const androidManifestPath = await AndroidConfig.Paths.getAndroidManifestAsync(projectDir);
+  if (!androidManifestPath) {
+    throw new Error(`Could not find AndroidManifest.xml in project directory: "${projectDir}"`);
+  }
+  const androidManifest = await AndroidConfig.Manifest.readAndroidManifestAsync(
+    androidManifestPath
+  );
+  const updatedAndroidManifest = AndroidConfig.Package.setPackageInAndroidManifest(
+    exp,
+    androidManifest
+  );
+  await AndroidConfig.Manifest.writeAndroidManifestAsync(
+    androidManifestPath,
+    updatedAndroidManifest
+  );
+
+  await AndroidConfig.Package.renamePackageOnDisk(exp, projectDir);
+}
--- a/packages/eas-cli/src/build/android/applicationId.ts
+++ b/packages/eas-cli/src/build/android/applicationId.ts
@@ -33,10 +33,13 @@ export async function configureApplicationIdAsync(
       log.warn(
         `We detected that your Android project is configured with a different application id than the one defined in ${configDescription}.`
       );
-      if (!(await hasApplicationIdInStaticConfigAsync(projectDir, exp))) {
+      const hasApplicationIdInStaticConfig = await hasApplicationIdInStaticConfigAsync(
+        projectDir,
+        exp
+      );
+      if (!hasApplicationIdInStaticConfig) {
         log(`If you choose the one defined in ${configDescription} we'll automatically configure your Android project with it.
-However, if you choose the one defined in the Android project you'll have to update ${configDescription} on your own.
-Otherwise, you'll see this prompt again in the future.`);
+However, if you choose the one defined in the Android project you'll have to update ${configDescription} on your own.`);
       }
       log.newLine();
       const { applicationIdSource } = await promptAsync({
@@ -62,22 +65,22 @@ Otherwise, you'll see this prompt again in the future.`);
           break;
         }
         case ApplicationIdSource.AndroidProject: {
-          await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+          if (hasApplicationIdInStaticConfig) {
+            await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+          } else {
+            throw new Error(missingPackageMessage(configDescription));
+          }
           break;
         }
       }
     }
   } else if (!applicationIdFromAndroidProject && !applicationIdFromConfig) {
-    throw new Error(
-      `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
-    );
+    throw new Error(missingPackageMessage(configDescription));
   } else if (applicationIdFromAndroidProject && !applicationIdFromConfig) {
     if (getConfigFilePaths(projectDir).staticConfigPath) {
       await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
     } else {
-      throw new Error(
-        `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
-      );
+      throw new Error(missingPackageMessage(configDescription));
     }
   } else if (!applicationIdFromAndroidProject && applicationIdFromConfig) {
     // This should never happen, adding warning just in case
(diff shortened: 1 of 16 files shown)


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -22,3 +25,72 @@ export async function getProjectIdAsync(projectDir: string): Promise<string> {
...
+}
+
+export async function ensureAppIdentifierAsync(
+  projectDir: string,
+  platform: Platform
+): Promise<string> {
+  const appIdentifier = await getAppIdentifierAsync(projectDir, platform);
+  if (!appIdentifier) {
+    const desc = getProjectConfigDescription(projectDir);
+    throw new Error(`Please define "ios.bundleIdentifier" in your ${desc}.`);
```
dsokal, line 72: Fix message.

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,153 @@
...
+    // This should never happen, adding warning just in case
+    log.warn(
+      'applicationId is not specified in your ./android/app/build.gradle file. Make sure your project is confgured correctly before building.'
+    );
+    return;
+  }
+  if (allowExperimental) {
+    // package name does not need to be updated
+    await updatePackageNameAsync(projectDir, exp);
+  }
```
dsokal, line 89: This seems unreachable when `!applicationIdFromAndroidProject && applicationIdFromConfig`. Is this intentional?

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,153 @@
...
+  } else if (applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    if (getConfigFilePaths(projectDir).staticConfigPath) {
+      await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+    } else {
+      throw new Error(
+        `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
+      );
+    }
+  } else if (!applicationIdFromAndroidProject && applicationIdFromConfig) {
+    // This should never happen, adding warning just in case
```
dsokal, line 80: Why?

packages/eas-cli/src/commands/build/configure.ts
```diff
@@ -15,15 +16,26 @@ export default class BuildConfigure extends Command {
...
+    }),
   };
 
   async run() {
     const { flags } = this.parse(BuildConfigure);
     const platform = flags.platform as BuildCommandPlatform;
+    const allowExperimental = flags['allow-experimental'];
+    if (allowExperimental) {
+      log.warn(
+        'Project confguration will execute some additional steps, that might fail if strucutre of your native project is significantly different from "expo eject" or "expo init"'
```
dsokal, line 31:
```suggestion
        'Project configuration will execute some additional steps, that might fail if the structure of your native project is significantly different from "expo eject" or "expo init"'
```

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+async function updateAppJsonConfigAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  newApplicationId: string
+): Promise<void> {
+  const paths = getConfigFilePaths(projectDir);
+  assert(paths.staticConfigPath, "can't update dynamic configs");
```
brentvatne, line 102: we should provide guidance on what the user needs to update specifically, for example https://github.com/expo/expo-cli/blob/9777c36cbe73af2f4e4f47f408f083ac96c6be8b/packages/expo-cli/src/commands/upgrade.ts#L220-L228

eg: `Dynamic configuration files can't be updated automatically. Please set android.package to "${newApplicationId}`" and then (suggest next action here, eg: run build)`

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+        }
+        case ApplicationIdSource.AndroidProject: {
+          await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+          break;
+        }
+      }
+    }
+  } else if (!applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    throw new Error(
+      `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
```
brentvatne, line 72: in the future we'll [support the `applicationId`](https://github.com/expo/expo-cli/pull/2962) field, so we may want to update this at some point to point there instead

packages/eas-cli/src/build/ios/bundleIdentifer.ts
```diff
@@ -1,76 +1,121 @@
...
+    ios: { ...rawStaticConfig.expo.ios, bundleIdentifier: newBundleIdentifier },
+  };
+  await fs.writeJson(paths.staticConfigPath, rawStaticConfig, { spaces: 2 });
+
+  exp.ios = { ...exp.ios, bundleIdentifier: newBundleIdentifier };
+}
+
+/**
+ * Check if static config exists and if ios.bundleIdentifier is defined there.
+ * It will return false if value in static confgi is different than exp.ios.bundleIdentifier
```
brentvatne, line 106:
```suggestion
 * It will return false if value in static config is different than exp.ios.bundleIdentifier
```

packages/eas-cli/src/commands/build/configure.ts
```diff
@@ -15,15 +16,26 @@ export default class BuildConfigure extends Command {
...
+    }),
   };
 
   async run() {
     const { flags } = this.parse(BuildConfigure);
     const platform = flags.platform as BuildCommandPlatform;
+    const allowExperimental = flags['allow-experimental'];
+    if (allowExperimental) {
+      log.warn(
+        'Project configuration will execute some additional steps, that might fail if structure of your native project is significantly different from "expo eject" or "expo init"'
```
brentvatne, line 31:
```suggestion
        'Project configuration will execute some additional steps that might fail if structure of your native project is significantly different from "expo eject" or "expo init"'
```

my only concern with this text is that it leaves users that don't have an app initialized this way without any understanding of how they should proceed if it does not work for them

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -22,3 +26,73 @@ export async function getProjectIdAsync(projectDir: string): Promise<string> {
     projectName: exp.slug,
   });
 }
+
+// TODO move to @expo/config
```
brentvatne, line 30: :+1:

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: it's should never happen

I did not add any comments that state that because use assert should already inform about it.

packages/eas-cli/src/commands/build/configure.ts
wkozyra95: There could be two types of operations hidden under this flag (required and optional)
- for optional (like a renaming package) in case of error we can just abort process and print message to user to remove `--alow-experimental` flag
- for required values behavior would be the same as optional when running with the flag `--alow-experimental`, but if flag is not specified configure should throw descriptive error what needs to be configured for configure to work.

I didn't want to introduce that distinction in this PR because for now we only have one optional operation that needs it.

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  } else if (!applicationIdFromAndroidProject && applicationIdFromConfig) {
+    // This should never happen, adding warning just in case
+    log.warn(
+      'applicationId is not specified in your ./android/app/build.gradle file. Make sure your project is confgured correctly before building.'
+    );
+    return;
+  }
+  if (allowExperimental) {
+    // package name does not need to be updated
```
dsokal, line 90: I think this comment is confusing, at least to me. 
Why?
1. I see a comment "package name does not need to be updated"
2. Just one line below, we're calling a function called `updatePackageNameAsync`
3. WTF 🤯

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+    android: { ...rawStaticConfig.expo.android, package: newApplicationId },
+  };
+  await fs.writeJson(paths.staticConfigPath, rawStaticConfig, { spaces: 2 });
+
+  exp.android = { ...exp.android, package: newApplicationId };
+}
+
+/**
+ * Check if static config exists and if expo.android.package is defined there.
+ * It will return false if value in static confgi is different than exp.android.package
```
dsokal, line 116:
```suggestion
 * It will return false if the value in static config is different than expo.android.package
```

packages/eas-cli/src/build/android/applicationId.ts
dsokal: So maybe we should have this check/assert outside of this function? It's weird it's named `updateAppJsonConfigAsync` and there is an assert (checking if it's not dynamic config after all) inside.

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+  const rawStaticConfig = await fs.readJson(paths.staticConfigPath);
+  return rawStaticConfig?.expo?.android?.package === exp.android.package;
+}
+
+async function updateApplicationIdInBuildGradleAsync(projectDir: string, exp: ExpoConfig) {
+  const buildGradlePath = AndroidConfig.Paths.getAppBuildGradle(projectDir);
+  let buildGradleContent = await fs.readFile(buildGradlePath, 'utf8');
+  buildGradleContent = AndroidConfig.Package.setPackageInBuildGradle(exp, buildGradleContent);
+  await fs.writeFile(buildGradlePath, buildGradleContent);
```
dsokal, line 137:
```suggestion
  const buildGradleContents = await fs.readFile(buildGradlePath, 'utf8');
  const updatedBuildGradleContents = AndroidConfig.Package.setPackageInBuildGradle(exp, buildGradleContents);
  await fs.writeFile(buildGradlePath, updatedBuildGradleContents);

File: packages/eas-cli/src/utils/git.ts [hunk 13]
```diff
@@ -8,8 +8,13 @@ async function gitStatusAsync({ showUntracked }: GitStatusOptions = {}): Promise
...
 }
 
-async function gitDiffAsync(): Promise<void> {
-  await spawnAsync('git', ['--no-pager', 'diff'], { stdio: ['ignore', 'inherit', 'inherit'] });
+async function gitDiffAsync({ withPager = false }: { withPager?: boolean } = {}): Promise<void> {
+  const options = withPager ? [] : ['--no-pager'];
+  await spawnAsync('git', [...options, 'diff'], { stdio: ['ignore', 'inherit', 'inherit'] });
+}
+
+async function gitDiffOutputAsync(): Promise<string> {
```
dsokal, line 16:
```suggestion
async function getGitDiffOutputAsync(): Promise<string> {
```
🤔

packages/eas-cli/src/prompts.ts
```diff
@@ -55,3 +55,18 @@ export async function toggleConfirmAsync(
...
 }
+
+export async function pressAnyKeyToContinueAsync() {
+  process.stdin.setRawMode(true);
+  process.stdin.resume();
+  process.stdin.setEncoding('utf8');
+
+  await new Promise(res => {
+    process.stdin.on('data', key => {
+      if (String(key) === '\u0003') {
```
dsokal, line 66: Why are we checking for `END OF TEXT` (https://www.fileformat.info/info/unicode/char/0003/index.htm) here?

packages/eas-cli/src/build/ios/configure.ts
```diff
@@ -24,7 +23,14 @@ export async function validateAndSyncProjectConfigurationAsync(
   projectDir: string,
   exp: ExpoConfig
 ): Promise<void> {
-  // TODO: check bundle identifier
+  const bundleIdentifierFromPbxproj = IOSConfig.BundleIdenitifer.getBundleIdentifierFromPbxproj(
+    projectDir
+  );
+  if (!bundleIdentifierFromPbxproj || bundleIdentifierFromPbxproj !== exp.ios?.bundleIdentifier) {
+    throw new Error(
+      'Bundle identifier is not configured correctly in your Xcode project. Please run "eas build:configure" first to configure it.'
```
dsokal, line 31:
```suggestion
      'Bundle identifier is not configured correctly in your Xcode project. Please run "eas build:configure" to configure it.'
```

packages/eas-cli/src/build/ios/bundleIdentifer.ts
```diff
@@ -1,76 +1,121 @@
...
+  }
+}
+
+async function updateAppJsonConfigAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  newBundleIdentifier: string
+): Promise<void> {
+  const paths = getConfigFilePaths(projectDir);
+  assert(paths.staticConfigPath, "can't update dynamic configs");
```
dsokal, line 92: Ditto about the check

packages/eas-cli/src/build/ios/bundleIdentifer.ts
```diff
@@ -1,76 +1,121 @@
...
+    ios: { ...rawStaticConfig.expo.ios, bundleIdentifier: newBundleIdentifier },
+  };
+  await fs.writeJson(paths.staticConfigPath, rawStaticConfig, { spaces: 2 });
+
+  exp.ios = { ...exp.ios, bundleIdentifier: newBundleIdentifier };
+}
+
+/**
+ * Check if static config exists and if ios.bundleIdentifier is defined there.
+ * It will return false if value in static config is different than exp.ios.bundleIdentifier
```
dsokal, line 106:
```suggestion
 * It will return false if the value in static config is different than expo.ios.bundleIdentifier
```

dsokal (approved):

packages/eas-cli/src/build/ios/bundleIdentifer.ts
EvanBacon: Depending on what editor you have @wkozyra95 (I use VSCode), you can get a autocorrect linter, I use this one https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

packages/eas-cli/src/prompts.ts
wkozyra95: that is a code for `ctrl-c`
there is a comment line bellow

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  } else if (!applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    throw new Error(
+      `Please define "android.package" in ${configDescription} and run "eas build:configure" again.`
+    );
+  } else if (applicationIdFromAndroidProject && !applicationIdFromConfig) {
+    if (getConfigFilePaths(projectDir).staticConfigPath) {
+      await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
+    } else {
+      throw new Error(
```
EvanBacon, line 78: Maybe you could just copy over the prompt (and also tests) we use in expo-cli for now https://github.com/expo/expo-cli/blob/master/packages/expo-cli/src/commands/eject/ConfigValidation.ts

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: why?
name `updateAppJsonConfigAsync` already states that we want to update a static config (`AppJson`)

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: it's optional

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+          },
+        ],
+      });
+      switch (applicationIdSource) {
+        case ApplicationIdSource.AppJson: {
+          await updateApplicationIdInBuildGradleAsync(projectDir, exp);
+          break;
+        }
+        case ApplicationIdSource.AndroidProject: {
+          await updateAppJsonConfigAsync(projectDir, exp, applicationIdFromAndroidProject);
```
EvanBacon, line 65: Maybe we should check if the user has an Expo config, and if so, quit the process and tell the user that the config doesn't match the native project, that they should re-run `expo eject` to sync automatically, or manually sync the config if eject doesn't work for some reason. This leans more towards encouraging the user to control their project via the app config. Also allows us to keep complex syncing logic in the expo/config ecosystem.

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: I'm referring here to variable exp passed as an argument

packages/eas-cli/src/submissions/ios/AppProduce.ts
```diff
@@ -33,16 +33,9 @@ export async function ensureAppStoreConnectAppExistsAsync(
...
-        displayAutoconfigMessage: false,
-        configDescription: description,
-      }));
-  } catch (e) {
+  const resolvedBundleId =
+    bundleIdentifier ?? (await getAppIdentifierAsync(ctx.projectDir, Platform.iOS));
+  if (!resolvedBundleId) {
     throw new Error(
       `Please define "expo.ios.bundleIdentifier" in your app config or provide it using --bundle-identifier param.
 Learn more here: https://expo.fyi/bundle-identifier`
```
EvanBacon, line 41: In the future we should make this more like a controlled, failing warning message like https://github.com/expo/expo-cli/blob/2ac0ea6834727f346d5c6e55fc458b73d01d5254/packages/expo-cli/src/commands/init.ts#L132-L142 and reserve errors for unexpected functionality.

packages/eas-cli/src/build/configure.ts
```diff
@@ -60,6 +63,11 @@ export async function configureAsync(options: {
         "Aborting, run the command again once you're ready. Make sure to commit any changes you've made."
       );
     }
+    log.newLine();
```
EvanBacon, line 66:
```suggestion
    log.addNewLineIfNone();
```
??

packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -0,0 +1,158 @@
...
+  const updatedAndroidManifest = AndroidConfig.Package.setPackageInAndroidManifest(
+    exp,
+    androidManifest
+  );
+  await AndroidConfig.Manifest.writeAndroidManifestAsync(
+    androidManifestPath,
+    updatedAndroidManifest
+  );
+
+  await AndroidConfig.Package.renamePackageOnDisk(exp, projectDir);
```
EvanBacon, line 157: In config plugins, we actually run this first (second because of configure-splash-screen race conditions). Found that running this first was the safest because of file resolution looking for application code using the package name.

packages/eas-cli/src/build/ios/bundleIdentifer.ts
```diff
@@ -1,76 +1,121 @@
...
+    );
+  }
+}
+
+async function updateAppJsonConfigAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  newBundleIdentifier: string
+): Promise<void> {
+  const paths = getConfigFilePaths(projectDir);
```
EvanBacon, line 91: nice

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: ok, but I'll do that in separate PR

packages/eas-cli/src/build/configure.ts
wkozyra95: explicit newline here was intended, I think that it looks better this way
2020-12-01-171100_1029x293_scrot

packages/eas-cli/src/build/android/applicationId.ts
wkozyra95: That sth similar i initially intended, there was log above but I forgot to add the same check inside switch

`updateAppJsonConfigAsync` should not be called if `hasApplicationIdInStaticConfigAsync` returns false

dsokal (approved):

PR #96, opened 2020-11-30
[eas-cli] add release:view

jkhales:
Implement `release:view`

Currently we list the last 1000 `updateGroup`s. Should this be truncated to last 10? If so, should we paginate? 
Screen Shot 2020-11-30 at 8 56 16 AM

### DIFF
--- a/packages/eas-cli/src/commands/release/view.ts
+++ b/packages/eas-cli/src/commands/release/view.ts
@@ -82,6 +82,7 @@ async function viewUpdateReleaseAsync({
 }
 
 export default class ReleaseView extends Command {
+  static hidden = true;
   static description = 'View a release.';
 
   static args = [
--- a/packages/eas-cli/src/commands/release/view.ts
+++ b/packages/eas-cli/src/commands/release/view.ts
@@ -3,7 +3,7 @@ import { Command, flags } from '@oclif/command';
 import chalk from 'chalk';
 import Table from 'cli-table3';
 import gql from 'graphql-tag';
-import _ from 'lodash';
+import { groupBy } from 'lodash';
 
 import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
 import { Update } from '../../graphql/generated';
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/release/view.ts
```diff
@@ -0,0 +1,171 @@
...
+        }
+      >(
+        gql`
+          query ViewRelease($appId: String!, $releaseName: String!) {
+            app {
+              byId(appId: $appId) {
+                updateReleaseByReleaseName(releaseName: $releaseName) {
+                  id
+                  releaseName
+                  updates(offset: 0, limit: 1000) {
```
jkhales, line 54: Should this be truncated to last 10? If so, should we paginate?

jkhales: Resolved in private discussion with @fson. We will return it all and allow `less` to handled it.

packages/eas-cli/src/config.ts
```diff
@@ -0,0 +1 @@
+export const PAGE_LIMIT = 10000; // ten thousand
```
jkhales, line 1: Is this the right filename/place for this?

fson: I'd just hard-code it in places that need this for now, easy to extract later. This also avoids the problem of `PAGE_LIMIT` being maybe too generic as a name (what pages does this apply to?).

```suggestion
export const PAGE_LIMIT = 10_000;
```

jkhales: oh this is a cool trick

packages/eas-cli/src/commands/release/view.ts
```diff
@@ -0,0 +1,171 @@
+import { getConfig } from '@expo/config';
+import { Command, flags } from '@oclif/command';
+import chalk from 'chalk';
+import Table from 'cli-table3';
+import gql from 'graphql-tag';
+import _ from 'lodash';
```
fson, line 6:
```suggestion
import groupBy from 'lodash/groupBy';
```
(We don't import `_`)

fson (approved):

PR #98, opened 2020-11-30
[eas-cli] add channel:create

jkhales:
Create a channel for a project.

Default pointing to no releases.

### DIFF
--- a/packages/eas-cli/src/commands/channel/create.ts
+++ b/packages/eas-cli/src/commands/channel/create.ts
@@ -1,0 +1,155 @@
+import { getConfig } from '@expo/config';
+import { Command, flags } from '@oclif/command';
+import chalk from 'chalk';
+import gql from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
+import { UpdateChannel } from '../../graphql/generated';
+import log from '../../log';
+import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
+import {
+  findProjectRootAsync,
+  getProjectAccountNameAsync,
+  getReleaseByNameAsync,
+} from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+import { createUpdateReleaseOnAppAsync } from '../release/create';
+
+async function createUpdateChannelOnAppAsync({
+  appId,
+  channelName,
+  releaseId,
+}: {
+  appId: string;
+  channelName: string;
+  releaseId: string;
+}): Promise<UpdateChannel> {
+  // Point the new channel at a release with its same name.
+  const releaseMapping = JSON.stringify({
+    data: [{ releaseId, releaseMappingLogic: 'true' }],
+    version: 0,
+  });
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation<
+        { updateChannel: { createUpdateChannelForApp: UpdateChannel } },
+        { appId: string; channelName: string; releaseMapping: string }
+      >(
+        gql`
+          mutation CreateUpdateChannelForApp(
+            $appId: ID!
+            $channelName: String!
+            $releaseMapping: String!
+          ) {
+            updateChannel {
+              createUpdateChannelForApp(
+                appId: $appId
+                channelName: $channelName
+                releaseMapping: $releaseMapping
+              ) {
+                id
+                channelName
+                releaseMapping
+              }
+            }
+          }
+        `,
+        {
+          appId,
+          channelName,
+          releaseMapping,
+        }
+      )
+      .toPromise()
+  );
+  return data.updateChannel.createUpdateChannelForApp;
+}
+
+export default class ChannelCreate extends Command {
+  static hidden = true;
+  static description = 'Create a channel on the current project.';
+
+  static args = [
+    {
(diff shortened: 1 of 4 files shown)


jkhales:
Adding @jonsamp as he may have ideas about the default `releaseMapping` for a new `channel`

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -0,0 +1,127 @@
...
+import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
+import log from '../../log';
+import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
+import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+
+type UpdateChannel = {
+  id: string;
+  channelName: string;
+};
```
fson, line 15: Use the types from `src/graphql/generated.ts`.

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -0,0 +1,127 @@
...
+    {
+      name: 'channelName',
+      required: false,
+      description: 'Name of the channel to create',
+    },
+  ];
+
+  static flags = {
+    json: flags.boolean({
+      description: 'return a json with the new channel ID, name and release mapping.',
```
fson, line 74:
```suggestion
      description: 'print output as a JSON object with the new channel ID, name and release mapping.',
```

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -0,0 +1,127 @@
...
+      exp: { slug },
+    } = getConfig(projectDir, { skipSDKVersionRequirement: true });
+    const projectId = await ensureProjectExistsAsync({
+      accountName,
+      projectName: slug,
+    });
+
+    if (!channelName) {
+      const validationMessage = 'Channel name may not be empty.';
+      if (jsonFlag) {
```
fson, line 100:
```suggestion
      if (flags.json) {
```
No need to assign this to a separate variable, I think.

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -0,0 +1,127 @@
...
+      channelName,
+    });
+
+    if (jsonFlag) {
+      log(newChannel);
+      return;
+    }
+
+    log.withTick(
+      `️Created a new channel: ${chalk.bold(newChannel.channelName)} on project ${chalk.bold(
```
fson, line 122:
```suggestion
      `️Created a new channel ${chalk.bold(newChannel.channelName)} on project ${chalk.bold(
```

jonsamp:
The rules for mapping should be:

A channel, by default, points at a release of the same name. So a channel named `main` will automatically point at the release `main`. If I change the mapping to make the channel `main` point to `version-new`, then that should be persisted.'

If you have issues with this @jkhales lmk and we can figure out something!

jkhales:
> A channel, by default, points at a release of the same name. So a channel named `main` will automatically point at the release `main`. If I change the mapping to make the channel `main` point to `version-new`, then that should be persisted.'

If the `release` does not yet exist what should the behavior be?

1. error
2. point to a non-existent release. manifest requests will send back a 400
3. create a new release with the channel name. manifest requests will send back a 400 since there are no updates on the release.

I'm tending towards 3 as it is easier on first time users and the only downside is clutter, which is easy to remedy by deletion.

jonsamp:
> > A channel, by default, points at a release of the same name. So a channel named `main` will automatically point at the release `main`. If I change the mapping to make the channel `main` point to `version-new`, then that should be persisted.'
> 
> If the `release` does not yet exist what should the behavior be?
> 
> 1. error
> 2. point to a non-existent release. manifest requests will send back a 400
> 3. create a new release with the channel name. manifest requests will send back a 400 since there are no updates on the release.
> 
> I'm tending towards 3 as it is easier on first time users and the only downside is clutter, which is easy to remedy by deletion.

3 sounds great!

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -1,6 +1,13 @@
-import { getConfig } from '@expo/config';
+import { getConfig, getConfigFilePaths } from '@expo/config';
```
jkhales, line 1: Ignore this file. It will be rebased away once this PR lands: https://github.com/expo/eas-cli/pull/107

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -0,0 +1,159 @@
...
+      releaseId = existingRelease.id;
+      releaseMessage = `We found a release with the same name`;
+    } catch (e) {
+      const newRelease = await createUpdateReleaseOnAppAsync({
+        appId: projectId,
+        releaseName: channelName,
+      });
+      releaseId = newRelease.id;
+      releaseMessage = `We also went ahead and made a release with the same name`;
+    }
```
jkhales, line 140: @jonsamp these are your requested changes.

jonsamp (approved):
I tried this out locally and it worked for me. We will want to make our success messages more descriptive with more help for new developers, but we can do that later!

fson (approved):

PR #99, opened 2020-12-01
[eas-cli] Implement update:list

jonsamp:
Adds `eas update:list` command.

# Screenshots

Screen Shot 2020-12-21 at 12 08 20 PM
Screen Shot 2020-12-21 at 11 59 44 AM
Screen Shot 2020-12-21 at 12 00 04 PM
Screen Shot 2020-12-21 at 12 00 25 PM
Screen Shot 2020-12-21 at 12 00 46 PM
Screen Shot 2020-12-21 at 12 01 09 PM
Screen Shot 2020-12-21 at 12 03 56 PM

# Test plan

Run the commands in the screenshots in a project with updates, make sure that the output matches the screenshots/your expectations. Commands include:

- `eas update:list`: should be prompted for a release name, then see a list of updates grouped by update group id
- `eas update:list default`: should see a list of updates grouped by update group id
- `eas update:list default --platform android`: should see a list of updates grouped by update group id and filtered by platform.
- `eas update:list default --all`: should see all updates, not grouped by update group id
- `eas update:list default --json`: should see all updates printed as JSON


packages/eas-cli/src/commands/update/list.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const updates = await getUpdates({
+      projectId,
+      releaseName,
+      platformFlag,
+      allFlag,
+    });
+
+    if (jsonFlag) {
+      log(updates);
```
jonsamp, line 77: We have this in a few spots around the CLI (to print JSON). I noticed that this can produce a `... ## more` message, like this:

Screen Shot 2020-12-21 at 12 10 42 PM

Would it be better to use `JSON.stringify(...)`?

packages/eas-cli/src/commands/update/list.ts
```diff
@@ -0,0 +1,102 @@
...
+    {
+      name: 'releaseName',
+      required: false,
+      description: 'Name of the release to view',
+    },
+  ];
+
+  static flags = {
+    platform: flags.string({
+      description: 'Platform-specifc updates to return: ios, android, web.',
```
jkhales, line 25:
```suggestion
      description: 'Platform-specifc updates to return: ios, android or both.',
```
(we don't support web now and won't in the near future either.)

packages/eas-cli/src/commands/update/list.ts
```diff
@@ -0,0 +1,102 @@
...
+  async run() {
+    let {
+      args: { releaseName },
+      flags: { json: jsonFlag, platform: platformFlag, all: allFlag },
+    } = this.parse(UpdateList);
+
+    const projectDir = await findProjectRootAsync(process.cwd());
+    if (!projectDir) {
+      throw new Error('Please run this command inside a project directory.');
+    }
```
jkhales, line 46: I see this flow a lot, I'm wondering if I should make a PR to just move this into `findProjectRootAsync` @fson do you have an opinion?

packages/eas-cli/src/commands/update/list.ts
jkhales: From what i understand, `--json` is primarily for piping commands so we should be sure that anything we do doesn't affect that. For example, I assume that if we piped this into a file that it would provide the whole JSON?

packages/eas-cli/src/update/updateUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  return data.app.byId.updateReleaseByReleaseName;
+}
+
+export async function getUpdates(options: {
+  projectId: string;
+  releaseName: string;
+  platformFlag?: string;
+  allFlag?: boolean;
+}) {
```
jkhales, line 87: Can we add a return type?

packages/eas-cli/src/update/updateUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+          releaseName,
+        }
+      )
+      .toPromise()
+  );
+
+  return data.app.byId.updateReleaseByReleaseName;
+}
+
+export async function getUpdates(options: {
```
jkhales, line 82:
```suggestion
export async function getUpdatesAsync(options: {
```

packages/eas-cli/src/update/updateUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+        [update.updateGroup]: {
+          ...rest,
+          platforms,
+        },
+      };
+    },
+    {} as {
+      [i: string]: TruncatedUpdate;
+    }
+  );
```
jkhales, line 126: I'm a bit worried giving the option to also return 'update groups' will confuse people.

I think we could simplify this command by making it "well typed" and only return `updates`?

Is it even possible that we want to make this into a `view update group` command? The flow could be that a user sees an `update group` in the `release:view` command and wants a more granular view?

Like, what is the use case for see a list of all of the updates instead of update groups?

jkhales (commented):
Curious what you thinking about the inline suggestions to simplify this command?

jonsamp:
> From what i understand, --json is primarily for piping commands so we should be sure that anything we do doesn't affect that. For example, I assume that if we piped this into a file that it would provide the whole JSON?
~ @jkhales 

When you pipe, you'll see the `... 10 more items`: 

Screen Shot 2020-12-22 at 9 19 17 AM

I changed the code to be: `log(JSON.stringify(updates, null, 2));`. That will format an print the entire object, and you can pipe it to a file or use something like `jq`.

packages/eas-cli/src/update/updateUtils.ts
jonsamp: > I think we could simplify this command by making it "well typed" and only return updates?
What do you mean by well typed? Can you provide an example?

> Is it even possible that we want to make this into a view update group command? The flow could be that a user sees an update group in the release:view command and wants a more granular view?

We do have the `update:show` command, which is not yet implemented. You're right that this command is nearly identical to the `release:view` command. The difference is that you can filter by platform and it does not print details about the release. I'm happy to remove this command if it's not needed, btw. @fson do you have an opinion about having `release:view` and `update:list`, which both require the developer to provide a release name and both print out a list of update groups?

> Like, what is the use case for see a list of all of the updates instead of update groups?

If people want to directly edit platform-specific updates, this would be needed. With that, I removed the `--all` flag from this PR because developers can use our API to do this work, or we can add it later. I'm not able to think of common flows that would use that information, and taking it out makes our CLI cleaner.

jkhales:
> > From what i understand, --json is primarily for piping commands so we should be sure that anything we do doesn't affect that. For example, I assume that if we piped this into a file that it would provide the whole JSON?
> > ~ @jkhales
> 
> If you pipe into a file, you'll see the `... 10 more items`, like this screenshot of a json file I piped the `--json` output into:
> 
> Screen Shot 2020-12-22 at 9 19 17 AM
> 
> I changed the code to be: `log(JSON.stringify(updates, null, 2));`. That will format and print the entire object, and you can pipe it to a file or use something like `jq`.

Oh wow, this sounds like something we may need to change throughout! thanks for noticing this.

packages/eas-cli/src/update/updateUtils.ts
jkhales: This is a good question, would love to hear what @fson thinks about it.

jonsamp:
This PR is stale. We can also revive it if we want it again. Closing for now.

PR #102, opened 2020-12-02
Hide work-in-progress commands from `--help`

fson:


brentvatne (approved):
:godmode:

packages/eas-cli/src/commands/release/create.ts
```diff
@@ -48,6 +48,7 @@ async function createUpdateReleaseOnAppAsync({
 }
 
 export default class ReleaseCreate extends Command {
+  static hidden = true;
```
brentvatne, line 51: alternatively we could do `static hidden = getenv.boolish('EXPO_DEVELOPER', false)` or pull that getenv out into a some `FeatureFlags.ts` as something like `FeatureFlags.UPDATES_ENABLED`

fson: 🧠

fson: Both sound good. An important detail related to this however, is that all these fields required to display the help pages or parse arguments get compiled into JSON manifests at build time. (This makes starting the CLI much faster because it doesn't need to import any commands other than then one that's running.) So when the users install the CLI from npm the `hidden` property will be based on what the env variable was at build time.

I think this is not bad though, we can make sure the `prepack` script sets `EXPO_DEVELOPER` (or `EAS_UPDATES_ENABLED` perhaps) to `false` when we build/publish releases. And later we might start publishing "updates preview" releases that are built/published with this enabled and people could install them with something like `npm i -g eas-cli@updates-preview`.

brentvatne: let's maybe land this and follow up with changes for that as needed then

PR #103, opened 2020-12-03
[eas-cli] Move build:create to top-level command

brentvatne:
This makes it possible to run `eas build` rather than `eas build:create`


dsokal (approved):

wkozyra95:
`static aliases = ['build:submit'];` I assume copy-paste ?

I added that change in https://github.com/expo/eas-cli/pull/105 and I would prefer to keep file in build/create.ts and make `build` alias
Closing PR for now

brentvatne:
oops 👀 yeah copy/paste

PR #107, opened 2020-12-03
[eas-cli] add getReleaseByNameAsync

jkhales:
# Why

For several of the eas-cli commands under development, it will be useful to translate the user supplied releaseName to a releaseId.

# How

Add a method to translate a releaseName into a releaseId.

# Test Plan

Tested in graphql playground

### DIFF
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -117,7 +117,7 @@ export async function getReleaseByNameAsync({
 }): Promise<UpdateRelease> {
   const data = await withErrorHandlingAsync(
     graphqlClient
-      .mutation<
+      .query<
         {
           app: {
             byId: {


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -97,3 +100,47 @@ export function getProjectConfigDescription(projectDir: string): string {
...
   return 'app.config.js/app.json';
 }
+
+export async function getReleaseByNameAsync({
+  appId,
+  releaseName,
+}: {
+  appId: string;
+  releaseName: string;
+}): Promise<UpdateRelease> {
```
fson, line 110: What's returned by this function if the release doesn't exist (e.g. there's a typo)? Or does it throw an error? I think this error should be handled in a documented way, for example throw an error like `error.code === 'RELEASE_NOT_FOUND` or similar. (Perhaps the error returned by the API is self descriptive? In that case we can just let it bubble up.)

jkhales: The error is pretty self descriptive, for example:

Screen Shot 2020-12-08 at 8 11 56 AM

fson: That's pretty good, but we should not show the `CombinedError: [GraphQL] ` cruft to the user. Additionally `@owner/slug` might be nicer to show as the app identifier.

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -97,3 +100,47 @@ export function getProjectConfigDescription(projectDir: string): string {
...
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation<
```
fson, line 113: Shouldn't this be `query` instead?

packages/eas-cli/src/project/projectUtils.ts
jkhales: OK, it seems like the the standard way of doing that is to handle it at the call site. I can do that in the upcoming `release:publish` PR.

fson (approved):

packages/eas-cli/src/project/projectUtils.ts
jkhales: Thinking about this more, it seems like the graphQL errors should all be formatted here: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/graphql/client.ts#L44

Regarding the `@owner/slug` point, I will put up a PR server side to fix that.

PR #108, opened 2020-12-03
[eas-cli] Add support for Robot users

byCedric:
# Why

This adds support for robot users, through `EXPO_TOKEN=<robot-user-token> eas ...`. 

# Future stuff

We could wait for the outcome of the `Actor.displayName` discussion. That could replace the `getActorUserName` helper and make our usernames easier to use.

# How

It's similar to [PR #2440](https://github.com/expo/expo-cli/pull/2440), but implemented in such way that it's heavily reliant on the `CurrentUserQuery` result. If we add properties, like `isExpoAdmin` to this query, they are propagated everywhere once you run `$ yarn generate-graphql-code`.

This has a lot of benefits, including easier to maintain and more type-related to the output of the API. But, relating it more the GraphQL API also has its downsides. For example, the `meActor.accounts` is typed in GraphQL as `accounts: [Account]` meaning:
- `meActor.accounts` can be `null`
- `meActor.accounts` can be an `array`
  - this array can contain `null`
  - this array can contain an object with `id` and `name`

This type of arrays are pure suffering in TypeScript, because we have to account for the possible "`null` or `array`" _AND_ "`null` within the `array`". Would be nicer to have this typed as `accounts: [Account!]!` where possible, it still may contain an empty array but it's way easier to not-have-to-account-for all those possible `null` values.

# Test Plan

- Updated all tests, run `yarn test`
- Add additional tests to both `project` and `user`

**To test manually**
- Create a robot user for your account (see GraphQL query below, not yet available on website)
- Create an access token for your robot user ([using this link](https://staging.expo.io/settings/access-tokens))
- Try smoke testing commands with `EXPO_TOKEN=<robot-token> eas ...` (e.g. `eas whoami`)

<details>
<summary>Create a robot with graphql mutation</summary>
<pre>
mutation {
  robot {
    createRobotForAccount(
      accountID: "{id}"
      robotData: { name: "GLaDOS" }
      permissions: PUBLISH
    ) {
      id
      firstName
    }
  }
}
</pre>
</details>

### DIFF
--- a/packages/eas-cli/src/build/android/UpdatesModule.ts
+++ b/packages/eas-cli/src/build/android/UpdatesModule.ts
@@ -3,12 +3,13 @@ import { AndroidConfig } from '@expo/config-plugins';
 import fs from 'fs-extra';
 
 import log from '../../log';
-import { getProjectAccountNameAsync } from '../../project/projectUtils';
+import { getProjectAccountName } from '../../project/projectUtils';
+import { ensureLoggedInAsync } from '../../user/actions';
 import { ensureValidVersions } from '../utils/updates';
 
 export async function configureUpdatesAsync(projectDir: string, exp: ExpoConfig): Promise<void> {
   ensureValidVersions(exp);
-  const accountName = await getProjectAccountNameAsync(projectDir);
+  const accountName = getProjectAccountName(exp, await ensureLoggedInAsync());
   const buildGradlePath = AndroidConfig.Paths.getAppBuildGradle(projectDir);
   const buildGradleContent = await fs.readFile(buildGradlePath, 'utf8');
 
@@ -41,7 +42,7 @@ export async function syncUpdatesConfigurationAsync(
   exp: ExpoConfig
 ): Promise<void> {
   ensureValidVersions(exp);
-  const accountName = await getProjectAccountNameAsync(projectDir);
+  const accountName = getProjectAccountName(exp, await ensureLoggedInAsync());
   try {
     await ensureUpdatesConfiguredAsync(projectDir, exp);
   } catch (error) {
--- a/packages/eas-cli/src/build/android/UpdatesModule.ts
+++ b/packages/eas-cli/src/build/android/UpdatesModule.ts
@@ -3,12 +3,12 @@ import { AndroidConfig } from '@expo/config-plugins';
 import fs from 'fs-extra';
 
 import log from '../../log';
-import { ensureLoggedInAsync } from '../../user/actions';
+import { getProjectAccountNameAsync } from '../../project/projectUtils';
 import { ensureValidVersions } from '../utils/updates';
 
 export async function configureUpdatesAsync(projectDir: string, exp: ExpoConfig): Promise<void> {
   ensureValidVersions(exp);
-  const { username } = await ensureLoggedInAsync();
+  const accountName = await getProjectAccountNameAsync(projectDir);
   const buildGradlePath = AndroidConfig.Paths.getAppBuildGradle(projectDir);
   const buildGradleContent = await fs.readFile(buildGradlePath, 'utf8');
 
@@ -29,8 +29,8 @@ export async function configureUpdatesAsync(projectDir: string, exp: ExpoConfig)
     androidManifestPath
   );
 
-  if (!AndroidConfig.Updates.isMainApplicationMetaDataSynced(exp, androidManifest, username)) {
-    const result = await AndroidConfig.Updates.setUpdatesConfig(exp, androidManifest, username);
+  if (!AndroidConfig.Updates.isMainApplicationMetaDataSynced(exp, androidManifest, accountName)) {
+    const result = await AndroidConfig.Updates.setUpdatesConfig(exp, androidManifest, accountName);
 
     await AndroidConfig.Manifest.writeAndroidManifestAsync(androidManifestPath, result);
   }
@@ -41,7 +41,7 @@ export async function syncUpdatesConfigurationAsync(
   exp: ExpoConfig
 ): Promise<void> {
   ensureValidVersions(exp);
-  const { username } = await ensureLoggedInAsync();
+  const accountName = await getProjectAccountNameAsync(projectDir);
   try {
     await ensureUpdatesConfiguredAsync(projectDir, exp);
   } catch (error) {
@@ -62,7 +62,7 @@ export async function syncUpdatesConfigurationAsync(
     await AndroidConfig.Manifest.writeAndroidManifestAsync(androidManifestPath, androidManifest);
   }
 
-  if (!AndroidConfig.Updates.isMainApplicationMetaDataSynced(exp, androidManifest, username)) {
+  if (!AndroidConfig.Updates.isMainApplicationMetaDataSynced(exp, androidManifest, accountName)) {
     log.warn(
       'Native project configuration is not synced with values present in your app.json, run "eas build:configure" to make sure all values are applied in the native project'
     );
--- a/packages/eas-cli/src/credentials/context.ts
+++ b/packages/eas-cli/src/credentials/context.ts
@@ -3,8 +3,9 @@ import chalk from 'chalk';
 import pick from 'lodash/pick';
 
 import log from '../log';
+import { getProjectAccountName } from '../project/projectUtils';
 import { confirmAsync } from '../prompts';
-import { RobotUser, User } from '../user/User';
+import { Actor } from '../user/User';
 import { ensureLoggedInAsync } from '../user/actions';
 import AndroidApi from './android/api/Client';
 import iOSApi from './ios/api/Client';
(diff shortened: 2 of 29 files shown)


packages/eas-cli/src/credentials/ios/IosCredentialsProvider.ts
```diff
@@ -147,7 +147,7 @@ export default class IosCredentialsProvider implements CredentialsProvider {
   private async fetchRemoteAsync(): Promise<PartialIosCredentials> {
     if (this.options.distribution === DistributionType.INTERNAL) {
       const { app } = this.options;
-      const account = findAccountByName(this.ctx.user.accounts, app.accountName);
+      const account = findAccountByName(this.ctx.user.accounts ?? [], app.accountName);
```
byCedric, line 150: This is a side-effect of the `accounts: [Account]` GraphQL typing, which is pretty annoying.

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -33,7 +33,8 @@ export class SetupBuildCredentials implements Action {
     let iosAppBuildCredentials: IosAppBuildCredentials | null = null;
     try {
       if (this.distribution === DistributionType.INTERNAL) {
-        const account = findAccountByName(ctx.user.accounts, this.app.accountName);
+        // todo(cedric): check if we can make `meActor.accounts` a non-nullable array (or always a non-nullable array)
+        const account = findAccountByName(ctx.user.accounts ?? [], this.app.accountName);
```
byCedric, line 37: This is a side-effect of the `accounts: [Account]` GraphQL typing, which is pretty annoying.

packages/eas-cli/src/devices/__tests__/manager-test.ts
```diff
@@ -41,38 +42,38 @@ describe(AccountResolver, () => {
 
         const resolver = new AccountResolver(projectDir, user);
         const account = await resolver.resolveAccountAsync();
-        expect(account).toEqual(user.accounts[1]);
+        expect(account).toEqual(user.accounts![1]);
```
byCedric, line 45: All these `user.accounts!` notations are also a side effect of the `accounts: [Account]` GraphQL typing.

packages/eas-cli/src/devices/manager.ts
```diff
@@ -82,8 +83,8 @@ export class AccountResolver {
   }
 
   private async promptForAccountAsync(): Promise<Account> {
-    const choices: Choice[] = this.user.accounts.map(account => ({
-      title: account.name,
+    const choices: Choice[] = (this.user.accounts ?? []).filter(Boolean).map(account => ({
```
byCedric, line 86: Another sideeffect

byCedric:
@fson I'm not sure why tests are erroring on `Cannot destructure property 'exp' of '(0 , _config(...).getConfig)(...)' as it is undefined.` from the `getProjectAccountNameAsync` method. I don't think that's possible if I look at the types from `@expo/config`. Maybe we are doing something wrong in a mock?

packages/eas-cli/src/credentials/ios/IosCredentialsProvider.ts
fson: This is an unfortunate choice of type in the schema indeed, we should open an issue in www to change the type to `accounts: [Account!]!` (unless there's an issue for that already).

In the meantime would could add a helper like this:
```ts
ensureAccountsArray(accounts: Actor['accounts']): Account[]
```
It would then be used like this:
```ts
findAccountByName(ensureAccountsArray(this.ctx.user.accounts), app.accountName);
```
This would put all the workarounds like `accounts = accounts ?? []` and `filter(account => account)` in one place rather than spreading across the codebase. The idea would be that once the schema has been improved this function could just return the argument unchanged and it would be easy to find all the places that use it and remove it.

packages/eas-cli/src/build/android/UpdatesModule.ts
```diff
@@ -3,12 +3,13 @@ import { AndroidConfig } from '@expo/config-plugins';
...
 
 import log from '../../log';
+import { getProjectAccountName } from '../../project/projectUtils';
 import { ensureLoggedInAsync } from '../../user/actions';
 import { ensureValidVersions } from '../utils/updates';
 
 export async function configureUpdatesAsync(projectDir: string, exp: ExpoConfig): Promise<void> {
   ensureValidVersions(exp);
-  const { username } = await ensureLoggedInAsync();
+  const accountName = getProjectAccountName(exp, await ensureLoggedInAsync());
```
fson, line 12: Since the user account from `ensureLoggedInAsync` is only used if the `owner` field is not specified, I wonder if it would be worth it to have a variant of `getProjectAccountName` (or `fetchProjectAccountNameAsync`) that would only fetch the user account if `owner` is not specified? (As an optimization.)

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -8,13 +8,28 @@ import pkgDir from 'pkg-dir';
 
 import { graphqlClient, withErrorHandlingAsync } from '../graphql/client';
 import { UpdateRelease } from '../graphql/generated';
+import { Actor } from '../user/User';
 import { ensureLoggedInAsync } from '../user/actions';
 import { ensureProjectExistsAsync } from './ensureProjectExists';
 
+export function getProjectAccountName(exp: ExpoConfig, user: Actor) {
```
fson, line 15:
```suggestion
export function getProjectAccountName(exp: ExpoConfig, user: Actor): string {
```

packages/eas-cli/src/credentials/context.ts
```diff
@@ -90,15 +91,18 @@ class CredentialsContext implements Context {
 
   public logOwnerAndProject() {
     if (this.hasProjectContext) {
+      const owner = getProjectAccountName(this.exp, this.user);
       // Figure out if User A is configuring credentials as admin for User B's project
-      const isProxyUser = this.exp.owner && this.exp.owner !== this.user.username;
+      const isProxyUser =
+        this.user.__typename === 'Robot' || (owner && owner !== this.user.username);
```
fson, line 97:
```suggestion
        this.user.__typename === 'Robot' || owner !== this.user.username;
```
I belive `owner` should be `String` at this point.

packages/eas-cli/src/devices/manager.ts
```diff
@@ -60,10 +61,10 @@ export class AccountResolver {
     assert(this.projectDir, 'project directory is not set ');
 
     const projectAccountName = await getProjectAccountNameAsync(this.projectDir);
-    const projectAccount = findAccountByName(this.user.accounts, projectAccountName);
+    const projectAccount = findAccountByName(this.user.accounts ?? [], projectAccountName);
     if (!projectAccount) {
       log.warn(
-        `Your user (${this.user.username}) doesn't have access to the ${chalk.bold(
+        `Your user (${getActorDisplayName(this.user)}) doesn't have access to the ${chalk.bold(
```
fson, line 67:
```suggestion
        `Your account (${getActorDisplayName(this.user)}) doesn't have access to the ${chalk.bold(
```

packages/eas-cli/src/user/actions.ts
```diff
@@ -40,3 +40,26 @@ export async function ensureLoggedInAsync(): Promise<User> {
...
+    case 'User':
+      return user.username;
+    case 'Robot':
+      return user.firstName ? `${user.firstName} (robot)` : 'robot';
+    default:
+      return 'anonymous';
+  }
+}
+
+export function ensureActorHasUsername(user: Actor) {
```
fson, line 60:
```suggestion
export function ensureActorHasUsername(user: Actor): string {
```

fson (changes requested):

packages/eas-cli/src/build/android/UpdatesModule.ts
byCedric: Yeah, that's a good one. We can definitely add that!

packages/eas-cli/src/credentials/ios/IosCredentialsProvider.ts
byCedric: This is a really good pattern, thanks for sharing. ❤️   I added it to the PR, with a `@todo` notice when the GraphQL types are stricter.

packages/eas-cli/src/build/android/UpdatesModule.ts
byCedric: Also, only in the `UpdatesModule` scenario, the user doesn't have to be authenticated. In all other scenarios, such as creating releases, the user _has_ to be authenticated regardless of the owner. We can still do it, but the eventual improvements are limited.

fson: Having to be logged in is different from having to fetch the user profile from the server, because the username is stored locally in the session file.

I think ideally we wouldn't fetch a user profile from the server unless we have to.

In Expo CLI we would use [`getCurrentUsernameAsync`](https://github.com/expo/expo-cli/blob/d60df2a1f7980b5c47c2c64bee825d5071ac077f/packages/xdl/src/User.ts#L289), which doesn't make an HTTP  request if you're a person, and if you're a robot, well in that case `owner` is mandatory, so we still wouldn't fetch the user profile.

fson (approved):

PR #118, opened 2020-12-04
[eas-cli] Use unified website route for all builds (no more /v2)

FiberJW:
# Why

As a follow-up to https://github.com/expo/universe/pull/6547, which unifies the URL structure of builds pages for Turtle v2, we should stop serving those URLs in our CLIs.

# How

- removed `/v2` from build link template


brentvatne:
oops! changed this on master already. thanks though

PR #126, opened 2020-12-07
Improve error messages for invalid name cases

EvanBacon:
# Why

- I found it not clear that I could submit with a different name than what was in my app config. 
- show that we are aware of these errors and have solutions for working around them.

# How

Log a warning about using the flag or changing the name in the app config.

# Test Plan

- (invalid characters) Submit an app with the name: "Speed 🏁"
- (name taken) Submit an app with the name: "Speedway"


brentvatne (approved):
we should circle back and add tests for error message cases, but looks good for now

PR #127, opened 2020-12-07
flip profile querying logic

EvanBacon:
# Why

- https://exponent-internal.slack.com/archives/C017N0N99RA/p1607342838383800
> I wonder if maybe the problem is: we query a `BundleId` then query the `profiles` on that object (URL `bundleIds/ID/profiles`) but maybe those are stale and apple doesn’t update them :thinking_face: we could try querying against all `profiles/` then filtering by bundle id locally. 

# Test Plan

- Unsure how to test this logic against the use-case that failed.


EvanBacon:
Not needed right now

PR #134, opened 2020-12-08
Fix the release action

fson:
# Why

The build was failing, e.g. https://github.com/expo/eas-cli/runs/1518995963

# How

Need to make sure all packages are built with `tsc` before running `oclif-dev pack`.

# Test Plan
 
Push a version tag and check the results.


fson:
Build passing here again: https://github.com/expo/eas-cli/actions/runs/409045877

Issue #7, opened 2020-09-15
Add action to report bundle size changes on PR


EvanBacon:
this is something I've implemented on CRNA and it works nicely for detecting when updating packages causes the bundle size to increase in an unexpected way https://github.com/expo/create-react-native-app

PR #139, opened 2020-12-09
Add GitHub Actions workflow measuring the difference in size of the CLI

fson:
# Why

We should be intentional about adding dependencies and be aware of the effect they have on the download size of EAS CLI, because making it too large will make installing the tools too slow. The target time to install our tools is 30 seconds. To be able to achieve this, the full EAS CLI download should probably not be larger than ~40 MB it is at the moment.

# How

Added a GitHub Action inspired by https://github.com/expo/eas-cli/issues/7.

# Test Plan

Make sure the action succeeds.

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -1,0 +1,18 @@
+name: Size
+on:
+  pull_request:
+    types: [synchronize, opened]
+jobs:
+  compare:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14'
+      - uses: preactjs/compressed-size-action@v2
+        with:
+          build-script: 'build:tarballs:linux'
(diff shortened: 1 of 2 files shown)


dsokal (approved):
LGTM

Just one comment:
I know that we're not planning on adding any platform-specific (native) dependencies to EAS CLI but maybe we should add similar actions for macOS and Windows builds? 🤔

fson:
Yeah, I thought of adding Windows and macOS, but the problem is that in order to get the platform specific dependencies installed we would need to run a Windows build on Windows and macOS build on macOS. `preactjs/compressed-size-action` doesn't really support this, because it expects a single command it can then run against both the PR branch and its base branch.

We also should absolutely avoid platform specific dependencies so there shouldn't be a difference in these sizes – just measuring the size on Linux (I've seen more network issues like failed Yarn installs on macOS runners recently, `ubuntu` seems more reliable) is a good enough compromise for now, I think. Also this runs on every PR so it needs to be fast. For release builds we always build the three platforms so we would see the sizes, in case something was amiss.

PR #144, opened 2020-12-10
[eas submit] UX improvements

barthap:
# Why

Addresses internal feedback, a little improves user experience.

# How

- After successful iOS submission, a message with app store link is displayed along with information about processing by Apple. For Android, unfortunately, there's no easy way to achieve link to Google Play Console.
- Changed Submission Summary table - this was done before #157 (had to rebase due to duplication of my changes)
- Removed `App Specific Password` row - it was useless (always displayed `[hidden]`)
- Reordered table values for  both platforms (see below)
- Instead of displaying `Archive URL`, the real user choice is now displayed (path / build id). This needed modification of `ArchiveFileSource.ts` to return also real user choice.
   > Maybe, in case of uploading local file, an archive URL should also be displayed - sometimes it may be useful

## Further Steps

- Ability for user to choose build from the list (extend current choice: id / latest)
- Improve iOS App Store Connect app creation - Ask for confirmation about App Name, optional ask for sku etc.
- Maybe ask for confirmation before starting submission.
- Find a way to check if specific binary has already been submitted.

# Test Plan

Tested manually, changes visible on screens below (mocked some values).

---

ff

Screenshot 2020-12-15 at 10 42 27

### DIFF
--- a/packages/eas-cli/src/submissions/android/AndroidSubmitter.ts
+++ b/packages/eas-cli/src/submissions/android/AndroidSubmitter.ts
@@ -8,7 +8,12 @@ import {
   ArchiveType,
   SubmissionPlatform,
 } from '../types';
-import { printSummary } from '../utils/summary';
+import {
+  ArchiveSourceSummaryFields,
+  breakWord,
+  formatArchiveSourceSummary,
+  printSummary,
+} from '../utils/summary';
 import { AndroidPackageSource, getAndroidPackageAsync } from './AndroidPackageSource';
 import { AndroidSubmissionConfig, ReleaseStatus, ReleaseTrack } from './AndroidSubmissionConfig';
 import { ServiceAccountSource, getServiceAccountAsync } from './ServiceAccountSource';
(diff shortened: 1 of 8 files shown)


packages/eas-cli/src/submissions/android/AndroidSubmitter.ts
```diff
@@ -53,71 +62,66 @@ class AndroidSubmitter extends BaseSubmitter<AndroidSubmissionContext, AndroidSu
...
+      ...formatArchiveSourceSummary(archive),
       serviceAccountPath,
-    });
-    return { ...submissionConfig, serviceAccount };
+    };
   }
 }
 
-/**
- * Log the summary as a table. Exported for testing locally.
```
EvanBacon, line 80: has a new way to test this been added?

packages/eas-cli/src/submissions/utils/summary.ts
```diff
@@ -1,31 +1,53 @@
...
+      // TODO: Resolve real build ID here
+      summarySlice.buildId = '[latest]';
+      break;
+  }
+  return summarySlice;
+}
 
 export function printSummary<T>(
   summary: T,
+  title: string,
```
EvanBacon, line 38: Since you can only upload one platform at a time, the header seems unneeded.

packages/eas-cli/src/submissions/ios/IosSubmitter.ts
```diff
@@ -35,9 +44,25 @@ class IosSubmitter extends BaseSubmitter<IosSubmissionContext, IosSubmissionOpti
...
+
+    if (result === SubmissionStatus.FINISHED) {
+      log(
+        'Your binary has been successfully uploaded to App Store Connect!\n' +
+          'It is now being processed by Apple - you will receive an e-mail when the processing finishes.\n' +
+          'It usually takes about 5-10 minutes depending on how busy Apple servers are.\n' +
+          'When it’s done, you can see your build here: ' +
+          chalk.dim.underline(
+            `https://appstoreconnect.apple.com/apps/${this.options.ascAppId}/appstore/ios`
+          )
```
EvanBacon, line 63: This message is pretty verbose and the format doesn't seem to match the rest of the command. 
Maybe newline it and make it more like a bullet list. Use terminal link for the link. 
Generally this message also seems too long.

EvanBacon (approved):

brentvatne (approved):

PR #146, opened 2020-12-10
[eas-cli] add release:publish 4/4

jkhales:
# Why

This is the final of 4 PR's to add `eas release:publish`. 

It adds the command.

# How

Ask for: `inputDir`, `releaseName`, `runtimeVersion`, and `updateMessage`.

# Test Plan

yarn test

Also ran this in the alpha.

Screen Shot 2020-12-13 at 3 51 37 PM

### DIFF
--- a/packages/eas-cli/src/commands/release/publish.ts
+++ b/packages/eas-cli/src/commands/release/publish.ts
@@ -1,6 +1,7 @@
 import { getConfig } from '@expo/config';
 import { Command, flags } from '@oclif/command';
 import chalk from 'chalk';
+import Table from 'cli-table3';
 import ora from 'ora';
 
 import { PublishMutation } from '../../graphql/mutations/PublishMutation';
@@ -98,14 +99,8 @@ export default class ReleasePublish extends Command {
       releaseName: releaseName!,
     });
 
-    log.withTick(
-      `️Publishing "${updateMessage}" to ${releaseName}@${runtimeVersion} to on project ${chalk.bold(
-        `@${accountName}/${slug}`
-      )}.`
-    );
-
     let updateInfoGroup;
-    const assetSpinner = ora('Uploading assets').start();
+    const assetSpinner = ora('Uploading assets...').start();
     try {
       const assets = collectAssets(inputDir!);
       await uploadAssetsAsync(assets);
@@ -117,7 +112,7 @@ export default class ReleasePublish extends Command {
     }
 
     let newUpdateGroup;
-    const publishSpinner = ora('Publishing updates').start();
+    const publishSpinner = ora('Publishing...').start();
     try {
       newUpdateGroup = await PublishMutation.publishUpdateGroupAsync({
         releaseId,
@@ -125,7 +120,7 @@ export default class ReleasePublish extends Command {
         runtimeVersion,
         updateMessage,
       });
-      publishSpinner.succeed('Published updates!');
+      publishSpinner.succeed('Published!');
     } catch (e) {
       publishSpinner.fail('Failed to published updates');
       throw e;
@@ -133,14 +128,36 @@ export default class ReleasePublish extends Command {
 
     if (jsonFlag) {
       log(newUpdateGroup);
-      return;
+    } else {
+      const outputMessage = new Table({
+        wordWrap: true,
+        chars: {
+          top: '',
+          'top-mid': '',
+          'top-left': '',
+          'top-right': '',
+          bottom: '',
+          'bottom-mid': '',
+          'bottom-left': '',
+          'bottom-right': '',
+          left: '',
+          'left-mid': '',
+          mid: '',
+          'mid-mid': '',
+          right: '',
+          'right-mid': '',
+          middle: ' ',
+        },
+        style: { 'padding-left': 0, 'padding-right': 0 },
+      });
+      outputMessage.push(
+        [chalk.dim('project'), `@${accountName}/${slug}`],
+        [chalk.dim('release'), releaseName],
+        [chalk.dim('runtimeVersion'), runtimeVersion],
+        [chalk.dim('updateGroupID'), newUpdateGroup.updateGroup],
+        [chalk.dim('message'), updateMessage]
+      );
+      log(outputMessage.toString());
     }
-    log.withTick(
-      `Published update group ${
-        newUpdateGroup.updateGroup
-      } to release ${releaseName}@${runtimeVersion} on app ${chalk.bold(
-        `@${accountName}/${slug}`
-      )}!`
-    );
   }
 }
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+    );
+
+    log.withTick('Collecting assets...');
+    const assets = collectAssets(inputDir!);
+
+    log.withTick('Uploading assets...');
+    await uploadAssetsAsync(assets);
+    const updateInfoGroup = buildUpdateInfoGroup(assets);
+
+    log.withTick('Publishing...');
```
fson, line 132: If `--json` is specified, these logs break the output, which then can't be parsed as JSON.

I think the correct way to do this would be to only print the final JSON output to `stdout` and print these status messages to `stderr` (idea from https://clig.dev/#the-basics).

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+      releaseName: releaseName!,
+    });
+
+    log.withTick(
+      `️Publishing "${publishMessage}" to ${releaseName}@${runtimeVersion} to on project ${chalk.bold(
+        `@${accountName}/${slug}`
+      )}.`
+    );
+
+    log.withTick('Collecting assets...');
```
fson, line 125: This makes it seem like collecting assets is already done. Because these things might take a while, I think you'll want to use `ora` instead of `log.withTick`. An example of this:
https://github.com/expo/eas-cli/blob/2ec4279ac78215e7c07a96cb41e6daf6a117fedf/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts#L68-L85

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    log.withTick('Uploading assets...');
+    await uploadAssetsAsync(assets);
+    const updateInfoGroup = buildUpdateInfoGroup(assets);
+
+    log.withTick('Publishing...');
+    const newUpdateGroup = await PublishMutation.publishUpdateGroupAsync({
+      releaseId,
+      updateInfoGroup,
+      runtimeVersion: runtimeVersion!,
```
fson, line 136:
```suggestion
      runtimeVersion: runtimeVersion,
```
Need to fix the types above to make this work. It's not advisable to use `!` like this because it could hide an actual type error if the code above gets changed later.

Maybe something like this?
```ts
// ...
const releaseName: string = flags.release ?? await promptReleaseNameAsync();
const runtimeVersion: string = flags.runtimeVersion ?? await promptRuntimeVersionAsync();
// ...
```

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+      if (jsonFlag) {
+        throw new Error(validationMessage);
+      }
+      ({ inputDir } = await promptAsync({
+        type: 'text',
+        name: 'inputDir',
+        message: 'Please enter the name of the directory created by the bundler:',
+        initial: 'dist',
+        validate: value => (value ? true : validationMessage),
+      }));
```
fson, line 71: Instead of prompting this (which is not a very good experience because the prompted name is not saved anywhere), I think `"dist"` should be the default value for the `--input-dir` param. This way if you can just run `expo export` with the default flags, followed by `eas release:publish` with default flags.

If the directory doesn't exist we can print something like:
```
The input directory "dist" does not exist.
You'll need to run a command to build the JS bundle first (e.g. `expo export`).
You can use `--input-dir` to specify a different input directory.
```

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    if (!runtimeVersion) {
+      const validationMessage = 'runtimeVersion name may not be empty.';
+      if (jsonFlag) {
+        throw new Error(validationMessage);
+      }
+      ({ runtimeVersion } = await promptAsync({
```
fson, line 92: Shouldn't this be read from `app.config.ts` or other configuration file?

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,152 @@
...
+      }));
+    }
+
+    const { id: releaseId } = await getReleaseByNameAsync({
+      appId: projectId,
+      releaseName: releaseName!,
+    });
+
+    log.withTick(
+      `️Publishing "${publishMessage}" to ${releaseName}@${runtimeVersion} to on project ${chalk.bold(
```
fson, line 120: Not sure what kind of messages people write, but if they're anything like commit messages (ranging from short to long), inlining the message here might not look great.

Because there are so many fields, maybe consider printing a differently formatted summary instead? e.g.
```
------
message             Hello world
release             1.0.0
runtime version     49
------
Publishing...
```

fson (changes requested):

packages/eas-cli/src/commands/release/publish.ts
jkhales: nice!

packages/eas-cli/src/commands/release/publish.ts
jkhales: How does this look?

Screen Shot 2020-12-22 at 7 47 06 PM

packages/eas-cli/src/commands/release/publish.ts
jkhales: I removed all of the log.ticks and replaced them with `ora` spinners or only output if the json flag is not set.

packages/eas-cli/src/commands/release/publish.ts
```diff
@@ -0,0 +1,163 @@
...
+        style: { 'padding-left': 0, 'padding-right': 0 },
+      });
+      outputMessage.push(
+        [chalk.dim('project'), `@${accountName}/${slug}`],
+        [chalk.dim('release'), releaseName],
+        [chalk.dim('runtimeVersion'), runtimeVersion],
+        [chalk.dim('updateGroupID'), newUpdateGroup.updateGroup],
+        [chalk.dim('message'), updateMessage]
+      );
+      log(outputMessage.toString());
```
jkhales, line 160: This is a combo of evan and satya's output in this discussion: https://exponent-internal.slack.com/archives/C017N0N99RA/p1607720823430200
If we are happy with how this looks, it maybe worth standardizing this throughout eas in a separate PR?

packages/eas-cli/src/commands/release/publish.ts
fson: Looks great! Maybe add an empty line before the summary?

packages/eas-cli/src/commands/release/publish.ts
fson: Yeah makes sense to pull it to a separate function and use the same formatting everywhere summaries like this are shown. (In a new PR.)

fson (approved):
Thanks, looks good 🎅

PR #147, opened 2020-12-10
[eas-cli] add publish.ts 3/4

jkhales:
# Why

This is the third of 4 PR's to add `eas release:publish`. 

It adds all of the helper functions used by the command.

It is the most complex of the 4. It maybe worth cross referencing the final https://github.com/expo/eas-cli/pull/146 to see how these functions are used.

# How

* collect assets from bundle directory.
* get signed URLs and upload assets to GCS. Query expo servers every second to monitor the progress and fail if it takes more than 60 seconds.
* If successful publish manifest.

# Test Plan

yarn test

and also 


packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+  buffer: Buffer;
+};
+type CollectedAssets = {
+  [platform in PublishPlatforms]?: {
+    launchAsset: RawAsset;
+    assets: RawAsset[];
+  };
+};
+
+export function guessContentTypeFromExtension(ext?: string): string {
```
jkhales, line 37: The content type is important for asset storage. In the future we may ask the user to specify the contentType if we don't recognize it. For now we default to 'application/octet-stream'

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+        ...asset,
+        storageKey: getStorageKeyForAsset(asset),
+      };
+    }),
+    'storageKey'
+  );
+
+  let missingAssets = await filterOutAssetsThatAlreadyExistAsync(uniqueAssets);
+  const { specifications } = await PublishMutation.getUploadURLsAsync(
+    missingAssets.map((ma: any) => ma.contentType)
```
fson, line 286:
```suggestion
    missingAssets.map((asset: RawAsset) => asset.contentType)
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+  // Wait up to TIMEOUT_LIMIT for assets to be uploaded and processed
+  const start = Date.now();
+  while (missingAssets.length > 0) {
+    if (process.env.NODE_ENV !== 'test') {
+      await new Promise(resolve => setTimeout(resolve, 1000));
+    }
+    missingAssets = await filterOutAssetsThatAlreadyExistAsync(missingAssets);
+
+    if (Date.now() - start > TIMEOUT_LIMIT) {
+      throw new Error('Failed to upload all assets. Please try again.');
```
fson, line 305: Maybe add something here that makes it clearer that uploading *timed out* (rather than some other kind of error).

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+      const presignedPost: PresignedPost = JSON.parse(specifications[i]);
+      return uploadWithPresignedPostAsync(missingAsset.buffer, presignedPost);
+    })
+  );
+
+  // Wait up to TIMEOUT_LIMIT for assets to be uploaded and processed
+  const start = Date.now();
+  while (missingAssets.length > 0) {
+    if (process.env.NODE_ENV !== 'test') {
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
fson, line 300: I think it might be nice to slow down the pace of polling (e.g. exponential backoff) when more time has passed. For example if it's already taken 30 seconds, we might as well just check every few seconds rather than once per second. This way the total number of requests that hit the server could be reduced, while the time to wait would still be roughly the same.

This is just a nice to have, though, feel free to not implement this suggestion.

fson (changes requested):

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+  [platform in PublishPlatforms]?: {
+    launchAsset: RawAsset;
+    assets: RawAsset[];
+  };
+};
+
+export function guessContentTypeFromExtension(ext?: string): string {
+  // copied from https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+  const CONTENT_TYPE: { [keyof: string]: string } = {
+    aac: 'AACaudioaudio/aac',
```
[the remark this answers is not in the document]
jkhales: That's wonderful, thank you.

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+  return {
+    fileHash,
+    contentType,
+    storageBucket: STORAGE_BUCKET,
+    storageKey,
+    bundleKey,
+  };
+}
+
+export function buildUpdateInfoGroup(assets: CollectedAssets) {
```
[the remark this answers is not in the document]
jkhales: Good point, I made this PR to add `noImplicitReturns` to the tsconfig to help catch this in the future: https://github.com/expo/eas-cli/pull/166

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+  const distRoot = path.join(projectRoot, customDist);
+  if (!fs.existsSync(distRoot)) {
+    throw new Error(`${distRoot} does not exist. Please create it with your desired bundler.`);
+  }
+  return distRoot;
+}
+
+export function collectUserDefinedAssets(distRoot: string): RawAsset[] {
+  const assetRoot = path.join(distRoot, 'assets');
+  const assetPointers = Platforms.map(platform => {
```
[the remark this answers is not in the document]
jkhales: This is a good question and one that is still being discussed. @jonsamp has argued that a less error prone flow would be to make use of the new `branch` feature and make an `ios` branch that we publish to. This keeps the branches well typed and makes the reasoning a bit simpler. 

If we decide to add the parametrization, it shouldn't be too difficult, but at the very least should be done in a separate PR as it would require some server side changes.

packages/eas-cli/src/project/publish.ts
jkhales: I added a linear back off as exponential seemed a bit aggressive.

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+import path from 'path';
+
+import { PartialManifestAsset } from '../graphql/generated';
+import { PublishMutation } from '../graphql/mutations/PublishMutation';
+import { PublishQuery } from '../graphql/queries/PublishQuery';
+import { PresignedPost, uploadWithPresignedPostAsync } from '../uploads';
+
+export const TIMEOUT_LIMIT = 60_000; // 1 minute
+let STORAGE_BUCKET: string;
+export const Platforms: PublishPlatforms[] = ['android', 'ios']; // TODO-JJ allow users to specify this in app.js
```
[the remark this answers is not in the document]
jkhales: There is different capitalization between `AppPlatform` and `import { Platform } from '@expo/config';`.

On the backend we've been using the lowercase typing supplied by the latter.

It seems like these should be unified, but that is out of the scope of this PR. For now to keep consistency with the backend will prefer the lowercase type.

packages/eas-cli/src/uploads.ts
```diff
@@ -57,26 +57,30 @@ async function obtainS3PresignedPostAsync(
 export async function uploadWithPresignedPostAsync(
   stream: Readable | Buffer,
   presignedPost: PresignedPost,
-  handleProgressEvent: ProgressHandler
-) {
+  handleProgressEvent?: ProgressHandler
```
jkhales, line 60: Allow handleProgressEvent to be optional. Uploading 100's of assets would look chaotic.

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,308 @@
...
+    // load JSON with fs instead of require for easier mocking in tests
+    return JSON.parse(fs.readFileSync(assetJsonPath).toString()).bundledAssets;
+  }).flat();
+
+  return [...new Set(assetPointers)].map(pointer => {
+    const [filename, ext] = pointer.split('_').pop().split('.');
+    return {
+      type: ext ?? '',
+      contentType: guessContentTypeFromExtension(ext),
+      buffer: fs.readFileSync(path.join(assetRoot, filename)),
```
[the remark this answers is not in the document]
jkhales: Great points. 
* I changed it to pass around the `filename` instead of a `buffer`.
* Uploads and hashing are parallelized.
* Not sure how to handle progress bar for multiple uploads. Left it out for now.

packages/eas-cli/src/graphql/queries/PublishQuery.ts
```diff
@@ -1,17 +1,13 @@
 import gql from 'graphql-tag';
 
 import { graphqlClient, withErrorHandlingAsync } from '../client';
+import { AssetMetadataResult } from '../generated';
```
jkhales, line 4: Changed to generated type.

packages/eas-cli/src/project/publish.ts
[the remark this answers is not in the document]
fson: Looks like these are used inconsistently between different parts of the API as well – compare `Build`, which uses the `AppPlatform` enum (`IOS`, `ANDROID`):
https://github.com/expo/eas-cli/blob/58919d37ab3b42585b12ac593159f5e004bd2d6b/packages/eas-cli/src/graphql/generated.ts#L695
and `Update`, which uses string (expecting lowercase `'ios'` or `'android'` I guess):
https://github.com/expo/eas-cli/blob/58919d37ab3b42585b12ac593159f5e004bd2d6b/packages/eas-cli/src/graphql/generated.ts#L971
There should prefereably be only one way to represent platforms in the GraphQL API, and I think the enum is better because it's less prone to error (no confusion between `'iOS'`, `'ios'` or `'IOS'`, values validated by the schema). But you are right that unifying this doesn't belong in this PR.

packages/eas-cli/src/uploads.ts
```diff
@@ -57,26 +57,30 @@ async function obtainS3PresignedPostAsync(
...
   if (handleProgressEvent) {
     uploadPromise = uploadPromise.on('uploadProgress', progress =>
       handleProgressEvent({ progress })
     );
+    try {
+      const response = await uploadPromise;
+      handleProgressEvent({ isComplete: true });
+      return String(response.headers.location);
+    } catch (error) {
+      handleProgressEvent({ isComplete: true, error });
```
fson, line 79: Not sure if we need these synthetic events after success/error because the caller can as easily trigger the necessary actions after the promise has resolved. Removing them (and additionally using the `Progress` event from got without wrapping it) would also greatly simplify this code:
```ts
const uploadPromise = got.post(presignedPost.url, { body: form, headers: { ...formHeaders } });
if (handleProgressEvent) {
  uploadPromise.on('uploadProgress', handleProgressEvent);
}
const response = await uploadPromise;
return String(response.headers.location);
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export function collectAssets(inputDir: string): CollectedAssets {
+  const distRoot = resolveInputDirectory(inputDir);
+  const assets = collectUserDefinedAssets(distRoot);
+  const bundles = collectBundles(distRoot);
+  const assetsFinal: CollectedAssets = {};
+
+  let platform: PublishPlatforms;
+  for (platform of Platforms) {
```
fson, line 176:
```suggestion
  for (const platform of Platforms) {
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,255 @@
...
+  if (process.env.NODE_ENV === 'test') {
+    return 'update-assets-testing';
+  } else if (process.env.EXPO_STAGING || process.env.EXPO_LOCAL) {
+    return 'update-assets-staging';
+  } else {
+    return 'update-assets-production';
+  }
+}
+
+export type PublishPlatforms = Extract<'android' | 'ios', Platform>;
```
fson, line 28:
```suggestion
export type PublishPlatform = Extract<'android' | 'ios', Platform>;
```
(The prevailing convention is to use singular names for unions.)

fson (changes requested):
Thanks for making the changes! I did another round of review and noticed a few more things below.

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,255 @@
...
+import { PublishMutation } from '../graphql/mutations/PublishMutation';
+import { PublishQuery } from '../graphql/queries/PublishQuery';
+import { PresignedPost, uploadWithPresignedPostAsync } from '../uploads';
+
+export const TIMEOUT_LIMIT = 60_000; // 1 minute
+const STORAGE_BUCKET = getStorageBucket();
+export const Platforms: PublishPlatforms[] = ['android', 'ios']; // TODO-JJ allow users to specify this in app.js
+
+function getStorageBucket(): string {
+  if (process.env.NODE_ENV === 'test') {
```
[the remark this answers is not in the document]
jkhales: What do you think `test` should point to, production or staging?

packages/eas-cli/src/uploads.ts
jkhales: This is much more concise, but looking through the code for a progress handler, it seems to expect an `isComplete` argument.

See this recent PR in particular: https://github.com/expo/eas-cli/pull/153/files#diff-54650d756e5f6628f0b3c7c2564229e97085d05dacc0fac8263e0fcb9647da47R70

Simplifying this code seems desirable, but perhaps worth consulting @EvanBacon first and doing in a separate PR.

packages/eas-cli/src/project/publish.ts
```diff
@@ -0,0 +1,255 @@
...
+      const presignedPost: PresignedPost = JSON.parse(specifications[i]);
+      return uploadWithPresignedPostAsync(fs.createReadStream(missingAsset.path), presignedPost);
+    })
+  );
+
+  // Wait up to TIMEOUT_LIMIT for assets to be uploaded and processed
+  const start = Date.now();
+  let timeout = 0;
+  while (missingAssets.length > 0) {
+    if (process.env.NODE_ENV !== 'test') {
```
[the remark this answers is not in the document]
jkhales: nice.

packages/eas-cli/src/uploads.ts
fson: I think it might be still best to keep this function as simple as possible and then implement the progress handler interface in `publish.ts`, because it will need to keep track of all the uploads in order to show a combined progress bar. And I think the progress bar can be added in another PR.

fson (approved):
Looks good! Thanks.

PR #149, opened 2020-12-10
[eas-cli] add graphql calls 2/4

jkhales:
# Why

This is the second of 4 PR's to add `eas release:publish`

# How

Update the generate graphQL and add 3 graphQL calls.

# Test Plan

### DIFF
--- a/packages/eas-cli/src/graphql/mutations/PublishMutation.ts
+++ b/packages/eas-cli/src/graphql/mutations/PublishMutation.ts
@@ -1,0 +1,57 @@
+import publishGraphQL from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../client';
+import { PublishUpdateGroupInput, Update } from '../generated';
+
+const PublishMutation = {
+  async getUploadURLsAsync(contentTypes: string[]): Promise<{ specifications: string[] }> {
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .mutation<
+          { asset: { getSignedAssetUploadSpecifications: { specifications: string[] } } },
+          { contentTypes: string[] }
+        >(
+          publishGraphQL` 
+        mutation {
+          asset {
+            getSignedAssetUploadSpecifications(assetContentTypes: ${JSON.stringify(contentTypes)}) {
+              specifications
+            }
+          }
+        }`,
+          {
+            contentTypes,
+          }
+        )
+        .toPromise()
+    );
+    return data.asset.getSignedAssetUploadSpecifications;
+  },
+
+  async publishUpdateGroupAsync(
+    publishUpdateGroupInput: PublishUpdateGroupInput
+  ): Promise<Pick<Update, 'updateGroup'>> {
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .mutation<
+          { updateRelease: { publishUpdateGroup: Pick<Update, 'updateGroup'>[] } },
+          { publishUpdateGroupInput: PublishUpdateGroupInput }
+        >(
+          publishGraphQL`
+          mutation publish($publishUpdateGroupInput: PublishUpdateGroupInput) {
+            updateRelease {
+              publishUpdateGroup(publishUpdateGroupInput: $publishUpdateGroupInput) {
+                updateGroup
+              }
+            }
+          }
+        `,
+          { publishUpdateGroupInput }
+        )
+        .toPromise()
+    );
+    return data.updateRelease.publishUpdateGroup[0];
+  },
+};
+
+export { PublishMutation };
--- a/packages/eas-cli/src/graphql/mutations/PublishMutation.ts
+++ b/packages/eas-cli/src/graphql/mutations/PublishMutation.ts
@@ -38,14 +38,14 @@ const PublishMutation = {
           { publishUpdateGroupInput: PublishUpdateGroupInput }
         >(
           gql`
-          mutation publish($publishUpdateGroupInput: PublishUpdateGroupInput) {
-            updateRelease {
-              publishUpdateGroup(publishUpdateGroupInput: $publishUpdateGroupInput) {
-                updateGroup
+            mutation PublishMutation($publishUpdateGroupInput: PublishUpdateGroupInput) {
+              updateRelease {
+                publishUpdateGroup(publishUpdateGroupInput: $publishUpdateGroupInput) {
+                  updateGroup
+                }
               }
             }
-          }
-        `,
+          `,
           { publishUpdateGroupInput }
         )
         .toPromise()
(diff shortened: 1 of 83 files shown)


packages/eas-cli/src/graphql/mutations/PublishMutation.ts
```diff
@@ -0,0 +1,57 @@
+import publishGraphQL from 'graphql-tag';
```
fson, line 1:
```suggestion
import gql from 'graphql-tag';
```

packages/eas-cli/src/graphql/mutations/PublishMutation.ts
```diff
@@ -0,0 +1,57 @@
...
+    publishUpdateGroupInput: PublishUpdateGroupInput
+  ): Promise<Pick<Update, 'updateGroup'>> {
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .mutation<
+          { updateRelease: { publishUpdateGroup: Pick<Update, 'updateGroup'>[] } },
+          { publishUpdateGroupInput: PublishUpdateGroupInput }
+        >(
+          publishGraphQL`
+          mutation publish($publishUpdateGroupInput: PublishUpdateGroupInput) {
```
fson, line 41:
```suggestion
          mutation PublishMutation($publishUpdateGroupInput: PublishUpdateGroupInput) {
```

packages/eas-cli/src/graphql/mutations/PublishMutation.ts
```diff
@@ -0,0 +1,57 @@
...
+                updateGroup
+              }
+            }
+          }
+        `,
+          { publishUpdateGroupInput }
+        )
+        .toPromise()
+    );
+    return data.updateRelease.publishUpdateGroup[0];
```
fson, line 53: Is there a reason why we only return the first `Update`?

packages/eas-cli/src/graphql/queries/PublishQuery.ts
```diff
@@ -0,0 +1,34 @@
+import publishGraphQL from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../client';
+
+const PublishQuery = {
```
fson, line 5: Putting all GraphQL queries in the `src/graphql/queries` folder and mutations in `src/graphql/mutations` folder was a pattern we tried initially, but abandoned it. It would be better to put the queries inside (or near) the modules where they're used.

packages/eas-cli/src/graphql/queries/PublishQuery.ts
```diff
@@ -0,0 +1,34 @@
+import publishGraphQL from 'graphql-tag';
```
fson, line 1:
```suggestion
import gql from 'graphql-tag';
```

packages/eas-cli/src/graphql/queries/PublishQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+  ): Promise<{ storageKey: string; status: string; __typename: string }[]> {
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .query<
+          { asset: { metadata: { status: string; storageKey: string; __typename: string }[] } },
+          { assetHashes: string[] }
+        >(
+          publishGraphQL` query {
+          asset {
+            metadata(storageKeys: ${JSON.stringify(assetHashes)}) {
```
fson, line 17:
```suggestion
          gql`query AssetMetadataQuery($assetHashes: [String!]!) {
          asset {
            metadata(storageKeys: $assetHashes) {
```
Please use GraphQL variables instead of string interpolation.

packages/eas-cli/src/graphql/queries/PublishQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+              storageKey
+              status
+            }
+          }
+        }`,
+          {
+            assetHashes,
+          },
+          { requestPolicy: 'network-only' } // Since we reptitively query this to monitor the asset upload, we need to ensure it is not cached.
```
fson, line 26: 👍

packages/eas-cli/src/utils/git.ts
```diff
@@ -56,6 +56,12 @@ async function getBranchNameAsync(): Promise<string | undefined> {
   } catch (e) {}
 }
 
+async function getLastCommitMessageAsync(): Promise<string | undefined> {
+  try {
+    return (await spawnAsync('git', ['--no-pager', 'log', '-1', '--pretty=%B'])).stdout;
```
fson, line 61:
```suggestion
    return (await spawnAsync('git', ['--no-pager', 'log', '-1', '--pretty=%B'])).stdout.trim();
```

packages/eas-cli/src/utils/git.ts
```diff
@@ -56,6 +56,12 @@ async function getBranchNameAsync(): Promise<string | undefined> {
...
+  } catch (e) {}
```
fson, line 62:
```suggestion
  } catch {
    return null;
  }
```

packages/eas-cli/src/utils/git.ts
```diff
@@ -56,6 +56,12 @@ async function getBranchNameAsync(): Promise<string | undefined> {
   } catch (e) {}
 }
 
+async function getLastCommitMessageAsync(): Promise<string | undefined> {
```
fson, line 59:
```suggestion
async function getLastCommitMessageAsync(): Promise<string | null> {
```

fson (changes requested):

packages/eas-cli/src/graphql/queries/PublishQuery.ts
jkhales: Ah woops, thanks for catching that.

packages/eas-cli/src/graphql/mutations/PublishMutation.ts
jkhales: We only display the `updateGroup` and `updateMessage` when logging a success error message. These are the same between the updates.

packages/eas-cli/src/graphql/queries/PublishQuery.ts
jkhales: OK, I will do that when landing the following (3/4) stacked PR https://github.com/expo/eas-cli/pull/147 to make rebasing simpler.

packages/eas-cli/src/utils/git.ts
jkhales: Ah, this is leftover from a bad rebase. Updated, thanks.

packages/eas-cli/src/graphql/mutations/PublishMutation.ts
fson: If it's not too late to change the GraphQL API, perhaps we could return an object instead of an array? So instead of the current return type `[Update]!`, we would return something like this:
```graphql
type PublishUpdateGroupResult {
  updateGroup: String!
  updates: [Update!]!
}
```
This also makes it possible to add fields to the result later without breaking backwards compatibility (impossible if we return an array).

fson (approved):

PR #153, opened 2020-12-14
Implement new progress logs

EvanBacon:
# Why

Replace the loading bar with a percentage ticker (also fix the loading progress).

> ✔ Uploaded to EAS 1s

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -89,7 +89,9 @@ export async function prepareBuildRequestForPlatformAsync<
     try {
       return await withAnalyticsAsync(
         async () => {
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/build/build.ts
```diff
@@ -89,7 +89,7 @@ export async function prepareBuildRequestForPlatformAsync<
     try {
       return await withAnalyticsAsync(
         async () => {
-          log(`Starting ${platformDisplayNames[job.platform]} build`);
+          // log(`Starting ${platformDisplayNames[job.platform]} build`);
```
brentvatne, line 92: delete

brentvatne (approved):
looks good for now. we should add a note to circle back and add some test coverage

PR #156, opened 2020-12-14
Simplify the prompts for auto generation

EvanBacon:
# Why

Prompts for auto generation are too complex:

```
Would you like to upload a Keystore or have us generate one for you?\nIf you don't know what this means, let us generate it! :) [Use arrow keys ...]

- Generate new keystore
- I want to upload my own file
```

Is now a confirm prompt

```
? Generate a new Android Keystore? (Y/n)
```

# Test Plan

```
✔ Do you want to generate new credentials? … yes
✔ Generate a new Android Keystore? … no

In this mode, we won't be able to make sure that your credentials are valid.
Please double check that you're uploading valid files for your app otherwise you may encounter strange errors!
When building for IOS make sure you've created your App ID on the Apple Developer Portal, that your App ID
is in app.json as `bundleIdentifier`, and that the provisioning profile you
upload matches that Team ID and App ID.

✖ Path to the Keystore file. … 
demo 𝝠 eas build -p android
✔ Ensuring @bacon/demo2 is created on Expo
✔ Do you want to generate new credentials? … yes
✔ Generate a new Android Keystore? … yes
```

### DIFF
--- a/packages/eas-cli/src/credentials/android/actions/__tests__/UpdateKeystore-test.ts
+++ b/packages/eas-cli/src/credentials/android/actions/__tests__/UpdateKeystore-test.ts
@@ -38,7 +38,7 @@ describe('run UpdateKeystore when keystore exist on www', () => {
     });
     const manager = createManagerMock();
 
-    (prompts as any).mockImplementationOnce(() => ({ answer: false })); // promptForCredentials: let expo handle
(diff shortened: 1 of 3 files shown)


brentvatne (approved):

brentvatne (changes requested):
test failure looks like it may actually be related to changes

brentvatne (approved):

PR #158, opened 2020-12-14
[submit] Unify app creation code

EvanBacon:
# Why

- unify logging and spinners


brentvatne (approved):

brentvatne:
expediting merging this to move forward with release but lmk if you have any issues upon review @barthap :)

barthap:
I'm afraid that moving all app creation code out of submission command into single `ensureAppExists` is very limiting, because I don't have a way to check if the app was already existing or has just been created.
I have some plans to prompt user for some additional information when creating app - confirmation for name, bundle ID, optionally provide a SKU etc.
Anyway, for now its LGTM, I'll rework it myself when doing these improvements.

PR #166, opened 2020-12-16
[expo-cli] set noImplicitReturns to true

jkhales:
# Why

Enforcing no implicit returns has been useful in catching errors in `www` and we seem to have an explicit is better than implicit philosophy in this repo already.

# How

added tsconfig and fixed two violations.

# Test Plan

yarn tsc


fson (approved):
👍

PR #176, opened 2020-12-23
[eas cli] Surface Apple contract messages

EvanBacon:
# Why

When Apple has contract messages, things tend to break in cryptic and unfortunate ways. Contract messages are hard to test against because they come at unexpected times. Often the issue is resolved when the user gives up and goes to app store connect manually, at which point they see the warning messages, resolve them, and get back to business. 

This PR implements our new [ITCAgreements](https://github.com/expo/third-party/pull/34) module and surfaces the contractual warning messages to the user directly so they can quickly resolve any new Apple issues. This also helps users know that we're aware of contract changes and account for them gracefully.

# How

- Assert contract errors when the user creates an app, or when bundle ID queries fail due to security reasons.
- The contract messages are returned in HTML format, we use a best-effort approach to convert them to ASCII.
- If an unknown contract state is returned (and no contract messages are present), the program will attempt to create the app, and fail with whatever Apple returns.
  - Currently we account for paid, and free contracts being active or outdated, and expired accounts (can only see states based on my accounts).
  - This helps prevent false positives in case there's a non-fatal contract message that apple chooses to show.
- Added an extra part to the app creation error informing the user to check ASC just in case we missed something (unlikely).

# Test Plan

- Testing is difficult because it depends on the state of a paid apple developer account. Right now if you have an account with an expired license you can run `eas submit -p ios` on a new app (new bundle id), it should fail with the following:

Screen Shot 2020-12-22 at 5 06 56 PM

Before, the message was:
```
✖ Failed to create App Store app HackerWebby
    UnexpectedAppleResponse: An attribute value is invalid. - Apps of this type cannot be created.
```

- If you have an expired Apple developer account (after 1 year you don't pay the membership or your card fails to process), running `eas submit -p ios` should show the following:

```
✖ Failed to register bundle identifier com.bacon.fastapp2

Contract messages from App Store Connect:

› Payment Returned  
Your payment was returned by your bank. The reason your payment was returned is:  

Your bank account number is formatted incorrectly.  

You can review your banking information and make any necessary changes in Agreements, Tax, and Banking
(https://appstoreconnect.apple.com/agreements/#) for:  
CHASE BANK - \*\*\*\*1234  
Bank Payment Reference Number: 123456789

› Developer Program Membership Expired  
Your membership has expired, and your apps have been removed from the App Store until you renew your membership. To renew, a user
with the Account Holder role must sign in (http://developer.apple.com/membercenter/index.action) and renew the
membership on the Apple Developer website. Learn More (https://developer.apple.com/support/renewal/).

    Error: App Store Connect has contract issues that must be resolved before apps can be created
```

Before, the message was:

```
✖ Failed to register bundle identifier com.bacon.fastapp2
    UnexpectedAppleResponse: This request is forbidden for security reasons - Team ID: '7T5636VC39' is not associated with an 
    active membership. To check your teams membership status, sign in your account on the developer website. 
    https://developer.apple.com/account/
```

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/contractMessages.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/contractMessages.ts
@@ -51,7 +51,9 @@ const rootUrl = 'https://appstoreconnect.apple.com';
 
 export function formatContractMessage(message: ITCAgreements.ITCContractMessage): string {
   return convertHTMLToASCII({
-    content: [message.subject, message.message].filter(Boolean).join('<br />'),
+    content: [message.subject && `<b>${message.subject}</b>`, message.message]
+      .filter(Boolean)
+      .join('<br />'),
     rootUrl,
   });
 }
--- a/packages/eas-cli/src/credentials/ios/appstore/contractMessages.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/contractMessages.ts
@@ -27,25 +27,34 @@ async function getContractMessagesAsync(context: RequestContext) {
 }
 
 async function getRequiredContractMessagesAsync(context: RequestContext) {
+  // This emulates the check that's performed on the ASC website's "apps"
+  // page before presenting the (+) create app button.
   const status = await getContractStatusAsync(context);
 
-  switch (status) {
-    // The user can freely create an app
-    case 'FREE_APP_AGREEMENT_ACTIVE':
-    case 'PAID_APP_AGREEMENT_ACTIVE':
-      return false;
-    // The user cannot create an app until they've resolved contract issues on ASC.
-    case 'FREE_APP_AGREEMENT_OUTDATED':
-    case 'PAID_APP_AGREEMENT_OUTDATED':
-    case 'EXPIRED_MEMBERSHIP':
-      return (await getContractMessagesAsync(context)) ?? [];
-    default:
-      // TODO: Maybe a silent analytic would be better
-      log.warn(
-        `Unknown Apple developer contract status "${status}". Please open an issue on https://github.com/expo/eas-cli`
-      );
-      return (await getContractMessagesAsync(context)) ?? [];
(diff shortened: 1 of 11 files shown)


packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -117,7 +127,8 @@ export async function ensureAppExistsAsync(
       }
 
       spinner.fail(`Failed to create App Store app ${chalk.dim(name)}`);
-
+      error.message +=
+        '\nPlease visit https://appstoreconnect.apple.com and ensure all warnings are resolved before trying again.';
```
brentvatne, line 131:
```suggestion
        '\nPlease visit https://appstoreconnect.apple.com and resolve any warnings, then try again.';
```

packages/eas-cli/src/credentials/ios/appstore/contractMessages.ts
```diff
@@ -0,0 +1,85 @@
...
+    case 'FREE_APP_AGREEMENT_ACTIVE':
+    case 'PAID_APP_AGREEMENT_ACTIVE':
+      return false;
+    // The user cannot create an app until they've resolved contract issues on ASC.
+    case 'FREE_APP_AGREEMENT_OUTDATED':
+    case 'PAID_APP_AGREEMENT_OUTDATED':
+    case 'EXPIRED_MEMBERSHIP':
+      return (await getContractMessagesAsync(context)) ?? [];
+    default:
+      // TODO: Maybe a silent analytic would be better
```
brentvatne: yeah this would be useful to report. cc @fson / @animeeples for thoughts on what we should be using to track these sorts of errors

brentvatne (approved):
is there any way users could end up getting warnings that wouldn't actually block submission? probably not, if so we should maybe provide a flag that lets users try to proceed despite warnings.

EvanBacon:
@brentvatne I've added an error log to prevent false positives, if a code that we don't expect is found, then it'll prompt the user to report it as an issue, show the messages, and continue with the process regardless.
Screen Shot 2020-12-23 at 1 38 56 PM

PR #183, opened 2021-01-06
[eas-cli] more gql type cleanup

quinlanj:
# Why

- Cleaned up graphql types in www (https://github.com/expo/universe/pull/6773) to update everything to stricter types. For example, types like `Maybe<Array<Maybe<Foo>>>` get converted to `Array<Foo>` where possible.
- This PR updates the autogenerated schema
- Removed `@todos` that hack around `Maybe` types

# Test Plan

- [ ] current tests still pass?


dsokal (approved):

fson (approved):
This is great. Thank you very much 💯

PR #184, opened 2021-01-08
[eas-cli] generate fragment types

quinlanj:
# Why

- The graphql code generator won't parse AST strings unless it is in a `gql` tag or preceded by the `/* GraphQL*/ ` comment.
- This PR allows fragments types to be autogenerated so we can use them in the future

# Test Plan

### DIFF
--- a/packages/eas-cli/src/graphql/types/App.ts
+++ b/packages/eas-cli/src/graphql/types/App.ts
@@ -1,8 +1,10 @@
+import gql from 'graphql-tag';
+
 import { Fragment } from '../fragment';
 
 export const AppFragment: Fragment = {
   name: 'app',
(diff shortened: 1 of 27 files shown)


dsokal (approved):
🚀

packages/eas-cli/src/graphql/types/App.ts
```diff
@@ -2,7 +2,7 @@ import { Fragment } from '../fragment';
 
 export const AppFragment: Fragment = {
   name: 'app',
-  definition: `
+  definition: /* GraphQL*/ `
```
fson, line 5:
```suggestion
  definition: gql`
```
(Maybe? I think we use `graphql-tag` in other places.)

PR #186, opened 2021-01-08
[eas-cli] gql nits

quinlanj:
# Why

this pr makes sure that everything covered by the codegen tool is named

https://the-guild.dev/blog/graphql-codegen-best-practices
```
It's highly important to name your GraphQL operations, because otherwise it will be difficult for your GraphQL client to cache and manage it. It will also make it difficult for the codegen to create easy-to-use types, and it will fallback to Unnamed_Operation_.
```

# Test Plan

### DIFF
--- a/packages/eas-cli/src/graphql/generated.ts
+++ b/packages/eas-cli/src/graphql/generated.ts
@@ -2703,13 +2703,13 @@ export type PendingBuildsForAccountAndPlatformQuery = (
   ) }
 );
 
-export type Unnamed_1_QueryVariables = Exact<{
+export type ProjectByUsernameAndSlugQueryQueryVariables = Exact<{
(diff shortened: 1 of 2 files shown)


dsokal (approved):
I don't know if it's possible but maybe some lint rule that enforces naming graphql queries/mutations would be useful here? I'm afraid that only few people are going to remember doing this in the future and more PRs like this will be necessary.

fson (approved):
Nice! Looks like we could use the [Named Operations validation rule](https://github.com/apollographql/eslint-plugin-graphql#named-operations-validation-rule) from `eslint-graphql` to check for this in the future.

quinlanj:
https://github.com/expo/eas-cli/pull/189

PR #188, opened 2021-01-09
[eas-cli] 2/n apply template literals rule

quinlanj:
# Why

Apply the graphql `template-strings` rule which perform a bunch of validations, most notably disallowing template literals in certain part of the AST string. More info here: https://github.com/apollographql/eslint-plugin-graphql

- I've also refactored the `Fragments` as discussed here: https://github.com/expo/eas-cli/pull/185#issuecomment-756765674

# Test Plan

### DIFF
--- a/packages/eas-cli/src/credentials/ios/api/graphql/mutations/AppleAppIdentifierMutation.ts
+++ b/packages/eas-cli/src/credentials/ios/api/graphql/mutations/AppleAppIdentifierMutation.ts
@@ -3,7 +3,7 @@ import gql from 'graphql-tag';
 
 import { graphqlClient, withErrorHandlingAsync } from '../../../../../graphql/client';
 import { AppleAppIdentifier } from '../../../../../graphql/generated';
-import { AppleAppIdentifierFragment } from '../../../../../graphql/types/credentials/AppleAppIdentifier';
+import { AppleAppIdentifierFragmentDoc } from '../../../../../graphql/types/credentials/AppleAppIdentifier';
 
 const AppleAppIdentifierMutation = {
   async createAppleAppIdentifierAsync(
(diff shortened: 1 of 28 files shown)


dsokal (approved):
Looks good. Just one inline comment.

packages/eas-cli/src/credentials/ios/api/graphql/queries/AppQuery.ts
```diff
@@ -14,11 +14,11 @@ const AppQuery = {
             query($fullName: String!) {
               app {
                 byFullName(fullName: $fullName) {
-                  ...${AppFragment.name}
+                  ...AppFragment
                 }
               }
             }
-            ${print(AppFragment.definition)}
+            ${print(AppFragment)}
```
jkhales, line 21:
```suggestion
            ${print(AppFragmentDoc)}
```

packages/eas-cli/src/graphql/types/App.ts
```diff
@@ -1,12 +1,7 @@
...
-
-export const AppFragment: Fragment = {
-  name: 'app',
-  definition: gql`
-    fragment app on App {
-      id
-    }
-  `,
-};
+export const AppFragment = gql`
```
jkhales, line 3:
```suggestion
export const AppFragmentDoc = gql`
```

jkhales (approved):
Massive improvement. 🔥

packages/eas-cli/src/credentials/ios/api/graphql/mutations/AppleAppIdentifierMutation.ts
```diff
@@ -17,14 +17,20 @@ const AppleAppIdentifierMutation = {
...
+                createAppleAppIdentifier(
+                  appleAppIdentifierInput: $appleAppIdentifierInput
+                  accountId: $accountId
+                ) {
+                  ...AppleAppIdentifierFragment
                 }
               }
             }
-            ${print(AppleAppIdentifierFragment.definition)}
+            ${print(AppleAppIdentifierFragmentDoc)}
```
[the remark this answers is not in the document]
quinlanj: I added the suffix for a couple reasons:
1. It's the suffix I've seen the `typescript-urql` plugin add to fragment [variables](https://graphql-code-generator.com/docs/plugins/typescript-urql) (fragmentVariableSuffix)
2. Most queries that use a fragment will do something like this:
```
async fooQuery(): Promise<FooFragment> {
    urqlClient.query(
       gql`<query ast>
           ... ${FooFragmentDoc}
       `
    )
}
```
The important part is that FooFragmentDoc is named something different from the autogenerated type, FooFragment. Otherwise you'll have to do `import as` to resolve the duplicate name. 

I dont really know if the `Doc` suffix is a common convention (other than the one time i saw a plugin do it somewhere), but if `*Node` gives you a better idea about what the object is, let's just go with that.

PR #189, opened 2021-01-09
[eas-cli] 3/n apply  named operations rule

quinlanj:
# Why

Ensure all graphql operations have a name. More info here: https://github.com/apollographql/eslint-plugin-graphql

# Test Plan


dsokal (approved):
❤️

PR #192, opened 2021-01-10
[eas-cli] 4/n apply required id field rule

quinlanj:
# Why

Require that we get the `id` field whenever possible in our graphql queries and mutations. I have mixed feelings about this rule because:
- it has a huge upside. Most graphql cachers will rely on the presence of `id` to figure whether to invalidate or add to the cache. I have personally experienced many painful bugs because of this 😭 
- downside: this rule doesn't look for `id` in referenced fragments, leading to a lot of redundancy. This is currently an unresolved bug: https://github.com/apollographql/eslint-plugin-graphql/issues/248

# Test Plan

### DIFF
--- a/packages/eas-cli/.eslintrc.js
+++ b/packages/eas-cli/.eslintrc.js
@@ -18,6 +18,14 @@ module.exports = {
         schemaJson: require('./graphql.schema.json'),
       },
     ],
+    'graphql/required-fields': [
+      'error',
+      {
+        env: 'apollo',
+        schemaJson: require('./graphql.schema.json'),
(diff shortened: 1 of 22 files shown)


dsokal (approved):
Personally, I haven't run into problems/bugs with missing cache fields.
However, having duplicated fields is not that big a problem for me if there are obvious benefits.

jkhales (approved):
This seems like it would ideally/eventually be a lint rule for the `Global Object Identification` https://graphql.org/learn/global-object-identification/

But since we don't have that yet this PR seems like the next best thing.

PR #193, opened 2021-01-10
[eas-cli] initial fetch for credentials manager

quinlanj:
# Why

This is the start of an effort to move all credential api requests from apiv2 to graphql, and to bring the UX of the credentials manager closer to what we see in the website. This PR fetches all the initial credentials required to print out the user's credentials when they first run `eas credentials`. The excalidraw wireframes are [here](https://excalidraw.com/#json=5197280390938624,q3HFlStPaD1zjYI2WdG_FQ)

Screen Shot 2021-01-12 at 8 34 08 PM

# How

- We want the credentials manager to show information in the same way the website's UI does
- I'll be committing these incremental changes in `beta` files. It's a bit difficult to integrate the new API changes to the existing code because of caching issues. The current credentials manager maintains a cache of credentials from apiv2, and it'll be hard to keep the graphql and apiv2 caches in sync. It'll be easier to just build a `beta` credentials manager and switch over to that when it's complete.

### DIFF
--- a/packages/eas-cli/src/credentials/ios/utils/printCredentialsBeta.ts
+++ b/packages/eas-cli/src/credentials/ios/utils/printCredentialsBeta.ts
@@ -1,0 +1,124 @@
+import chalk from 'chalk';
+import dateformat from 'dateformat';
+
+import {
+  CommonIosAppCredentialsFragment,
+  IosAppBuildCredentialsFragment,
+  IosDistributionType,
+} from '../../../graphql/generated';
+import log from '../../../log';
+import { fromNow } from '../../../utils/date';
+import { AppLookupParams } from '../api/GraphqlClient';
+
+export function prettyIosDistributionType(distributionType: IosDistributionType): string {
+  switch (distributionType) {
+    case IosDistributionType.AppStore:
+      return 'App Store';
+    case IosDistributionType.AdHoc:
+      return 'Ad Hoc';
+    case IosDistributionType.Development:
+      return 'Development';
+    case IosDistributionType.Enterprise:
+      return 'Enterprise';
+    default:
+      return 'Unknown';
+  }
+}
+
+export function displayEmptyIosCredentials(appLookupParams: AppLookupParams): void {
+  const { projectName, bundleIdentifier } = appLookupParams;
+  log(chalk.bold(`iOS Credentials`));
+  log(`  Project: ${projectName}`);
+  log(`  Bundle Identifier: ${bundleIdentifier}`);
+  log(`  No credentials set up yet!`);
+}
+
+/**
+ * sort a build credentials array in descending order of preference
+ */
+function sortBuildCredentialsByDistributionType(
+  iosAppBuildCredentialsArray: IosAppBuildCredentialsFragment[]
+): IosAppBuildCredentialsFragment[] {
+  // The order in which we choose the distribution type from least to most preferred
+  const typePriority = [
+    IosDistributionType.Development,
+    IosDistributionType.AdHoc,
+    IosDistributionType.Enterprise,
+    IosDistributionType.AppStore,
+  ];
+  return iosAppBuildCredentialsArray
+    .sort(
+      (buildCredentialsA, buildCredentialsB) =>
+        typePriority.indexOf(buildCredentialsA.iosDistributionType) -
+        typePriority.indexOf(buildCredentialsB.iosDistributionType)
+    )
+    .reverse();
+}
+
+export function displayIosAppCredentials(credentials: CommonIosAppCredentialsFragment): void {
+  log(chalk.bold(`iOS Credentials`));
+  log(`  Project: ${credentials.app.fullName}`);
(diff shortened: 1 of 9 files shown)


packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
```diff
@@ -0,0 +1,125 @@
...
+    const accountName = getProjectAccountName(ctx.exp, ctx.user);
+
+    const account = findAccountByName(ctx.user.accounts, accountName);
+    if (!account) {
+      throw new Error(`You do not have access to account: ${accountName}`);
+    }
+
+    const bundleIdentifier = ctx.exp.ios?.bundleIdentifier;
+    if (!bundleIdentifier) {
+      throw new Error(`ios.bundleIdentifier needs to be defined in your eas.json file`);
```
wkozyra95, line 111:
```suggestion
      throw new Error(`ios.bundleIdentifier needs to be defined in your app.json file`);
```

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
```diff
@@ -0,0 +1,125 @@
...
+  RemoveSpecificProvisioningProfile,
+  CreateDistributionCertificate,
+  UpdateDistributionCertificate,
+  RemoveDistributionCertificate,
+}
+
+export class ManageIosBeta implements Action {
+  private _currentDistributionType = null as IosDistributionType | null;
+
+  chooseDistributionType(
```
wkozyra95, line 30: If we want to keep this behavior this function should be part of the displayAllIosCredentials implementation. (an argument for changing that behavior in comments below)

packages/eas-cli/src/credentials/ios/utils/printCredentialsBeta.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export function displayEmptyIosCredentials(appLookupParams: AppLookupParams): void {
+  const { projectName, bundleIdentifier } = appLookupParams;
+  log(chalk.bold(`iOS Credentials`));
+  log(`  Project: ${projectName}`);
+  log(`  Bundle Identifier: ${bundleIdentifier}`);
+  log(`  No credentials set up yet!`);
+}
+
+export function displayAllIosCredentials(
```
wkozyra95, line 32: This function displays credentials for one app and for one specific built type, using `All` in the name is confusing here.

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (ctx.hasProjectContext) {
+      const appLookupParams = this.getAppLookupParamsFromContext(ctx);
+      const iosAppCredentials = await ctx.newIos.getIosAppCredentialsWithBuildCredentialsAsync(
+        appLookupParams
+      );
+      if (!iosAppCredentials) {
+        displayEmptyIosCredentials(appLookupParams);
+      } else {
+        this._currentDistributionType = this.chooseDistributionType(iosAppCredentials);
+        displayAllIosCredentials(iosAppCredentials, this._currentDistributionType);
```
wkozyra95, line 75: I think it would be better if we display all credentials, but if you feel strongly that only one set of credentials should be displayed I suggest limiting it to only the app store and enterprise. I assume there will be a way to display all other credentials via the credentials manager actions

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
wkozyra95: Or use https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/project/projectUtils.ts#L112
to get an actual name of the config file or files

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
quinlanj: Thinking about it a bit more, it makes sense to display all the credentials (app store/enterprise, adhoc, etc), instead of having an action to navigate to another configuration which is easy to miss.

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
quinlanj: yeah i ended up getting rid of this

packages/eas-cli/src/graphql/types/credentials/IosAppBuildCredentials.ts
```diff
@@ -36,17 +39,5 @@ export const IosAppBuildCredentialsFragmentNode = gql`
         appleTeamName
       }
     }
-    appleDevices {
```
quinlanj, line 39: i removed this because i deprecated it on the server (https://github.com/expo/universe/blob/master/server/www/src/graphql/types/credentials/IosAppBuildCredentials.ts#L27). We don't actually resolve `appleDevices` any more

wkozyra95 (approved):

PR #196, opened 2021-01-11
[eas-cli] rework expected bundle format for eas publish

jkhales:
# Why

Previously we were relying on `expo export` to build the bundles published by `eas publish`. `expo export` turns out to do a lot of unrelated stuff with the assumption the developer is going to use the output for privately hosted static endpoints. The core elements of `expo export` were factored out and put into a new command `expo bundle` https://github.com/expo/expo-cli/pull/3074. 

`expo bundle` outputs the bundle in a new format. This PR updates `eas publish` to expect this new format.

# How

Rewrote `collectAssets`.

# Test Plan

Confirmed the `collectAssets` output test still passed after updating it to the new formate.

Made a test publish after building the bundle in the new format.

### DIFF
--- a/packages/eas-cli/src/project/__tests__/publish-test.ts
+++ b/packages/eas-cli/src/project/__tests__/publish-test.ts
@@ -160,7 +160,6 @@ describe(resolveInputDirectory, () => {
 describe(collectAssets, () => {
   it('builds an update info group', () => {
     const fakeHash = 'md5-hash-of-jpg';
-    const fakeJson = { bundledAssets: [`asset_${fakeHash}.jpg`] };
     const bundles = { android: 'android-bundle-code', ios: 'ios-bundle-code' };
     const inputDir = uuidv4();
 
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/project/__tests__/publish-test.ts
```diff
@@ -177,31 +176,41 @@ describe(collectAssets, () => {
...
     Platforms.forEach(platform => {
-      fs.writeFileSync(
-        path.resolve(`${inputDir}/${platform}-index.json`),
-        JSON.stringify(fakeJson)
-      );
-      fs.writeFileSync(
-        path.resolve(`${inputDir}/bundles/${platform}-randomHash.js`),
-        bundles[platform]
-      );
+      fs.writeFileSync(path.resolve(`${inputDir}/bundles/${platform}.js`), bundles[platform]);
```
fson, line 179:
```suggestion
      fs.writeFileSync(path.resolve(inputDir, `bundles/${platform}.js`), bundles[platform]);
```

packages/eas-cli/src/project/__tests__/publish-test.ts
```diff
@@ -177,31 +176,41 @@ describe(collectAssets, () => {
...
     });
     fs.writeFileSync(path.resolve(`${inputDir}/assets/${fakeHash}`), dummyFileBuffer);
+    fs.writeFileSync(
+      path.resolve(`${inputDir}/metadata.json`),
```
fson, line 183:
```suggestion
      path.resolve(inputDir, 'metadata.json'),
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -127,55 +134,36 @@ export function resolveInputDirectory(customInputDirectory: string): string {
...
-    };
-  });
-  return bundleBuffers;
+export function loadMetadata(distRoot: string): Metadata {
+  const metadata: Metadata = JsonFile.read(path.join(distRoot, 'metadata.json'));
+  if (metadata.version !== 0) {
+    throw new Error('Only bundles with metadata version 0 are supported');
+  }
+  if (metadata.bundler !== 'metro') {
+    throw new Error('Only bundles created by metro are supported');
```
fson, line 143:
```suggestion
    throw new Error('Only bundles created with Metro are currently supported');
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -127,55 +134,36 @@ export function resolveInputDirectory(customInputDirectory: string): string {
...
 
+  const assetsFinal: CollectedAssets = {};
   for (const platform of Platforms) {
     assetsFinal[platform] = {
-      launchAsset: bundles[platform],
-      assets,
+      launchAsset: {
+        type: 'bundle',
+        contentType: 'application/javascript',
+        path: path.resolve(distRoot, metadata.fileMetadata[platform].bundle),
```
fson, line 158: Looks like this might result in `Cannot read property 'bundle' of undefined`, if the bundle doesn't include all platforms? I wonder if we should check that the key exist?

fson (commented):
Added a few comments below, nothing major – the PR looks pretty good!

packages/eas-cli/src/project/publish.ts
jkhales: good point, added validation with @hapi/joi

PR #207, opened 2021-01-15
Added support for EXPO_APPLE_TEAM_ID

EvanBacon:
# Why

In order for `expo prebuild` to have parity with `expo build` we need a way for users to define their team ID so that config plugins which require Team ID can be setup. An example of this is [expo-document-picker plugin](https://github.com/expo/expo/blob/b8e143593c8ca65966930e0b3d901aabbf8ca2ca/packages/expo-document-picker/plugin/src/withDocumentPicker.ts#L9) which uses the team ID in the entitlements when `ios.usesIcloudStorage` is enabled. For now, we'll opt towards using an environment variable. 
~~Before introducing that in an expo config plugin, we should ensure EAS supports it.~~ The config plugin just expects it now and we should set it automatically for users.

# How

- Resolve `EXPO_APPLE_TEAM_ID` as the Apple team ID if defined.

# Test Plan

```
export EXPO_APPLE_TEAM_ID=FOOBAR
eas credentials
- iOS
...
AssertionError [ERR_ASSERTION]: Your account is not associated with Apple Team ID: FOOBAR. Valid team IDs are: QQ57RJ5UTD
```

### DIFF
--- a/packages/eas-cli/src/commands/submit.ts
+++ b/packages/eas-cli/src/commands/submit.ts
@@ -112,7 +112,8 @@ export default class BuildSubmit extends Command {
       helpLabel: IOS_FLAGS,
     }),
     'apple-team-id': flags.string({
-      description: 'Your Apple Developer Team ID',
+      description:
+        'Your Apple Developer Team ID (you can also set EXPO_APPLE_TEAM_ID env variable)',
       helpLabel: IOS_FLAGS,
     }),
     'app-name': flags.string({
--- a/packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
@@ -67,7 +67,7 @@ async function loginAsync(
   }
 
   // Resolve the user credentials, optimizing for password-less login.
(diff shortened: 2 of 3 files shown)


wkozyra95:
> For expo prebuild we need a way for users to define their team ID. For now, we'll opt towards using an environment variable. Before introducing that in an expo config plugin, we should ensure EAS supports it.

Don't really see how those two actions are related and why "we should ensure EAS supports it".

I'm not really sure what `expo prebuild` does, but based on a quick look I'm guessing that this code will be run on workers to configure managed projects before the build. If I'm correct about that then those changes are not necessary and they are not affecting in any way how EAS Build works.

Alternatively, if you are adding it so the code from `config-plugins` we are using in eas-cli could see those envs then it also won't work because there is still an option to provide this value via interactive prompt.

Sorry if I'm missing a point of these changes. Can you clarify?

wkozyra95:
We talked about this during the last sync meetings, when originally reviewing this PR I didn't understand what the goal is here.
No changes in eas-cli are necessary to pass team id to the worker, that value can be read from the provisioning profile.

Closing PR, it will be handled in turtle code

PR #208, opened 2021-01-15
[eas-cli] return non-zero exit code if prompt is displayed on CI

wkozyra95:
# Why

When prompts is called on ci it exits from the process with 0 exit code

The error does not need to be descriptive because all commands that might be run on CI should support `--non-interactive` flag, this change is only to ensure non-zero exit code.

# How

- if stdin is not readable before prompts call, then throw an error.
- skip check if running in test `!global.test` (explained in a comment below)

# Test Plan

Reproduced with `eas build -p andorid < /dev/null`

### DIFF
--- a/packages/eas-cli/src/prompts.ts
+++ b/packages/eas-cli/src/prompts.ts
@@ -9,6 +9,9 @@ export async function promptAsync<T extends string = string>(
   questions: Question<T> | Question<T>[],
   options: Options = {}
 ): Promise<Answers<T>> {
+  if (process.stdin.readableFlowing === null && !global.test) {
+    throw new Error('Input is required, but stdin is not readable.');
+  }
   return await prompts<T>(questions, {
     onCancel() {
       process.exit(constants.signals.SIGINT + 128); // Exit code 130 used when process is interrupted with ctrl+c.


dsokal (approved):

packages/eas-cli/src/prompts.ts
```diff
@@ -9,6 +9,9 @@ export async function promptAsync<T extends string = string>(
   questions: Question<T> | Question<T>[],
   options: Options = {}
 ): Promise<Answers<T>> {
+  if (process.stdin.readableFlowing === null) {
+    throw new Error('Input is required, but stdin is not readable.');
```
byCedric, line 13: Maybe we should just use something like "Input is required, but command is in non-interactive mode"? Mentioning non-interactive could be more understandable compared to `stdin is not readable`?

wkozyra95: non-interactive is not exactly the same as the case where stdin is not readable

wkozyra95:
I had to check for `global.test` because otherwise all test that mock prompts are failing. The only alternative I see is to mock `./packages/eas-cli/src/promts.ts` functions in tests instead of `prompts`, but it would complicate testing because we would need to mock all those functions in tests.

I realize that checking for test env in code is not a great solution, but in this case, I think it's better than alternatives.

Issue #168, opened 2020-12-16
[ios build] Cannot find proper provisioning profile when using remote credentials


cruzach:
After letting Expo set up credentials, and verifying that those credentials are correct according to `expo credentials:manager -p ios`, the `eas build` command will fail on the "fetching provisioning profile" step:

![image](https://user-images.githubusercontent.com/35579283/102403715-b055ac80-3fb4-11eb-976c-ab39dc50054d.png)

If you sync your credentials with `eas credentials` and run another build, this time relying on your local credentials.json file, the build succeeds

byCedric:
+1

brentvatne:
@cruzach - do you have any more details on the repro steps? I can't repro by just setting up credentials, verifying, then running build

cruzach:
The problem occurs when the existing provisioning profile has been invalidated (i.e. revoked or expired), and you generate a new one (with `expo credentials:manager` for instance), eas-cli still tries to grab the old (invalid) one instead of the new one

brentvatne:
if you run into this, can you please run `eas build` with the `EXPO_APP_STORE_DEBUG=true` set?

eg, if you use macos/linux: `EXPO_APP_STORE_DEBUG=true eas build` 

once you do this, please paste the log output here

cruzach:
Results when running with `EXPO_APP_STORE_DEBUG`:
```
% EXPO_APP_STORE_DEBUG=true eas build --profile preview --platform ios
-- Enabled debug logging --
✔ Linked to project @instamenu/instamenu-mobile
› Log in to your Apple Developer account to continue
✔ Apple ID: … <APPLE ID>
› Restoring session XXX
POST https://developer.apple.com/services-account/XXX/account/listTeams.action
200 OK (POST https://developer.apple.com/services-account/XXX/account/listTeams.action)
› Team NAME (XXX)
GET https://appstoreconnect.apple.com/olympus/v1/session
200 OK (GET https://appstoreconnect.apple.com/olympus/v1/session)
› Provider NAME (XXX)
✔ Logged in Local session
POST https://developer.apple.com/services-account/XXX/account/listTeams.action
200 OK (POST https://developer.apple.com/services-account/XXX/account/listTeams.action)
✔ Using remote iOS credentials (Expo server)
⠋ Linking bundle identifier com.instamenu.instamenuPOST https://developer.apple.com/services-account/v1/bundleIds
[...]

PR #215, opened 2021-01-20
[eas-cli] filter stale profiles

quinlanj:
# Why

Users can sometimes find themselves with a poisoned Apple cache. Once the Apple cache is poisoned, they will always get stale data for the forseeable future. We aren't really sure how this happens, but most users get in this state by going through the adhoc credentials build flow (internal distribution build). Getting into a 'poisoned cache state' is non deterministic, which makes repro difficult.

fixes https://github.com/expo/eas-cli/issues/168

# How

- call an arbitrary method on a profile once we receive it from the Apple API
- if it errors, we filter it out before we return them to the user

# Test Plan
Because this bug is hard to reproduce, I have a branch that will fetch the profiles, ask you to delete a profile from your Developer portal, then continue to filter out the stale profile. You can 'repro' the bug and see the fix in action on the [@quin/reproBrentBug](https://github.com/expo/eas-cli/tree/%40quin/reproBrentBug) branch

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/bundleId.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/bundleId.ts
@@ -1,6 +1,6 @@
 import { BundleId, Profile, RequestContext } from '@expo/apple-utils';
 
-export async function getProfilesForBundleIdAsync(
+async function getProfilesForBundleIdDangerousAsync(
   context: RequestContext,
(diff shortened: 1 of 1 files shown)


EvanBacon (approved):
We should start tracking these events with analytics. At the very least, add a TODO comment so we can come back to it.

PR #218, opened 2021-01-21
[eas-cli] add tracking analytics for stale profile

quinlanj:
# Why

- This would log the event to our hypothetical Analytics, but it's missing the `trackingCtx`.
- I noticed in the `build` command, we are wrapping calls with a `withAnalytics` method and then passing in a trackingCtx. When an error occurs, we catch it and feed in the `trackingCtx`. 
- In this case, we aren't actually throwing an error, and this method has a variety of callers. It'd be nice if we could import some default tracking context (maybe initialized with some basic info at the start of every command). The alternative would be to plumb down a `trackingCtx` in the method parameter .

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/bundleId.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/bundleId.ts
@@ -1,5 +1,8 @@
 import { BundleId, Profile, RequestContext } from '@expo/apple-utils';
 
+import { AnalyticsEvent } from '../../../build/types';
+import Analytics from '../../../build/utils/analytics';
+
 async function getProfilesForBundleIdDangerousAsync(
   context: RequestContext,
   bundleIdentifier: string
@@ -30,7 +33,7 @@ export async function getProfilesForBundleIdAsync(
           e.name === 'UnexpectedAppleResponse' &&
           e.message.includes('The specified resource does not exist - There is no resource of type')
         ) {
-          // TODO: add tracking analytics here
+          Analytics.logEvent(AnalyticsEvent.GATHER_CREDENTIALS_FAIL, { reason: e.message });
           return;
         }
         throw e;


EvanBacon (approved):
Let's add some info about the request as well.

packages/eas-cli/src/credentials/ios/appstore/bundleId.ts
```diff
@@ -30,7 +33,7 @@ export async function getProfilesForBundleIdAsync(
           e.name === 'UnexpectedAppleResponse' &&
           e.message.includes('The specified resource does not exist - There is no resource of type')
         ) {
-          // TODO: add tracking analytics here
+          Analytics.logEvent(AnalyticsEvent.GATHER_CREDENTIALS_FAIL, { reason: e.message });
```
wkozyra95, line 36: Event from build flow are specifically tied to build flow (according to docs that @animeeples prepared), so we shouldn't use it in other contexts, because it will skew the data.

The second point is those analytics here are not implemented yet, `../../../build/utils/analytics` is empty implementation.  There is PR with implementation https://github.com/expo/eas-cli/pull/162 , but we decided to hold on with that for now, some privacy policy changes and docs might be required + info to the user that we track some stuff.

brentvatne: oh my bad, i didn't realize that was the exact context this event was meant for. i'll chat more with @animeeples

quinlanj: In this particular use case, where:
- we arent throwing an error
- we want to track that this 'event' occurs - it comes from a variety of callers, so knowing which command it came from would be helpful

It'd be nice if we could import some default tracking context (maybe initialized with some basic info at the start of every command). For example, if we could do something like:

```
import Analytics from 'analytics';

Analytics.logEvent(CREDENTIALS_EVENT, Analytics.getContext());
```

PR #223, opened 2021-02-02
Refactor log module so that weird __esModule mocks are not needed

wschurman:
# Why

Mocking `log` and verifying results is a common way to unit test a CLI. Until this PR, the only way to mock the `log` module was to do the following due to the way it was set up (custom properties on a function instance, which was copied from expo-cli):
```typescript
const mockLog = {	
  __esModule: true, // this property makes it work	
  default: jest.fn(toLog => toLog),	
};	
(mockLog.default as any).newLine = jest.fn();	
jest.mock('../../../../log', () => mockLog);
```

This PR makes it a better default export (class) for cleaner mocking. I may do a similar refactor for `expo-cli` where this was copied from.

# How

Two commits:
1. Refactor the `Log` module to be a class with all static members. Update all callsites to use the class. `yarn tsc` to verify.
2. Update the tests to use simpler mocks.

# Test Plan

`yarn tsc`
`yarn test`

### DIFF
--- a/packages/eas-cli/src/build/android/applicationId.ts
+++ b/packages/eas-cli/src/build/android/applicationId.ts
@@ -5,7 +5,7 @@ import chalk from 'chalk';
 import fs from 'fs-extra';
 import path from 'path';
 
-import log from '../../log';
+import Log from '../../log';
 import {
   ensureAppIdentifierIsDefinedAsync,
   getAndroidApplicationIdAsync,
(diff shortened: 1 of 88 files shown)


packages/eas-cli/src/build/android/applicationId.ts
```diff
@@ -5,7 +5,7 @@ import chalk from 'chalk';
 import fs from 'fs-extra';
 import path from 'path';
 
-import log from '../../log';
+import Log from '../../log';
```
EvanBacon, line 8: I dont have a strong preference but I think lowercase log is more common in CLI related tools, also this was lifted from expo-cli so it might make sense to keep them consistent (even if that means updating it over there).

wschurman: I think it depends on what is being exported. I made it a class so exporting uppercase made the most sense to me. In the old version it exported a function so lowercase made sense.

As a completely different approach to this PR I'd be okay removing the default export from the module entirely and just have all exports be named exports, but for some reason I like the android style `Log.level` a little bit better than naked `warn` calls.

wschurman: We could also call the module something like `Stdout`.

quinlanj (approved):
lgtm, thanks a bunch. I was the one who originally wrote the weird `__esModule` mocks stuff, and this pr makes it considerably easier to test 😅

PR #237, opened 2021-02-16
[eas-cli] add default bundler for branch:publish

jkhales:
# Why

Currently in order to publish we have to go through a 2 step process:
 1. manually create the bundles with `expo-cli export --experimental-bundle --force`.
 2. publish with `eas branch:publish`

The commands are seperate so that a developer can select their preferred bundler. However, we also want the `publish` command to be a short and simple as possible. This PR makes usage of `expo-cli` the default behavior: `eas branch:publish` will automatically check to see if `expo-cli` is installed and then run step (1). 

We use the same expo-cli verison as yarn.

If expo-cli is not installed, the option to install it will be offered via the wrapper in `node_modules/.bin/cli.js`.

If the developer wishes to avoid this behavior and handle bundling on their own, they can set the new `--skip-bundler` flag.

# How

Check in the package.json for `expo-cli` and run it with `@expo/spawn-async`

The output of the `expo-cli` bundling process is prefixed with `[expo-cli]` and passed through:
Screen Shot 2021-02-16 at 10 27 09 AM

CC: @jon @brentvatne 
# Test Plan

confirmed bundles are build and published.
confirmed --experimental-bundle flag prevents expo-cli from being used.

### DIFF
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -12,7 +12,12 @@ import {
   getBranchByNameAsync,
   getProjectAccountNameAsync,
 } from '../../project/projectUtils';
-import { buildUpdateInfoGroupAsync, collectAssets, uploadAssetsAsync } from '../../project/publish';
+import {
+  buildBundlesAsync,
+  buildUpdateInfoGroupAsync,
+  collectAssets,
+  uploadAssetsAsync,
+} from '../../project/publish';
 import { promptAsync } from '../../prompts';
 import { getLastCommitMessageAsync } from '../../utils/git';
 
@@ -37,11 +42,21 @@ export default class BranchPublish extends Command {
       description: `return a json with the edited branch's ID and name.`,
       default: false,
     }),
+    'skip-bundler': flags.boolean({
+      description: `skip running Expo CLI to bundle the app before publishing`,
+      default: false,
+    }),
   };
 
   async run() {
     let {
-      flags: { json: jsonFlag, branch: name, message, 'input-dir': inputDir },
+      flags: {
+        json: jsonFlag,
+        branch: name,
+        message,
+        'input-dir': inputDir,
+        'skip-bundler': skipBundler,
+      },
     } = this.parse(BranchPublish);
 
     const projectDir = await findProjectRootAsync(process.cwd());
@@ -94,6 +109,10 @@ export default class BranchPublish extends Command {
       appId: projectId,
       name: name!,
     });
+
+    if (!skipBundler) {
+      await buildBundlesAsync({ projectDir, inputDir });
+    }
 
     let updateInfoGroup;
     const assetSpinner = ora('Uploading assets...').start();
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -1,5 +1,6 @@
 import { Platform } from '@expo/config';
-import JsonFile from '@expo/json-file';
+import JsonFile, { JSONObject } from '@expo/json-file';
+import spawnAsync from '@expo/spawn-async';
 import Joi from '@hapi/joi';
 import crypto from 'crypto';
 import fs from 'fs';
@@ -10,6 +11,7 @@ import path from 'path';
 import { AssetMetadataStatus, PartialManifestAsset } from '../graphql/generated';
 import { PublishMutation } from '../graphql/mutations/PublishMutation';
 import { PublishQuery } from '../graphql/queries/PublishQuery';
+import Log from '../log';
 import { PresignedPost, uploadWithPresignedPostAsync } from '../uploads';
 
 export const TIMEOUT_LIMIT = 60_000; // 1 minute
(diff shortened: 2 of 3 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -140,11 +143,60 @@ export async function buildUpdateInfoGroupAsync(assets: CollectedAssets): Promis
...
+    !!packageJSON.devDependencies['expo-cli'] || !!packageJSON.dependencies['expo-cli'];
+  if (!isCLIInstalled) {
+    throw new Error(
+      'expo-cli is not installed. If you would like us to handle bundling your app, please install it with: yarn add -D expo-cli'
+    );
+  }
+
+  Log.withTick(`Building bundle with expo-cli...`);
+  const spawnPromise = spawnAsync('yarn', [
+    'expo-cli',
```
fson, line 167:
```suggestion
    'expo',
```
Both commands exists, but `expo` is what people mostly use and what we document, so I think it's less obscure.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -37,11 +42,21 @@ export default class BranchPublish extends Command {
       description: `return a json with the edited branch's ID and name.`,
       default: false,
     }),
+    'custom-bundler': flags.boolean({
```
fson, line 45: `--custom-bundler` sounds like something that takes a name / command as an argument. Maybe `--skip-bundle` or `--skip-export` would be more self descriptive?

packages/eas-cli/src/project/publish.ts
```diff
@@ -140,11 +143,60 @@ export async function buildUpdateInfoGroupAsync(assets: CollectedAssets): Promis
...
 }
 
+export async function buildBundlesAsync({
+  projectDir,
+  inputDir,
+}: {
+  projectDir: string;
+  inputDir: string;
+}) {
+  const packageJSON = await readJSON(path.resolve(projectDir, 'package.json'));
```
fson, line 153:
```suggestion
  const packageJSON = JsonFile.read(path.resolve(projectDir, 'package.json'));
```
Already imported in this file.

packages/eas-cli/src/project/publish.ts
```diff
@@ -140,11 +143,60 @@ export async function buildUpdateInfoGroupAsync(assets: CollectedAssets): Promis
...
+  if (!packageJSON) {
+    throw new Error('Could not locate package.json');
+  }
+  const isCLIInstalled =
+    !!packageJSON.devDependencies['expo-cli'] || !!packageJSON.dependencies['expo-cli'];
+  if (!isCLIInstalled) {
+    throw new Error(
+      'expo-cli is not installed. If you would like us to handle bundling your app, please install it with: yarn add -D expo-cli'
```
fson, line 161:
```suggestion
      'expo-cli is not installed. To bundle your app, install expo-cli, or skip bundling with --skip-bundle'
```

fson: Also note that we currently ask people to install `expo-cli` globally, in the documentation https://docs.expo.io/get-started/installation/#installing-expo-cli

packages/eas-cli/src/commands/branch/publish.ts
fson: Or perhaps don't have this flag, and skip bundling if `--input-dir` is given. This would make it very clear that you're either going to use expo-cli or provide a path to an already bundled app. It would also make it explicit where we look for the custom bundled output.

fson (changes requested):

packages/eas-cli/src/commands/branch/publish.ts
jkhales: the `--skip-bundler` command makes sense to me.

I think we should avoid overloading the `--input-dir`. It seems OK to allow the developer to both specify a custom input dir *and* have us bundle things automatically.

packages/eas-cli/src/project/publish.ts
jkhales: It still seems like a good idea to give them a suggestive error message, even if they have ignored our good advice... 😅

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -37,11 +42,21 @@ export default class BranchPublish extends Command {
       description: `return a json with the edited branch's ID and name.`,
       default: false,
     }),
+    'skip-bundler': flags.boolean({
+      description: `do not run the default expo bundler`,
```
jkhales, line 46: Updated this from `do not overwrite the bundles in the input-dir`

packages/eas-cli/src/project/publish.ts
fson: Sure! The message initially included `yarn add -D expo-cli`, which installs it locally, that's what I meant.

packages/eas-cli/src/commands/branch/publish.ts
fson, line 46:
```suggestion
      description: `skip running Expo CLI to bundle the app before publishing`,
```
Maybe?

packages/eas-cli/src/project/publish.ts
```diff
@@ -140,11 +142,61 @@ export async function buildUpdateInfoGroupAsync(assets: CollectedAssets): Promis
...
+  inputDir: string;
+}) {
+  const packageJSON = JsonFile.read(path.resolve(projectDir, 'package.json'));
+  if (!packageJSON) {
+    throw new Error('Could not locate package.json');
+  }
+  const isCLIInstalled =
+    !!(packageJSON.dependencies as JSONObject)['expo-cli'] ||
+    !!(packageJSON.devDependencies as JSONObject)['expo-cli'];
+  if (!isCLIInstalled) {
```
fson, line 159: Just realized this won't work correctly for people who don't have Expo CLI as a local dependency. (We recommend installing it globally, so most people have it installed globally.)

packages/eas-cli/src/commands/branch/publish.ts
jkhales: great.

packages/eas-cli/src/project/publish.ts
jkhales: oh great catch. Lemme think about how to fix that.

fson (approved):

PR #242, opened 2021-02-19
[eas-cli] rename updateGroup, updateMessage, branchName, channelName

jkhales:
# Why

Following @fson observation about excessively verbose attribute names, we refactored a bunch of the graphQL methods server side.

This PR brings the eas-cli up to speed.

# How

renamed and ran `yarn generate-graphql-code`

# Test Plan

tested that all of the `eas branch:*` and `eas channel:create` commands work.

### DIFF
--- a/packages/eas-cli/src/commands/channel/create.ts
+++ b/packages/eas-cli/src/commands/channel/create.ts
@@ -17,11 +17,11 @@ import { createUpdateBranchOnAppAsync } from '../branch/create';
 
 async function createUpdateChannelOnAppAsync({
   appId,
-  name,
+  channelName,
   branchId,
 }: {
   appId: string;
-  name: string;
+  channelName: string;
   branchId: string;
 }): Promise<UpdateChannel> {
   // Point the new channel at a branch with its same name.
@@ -48,7 +48,7 @@ async function createUpdateChannelOnAppAsync({
         `,
         {
           appId,
-          name,
+          name: channelName,
           branchMapping,
         }
       )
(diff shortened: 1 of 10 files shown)


packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -122,22 +114,22 @@ export default class ChannelCreate extends Command {
     try {
       const existingBranch = await getBranchByNameAsync({
         appId: projectId,
-        branchName: channelName,
```
quinlanj, line 125: branchName and channelName are different things right? shouldn't this be the name of a branch instead of a channel?

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -17,11 +17,11 @@ import { createUpdateBranchOnAppAsync } from '../branch/create';
 
 async function createUpdateChannelOnAppAsync({
   appId,
-  channelName,
+  name,
   branchId,
```
quinlanj, line 21: i think this is fine, but it's less obvious that `name` is `channelName` in `createUpdateChannelOnAppAsync`, compared to the other rename, where it's pretty clear that `name` is `branchName` in `createBranchAsync`

quinlanj (approved):
there's some confusion for me with `branchName` and `channelName` being used in the same spots, but other than that, lgtm

packages/eas-cli/src/commands/channel/create.ts
jkhales: yea that's a good point, Lemme see if it makes sense to switch it back.

jkhales: switched!

PR #248, opened 2021-02-24
[eas-cli] add branch:delete command

jkhales:
# Why

Add `eas branch:delete <name>`

# How

delete the branch name.

Added an additional `no/yes` toggle as this is a sensitive operation.

# Test Plan

created and deleted branches.

Confirmed it doesn't delete a branch which a channel is pointing at.
Confirmed it deletes a branch with/without updates on it.

### DIFF
--- a/packages/eas-cli/src/commands/branch/delete.ts
+++ b/packages/eas-cli/src/commands/branch/delete.ts
@@ -163,12 +163,12 @@ export default class BranchDelete extends Command {
       }
     }
 
-    const { id: deletedBranchId } = await deleteBranchOnAppAsync({
+    const deletionResult = await deleteBranchOnAppAsync({
       branchId,
     });
 
     if (jsonFlag) {
-      Log.log(deletedBranchId);
+      Log.log(JSON.stringify(deletionResult));
     }
 
     Log.withTick(
--- a/packages/eas-cli/src/commands/branch/delete.ts
+++ b/packages/eas-cli/src/commands/branch/delete.ts
@@ -4,7 +4,13 @@ import chalk from 'chalk';
 import gql from 'graphql-tag';
 
 import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
-import { DeleteUpdateBranchResult, Update } from '../../graphql/generated';
+import {
+  DeleteUpdateBranchMutation,
+  DeleteUpdateBranchMutationVariables,
+  DeleteUpdateBranchResult,
+  GetBranchInfoQuery,
+  GetBranchInfoQueryVariables,
+} from '../../graphql/generated';
 import Log from '../../log';
 import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
 import { findProjectRootAsync, getProjectAccountName } from '../../project/projectUtils';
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+      }));
+    }
+
+    const { branchId } = await getBranchInfoAsync({ appId: projectId, name });
+
+    if (!jsonFlag) {
+      Log.addNewLineIfNone();
+      Log.warn(
+        `You are about to permamently delete branch: "${name}" and all of the updates published on it.` +
+          `\nThis action is irreversable.`
```
byCedric, line 156: I'm not sure if we still have it in Expo CLI, but making this text might help on the emphasis?

byCedric (approved):

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+async function getBranchInfoAsync({
+  appId,
+  name,
+}: {
+  appId: string;
+  name: string;
+}): Promise<{ branchId: string }> {
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation<
```
fson, line 23:
```suggestion
      .query<
```

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+  );
+
+  const {
+    app: {
+      byId: {
+        updateBranchByName: { id: branchId },
+      },
+    },
+  } = data;
+  return { branchId };
```
fson, line 68:
```suggestion
  return { branchId: data.app.byId.updateBranchByName.id };
```
Seems less convoluted than destructuring in this case.

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+                name: string;
+                updates: Pick<Update, 'id'>[];
+              };
+            };
+          };
+        },
+        {
+          appId: string;
+          name: string;
+        }
```
fson, line 38:
```suggestion
        ViewBranchQuery,
        ViewBranchQueryVariables
```
`graphql-codegen` will generate these types automatically for you, no need to write and maintain them by hand.

Add this file to the paths of files with GraphQL queries and then re-run `yarn generate-graphql-code`. The config is here:
https://github.com/expo/eas-cli/blob/2e25ed81ca52d9d25afeadd751248d885ed152a9/packages/eas-cli/graphql-codegen.yml#L3-L5

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+  branchId: string;
+}): Promise<DeleteUpdateBranchResult> {
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation<
+        { updateBranch: { deleteUpdateBranch: DeleteUpdateBranchResult } },
+        {
+          branchId: string;
+        }
+      >(
```
fson, line 83:
```suggestion
      .mutation<DeleteUpdateBranchMutation, DeleteUpdateBranchMutationVariables>(
```

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+  static args = [
+    {
+      name: 'name',
+      required: false,
+      description: 'Name of the branch to delete',
+    },
+  ];
+  static flags = {
+    json: flags.boolean({
+      description: `return a json with the edited branch's ID and name.`,
```
fson, line 115:
```suggestion
      description: `return JSON with the edited branch's ID and name.`,
```

fson (changes requested):

packages/eas-cli/src/commands/branch/delete.ts
jkhales: Nice!
I added the whole commands sub-directory:
`'src/commands/**/*.ts'`

Will go through and convert some more types in a follow up PR.

packages/eas-cli/src/commands/branch/delete.ts
jkhales: That's a fair point, for now will keep like this for consistency as other parts of the CLI output like this.

There is this general issue here as well: https://linear.app/expo/issue/ENG-205/standardize-eas-cli-output

packages/eas-cli/src/commands/channel/create.ts
```diff
@@ -36,7 +36,7 @@ async function createUpdateChannelOnAppAsync({
         { appId: string; name: string; branchMapping: string }
       >(
         gql`
-          mutation CreateUpdateChannelForApp($appId: ID!, $name: String!, $branchMapping: String!) {
+          mutation CreateUpdateChannelOnApp($appId: ID!, $name: String!, $branchMapping: String!) {
```
jkhales, line 39: fix namespace conflict

fson (approved):

PR #250, opened 2021-02-25
[eas-cli] Fix json output for branch and channel commands

byCedric:
# Why

Logging JS objects/arrays will return a format provided by `util`, which isn't JSON. This fixes that for:

- `branch:publish`
- `branch:rename`
- `branch:view`
- `channel:create`

# How

- Added `JSON.stringify()` around the content

# Test Plan

You can run the command, with `--json` flag, and parse it with `jq`. E.g.:

- `$ eas channel:create <name> --json | jq '.'`

This should not throw "invalid token" errors.

### DIFF
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -142,7 +142,7 @@ export default class BranchPublish extends Command {
     }
 
     if (jsonFlag) {
-      Log.log(newUpdateGroup);
+      Log.log(JSON.stringify(newUpdateGroup));
     } else {
       const outputMessage = new Table({
         wordWrap: true,
--- a/packages/eas-cli/src/commands/branch/rename.ts
+++ b/packages/eas-cli/src/commands/branch/rename.ts
@@ -119,7 +119,7 @@ export default class BranchRename extends Command {
     });
 
     if (jsonFlag) {
-      Log.log(editedBranch);
+      Log.log(JSON.stringify(editedBranch));
       return;
     }
 
--- a/packages/eas-cli/src/commands/branch/view.ts
+++ b/packages/eas-cli/src/commands/branch/view.ts
@@ -147,7 +147,7 @@ export default class BranchView extends Command {
     );
 
     if (jsonFlag) {
-      Log.log({ ...UpdateBranch, updates });
+      Log.log(JSON.stringify({ ...UpdateBranch, updates }));
       return;
     }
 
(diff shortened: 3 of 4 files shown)


packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -22,7 +22,7 @@ import { promptAsync } from '../../prompts';
 import { getLastCommitMessageAsync } from '../../utils/git';
 
 export default class BranchPublish extends Command {
-  static hidden = true;
+  static hidden = false;
```
jkhales, line 25: keep as true

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -118,11 +118,13 @@ export default class BranchPublish extends Command {
     const assetSpinner = ora('Uploading assets...').start();
     try {
       const assets = collectAssets(inputDir!);
+      console.log('[debug] assets -', 'uploading these assets:', JSON.stringify(assets, null, 2));
```
jkhales, line 121:
```suggestion
      Log.log('[debug] assets -', 'uploading these assets:', JSON.stringify(assets, null, 2));
```

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -118,11 +118,13 @@ export default class BranchPublish extends Command {
...
       await uploadAssetsAsync(assets);
       updateInfoGroup = await buildUpdateInfoGroupAsync(assets);
       assetSpinner.succeed('Uploaded assets!');
     } catch (e) {
       assetSpinner.fail('Failed to upload assets');
+      console.log('[debug] assets -', 'failed because of', e);
```
jkhales, line 127:
```suggestion
      Log.log('[debug] assets -', 'failed because of', e);
```

jkhales: Also, maybe just put this error message into the assetSpinner.fail message on the line above?

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Actually did you mean to delete these?

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Actually did you mean to delete these?

jkhales (approved):
Thank you!

packages/eas-cli/src/commands/branch/publish.ts
byCedric: oh my bad, didn't intend to push this. Thanks for catching it!

packages/eas-cli/src/commands/branch/publish.ts
byCedric: yeah, my bad, this shouldn't have been committed. Will pull this out!

PR #253, opened 2021-02-26
[eas-cli] add --republish and  --group flags to branch:publish command.

jkhales:
This adds a `--republish`/`--group`  flag to the `eas branch:publish` command.

These commands add a duplicate of a previous update group to the head of a given branch.
The `--republish` (boolean flag) is convenient and interactive, the `--group` (string flag) command is more flexible.

* `--republish` provides an interactive list of previous update groups on the *given* branch that the user may select from is shown:

Screen Shot 2021-03-08 at 8 44 34 AM

The `--group` flag allows a little more flexibility in that the user may copy the update group from *any* branch.

# testing

confirmed the commands work on a test project.

CC @jonsamp


packages/eas-cli/src/commands/branch/list.ts
```diff
@@ -18,6 +18,59 @@ import { findProjectRootAsync, getProjectFullNameAsync } from '../../project/pro
...
+        {
+          fullName,
+          limit: BRANCHES_LIMIT,
+        }
+      )
+      .toPromise()
+  );
+
+  return data?.app?.byFullName.updateBranches ?? [];
+}
```
jkhales, line 73: this is unchanged, just pulled out of the class below so that it can be exported.

packages/eas-cli/src/graphql/mutations/PublishMutation.ts
```diff
@@ -31,7 +31,7 @@ const PublishMutation = {
 
   async publishUpdateGroupAsync(
     publishUpdateGroupInput: PublishUpdateGroupInput
-  ): Promise<Pick<Update, 'id' | 'group'>> {
+  ): Promise<Pick<Update, 'group'>> {
```
jkhales, line 34: We should encourage the use of `group` instead of `id` when dealing with updates.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -90,6 +143,66 @@ export default class BranchPublish extends Command {
...
+      const assetSpinner = ora('Uploading assets...').start();
+      try {
+        const assets = collectAssets(inputDir!);
+        await uploadAssetsAsync(assets);
+        updateInfoGroup = await buildUpdateInfoGroupAsync(assets);
+        assetSpinner.succeed('Uploaded assets!');
+      } catch (e) {
+        assetSpinner.fail('Failed to upload assets');
+        throw e;
+      }
```
jkhales, line 204: this is section deleted below moved into an `else` clause.

packages/eas-cli/src/commands/branch/list.ts
```diff
@@ -18,6 +18,59 @@ import { findProjectRootAsync, getProjectFullNameAsync } from '../../project/pro
 import { getActorDisplayName } from '../../user/actions';
 
 const BRANCHES_LIMIT = 10_000;
+export async function listBranchesAsync({
```
fson, line 21:
```suggestion

async function listBranchesAsync({
```
I think this was unintentionally exported from this modules?

packages/eas-cli/src/commands/branch/list.ts
```diff
@@ -18,6 +18,59 @@ import { findProjectRootAsync, getProjectFullNameAsync } from '../../project/pro
...
+  fullName,
+}: {
+  fullName: string;
+}): Promise<
+  (Pick<UpdateBranch, 'id' | 'name'> & {
+    updates: (Pick<Update, 'id' | 'updatedAt' | 'message'> & {
+      actor?: Maybe<Pick<User, 'username' | 'id'> | Pick<Robot, 'firstName' | 'id'>>;
+    })[];
+  })[]
```
fson, line 30: Not sure what to make of this type. It seems like something that's been copy pasted from the generated types and it's bound to change whenever the query below is changed.

I believe TypeScript could actually infer this return type without having it written out here. I know we have a convention to explicitly write the return types in functions, but I think in this case it would be better / more maintainable to let TypeScript infer the return type.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -1,51 +1,100 @@
 import { getConfig } from '@expo/config';
 import { Command, flags } from '@oclif/command';
+import { uniqBy } from '@oclif/plugin-help/lib/util';
```
fson, line 3: Not sure why this is imported from `@oclif/plugin-help`, maybe this was added automatically?

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -1,51 +1,100 @@
...
 import { getLastCommitMessageAsync } from '../../utils/git';
+import { viewUpdateBranchAsync } from './view';
+
+type PublishPlatforms = 'android' | 'ios';
 
+export async function getUpdateGroupAsync({
+  group,
+}: RootQueryUpdatesByGroupArgs): Promise<
+  Pick<Update, 'group' | 'runtimeVersion' | 'manifestFragment' | 'platform' | 'message'>[]
+> {
```
fson, line 41:
```suggestion
async function getUpdateGroupAsync({
  group,
}: RootQueryUpdatesByGroupArgs): GetUpdateGroupAsyncQuery
['updatesByGroup'] {
```

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -176,3 +270,14 @@ export default class BranchPublish extends Command {
     }
   }
 }
+
+function formatUpdateTitle(
+  update: Pick<Update, 'message' | 'createdAt' | 'runtimeVersion'> & {
+    actor?: Maybe<Pick<User, 'firstName' | 'id'> | Pick<Robot, 'firstName' | 'id'>>;
```
fson, line 276:
```suggestion
    actor?: { firstName: string | null };
```
This type covers all the fields used by this function.

fson (changes requested):

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Yea good point.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: It would be really convenient, but I don't think we can use types like this: `GetUpdateGroupAsyncQuery['updatesByGroup']`

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -1,51 +1,99 @@
...
+import { promptAsync, selectAsync } from '../../prompts';
 import { getLastCommitMessageAsync } from '../../utils/git';
+import { viewUpdateBranchAsync } from './view';
+
+type PublishPlatforms = 'android' | 'ios';
 
+async function getUpdateGroupAsync({
+  group,
+}: RootQueryUpdatesByGroupArgs): GetUpdateGroupAsyncQuery
+['updatesByGroup'] {
```
fson, line 40:
```suggestion
async function getUpdateGroupAsync({
  group,
}: RootQueryUpdatesByGroupArgs): Promise<GetUpdateGroupAsyncQuery['updatesByGroup']> {
```
This should work.

packages/eas-cli/src/commands/branch/list.ts
fson: Screen Shot 2021-03-09 at 17 16 17

If you remove the return type, it's still inferred correctly.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: oh wow, nice!

packages/eas-cli/src/commands/branch/list.ts
jkhales: Good catch, I am using it now: https://github.com/expo/eas-cli/pull/253/files#diff-2fb981f3a86534f69bea6c5560ce229c420628a9969ee0a07169d7623a21b1e0R139

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -83,12 +135,79 @@ export default class BranchPublish extends Command {
...
+        branches.map(branch => {
+          return {
+            title: `${branch.name} ${chalk.grey(
+              `- current update: ${formatUpdate(branch.updates[0])}`
+            )}`,
+            value: branch.name,
+          };
+        })
+      );
+    }
```
jkhales, line 151: This prompt looks like:
Screen Shot 2021-03-09 at 8 13 47 AM

fson (approved):

PR #261, opened 2021-03-02
[eas-cli] Add update delete command

byCedric:
# Why

See [ENG-251](https://linear.app/expo/issue/ENG-251/implement-updatedelete)

# How

- Added command `update:delete`
- Used mutation `update.deleteUpdate`
- Added additional info why users might want `branch:republish` instead (as mentionedin Linear)
- Added `--force` flag, mostly for JSON output but also a way to delete something with prior confirmation.

# Examples

**Running `$ eas update:delete` and aborting**
Screenshot 2021-03-02 at 13 42 02

**Running `$ eas update:delete abc-123 --json`**
Screenshot 2021-03-02 at 13 42 24

**Running `$ eas update:delete` and confirming**
Screenshot 2021-03-02 at 13 44 39

# Test Plan

- `$ eas update:delete <id>`
- `$ eas update:delete <id> --json`
- `$ eas update:delete <id> --json --force`

### DIFF
--- a/packages/eas-cli/src/commands/update/delete.ts
+++ b/packages/eas-cli/src/commands/update/delete.ts
@@ -93,7 +93,7 @@ export default class UpdateDelete extends Command {
       chalk`⚠️  This will delete all assets from the update, {red this action is IRREVERSIBLE}.`
     );
     Log.log(
-      chalk`Users with this update installed will still use this update, {bold branch:republish} is likely what you want.`
+      chalk`Users with the update installed will continue to use this update, {bold branch:republish} is likely what you want.`
     );
 
     if (!forceFlag) {
--- a/packages/eas-cli/src/commands/update/delete.ts
+++ b/packages/eas-cli/src/commands/update/delete.ts
@@ -1,0 +1,123 @@
+import { getConfig } from '@expo/config';
+import { Command, flags } from '@oclif/command';
+import chalk from 'chalk';
+import gql from 'graphql-tag';
+
+import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
+import { Update } from '../../graphql/generated';
+import Log from '../../log';
+import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
+import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
+import { promptAsync } from '../../prompts';
+
+async function deleteUpdateByIdAsync({
+  updateId,
+}: {
+  updateId: Update['id'];
+}): Promise<Pick<Update, 'id'>> {
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .mutation(
+        gql`
+          mutation DeleteUpdateById($updateId: ID!) {
+            update {
+              deleteUpdate(updateId: $updateId) {
+                id
+              }
+            }
+          }
+        `,
+        { updateId }
+      )
+      .toPromise()
+  );
+  return data.update.deleteUpdate;
+}
+
+export default class UpdateDelete extends Command {
+  static hidden = true;
+  static description = 'Delete an update on the current project';
+
+  static args = [
+    {
+      name: 'updateId',
+      required: false,
+      description: 'ID of the update to delete',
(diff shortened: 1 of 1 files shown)


jkhales (approved):
Nice! ✨ 

One issue that I'm thinking about here is that we `delete` by `updateId` but `republish` by `updateGroup`.

Would be curious to get @jonsamp 's view on this.

jonsamp:
> Nice! ✨
> 
> One issue that I'm thinking about here is that we `delete` by `updateId` but `republish` by `updateGroup`.
> 
> Would be curious to get @jonsamp 's view on this.

I think making mutations by update groups is the way to go for all things, and if you want to mutate specific updates inside of an update group, you'd need to specify more arguments.

What are the steps before this? How can I get a `updateId` and how would I (as a developer) have come across this?

My assumption is that a developer creates an update with a secret key or something and they want to delete it. So, since they used `eas branch:publish`, they now have an update group with multiple platform specific updates with that secret. So, then, they'd run  `eas update:list` (or `eas branch:view ...`), get the most recent update group, then say `eas update:delete` (or similar). Also, maybe what I am describing is just another command and this is a more specific command. We could have both:

- The ability to delete an individual update (less common, and should be harder)
- The ability to delete an update group (more common, and should be easier)

Overall, I like the UX of this command though. Nice work!

byCedric:
Alright, so atm this command deletes a _single update_. We can switch it by deleting an update group instead, so all `update:*` commands are querying and mutating the update group?

jonsamp:
Okay, so we're going to need two things:
- The ability to delete and update group:
  - The command could be something like `eas branch:deleteUpdateGroup [update-group-id]`.
- The ability to delete an individual update: 
  - The command could be what this PR specifies `eas update:delete [update-id]`.

We could also make the user-facing word `update` really mean an update group. In that case we could have:
- To delete an update group, would be: `eas update:delete [update-group-id]`.
- To delete an individual update, would be: `eas update:delete --individual [update-id]` (or something like that).

From looking at these options, I'm leaning toward this second option, since it allows both abilities and makes deleting a single update harder than deleting an update group. What do you both think? This might have some implications for other `eas update:*` commands? @byCedric @jkhales

jkhales:
> The ability to delete and update group:

I strongly agree with the second. I can't think of a net positive situation of letting a user deal with individual updates.

jonsamp:
In that case, we need to:
- Change this PR to affect an update group
- Make it when you run `eas branch:view [...]`, that it shows an "updates"  header in the table.
- Add an `--individual` flag to allow deleting an individual update.
- Update any other `eas update:*` commands to affect update groups, and also give most of them an `--individual` flag.

jkhales:
> * Make it when you run `eas branch:view [...]`, that it shows an "updates"  header in the table.

What did you want to change about this?

Screen Shot 2021-03-09 at 8 32 55 AM

jonsamp:
> > * Make it when you run `eas branch:view [...]`, that it shows an "updates"  header in the table.
> 
> What did you want to change about this?
> 
> Screen Shot 2021-03-09 at 8 32 55 AM

Screen Shot 2021-03-09 at 11 37 23 AM

jkhales:
So you want to refer to the `group` id as the `update` id? 

🤔 seems potentially buggy since we already have an update id.

jonsamp:
> So you want to refer to the `group` id as the `update` id?
> 
> 🤔 seems potentially buggy since we already have an update id.

We need two user-facing indentifiers. I was thinking our internal "update group" would be "update" to the user, and the our "update" (which is platform-specific) would be "individual update" to the user. This is just off the top of my head.

jkhales:
> > So you want to refer to the `group` id as the `update` id?
> > 🤔 seems potentially buggy since we already have an update id.
> 
> We need two user-facing indentifiers. I was thinking our internal "update group" would be "update" to the user, and the our "update" (which is platform-specific) would be "individual update" to the user. This is just off the top of my head.

We should decide on this and refactor the graphQL calls before release if so.

jkhales:
closing this as it was replaced by #332

PR #289, opened 2021-03-23
[eas-cli] add support for rollouts to channel:view command

jkhales:
# Why

Add support for rollouts

# How

If branch mapping data has a length of 2, display both branches and the rollout percentage.

# Test Plan

Tested locally:

Screen Shot 2021-03-23 at 1 28 03 PM

### DIFF
--- a/packages/eas-cli/src/commands/channel/view.ts
+++ b/packages/eas-cli/src/commands/channel/view.ts
@@ -14,6 +14,68 @@ import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
 import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
 
+type BranchMapping = {
+  version: number;
+  data: {
+    branchId: string;
+    branchMappingLogic: {
+      operand: number;
+      clientKey: string;
+      branchMappingOperator: string;
+    } & string;
+  }[];
+};
+
+/**
+ * Get the branch mapping and determine whether it is a rollout.
+ * Ensure that the branch mapping is properly formatted.
+ */
+function getBranchMapping(
+  getChannelByNameForAppQuery: GetChannelByNameForAppQuery
+): { branchMapping: BranchMapping; isRollout: boolean; rolloutPercent?: number } {
+  const branchMappingString =
+    getChannelByNameForAppQuery.app?.byId.updateChannelByName.branchMapping;
+  if (!branchMappingString) {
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/channel/view.ts
```diff
@@ -14,6 +14,68 @@ import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
...
+      break;
+    case 2:
+      if (branchMapping.data[0].branchMappingLogic.clientKey !== 'rolloutToken') {
+        throw new Error('Client key of initial branch mapping must be "rolloutToken"');
+      }
+      if (branchMapping.data[0].branchMappingLogic.branchMappingOperator !== 'hash_lt') {
+        throw new Error('Branch mapping operator of initial branch mapping must be "hash_lt"');
+      }
+      if (!rolloutPercent) {
+        throw new Error('"rolloutPercent" must be defined.');
```
byCedric, line 66: Maybe it's good to mention `"Branch mapping"` in this error as well. Whenever it pops up, it  should be clear where the `rolloutPercent` is missing without checking the full stack trace 😄

packages/eas-cli/src/commands/channel/view.ts
```diff
@@ -124,17 +187,42 @@ export default class ChannelView extends Command {
...
+      ],
       wordWrap: true,
     });
 
-    for (const branch of channel.updateBranches) {
-      // TODO (cedric): refactor when multiple branches per channel are available
-      const update = branch.updates[0];
+    for (const index in branchMapping.data) {
+      if (parseInt(index, 10) > 1) {
+        throw new Error('Branch Mapping data must have lenght less than or equal to 2.');
```
byCedric, line 208: I'm a bit concerned about this "duplicate" validation logic, compared to the other validation in `getBranchMapping`. If this needs to ship asap, maybe add a note that this is similar to that file? Or preferably, if we have time, maybe export a validation method to validate an arbitrary object from somewhere (like utils?).

byCedric (approved):
Overall looks good, added 2 comments but nothing majoy! ❤️

packages/eas-cli/src/commands/channel/view.ts
jkhales: I was thinking that the validation in `getBranchMapping` could expand with this still being restricted to just 2. 

Like what if we did more complex branch mappings, I would want the developer to realize that this output loop was only capable of handling the case of `n=2`.  In that case we may want to keep this loop as is, but write a different function for printing `n>2`.

PR #299, opened 2021-04-01
[eas-cli] configure provisioning profile using new api

quinlanj:
# Why
This is part of the effort to move all credentials apiv2 calls to our graphql api. This PR allows you to create a provisioning profile using the new Graphql Api. This will eventually replace the old `ConfigureProvisioningProfile` here:  https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/credentials/ios/actions/ConfigureProvisioningProfile.ts

# Notable Changes
- Throw a specialized error if provisioning profile on Expo servers isn't found on Apple Developer portal. We can recover from this particular error in `SetupProvisioningProfile` later.

# Test Plan

- [x] New tests pass

### DIFF
--- a/packages/eas-cli/src/credentials/ios/actions/new/ConfigureProvisioningProfile.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/new/ConfigureProvisioningProfile.ts
@@ -1,0 +1,123 @@
+import assert from 'assert';
+import chalk from 'chalk';
+import ora from 'ora';
+
+import {
+  AppleDistributionCertificateFragment,
+  AppleProvisioningProfileFragment,
+} from '../../../../graphql/generated';
+import Log from '../../../../log';
+import { Action, CredentialsManager } from '../../../CredentialsManager';
+import { Context } from '../../../context';
+import { AppLookupParams } from '../../api/GraphqlClient';
+import { AppleProvisioningProfileMutationResult } from '../../api/graphql/mutations/AppleProvisioningProfileMutation';
+import { ProvisioningProfileStoreInfo } from '../../appstore/Credentials.types';
+import { AuthCtx } from '../../appstore/authenticate';
+import {
+  MissingCredentialsNonInteractiveError,
+  ProvisioningProfileNotFoundOnAppleServersError,
+} from '../../errors';
+
+export class ConfigureProvisioningProfile implements Action {
+  private _configuredProvisioningProfile?: AppleProvisioningProfileMutationResult;
+  constructor(
+    private app: AppLookupParams,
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/credentials/ios/actions/new/ConfigureProvisioningProfile.ts
```diff
@@ -0,0 +1,123 @@
...
+      );
+      const [matchingProfile] = profilesFromApple.filter(appleInfo =>
+        developerPortalIdentifier
+          ? appleInfo.provisioningProfileId === developerPortalIdentifier
+          : appleInfo.provisioningProfile === provisioningProfile
+      );
+      if (!matchingProfile) {
+        throw new ProvisioningProfileNotFoundOnAppleServersError(
+          `Profile ${
+            developerPortalIdentifier ? developerPortalIdentifier + ' ' : null
```
wkozyra95, line 61:
```suggestion
            developerPortalIdentifier ? `${developerPortalIdentifier} ` : null
```

packages/eas-cli/src/credentials/ios/actions/new/ConfigureProvisioningProfile.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (ctx.appStore.authCtx) {
+      const profilesFromApple = await ctx.appStore.listProvisioningProfilesAsync(
+        this.app.bundleIdentifier
+      );
+      const [matchingProfile] = profilesFromApple.filter(appleInfo =>
+        developerPortalIdentifier
+          ? appleInfo.provisioningProfileId === developerPortalIdentifier
+          : appleInfo.provisioningProfile === provisioningProfile
+      );
+      if (!matchingProfile) {
```
wkozyra95, line 58: Shouldn't we just create it at this point if we already logged in to apple?

wkozyra95 (approved):

packages/eas-cli/src/credentials/ios/actions/new/ConfigureProvisioningProfile.ts
quinlanj: I've changed this so that `runAsync` returns `null` in a recoverable situation. SetupProvisioningProfile will call ConfigureProvisioningProfile, and if it returns `null`, we will try to create a new Provisioning Profile. Example of this are if the current Provisioning Profile doesnt exist on Apple Portal anymore. 

In a non recoverable situation, we throw an error and SetupProvisioningProfile will fail. Examples of this are if we try to run in non interactive mode, and we don't have all the build credentials setup.

PR #300, opened 2021-04-01
[eas-cli] use flags for secrets:create and rename target to scope

FiberJW:
# Checklist

# Why

@brentvatne had really good feedback stating that the CLI command should take flags rather than positional args for better UX.

Also Closes ENG-758

# How

- used the flags feature in oclif for user input
- renamed target to scope since that is simpler to understand
- reordered secrets:list table columns to show ID before the timestamp

# Test Plan

Screen Shot 2021-03-31 at 23 59 27


dsokal (approved):

brentvatne (approved):
🎉

PR #304, opened 2021-04-01
[eas-cli] Add branch mapping to channel list command

byCedric:
# Checklist

# Why

This adds the branch mapping description to the `channel:list` commands. As JJ mentioned, I reused the existing code from #289. I don't fully like the code in it's current shape, there is some copied code and a really hard-to-read/use type for `updates`. If it's possible, we should probably validate the `branchMapping` in a shared method, and convert it to something a bit more useful to display the rollout info.

Currently it's displayed like this:

Screenshot 2021-04-01 at 14 28 50

> Note the `\n`, omitting this causes the table to be extremely wide and break. We might need to truncate the `message` column too.

# How

- Reused code from #289
- Iterate over the channels and determine how to present the branch mapping
- Fetch the update from the branches in the same process

# Test Plan

- `$ eas channel:list` (with a rollout channel)

### DIFF
--- a/packages/eas-cli/src/commands/channel/list.ts
+++ b/packages/eas-cli/src/commands/channel/list.ts
@@ -6,14 +6,77 @@ import gql from 'graphql-tag';
 
 import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
 import {
+  Actor,
   GetAllChannelsForAppQuery,
   GetAllChannelsForAppQueryVariables,
+  Update,
 } from '../../graphql/generated';
 import Log from '../../log';
 import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
 import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
 
 const CHANNEL_LIMIT = 10_000;
+
+// TODO(cedric): refactor and reuse original code from channel view commands
+type BranchMapping = {
+  version: number;
+  data: {
+    branchId: string;
+    branchMappingLogic: {
+      operand: number;
+      clientKey: string;
+      branchMappingOperator: string;
+    } & string;
+  }[];
+};
+
+/**
+ * Get the branch mapping and determine whether it is a rollout.
+ * Ensure that the branch mapping is properly formatted.
+ */
+function getBranchMapping(
+  branchMappingString: string
+): { branchMapping: BranchMapping; isRollout: boolean; rolloutPercent?: number } {
+  if (!branchMappingString) {
+    throw new Error('Missing branch mapping.');
+  }
+
+  const branchMapping: BranchMapping = JSON.parse(branchMappingString);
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -108,13 +173,41 @@ export default class ChannelList extends Command {
...
+          rolloutMapping.push(branch.name);
+        } else {
+          rolloutMapping.push(
+            parseInt(index, 10) === 0
+              ? `${branch.name} (${rolloutPercent! * 100}%)`
+              : `${branch.name} (${(1 - rolloutPercent!) * 100}%)`
+          );
+        }
+
+        update = branch.updates[0];
```
byCedric, line 205: @jkhales is this correct?

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -6,15 +6,78 @@ import gql from 'graphql-tag';
...
   GetAllChannelsForAppQueryVariables,
+  Update,
 } from '../../graphql/generated';
 import Log from '../../log';
 import { ensureProjectExistsAsync } from '../../project/ensureProjectExists';
 import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
 
 const CHANNEL_LIMIT = 10_000;
 
+// TODO(cedric): refactor and reuse original code from channel view commands
```
jkhales, line 20: delete?

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -6,15 +6,78 @@ import gql from 'graphql-tag';
...
+type BranchMapping = {
```
jkhales, line 21: import this from `./view.ts`

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -6,15 +6,78 @@ import gql from 'graphql-tag';
...
+      branchMappingOperator: string;
+    } & string;
+  }[];
+};
+
+/**
+ * Get the branch mapping and determine whether it is a rollout.
+ * Ensure that the branch mapping is properly formatted.
+ */
+function getBranchMapping(
```
jkhales, line 37: import this from `./view.ts`

packages/eas-cli/src/commands/channel/list.ts
jkhales: Seems like it should be fine, if no update we just get undefined/blank in the output.

jkhales (changes requested):
Would it be ok to change the column names to match the more informative ones here?
image

Also, I liked your suggestion of displaying an update's runtime version/ possibly filtering by a `--runtime-version` flag. Can we put that in here?

jkhales:
Closing this as it was replaced by: https://github.com/expo/eas-cli/pull/331

PR #315, opened 2021-04-08
[eas-cli] make secrets commands follow styleguide

FiberJW:
# Checklist

# Why

I was recently made aware that we have a styleguide for CLI commands, so this PR makes the secrets command "topic" singular.

# How

- renamed `secrets` folder to `secret`
- changed delete ID argument to a flag

# Test Plan

Running `easd secret:<command>` should succeed.


dsokal (approved):

packages/eas-cli/src/commands/secret/create.ts
```diff
@@ -34,12 +34,16 @@ export default class EnvironmentSecretCreate extends Command {
     value: flags.string({
       description: 'Value of the secret',
     }),
+    json: flags.boolean({
```
wkozyra95, line 37: I'm not sure how useful that option would be, if there is some plain text output printed to the terminal.

wkozyra95 (approved):

PR #320, opened 2021-04-09
[eas-cli] upgrade @urql/core and add retry

jkhales:
# Checklist

# Why

Retry graphQL calls if they fail due to network or transient server errors.

# How

Added `@urql/exchange-retry` package and upgraded `@urql/core` to latest version (strictly speaking could have gotten away with upgrading to `1.14.0`).

Following their suggested upgrade process, ran `npx yarn-deduplicate` in root dir to: " ensure that your package manager hasn't accidentally duplicated the @urql/core package."

Read through the changelog for `@urql/core`:
https://github.com/FormidableLabs/urql/blob/main/packages/core/CHANGELOG.md

Found no relevant changes.

# Test Plan

Tested manually by looking at the `ensureLoggedInAsync` network calls pointed at a local server.

1. confirmed the command went smoothly, when the server was running.
2. confirmed there were 2 retries before the command failed when the server was off. (added a temporary console.log for observability):
Screen Shot 2021-04-09 at 4 27 35 AM

3. confirmed that a retry a method would eventually work by setting the number of retries very high (temporarily) and starting the server half way through the retry cycle:

Screen Shot 2021-04-09 at 4 27 22 AM


packages/eas-cli/src/graphql/client.ts
```diff
@@ -19,6 +23,16 @@ type SessionHeaders = {
...
+  exchanges: [
+    dedupExchange,
+    cacheExchange,
+    retryExchange({
+      maxDelayMs: 4000,
+      retryIf: err =>
+        !!(err && (err.networkError || err.graphQLErrors.some(e => e?.extensions?.isTransient))),
+    }),
+    fetchExchange,
+  ],
```
jkhales, line 35: `dedupExchange`, `cacheExchange`, and `fetchExchange` are implicit defaults that we need to explicitly specify now that we are adding exchanges:
Screen Shot 2021-04-09 at 8 11 31 am

https://formidable.com/open-source/urql/docs/architecture/

jkhales: The `retryExchange` must be placed prior to the fetchExchange and after the cacheExchange: https://formidable.com/open-source/urql/docs/advanced/retry-operations/

packages/eas-cli/src/graphql/client.ts
```diff
@@ -19,6 +23,16 @@ type SessionHeaders = {
 
 export const graphqlClient = createUrqlClient({
   url: getExpoApiBaseUrl() + '/--/graphql',
+  exchanges: [
+    dedupExchange,
+    cacheExchange,
+    retryExchange({
+      maxDelayMs: 4000,
```
jkhales, line 30: The default options for `retryExchange` are:
```
{
  initialDelayMs: 1000,
  maxDelayMs: 15000,
  randomDelay: true,
  maxNumberAttempts: 2,
  retryIf: err => err && err.networkError,
};
```

The potential 31 = 2*15+1 second delay seemed excessive. Went with a max potential 9 = 2*4 +1 second delay.

The randomization of the delay and the high max is for dealint with https://en.wikipedia.org/wiki/Thundering_herd_problem which does not seem like it will be as large of a problem when dealing with CLI calls.

fson (approved):

PR #322, opened 2021-04-09
[eas-cli] clean up formatting of list/view commands

jkhales:
# Why

Standardize and clean up formatting of channel/branch list/view display of their associated `updates`.

# How

 * Consolidated representation of update to three columns: `UPDATE_COLUMNS = ['update description', 'update runtime version', 'update group ID'];`

 * Used `formatUpdate` to make a compact description in all 4 commands.

  Chose to have a separate column for 
  1. `update runtime version` to highlight it to the user and avoid confusion.  
  2. `update group ID` due to its importance for further investigation with `eas update:view <ID>`

* fixed graphQL calls to correct `username` and other misc typos.

# Test Plan

Screen Shot 2021-04-09 at 7 36 23 AM

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -519,6 +519,22 @@
             "deprecationReason": null
           },
           {
+            "name": "submissions",
+            "description": "",
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "OBJECT",
+                "name": "SubmissionQuery",
+                "ofType": null
+              }
+            },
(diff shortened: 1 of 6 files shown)


byCedric (approved):
Looks good to me, the only concern I might have is the order of the exported `UPDATE_COLUMNS`. If we change this, there won't be any errors but the content can be mixed-up in the tables.

jkhales:
> Looks good to me, the only concern I might have is the order of the exported `UPDATE_COLUMNS`. If we change this, there won't be any errors but the content can be mixed-up in the tables.

that's true, I can't think of a clean way to enforce this though. Will merge for now if I can think of something smart will do in a future PR.

PR #326, opened 2021-04-12
[eas-cli] make branch name a positional instead of a named arg

jkhales:
# Why

Change publish syntax from:

`eas branch:publish --branch [name]`

to

`eas branch:publish [name]`

to match the syntax in the other commands.

# Test Plan

tested a publish.


jonsamp (approved):
Thanks!!

PR #331, opened 2021-04-13
[eas-cli] update channel:list to handle rollouts

jkhales:
# Why

Currently we only display a single branch for each channel. This PR updates the command to display multiple branches when a rollout is present:

Screen Shot 2021-04-13 at 11 15 45 AM

# How

We iterate of the channels and use the same function to display rollouts as is used in `eas channel:view`

# Test Plan

Tested on sample repo.

# Note
This PR replaces this in progress PR: https://github.com/expo/eas-cli/pull/304

### DIFF
--- a/packages/eas-cli/src/commands/branch/list.ts
+++ b/packages/eas-cli/src/commands/branch/list.ts
@@ -18,6 +18,10 @@ import Log from '../../log';
 import { findProjectRootAsync, getProjectFullNameAsync } from '../../project/projectUtils';
 import { getActorDisplayName } from '../../user/User';
 
+export type FormatUpdateParameter = Pick<Update, 'id' | 'createdAt' | 'message'> & {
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/commands/branch/list.ts
```diff
@@ -18,6 +18,10 @@ import Log from '../../log';
 import { findProjectRootAsync, getProjectFullNameAsync } from '../../project/projectUtils';
 import { getActorDisplayName } from '../../user/User';
 
+export type FormatUpdateParameter = Pick<Update, 'id' | 'createdAt' | 'message'> & {
+  actor?: Maybe<Pick<User, 'username' | 'id'> | Pick<Robot, 'firstName' | 'id'>>;
```
byCedric, line 22: This is a pain to write in TS 😅

byCedric (approved):

packages/eas-cli/src/commands/branch/list.ts
jkhales: indeed

PR #332, opened 2021-04-14
[eas-cli] add update:delete command

jkhales:
# Why

Add update:delete

# How

Since this is a dangerous operation that a user should avoid unless in situations like accidental secret publication an extensive warning message is included.

Screen Shot 2021-04-13 at 7 42 35 PM

Note the default is to abort.

# Test Plan

Tested on sample repo.

#Note 
This replaces PR https://github.com/expo/eas-cli/pull/261 which was targeted at deleting a single update. However after discussion we decided that this would lead to confusing errors and thus we should only expose the update *group* to developers.

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -16939,11 +16939,11 @@
         "description": "",
         "fields": [
           {
-            "name": "deleteUpdate",
-            "description": "Delete an EAS update",
+            "name": "deleteUpdateGroup",
+            "description": "Delete an EAS update group",
             "args": [
               {
-                "name": "updateId",
+                "name": "group",
                 "description": null,
                 "type": {
                   "kind": "NON_NULL",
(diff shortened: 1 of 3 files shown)


jkhales:
CC @jonsamp regarding the wording of the warning.

byCedric:
I think it's better if we keep te same "warning"/permanent action structure/design related to others too. Something [like this](https://github.com/expo/eas-cli/blob/24bc29b3cd35042e91b14da6ddc3fe54e55197d3/packages/eas-cli/src/credentials/android/actions/UpdateKeystore.ts#L32-L45) for example.

- First-line yellow with the description it _will_ do (followed by the "PERMANENT and IRREVERSIBLE" in red)
- Other lines are great, they give more context to the user. Don't think it will benefit users if that's also red or yellow

byCedric (approved):
Still approving, it looks solid but may need a UI tweak.

jkhales:
> * Other lines are great, they give more context to the user. Don't think it will benefit users if that's also red or yellow

Screen Shot 2021-04-14 at 11 03 03 AM

PR #341, opened 2021-04-16
add sdkVersion as default key for cache

wkozyra95:
# Checklist

# Why

When upgrading to sdk 41 default caching is breaking builds because of a updated packages. This is just a quickfix, better solution will be necessary.

# How

add sdkVersion as cache key

# Test Plan

Run build on prod and checks in  logs that cache key was set
.


brentvatne (approved):

wkozyra95:
Closing in favor of different approach
- add `--clear-cache` flag when building
- error message that suggests to change cache config when  incompatible pods are found

PR #351, opened 2021-04-20
[eas-cli][eas-json] add getBuildProfileNamesAsync helper

quinlanj:
# Checklist

# Why

Add a convenience method, `getBuildProfileNamesAsync ` to eas-json, based on wojtek's feedback here: https://github.com/expo/eas-cli/pull/348#discussion_r616548554

# Test Plan

- [ ] added new tests

### DIFF
--- a/packages/eas-json/src/EasJsonReader.ts
+++ b/packages/eas-json/src/EasJsonReader.ts
@@ -20,12 +20,12 @@ interface BuildProfilePreValidation {
   extends?: string;
 }
 
+function intersect<T>(setA: Set<T>, setB: Set<T>): Set<T> {
+  return new Set([...setA].filter(i => setB.has(i)));
+}
+
 export class EasJsonReader {
   constructor(private projectDir: string, private platform: 'android' | 'ios' | 'all') {}
-
-  private intersect<T>(setA: Set<T>, setB: Set<T>): Set<T> {
-    return new Set([...setA].filter(i => setB.has(i)));
-  }
 
   /**
    * Return build profile names for a particular platform.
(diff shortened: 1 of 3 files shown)


packages/eas-json/src/EasJsonReader.ts
```diff
@@ -23,6 +23,20 @@ interface BuildProfilePreValidation {
 export class EasJsonReader {
   constructor(private projectDir: string, private platform: 'android' | 'ios' | 'all') {}
 
+  public async getBuildProfileNamesAsync(): Promise<string[]> {
```
wkozyra95, line 26: if platform is all it should return only common profiles for both platforms

wkozyra95 (commented):
Thanks for adding that

packages/eas-json/src/EasJsonReader.ts
```diff
@@ -23,6 +23,31 @@ interface BuildProfilePreValidation {
...
+      const profileNames = new Set(Object.keys(profiles));
+      if (this.platform === platform) {
+        return Array.from(profileNames);
+      } else if (this.platform === 'all') {
+        intersectingProfileNames = intersectingProfileNames
+          ? this.intersect(intersectingProfileNames, profileNames)
+          : profileNames;
+      }
+    }
+    return intersectingProfileNames ? Array.from(intersectingProfileNames) : [];
```
wkozyra95, line 48:
```suggestion
    if (this.platform === 'android') {
      return Object.keys(easJson?.builds?.android ?? {});
    } else if (this.platform === 'ios') {
      return Object.keys(easJson?.builds?.ios ?? {});
    } else {
       const intersectingProfileNames = this.intersect(
         Object.keys(easJson?.builds?.ios ?? {}),
         Object.keys(easJson?.builds?.android ?? {})
       );
       return Array.from(intersectingProfileNames)
    }
```
It's correct behavior, but it's hard to read it

packages/eas-json/src/EasJsonReader.ts
```diff
@@ -23,6 +23,31 @@ interface BuildProfilePreValidation {
 export class EasJsonReader {
   constructor(private projectDir: string, private platform: 'android' | 'ios' | 'all') {}
 
+  private intersect<T>(setA: Set<T>, setB: Set<T>): Set<T> {
```
wkozyra95, line 26: consider moving it outside this class

wkozyra95 (approved):

PR #357, opened 2021-04-22
[eas-cli] add CreateDistCert and UseExistingDistCert to beta manager

quinlanj:
# Checklist

# Why

The current credentials manager has CreateDistCert and UseExistingDistCert functionalities. This PR adds this functionality to the beta manager. `CreateDistCert` is slightly different here, as we made an optimization in `expo-cli` where if we were in a project context, we would also assign it to the project. We do something similar, but we setup build credentials with the new dist cert if we detect the user to be in project context.

# Test Plan

- [x] do manual testing


dsokal (commented):
This looks good to me!

PR #361, opened 2021-04-23
[eas-cli] add RemoveProvisioningProfile to beta manager

quinlanj:
# Checklist

# Why

Add the equivalent of the RemoveProvisioningProfile Action in the current credentials manager to the beta manager

# Test Plan

- [ ] manual testing


dsokal (approved):

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
```diff
@@ -274,6 +280,26 @@ export class ManageIosBeta implements Action {
         }
         return;
       }
+      case ActionType.RemoveProvisioningProfile: {
```
wkozyra95, line 283: We should keep this file as small as possible, so maybe consider either
- extracting cases like this as separate classes
- converting runAccountSpecificActionAsync into separate class

not necessarily in this PR, but I think this file is reaching the size that we should consider that

wkozyra95 (approved):

packages/eas-cli/src/credentials/manager/ManageIosBeta.ts
quinlanj: Makes sense :)

PR #384, opened 2021-05-05
Dynamically toggle capabilities before build

EvanBacon:
# Checklist

# Why

If a capability is defined in the entitlements file but not registered with Apple, the build will fail. This PR aims to reduce the possibility of that happening by checking the native entitlements file and enabling/disabling capabilities.

Fixes https://github.com/expo/expo-cli/issues/3448 for `eas build`

# How

- If the project is specified to run as a bare workflow build, then read the entitlements file directly if one exists. 
  - Otherwise if it's a managed project, read the Expo config and use the `ios.entitlements` object, defaulting to the template object (which always enables push notifications). This needs to be improved further in expo/config to allow reading of evaluated entitlements automatically (can implement in another PR).
- Given a JSON representation of an entitlements object, match against our supported capabilities. Supported capabilities are based on entitlements, and iOS platform support.
- If a capability that we manage is enabled for app, but not defined in the entitlements file, we'll disable it. This allows for users to enable future capabilities without us needing to explicitly support them.
- Basic capabilities (boolean) can be omitted from the request if they already exist on a bundle identifier, this helps to reduce the amount of invocations in simpler projects. If a capability has settings, we'll always update it to ensure the settings are in sync. This can be optimized further in the future but works for now.
- Using debug mode will show more info on how the request is built.
- We surface the user friendly names for enabled and disabled capabilities after syncing. 

### Example

If the user has associatedDomains defined in their entitlements file, then in their app.json then enable/disable the capability remotely before building. Without this feature the [build fails](https://expo.io/accounts/bacon/projects/yolo22/builds/0485dd3c-f16d-4ea8-b998-1ce3547d0e30).

```log
❌  error: Provisioning profile "*[expo] com.bacon.yolo22 AppStore 2021-05-04T22:29:48.816Z" doesn't support the Associated Domains capability. (in target 'yolo22' from project 'yolo22')

❌  error: Provisioning profile "*[expo] com.bacon.yolo22 AppStore 2021-05-04T22:29:48.816Z" doesn't include the com.apple.developer.associated-domains entitlement. (in target 'yolo22' from project 'yolo22')
```

With this feature the [build works](https://expo.io/accounts/bacon/projects/yolo22/builds/0a216281-06e1-4427-adee-3c6d01e571d1).

# Test Plan

- Wrote unit tests for the request and assembly.

- In a managed project's `app.json`:
```js
{
  "ios": {
    "entitlements": {
      "com.apple.developer.associated-domains": [
        "applinks:evanbacon.dev",
        "webcredentials:evanbacon.dev"
      ]
    }
  }
}
```
- Then run `eas build -p ios`, after capabilities sync, check to see if `889YZVZQDS_ASSOCIATED_DOMAINS` is defined (my app's opaque id is `889YZVZQDS`).
- After the build completes, submit to the app store, it should finish processing without any warnings.
- Then remove the `entitlements` and rerun `eas build`, the syncing step should remove the capability. 

> This feature doesn't currently support `ios.usesAppleSignIn`, `ios.associatedDomains`, etc. because the mods must be statically evaluated for this to work. We can add this functionality in a future PR.

Visually, this is the scope of automatic functionality:

Screen Shot 2021-05-05 at 6 02 35 PM


brentvatne (approved):
this looks good to me. one concern i have about continuing to go down this road is that it's not clear how a third party lib using a config plugin to add a feature that depends on a capability would be able to work w/o modifying eas-cli like this. maybe we could add a more generic `ios.capabilities` key?

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -31,6 +35,31 @@ interface IosAppBuildCredentials {
...
+  // error: Provisioning profile "*[expo] com.bacon.myapp AppStore 2021-05-04T22:29:48.816Z" doesn't support the Associated Domains capability.
+  let isAssociatedDomainsDefined = isNativeAssociatedDomainsDefined(projectDir);
+  if (isAssociatedDomainsDefined == null) {
+    // Determine if the app.json has `ios.associatedDomains` or equivalent entitlements field defined.
+    isAssociatedDomainsDefined =
+      !!exp.ios?.associatedDomains ||
+      !!exp.ios?.entitlements?.['com.apple.developer.associated-domains'];
+  }
+  return isAssociatedDomainsDefined;
+}
```
dsokal, line 61: Maybe move this logic outside of the file, it's at a different level of abstraction than other functions in this file. 
I'd create `CapabilityUtils.ts` or similar in the same directory.

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -40,13 +69,17 @@ export class SetupBuildCredentials implements Action<IosAppBuildCredentials> {
     await ctx.bestEffortAppStoreAuthenticateAsync();
 
     if (ctx.appStore.authCtx) {
+      const enableAssociatedDomains = isAssociatedDomainsDefined(ctx.projectDir, ctx.exp);
```
wkozyra95, line 72: You don't know if exp is here or even if it is this interface does not guarantee that exp refers to the same app as `this.options.app`. I'm not sure if it's even possible to call this class in a way that would use different projects, but either way this code is not safe.

I'm not 100% how to handle this approach, you could do this by
```suggestion
      const enableAssociatedDomains = ctx.hasProjectContext && ctx.exp.slug === this.options.app.projectName ? isAssociatedDomainsDefined(ctx.projectDir, ctx.exp) : undefined;
```
and changing ensureBundleIdExistsAsync to not do anything in case of undefined (neither set nor unset)

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -83,18 +84,30 @@ export async function syncCapabilities(
...
-    const capability = await bundleId.hasCapabilityAsync(CapabilityType.PUSH_NOTIFICATIONS);
-    if (!capability && options.enablePushNotifications) {
+    const notifications = await bundleId.hasCapabilityAsync(CapabilityType.PUSH_NOTIFICATIONS);
+    if (!notifications && options.enablePushNotifications) {
       await bundleId.updateBundleIdCapabilityAsync({
         capabilityType: CapabilityType.PUSH_NOTIFICATIONS,
         option: CapabilityTypeOption.ON,
       });
-    } else if (capability && !options.enablePushNotifications) {
+    } else if (notifications && !options.enablePushNotifications) {
```
wkozyra95, line 93: maybe in case of undefined in options you should not modify anything(both here and for associated domains)

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
wkozyra95: cc @quinlanj 
what do you think about it?

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -83,18 +84,30 @@ export async function syncCapabilities(
   const spinner = ora(`Syncing capabilities`).start();
 
   try {
-    const capability = await bundleId.hasCapabilityAsync(CapabilityType.PUSH_NOTIFICATIONS);
-    if (!capability && options.enablePushNotifications) {
+    const notifications = await bundleId.hasCapabilityAsync(CapabilityType.PUSH_NOTIFICATIONS);
```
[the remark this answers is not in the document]
EvanBacon: I don't think they can be run in parallel.

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -31,6 +35,31 @@ interface IosAppBuildCredentials {
...
+  } catch {}
+  return null;
+}
+
+function isAssociatedDomainsDefined(projectDir: string, exp: Pick<ExpoConfig, 'ios'>) {
+  // Get the native entitlements file and determine if associated domains is defined in it.
+  // If we don't enable this capability xcode will fail to build:
+  // error: Provisioning profile "*[expo] com.bacon.myapp AppStore 2021-05-04T22:29:48.816Z" doesn't support the Associated Domains capability.
+  let isAssociatedDomainsDefined = isNativeAssociatedDomainsDefined(projectDir);
+  if (isAssociatedDomainsDefined == null) {
```
quinlanj, line 54:
```suggestion
  if (isAssociatedDomainsDefined === null) {
```

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
quinlanj: 1. On project context: This class is called by `build:ios` and the manager, afaik its always called in a project context, unless it's possible to call `build:ios` outside your project directory.
2. On `this.options.app` not referring to the same app as `ctx.exp`: this is possible but it was not a usecase I thought of 😅  We currently derive `this.options.app` from whatever is declared `ctx.exp`.

That being said, @wkozyra95  raises a great point and it's good to assert these assumptions in the class. I'd make these assertions at the beginning of `runAsync`, see here:

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -40,13 +69,17 @@ export class SetupBuildCredentials implements Action<IosAppBuildCredentials> {
     await ctx.bestEffortAppStoreAuthenticateAsync();
```
quinlanj, line 69:
```suggestion
   assert(ctx.hasProjectContext, 'SetupBuildCredentials must be run in a project context');
   assert(this.options.app.projectName === ctx.exp.slug, `App mismatch: cannot modify ${this.options.app.projectName} in the project directory of ${ctx.exp.slug}`);

    await ctx.bestEffortAppStoreAuthenticateAsync();
```

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -46,7 +47,10 @@ export class SetupBuildCredentials implements Action<IosAppBuildCredentials> {
           bundleIdentifier: app.bundleIdentifier,
           projectName: app.projectName,
         },
-        { enablePushNotifications: true }
+        {
+          // TODO: Get correct workflow
+          entitlements: await resolveEntitlementsJsonAsync(ctx.projectDir, 'generic'),
```
EvanBacon, line 52: Not sure how to do this "correctly"

EvanBacon: still relavent

brentvatne:
also, i spoke with @EvanBacon about this stuff today and i'm generally on board with changes. there's additional work being on on the expo-cli repo but this can ship independently.

packages/eas-cli/src/credentials/ios/actions/SetupBuildCredentials.ts
```diff
@@ -46,7 +48,10 @@ export class SetupBuildCredentials implements Action<IosAppBuildCredentials> {
           bundleIdentifier: app.bundleIdentifier,
           projectName: app.projectName,
         },
-        { enablePushNotifications: true }
+        {
+          // TODO: Get correct workflow
+          entitlements: await resolveEntitlementsJsonAsync(ctx.projectDir, Workflow.GENERIC),
```
wkozyra95, line 53: this code should not be aware of workflow because it depended on the way it's executed, users potentially can have build profiles with both generic managed workflow.

This class can be executed when building or from credentials manager. I suggest
- reading running resolveEntitlementsJsonAsync in here https://github.com/expo/eas-cli/blob/5e5badf11f2699a4e855cc15a1105547f72bf9aa/packages/eas-cli/src/build/ios/credentials.ts#L22 and passing entitlements to this class
- assume it's managed here https://github.com/expo/eas-cli/blob/5e5badf11f2699a4e855cc15a1105547f72bf9aa/packages/eas-cli/src/credentials/manager/ManageIos.ts#L226

packages/eas-cli/src/credentials/ios/appstore/entitlements.ts
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    const entitlementsPath = IOSConfig.Paths.getEntitlementsPath(projectDir);
+    if (entitlementsPath) {
+      return plist.parse(fs.readFileSync(entitlementsPath, 'utf8'));
+    }
+  } catch {}
+  return null;
+}
+
+export function resolveEntitlementsJsonAsync(projectDir: string, workflow: Workflow) {
```
wkozyra95, line 17: add return type

packages/eas-cli/src/credentials/ios/appstore/entitlements.ts
```diff
@@ -0,0 +1,32 @@
+import { getConfig } from '@expo/config';
+import { IOSConfig } from '@expo/config-plugins';
+import { Workflow } from '@expo/eas-build-job';
+import plist from '@expo/plist';
+import fs from 'fs';
+
+function getEntitlementsJson(projectDir: string) {
```
wkozyra95, line 7: add return type

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+  const request: { capabilityType: CapabilityType; option: any }[] = [];
+  for (const [key, value] of Object.entries(entitlements)) {
+    if (value) {
+      const capability = CapabilityMapping.find(capability => capability.entitlement === key);
+      if (capability) {
+        const existingIndex = capabilities.findIndex(existing =>
+          existing.isType(capability.capability)
+        );
+        const existing = existingIndex > -1 ? capabilities[existingIndex] : null;
+        // Only skip if the existing capability is a simple boolean value,
```
dsokal, line 59: I assume the capability is a simple boolean value when `existing?.attributes.settings == null` but IMO this is not easy to deduce that. Maybe clarify this in the comment.

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+        // if it has more complex settings then we should always update it.
+        if (existing && existing?.attributes.settings == null) {
```
dsokal, line 61:
```suggestion
        if (existing?.attributes.settings == null) {
```

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+
+const getBooleanOptions: GetOptionsMethod = entitlement => {
+  return entitlement === true ? CapabilityTypeOption.ON : CapabilityTypeOption.OFF;
+};
+
+const getDefinedOptions: GetOptionsMethod = entitlement => {
+  return entitlement ? CapabilityTypeOption.ON : CapabilityTypeOption.OFF;
+};
+
+export async function syncCapabilitiesForEntitlementsAsync(
```
dsokal, line 43: Can you please break this function down to a few smaller? It's really difficult to follow it.

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+  bundleId: BundleId,
+  entitlements: JSONObject
+) {
+  const capabilities = await bundleId.getBundleIdCapabilitiesAsync();
+  const enabled: string[] = [];
+  const disabled: string[] = [];
+  const request: { capabilityType: CapabilityType; option: any }[] = [];
+  for (const [key, value] of Object.entries(entitlements)) {
+    if (value) {
+      const capability = CapabilityMapping.find(capability => capability.entitlement === key);
```
dsokal, line 53:
```suggestion
      const capabilityDefinition = CapabilityMapping.find(capability => capability.entitlement === key);
```
or similar

It made me confused, because when I was reading the later part of this code I thought it's a member of `capabilities`.

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+        !!capability &&
+        !request.find(request => request.capabilityType && existing.isType(request.capabilityType))
+      ) {
+        request.push({
+          // @ts-ignore
+          capabilityType: adjustedType,
+          option: CapabilityTypeOption.OFF,
+        });
+
+        disabled.push(capability?.name || adjustedType);
```
dsokal, line 124:
```suggestion
        disabled.push(capability.name);
```
`?` (and the fallback value) shouldn't be necessary if you apply the previous suggestion.

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -0,0 +1,415 @@
...
+      }
+
+      // Only disable attributes that we handle,
+      // this enables devs to turn on capabilities outside of EAS without worrying about us disabling them.
+      const capability = CapabilityMapping.find(
+        capability => capability.capability === adjustedType
+      );
+      if (
+        !!capability &&
+        !request.find(request => request.capabilityType && existing.isType(request.capabilityType))
```
dsokal, line 116:
```suggestion
        !request.find(request => existing.isType(request.capabilityType))
```
as far as I can see `capabilityType` can't be falsy

dsokal (commented):
It generally looks good but please address my comments before merging to make the code easier to follow.

quinlanj (approved):
lgtm, but please get wojtek/dominik to approve this first before merging !

packages/eas-cli/src/build/ios/credentials.ts
```diff
@@ -7,6 +7,8 @@ import IosCredentialsProvider, {
   IosCredentials,
 } from '../../credentials/ios/IosCredentialsProvider';
 import { getAppLookupParamsFromContext } from '../../credentials/ios/actions/BuildCredentialsUtils';
+import { IOSCapabilitiesOptions } from '../../credentials/ios/appstore/ensureAppExists';
```
wkozyra95, line 10:
```suggestion
import { IosCapabilitiesOptions } from '../../credentials/ios/appstore/ensureAppExists';
```
With the exception of `Platform.IOS` we use `Ios` everywhere in cli

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -1,12 +1,15 @@
...
 import ora from 'ora';
 
+import Log from '../../../log';
 import { AuthCtx, getRequestContext } from './authenticate';
+import { syncCapabilitiesForEntitlementsAsync } from './bundleIdCapabilities';
 import { assertContractMessagesAsync } from './contractMessages';
 
-export interface EnsureAppExistsOptions {
-  enablePushNotifications?: boolean;
+export interface IOSCapabilitiesOptions {
```
dsokal, line 11:
```suggestion
export interface IosCapabilitiesOptions {
```

dsokal (approved):

wkozyra95 (approved):

flow-danny:
Is this supposed to work with `app.config.ts`? The sync basically disabled all our capabilities...

flow-danny:
Now I see EAS `"workflow": "managed"` still might behave as bare workflow in this case.

brentvatne:
@flow-danny - sorry can you elaborate on what you experienced? i'd like to understand what happened in your situation

flow-danny:
We have a regular managed Expo app and tried EAS for the first time.

We didn't have any custom entitlements configured, just `usesIcloudStorage` set to true as documented here https://docs.expo.io/versions/latest/sdk/document-picker/

(But in the typescript version if app.json)


iCloud entitlements get disabled during the syncing process until after we added these to the entitlements config. I used expo eject and Xcode to figure out the entitlement keys that we needed.

brentvatne:
if anyone runs into issues with capability syncing, you can disable it with `EXPO_NO_CAPABILITY_SYNC=1 eas build` (as described here: https://docs.expo.dev/build-reference/ios-capabilities/) and post an issue with some more information

PR #396, opened 2021-05-10
[eas-cli] add tests for unencrypted/empty password keystores

quinlanj:
# Checklist

# Why

From our conversation on slack, i thought it'd be a good idea to add testing for keystores with empty passwords and unencrypted passwords. I couldn't generate any of these using `keytool` because it doesn't allow you to generate keystores with an empty password, and had to use openssl instead.

https://stackoverflow.com/a/62863490/8611543

added tests for our validation code with the following:
- a keystore that has an empty string password
- a keystore that is unencrypted. **I dont know if this is the same thing as having no password though** -- i need to pass in `-passout pass:` to make the keystore which means that the password is basically an empty string, but i ran it with `openssl pkcs12 -info -in unencrypted.p12 -noout -passin pass:` after and it shows the data as being unencrypted.

# Test Plan

- [ ] new tests pass


wkozyra95 (commented):
I created new keystores using the same commands you provided in comment
- keystore with empty password does work
- unencrypted keystore fail (keytool does not see any entries) and build is failing with "Failed to read key test-alias from store "

quinlanj:
@wkozyra95 i tried several different ways but i wasn't able to find a good way to validate if the private key is encrypted... i just removed that test case 😭

wkozyra95 (approved):

PR #402, opened 2021-05-12
[eas-cli] add 'eas init' command

jkhales:
# Checklist

# Why

We want to store a project's ID locally.

This is important for the new `expo-updates` among other things: the EAS url for the manifest includes the projectId and will be stored in the app.json as well.

For previous discussions, see https://linear.app/expo/issue/ENG-1029/link-an-eas-project-to-an-expo-project and https://github.com/expo/universe/pull/7557 for some previous discussions.

# How

* create an `eas init` command that is essentially a wrapper for  `setProjectIdAsync`.
* `setProjectIdAsync` calls  `ensureProjectExistsAsync` and then saves the projectId to disk so we don't have to call it again. Previously `ensureProjectExistsAsync` was only callable as a side effect of some other command like `eas create:channel`.
* make `getProjectIdAsync` try to retrieve the ID from app.json before making a network call. 
* In a follow up PR almost all instances of `ensureProjectExistsAsync` will be replaced with calls to `getProjectIdAsync` This will reduce our network calls significantly and make the cli feel much snappier.

In keeping with the low attention cost philosophy/do things automatically for the user, any call for a project ID that does not find a the project ID in app.json will call `setProjectIdAsync` essentially running `eas init` for them. Thus the user doesn't need to ever call `eas init` directly.
 
 
# Test Plan

Tested in demo repo.

Screen Shot 2021-05-12 at 5 05 53 AM


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -52,13 +60,56 @@ export async function findProjectRootAsync(cwd?: string): Promise<string | null>
...
     projectName: exp.slug,
     privacy,
   });
+
+  const result = await modifyConfigAsync(projectDir, {
+    extra: { ...exp.extra, eas: { ...exp.extra?.eas, projectId } },
+  });
+
+  if (result.type !== 'success') {
+    throw new Error(result.message);
```
wkozyra95, line 78: maybe in the case of dynamic config, we should display instructions for the user, that either 
- suggest using both static and dynamic configs and rerurning command
- prints value of project id and location where should be in config

wkozyra95 (approved):

packages/eas-cli/src/project/projectUtils.ts
jkhales: * Even if we are using both, the method still throws: https://github.com/expo/expo-cli/blob/master/packages/config/src/Config.ts#L383
And perhaps this is the write behavior as we don't know if they would want this in the static or dynamic config.

* I added the following message and retested various other commands with a dynamic setup.
Screen Shot 2021-05-12 at 7 32 23 AM

PR #404, opened 2021-05-12
[eas-cli] replace ensureProjectExistsAsync with getProjectIdAsync

jkhales:
# Checklist

# Why

getProjectIdAsync is superior to ensureProjectExistsAsync as it will check to see if the ID is saved locally. This switch will reduce the number of network calls and make the CLI significantly "snappier",

# How

find/replace. Also made use of the `getFullName` util to further simplify code.

Would appreciate @FiberJW eyes on places I'm unfamiliar with usage in places like:
https://github.com/expo/eas-cli/blame/main/packages/eas-cli/src/commands/secret/create.ts#L50
in particular.

# Test Plan

tested various commands in a demo project.

### DIFF
--- a/packages/eas-cli/src/commands/branch/create.ts
+++ b/packages/eas-cli/src/commands/branch/create.ts
@@ -10,10 +10,12 @@ import {
   UpdateBranch,
 } from '../../graphql/generated';
 import Log from '../../log';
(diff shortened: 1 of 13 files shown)


FiberJW (approved):
thumbs-up-00001

PR #405, opened 2021-05-12
[eas-cli] query app by ID instead of fullName in listBranchesAsync

jkhales:
# Checklist

# Why

Now that we have easy access to the `projectId` switch over the branch:list command from `fullName`.

# How

* switched args and ran `yarn generate-graphql-code`
* reenabled optionality of branch name in `eas branch:publish`

# Test Plan

Tested `easd branch:list` still works in demo repo.
Tested `easd branch:publish` interactive branch selection prompt still works.


byCedric (approved):
LGTM

PR #442, opened 2021-06-04
[eas-cli] Deprecate --skip-credentials-check flag for build

brentvatne:
# Checklist

- [n/a] I've tagged the changelog entry with `[EAS BUILD API]` if it's a part of a breaking change to EAS Build API (only possible when updating `@expo/eas-build-job` package).

# Why

Context in https://linear.app/expo/issue/ENG-1103/remove-skip-credentials-check

# How

- Delete usage of `skipCredentialsCheck`
- Add warning to when the `--skip-credentials-check` flag is passed in to `eas build`
- Create a logging helper that we can use for future flag deprecations to ensure we have a consistent format (using the same format as expo-cli, eg: run `expo start --config example.json` anywhere to see a similar warning.
- Hide the flag from `eas build --help`

# Test Plan

### Flag is hidden from help

```
╰─$ easd build --help
Start a build

USAGE
  $ eas build

OPTIONS
  --clear-cache                     Clear cache before the build
  --local                           Run build locally [experimental]
  --non-interactive                 Run command in --non-interactive mode
  -p, --platform=(android|ios|all)
  --profile=profile                 [default: release] Name of the build profile from eas.json
  --skip-project-configuration      Skip project configuration
  --[no-]wait                       Wait for build(s) to complete

COMMANDS
  build:cancel     Cancel a build.
  build:configure  Configure the project to support EAS Build.
  build:list       list all builds for your project
  build:view       view a build for your project
```

### Warning is printed when flag is used

Screen Shot 2021-06-10 at 8 09 34 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Deprecate `--skip-credentials-check` flag because it doesn't do anything and is no longer needed.([#442](https://github.com/expo/eas-cli/pull/442) by [@brentvatne](https://github.com/brentvatne))
+
 ## [0.17.0](https://github.com/expo/eas-cli/releases/tag/v0.17.0) - 2021-06-02
 
 ### 🐛 Bug fixes
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -25,7 +25,7 @@ export default class Build extends EasCommand {
     platform: flags.enum({ char: 'p', options: ['android', 'ios', 'all'] }),
     'skip-credentials-check': flags.boolean({
       default: false,
-      description: 'Skip validation of build credentials',
+      hidden: true,
     }),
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Remove --skip-credentials-check flag, it doesn't do anything and is no longer needed.
```
wkozyra95, line 15: it's a breaking change, ci job that are using this flag will fail

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -27,10 +27,6 @@ export default class Build extends Command {
 
   static flags = {
     platform: flags.enum({ char: 'p', options: ['android', 'ios', 'all'] }),
-    'skip-credentials-check': flags.boolean({
```
wkozyra95, line 30: maybe it would be better to leave that flag here, mark it as deprecated and throw a better error informing that flag is not necessary anymore

wkozyra95 (approved):

PR #454, opened 2021-06-14
Record dev client builds created and downloaded to analytics

fson:
# Checklist

# Why

We track usage of dev client in order to to understand how it's used and to improve the experience.

# How

Added properties to `BuildContext.trackingCtx`.

# Test Plan

Because these properties get added to all build events, only did a spot check of a couple of events.
- Added logging to `analytics.ts`
- In a managed project with `expo-dev-client` installed, ran `eas build`
- Observed that the properties were included.
Screen Shot 2021-06-14 at 18 03 02

Tested `dev_client: false` case:
- Ran `yarn remove expo-dev-client`.
- Ran `eas build`.
- Checked that `dev_client` property was `false` and `dev_client_version` property was not defined.

### DIFF
--- a/packages/eas-cli/src/build/context.ts
+++ b/packages/eas-cli/src/build/context.ts
@@ -115,7 +115,7 @@ export function createBuildContext<T extends Platform>({
   }
 
   const accountId = findAccountByName(commandCtx.user.accounts, commandCtx.accountName)?.id;
-  const devClienProperties = getDevClientEventProperties(platform, commandCtx, buildProfile);
+  const devClientProperties = getDevClientEventProperties(platform, commandCtx, buildProfile);
   const trackingCtx = {
     tracking_id: uuidv4(),
     platform,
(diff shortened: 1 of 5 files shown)


kgc00:
looks good

packages/eas-cli/src/build/context.ts
```diff
@@ -122,3 +125,12 @@ export function createBuildContext<T extends Platform>({
     buildProfile,
   };
 }
+
+function getDevClientEventProperties(projectDir: string) {
```
wkozyra95, line 129: define return type

wkozyra95 (approved):

fson:
I've now added additional checks to the `dev_client` property:
- for managed projects, check that `buildType` is `development-client`
- for bare, check that it's a debug build

packages/eas-cli/src/build/context.ts
```diff
@@ -122,3 +136,43 @@ export function createBuildContext<T extends Platform>({
...
+  } else if (platform === Platform.ANDROID) {
+    const { gradleCommand } = buildProfile as AndroidGenericBuildProfile;
+    return {
+      dev_client: Boolean(devClientVersion && gradleCommand?.includes('Debug')),
+      dev_client_version: devClientVersion,
+    };
+  } else if (platform === Platform.IOS) {
+    const { schemeBuildConfiguration } = buildProfile as IosGenericBuildProfile;
+    return {
+      dev_client: Boolean(devClientVersion && schemeBuildConfiguration === 'Debug'),
```
wkozyra95, line 163: technically it's possible to have Debug as the default buildConfiguration in xcode, so it will build dev client if schemeBuildConfiguration is not defined

but it be would quite a weird thing to do so I don't think we need to handle that case just for analytics

wkozyra95 (approved):
remember to fix types before merging

PR #468, opened 2021-06-18
[eas-cli] make apple team optional in appropriate cases

quinlanj:
# why

Make Apple Team optional in the following cases:
- to create IosAppCredentials (see https://github.com/expo/universe/pull/7740 for more details)
- to create an explicit bundle identifier (AppleAppIdentifier). Bundle identifiers are like domain names -- once you claim `sdf.sdf.sdf`, no one else can use it. But in the case where someone has a wildcard bundle identifier like `sdf.sdf.*`, it only matches to bundle identifiers *your Apple Team* owns. For example, Alice's wildcard identifier `sdf.sdf.*`only applies to `sdf.alice` because it's owned by her, but it will not apply to `sdf.bob` because it's owned by Bob. Therefore, its important we know the Apple Team `sdf.sdf.*` belongs to, but it's not required for explicit identifiers like `sdf.sdf`.

# prereqs
- [ ] https://github.com/expo/universe/pull/7740 has been deployed


packages/eas-cli/src/credentials/ios/actions/AppleAppIdentifierUtils.ts
```diff
@@ -0,0 +1,4 @@
+export function isWildcardBundleIdentifier(bundleIdentifier: string): boolean {
+  const wildcardRegex = /^[A-Za-z0-9.-]+\*$/;
```
wkozyra95, line 2: maybe let's keep regexps like this in one place e.g. packages/eas-cli/src/project/ios/bundleIdentifier.ts

wkozyra95 (approved):

PR #470, opened 2021-06-22
[eas-cli] add 'remove push key' action

quinlanj:
# Checklist

# Why

Add the 'remove push key' action. It works the following way:
- user chooses a push key from their account to delete
- remove push key from eas servers
- ask if they want to revoke the push key on apple servers

# Test Plan

- [ ] new tests pass

### DIFF
--- a/packages/eas-cli/src/credentials/ios/actions/RemovePushKey.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/RemovePushKey.ts
@@ -26,6 +26,7 @@ export class RemovePushKey {
       throw new Error(`Cannot remove push keys in non-interactive mode`);
     }
 
+    // TODO(quin): show the projects that rely on this
     const confirm = await confirmAsync({
       message: `Removing this push key will disable push notifications for projects that rely on it. Do you want to continue?`,
     });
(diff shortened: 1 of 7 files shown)


packages/eas-cli/src/credentials/ios/actions/RemovePushKey.ts
```diff
@@ -0,0 +1,52 @@
...
+export class RemovePushKey {
+  constructor(private account: Account, private pushKey: ApplePushKeyFragment) {}
+
+  public async runAsync(ctx: Context): Promise<void> {
+    if (ctx.nonInteractive) {
+      throw new Error(`Cannot remove push keys in non-interactive mode`);
+    }
+
+    const confirm = await confirmAsync({
+      message: `Removing this push key will disable push notifications for projects that rely on it. Do you want to continue?`,
```
wkozyra95, line 30: shouldn't we show what project are using that specific push key?

wkozyra95 (approved):

packages/eas-cli/src/credentials/ios/actions/RemovePushKey.ts
quinlanj: Yup, i'll need to expose this in our graphql api in www first -- i've added a todo and a linear task for myself

PR #477, opened 2021-06-25
[eas-cli] Automatically migrate build credentials to EAS

brentvatne:
Hotfix for issue resulting in developers ending up in half-migrated credentials state


quinlanj (approved):
I'm debating whether it'd be a better user experience all around if we just scrapped all uses of `promptForNameAsync` and replaced it with `generateRandomName()` :/

PR #481, opened 2021-06-25
[eas-cli] Fix inverted conditional so we actually only prompt about commit when index is dirty

brentvatne:
# Checklist

# Why

I noticed that I was being prompted to commit my changes even when I had a clean git index:

```
╭─~/code/abcd ‹main›
╰─$ gs                                                                                                                                                                                                                                                  130 ↵
On branch main
nothing to commit, working tree clean
╭─~/code/abcd ‹main›
╰─$ easd build:configure
💡 The following process will configure your iOS and/or Android project to be compatible with EAS Build. These changes only apply to your local project files and you can safely revert them at any time.

✔ Which platforms would you like to configure for EAS Build? › All
Warning! Your repository working tree is dirty.
It's recommended to commit all your changes before proceeding, so you can revert the changes made by this command if necessary.
? Would you like to proceed? › (Y/n)
```

It looks like in the refactor to a VCS client interface we accidentally regressed this check.

# How

Add a strategic `!`

# Test Plan

Initialize a new project, and with a clean git index run `eas build:configure` - you should not be told that your git working tree is dirty.

### DIFF
--- a/packages/eas-cli/src/vcs/git.ts
+++ b/packages/eas-cli/src/vcs/git.ts
@@ -56,7 +56,7 @@ export default class GitClient extends Client {
 
   public async hasUncommittedChangesAsync(): Promise<boolean> {
     const changes = await gitStatusAsync({ showUntracked: true });
-    return changes.length !== 0;
+    return changes.length > 0;
   }
 
   public async getRootPathAsync(): Promise<string> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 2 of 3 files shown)


packages/eas-cli/src/vcs/git.ts
```diff
@@ -56,7 +56,7 @@ export default class GitClient extends Client {
 
   public async hasUncommittedChangesAsync(): Promise<boolean> {
     const changes = await gitStatusAsync({ showUntracked: true });
-    return changes.length !== 0;
+    return changes.length > 0;
```
brentvatne, line 59: this wasn't necessary to fix the bug but the double negative of "uncommitted" and !== is not well suited for the feeble human brain

PR #482, opened 2021-06-25
[eas-cli] Recommend bundle id and application id in the same way as Expo CLI

brentvatne:
# Checklist

# Why

This is a nice experience and consistent with Expo CLI.

https://user-images.githubusercontent.com/90494/123492409-40ad2380-d5ce-11eb-881c-38b439b8e07d.mov

# How

Mirrored the implementation from Expo CLI, using the utility methods available within EAS CLI.

# Test Plan

```
eas build:configure -p ios # suggests a value
eas build:configure -p android # suggests a value based on the ios config
# -- reset state in project

eas build:configure -p all # suggests a value for android based on username/slug, if you change it then the changed value will be used for ios suggestion

# etc..
```

### DIFF
--- a/packages/eas-cli/src/project/android/__tests__/applicationId-test.ts
+++ b/packages/eas-cli/src/project/android/__tests__/applicationId-test.ts
@@ -11,6 +11,15 @@ import {
 
 jest.mock('fs');
 jest.mock('../../../prompts');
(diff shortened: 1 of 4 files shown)


brentvatne:
closed in favor of #487

PR #497, opened 2021-07-06
[eas-cli] unify generic and manged workflows

wkozyra95:
# Checklist

# Why

unify generic and managed workflow

# How

support the same set of option for both workflows

# Test Plan

Run builds with few different combinations of options in eas.json against production


dsokal (approved):
🤞

PR #500, opened 2021-07-07
[eas-cli] un-hide project:*

jkhales:
# Checklist

# Why

This removes `fullName` from the log messages as well as the graphQL calls in everything *but* the credentials commands.

The credentials are somewhat involved and will be done in a separate PR to keep things easily reviewable.

# How

* remove fullName in message logs
* replace graphQL calls that use fullName with their `byId` analog.
* remove `hidden` from `project:*` commands.

Also cleaned up the `topic` display:

Before:
Screen Shot 2021-07-07 at 2 10 04 PM
After:
Screen Shot 2021-07-07 at 2 10 26 PM
(another option is to just remove the descriptions of these topics as they are pretty redundant)

Q: Are there any other blockers to removing the `EAS_ENABLE_PROJECT_ID` gating?

# Test Plan

Tested commands in demo repo.
* branch:*
* channel:*
* build:*
* secrets:*

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -141,7 +141,7 @@
         "description": "manage update channels"
       },
       "device": {
-        "description": "manage your Apple devices"
+        "description": "manage your Apple devices for internal distribution"
       },
       "project": {
         "description": "manage your project"
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -131,14 +131,26 @@
     ],
     "helpClass": "./build/help",
     "topics": {
+      "account": {
+        "description": "manage your account"
+      },
       "build": {
         "description": "build app binaries"
       },
       "channel": {
         "description": "manage update channels"
       },
+      "device": {
+        "description": "manage your Apple devices"
+      },
+      "project": {
+        "description": "manage your project"
+      },
       "release": {
         "description": "manage update releases"
+      },
+      "secret": {
+        "description": "manage project and account secrets"
       },
       "update": {
         "description": "manage individual updates"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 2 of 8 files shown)


packages/eas-cli/package.json
```diff
@@ -131,15 +131,27 @@
...
+      },
+      "project": {
+        "description": "manage your project"
+      },
       "release": {
         "description": "manage update releases"
       },
+      "secret": {
+        "description": "manage project and account secrets"
+      },
```
jkhales, line 154: another option is to just remove the descriptions of these topics as they are pretty redundant

packages/eas-cli/src/commands/secret/create.ts
```diff
@@ -121,9 +120,7 @@ export default class EnvironmentSecretCreate extends Command {
       }
 
       Log.withTick(
-        `️Created a new secret ${chalk.bold(name)} on project ${chalk.bold(
-          `@${accountName}/${slug}`
-        )}.`
+        `️Created a new secret ${chalk.bold(name)} on project with id ${chalk.bold(projectId)}.`
```
brentvatne, line 123: this feels like a meaningful regression in the quality of the output. i don't think anyone will learn their project uuid and so it seems better to not output it, or to print some human readable identifier for the project and the uuid dimmed beside it. we should look at other clis to see how they handle this, nothing comes to mind off of the top of my head 😂

jkhales: good point

jkhales:
reverted the removal of `fullName` logging. 

This PR now just activates `project:*` and updates a graphQL call to rely on `projectId` instead of `fullName`.

packages/eas-cli/package.json
```diff
@@ -131,15 +131,27 @@
...
+        "description": "manage your account"
+      },
       "build": {
         "description": "build app binaries"
       },
       "channel": {
         "description": "manage update channels"
       },
+      "device": {
+        "description": "manage your Apple devices"
```
brentvatne, line 144:
```suggestion
        "description": "manage your Apple devices for internal distribution"
```

brentvatne: seems like useful context

brentvatne (approved):

PR #513, opened 2021-07-14
[eas-cli] Add force option to overwrite existing secrets

byCedric:
# Checklist

# Why

This is a proposal to help users tie their app properties better into their git workflow. Someone in discord mentioned they want to use the pipeline ID as version code. In the most ideal scenario, people would be able to specify additional environment variables when they start a build. But this seems like a quick alternative.

### Example

```bash
$ eas secret:create --name=PIPELINE_ID --value=$GITHUB_RUN_ID --force
$ eas build ...
```

**Pros**

- Keep environment variables up to date, programmatically
- Allows for tighter integration with existing workflows

**Cons**

- Running multiple builds at once will not inherit the right env var

# How

Added force option to `secret:create` instead of a new `secret:update` command. This reflects better what it actually would do, delete and create.

# Test Plan

```bash
$ eas secret:create --name=PIPELINE_ID --value=1337

# Would fail, because secret already exists
$ eas secret:create --name=PIPELINE_ID --value=1337

# Would succeed, if it exists it gets deleted before creating again
$ eas secret:create --force --name=PIPELINE_ID --value=2337
$ eas secret:create --force --name=NON_EXISTING --value=2337
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Add `secret:create --force` command to overwrite existing secrets. ([#513](https://github.com/expo/eas-cli/pull/513) by [@bycedric](https://github.com/bycedric))
+
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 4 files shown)


FiberJW (approved):
This is dope! Thanks for adding

thumbs-up-00020

FiberJW:
We still need to update the changelog

byCedric:
> We still need to update the changelog

Sorry, was pushed to my fork. Done, will merge it!

PR #514, opened 2021-07-15
[eas-cli] Upload exp config as client config when publishing or republishing

wschurman:
# Why

As of https://github.com/expo/universe/pull/7893, the server will support storing and serving the `extra.expoClient` field in the eas update manifest. As part of that, we need to upload it during the update group publish process.

# How

Include the public `exp` project config in the `updateInfo` fields. This is the same thing that is uploaded in the `expo-cli` publish process as the "manifest" (https://github.com/expo/expo-cli/blob/master/packages/xdl/src/project/getPublishExpConfigAsync.ts#L35), which is what we want to put in the `extra.expoClient` field.

The format of the `extra.expoClient` field is defined in https://github.com/expo/expo-cli/pull/3606.

# Test Plan

Waiting on the dependent www PR to go to staging/prod, but after it does the test plan will be:
- `eas branch:publish` and then get a manifest for the update, ensure it contains the `extra.expoClient` field.

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -313,6 +313,22 @@
             "deprecationReason": null
           },
           {
+            "name": "easBuild",
+            "description": "Top-level query object for querying EAS Build configuration.",
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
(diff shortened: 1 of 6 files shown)


jkhales (approved):
🚀

PR #549, opened 2021-08-04
[eas-cli] add --auto flag to publish

jkhales:
# Checklist

# Why

We want a flag to automatically set the branch name and update message to the git branch and commit message.

Also made command create a branch if it doesn't already exist.

# Test Plan

Tested in demo repo.

### DIFF
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -72,12 +72,13 @@ async function ensureBranchExists({
 }: {
   appId: string;
   name: string;
-}): Promise<
-  Exclude<
+}): Promise<{
+  id: string;
+  updates: Exclude<
     Exclude<ViewBranchQuery['app'], null | undefined>['byId']['updateBranchByName'],
     null | undefined
-  >
-> {
+  >['updates'];
+}> {
   const { app } = await viewUpdateBranchAsync({
     appId,
     name: branchName,
@@ -85,17 +86,16 @@ async function ensureBranchExists({
   const updateBranch = app?.byId.updateBranchByName;
   if (updateBranch) {
     const { id, updates } = updateBranch;
-    return { id, name: branchName, updates };
+    return { id, updates };
   }
 
   const newUpdateBranch = await createUpdateBranchOnAppAsync({ appId, name: branchName });
-  return { id: newUpdateBranch.id, name: branchName, updates: [] };
+  return { id: newUpdateBranch.id, updates: [] };
 }
 
 export default class BranchPublish extends Command {
   static hidden = true;
   static description = 'Publish an update group to a branch.';
-  static aliases = ['publish', 'p'];
 
   static args = [
     {
@@ -106,7 +106,7 @@ export default class BranchPublish extends Command {
 
   static flags = {
     message: flags.string({
-      description: 'Short message describing the updates.',
+      description: 'short message describing the updates.',
       required: false,
     }),
     republish: flags.boolean({
@@ -128,7 +128,7 @@ export default class BranchPublish extends Command {
     }),
     platform: flags.enum({
       char: 'p',
-      description: `Only publish to a single platform`,
+      description: `only publish to a single platform`,
       options: [...defaultPublishPlatforms, 'all'],
       default: 'all',
       required: false,
@@ -139,7 +139,7 @@ export default class BranchPublish extends Command {
     }),
     auto: flags.boolean({
       description:
-        'Publish to an EAS branch with the same name as the current git branch and publish message equal to the last git commit message.',
+        'use the current git branch and commit message for the EAS branch and update message',
       default: false,
     }),
   };
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -10,11 +10,13 @@ import ora from 'ora';
 
 import { graphqlClient, withErrorHandlingAsync } from '../../graphql/client';
 import {
-  Actor,
   GetUpdateGroupAsyncQuery,
+  Robot,
   RootQueryUpdatesByGroupArgs,
   Update,
   UpdateInfoGroup,
+  User,
+  ViewBranchQuery,
 } from '../../graphql/generated';
 import { PublishMutation } from '../../graphql/mutations/PublishMutation';
 import Log from '../../log';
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -318,12 +362,29 @@ export default class BranchPublish extends Command {
...
+  let actorName: string;
+  switch (actor?.__typename) {
+    case 'User': {
+      actorName = (actor as Pick<User, 'username' | 'id'>).username;
+      break;
+    }
+    case 'Robot': {
+      const { firstName, id } = actor as Pick<Robot, 'firstName' | 'id'>;
+      actorName = firstName ?? `robot: ${id.slice(0, 4)}...`;
+      break;
```
jkhales, line 381: This is not strictly related to the `--auto` addition. Noticed we missed a case when fixing up the typing. Curious if this will fix that robot issue @jonsamp was having?

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -171,11 +212,10 @@ export default class BranchPublish extends Command {
         })
       );
     }
-    assert(name, 'branch name must be specified.');
-
-    const { id: branchId, updates } = await viewUpdateBranchAsync({
+    assert(branchName, 'branch name must be specified.');
+    const { id: branchId, updates } = await ensureBranchExists({
```
jkhales, line 216: No longer throw if branch doesn't exist. Create it instead.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -107,13 +137,19 @@ export default class BranchPublish extends Command {
       description: `return a json with the new update group.`,
       default: false,
     }),
+    auto: flags.boolean({
```
wkozyra95, line 140: maybe consider sth more descriptive, meaning of the auto in context of publishing is not really obvious

`--auto-branch`/`--auto-message`
`--managed`
or use `--non-interactive` flag and if values are not provided via `--message`/`--branch` us those defaults

Also if you are considering adding updates config to eas.json, options like this would be a good candidate for it

jkhales: The issue with `--non-interactive` is that is not clear that this means "apply an opinionated default" instead of "throw if essential args are not provided"

`--auto-branch/--auto-message` are nice, but maybe too verbose?

> Also if you are considering adding updates config to eas.json, options like this would be a good candidate for it

Interesting idea, I'm a bit hesitant to engage there right now as it is in such flux that things tend to get reverted. Maybe once things get more established?

@jonsamp curious what you think about these other flag suggestions.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -152,14 +188,19 @@ export default class BranchPublish extends Command {
     }
     const projectId = await getProjectIdAsync(exp);
 
-    if (!name) {
+    if (!branchName && autoFlag) {
+      branchName =
+        (await vcs.getBranchNameAsync()) || `branch-${Math.random().toString(36).substr(2, 4)}`;
```
jonsamp, line 193: We should create another task to make the `getBranchNameAsync` work inside GitHub Actions. The command this runs is `git rev-parse --abbrev-ref HEAD`, which works locally, but results in this error during a GitHub Action:

Screen Shot 2021-08-06 at 12 06 27 AM

This is not a blocker for this PR, and this should generally not be a problem for CI, since we can pass in the branch name/message programmatically. But it'd be great to run `eas branch:publish --auto` locally, and then write the same command for the CI action.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -171,11 +212,10 @@ export default class BranchPublish extends Command {
         })
       );
     }
-    assert(name, 'branch name must be specified.');
-
-    const { id: branchId, updates } = await viewUpdateBranchAsync({
+    assert(branchName, 'branch name must be specified.');
```
jonsamp, line 215: I think we could move this `assert` inside the conditional above it. It's a check to make sure that the user chose a branch via an interactive prompt, so it seems like it belongs in there, since the other flow will get the git branch or create a random branch.

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -271,6 +311,10 @@ export default class BranchPublish extends Command {
       }
     }
 
+    if (!message && autoFlag) {
+      message = (await vcs.getLastCommitMessageAsync())?.trim();
```
jonsamp, line 315: Similarly here, this fails when run on CI. It runs `git --no-pager log -1 --pretty=%B`, but since there's no git repo in an action, we'll see:

Screen Shot 2021-08-06 at 12 15 09 AM

Again, no need to update that logic in this PR, but let's make some tasks.

packages/eas-cli/src/commands/branch/publish.ts
jonsamp: Yeah, this should do it! Thanks for adding this along the way

packages/eas-cli/src/commands/branch/view.ts
```diff
@@ -131,10 +110,15 @@ export default class BranchView extends Command {
       }));
     }
 
-    const UpdateBranch = await viewUpdateBranchAsync({
+    // return data.app?.byId.updateBranchByName;
```
jonsamp, line 113:
```suggestion
```

jonsamp (approved):
Left some questions and suggestions but this code looks good to me. Excited to use this feature!

packages/eas-cli/src/commands/branch/publish.ts
wkozyra95: .git repository should exist after checkout, if it doesn't then I think it's bug in github action.

Seems like it might be cause by old git version in the image 
https://github.com/actions/checkout/issues/335
https://github.com/actions/checkout/issues/126

wkozyra95 (approved):

packages/eas-cli/src/commands/branch/publish.ts
jkhales: It's also meant as a defensive check in case someone changes the code where it's set automatically.

Seems costless to have it serve both purposes.

packages/eas-cli/src/commands/branch/view.ts
jkhales: whoops, thanks.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: @jonsamp let me know if @wkozyra95 suggestions for the github action don't work out.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Well, maybe you are right.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Thanks for the discussion, will go with `--auto` for now!

PR #550, opened 2021-08-05
[eas-cli] Pass through build settings metadata context to app version resolver

brentvatne:
# Checklist

# Why

Fixes https://forums.expo.dev/t/eas-build-fails-after-upload-ios/55210

# How

I noticed that we were properly extracting the build settings in the metadata context, but that we weren't passing it into the version resolver as expected, so I fixed that.

# Test Plan

I made a new project, went to Info.plist and changed the "Bundle version string (short)" to `$(MARKETING_VERSION)`:
Screen Shot 2021-08-05 at 3 03 55 PM

Next, I set the version in the general tab for the target (this sets the `MARKETING_VERSION` build setting):
Screen Shot 2021-08-05 at 3 04 21 PM

`eas build -p ios` will fail after uploading the project with this error: `ReferenceError: MARKETING_VERSION is not defined`

If you apply this patch, it will not fail and the version will be properly set on the build job (eg: https://expo.dev/accounts/notbrent/projects/attempt-repro/builds/02d01ceb-88ba-4b0e-9e63-59ca026732ea)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Pass through metadata contexct to app version resolver
+
 ### 🧹 Chores
 
 ## [0.22.4](https://github.com/expo/eas-cli/releases/tag/v0.22.4) - 2021-08-02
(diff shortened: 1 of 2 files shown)


brentvatne:
went ahead and merged without review because it's a small easy to grok diff and i am confident in it

PR #554, opened 2021-08-10
[eas-cli] Stop using project name in build details URL

brentvatne:
# Checklist

# Why

A user reported this issue:

Screen_Shot_2021-08-10_at_3 41 57_PM

Notice that where the slug should be, there is a string that would be a valid slug -- this is actually the project name.

# How

We don't actually have `project.slug` available on `BuildFragment` so I shortened the URL and just linked to `/accounts/${owner}/builds/${build.id}` -- without including the project segment. This will redirect to the appropriate place. That said, we likely want to properly fix this as I'm unsure what the downsides are of using this specific path, cc @jonsamp.

# Test Plan

Run a build and use a `name` that is different from `slug`, notice that it is used in the build details URL and  the link will be invalid. Run again with this PR, notice that the URL will work but won't include the project `name` or `slug`.


jonsamp (approved):
this is okay for a quick fix. we should get the slug on the build fragment in eas-cli. happy to take that on. https://linear.app/expo/issue/ENG-1793/add-project-slug-to-the-buildfragment-in-eas-cli

PR #555, opened 2021-08-12
[eas-cli] support submit profiles

wkozyra95:
# Checklist

# Why

Add submit profiles, one of the steps that prepare us to support automatic submit after build

# How

- add `submit` fields with eas.json, only fields related to where it should be submitted are included. 
  - fields that define what should be submitted(latest, url, id ...) are not part of the profile because it will depend on the context e.g. if you run automatic submit after building it is obvious from context what should be included
  - submit profiles do not support `extend` keyword
  - submit profiles do not have common values between platforms
  - when evaluating eas.json for build and submit we are checking the entire file, so invalid value in submit might make build fail and invalid value in build might fail in submit
- fix reading package name for generic projects
- cli params take precedence over profile values
- remove defaults from cli interface(we can't differentiate between default and user-provided value after parsing), defaults are now applied just before

# Test Plan

run submit for ios and android, it failed because I used fake credentials, but I was able to verify if correct values were passed to submission service


packages/eas-cli/src/submissions/android/AndroidSubmitCommand.ts
```diff
@@ -28,13 +29,13 @@ class AndroidSubmitCommand {
...
     Log.addNewLineIfNone();
-    const submissionOptions = this.getAndroidSubmissionOptions();
+    const submissionOptions = await this.getAndroidSubmissionOptions();
     const submitter = new AndroidSubmitter(this.ctx, submissionOptions);
     await submitter.submitAsync();
   }
 
-  private getAndroidSubmissionOptions(): AndroidSubmissionOptions {
-    const androidPackageSource = this.resolveAndroidPackageSource();
+  private async getAndroidSubmissionOptions(): Promise<AndroidSubmissionOptions> {
```
barthap, line 37:
```suggestion
  private async getAndroidSubmissionOptionsAsync(): Promise<AndroidSubmissionOptions> {
```

packages/eas-cli/src/submissions/android/AndroidSubmitCommand.ts
```diff
@@ -28,13 +29,13 @@ class AndroidSubmitCommand {
 
   async runAsync(): Promise<void> {
     Log.addNewLineIfNone();
-    const submissionOptions = this.getAndroidSubmissionOptions();
+    const submissionOptions = await this.getAndroidSubmissionOptions();
```
barthap, line 32:
```suggestion
    const submissionOptions = await this.getAndroidSubmissionOptionsAsync();
```

packages/eas-json/src/EasJsonSchema.ts
```diff
@@ -72,10 +74,36 @@ const EasJsonBuildProfileSchema = CommonBuildProfileSchema.concat(
...
+  releaseStatus: Joi.string().valid(...Object.values(ReleaseStatus)),
+  verbose: Joi.boolean(),
+});
+
+const IosSubmitProfileSchema = Joi.object({
+  appleId: Joi.string(),
+  ascAppId: Joi.string(),
+
+  appleTeamId: Joi.string(),
+  itcTeamId: Joi.string(),
```
barthap, line 89: The `itcTeamId` is not used anywhere. If you took it from here https://github.com/expo/eas-cli/blob/5c7862a685e550bffd5a6fa1b914774fc8143d59/packages/eas-cli/src/submissions/ios/AppProduce.ts#L21
it is a leftover and can be deleted. It is now automatically resolved based on `appleTeamId` by 🥓 Lane

packages/eas-cli/src/commands/submit.ts
```diff
@@ -87,13 +99,11 @@ export default class BuildSubmit extends EasCommand {
 
     track: flags.enum({
       description: 'The track of the application to use',
-      default: 'internal',
```
barthap, line 90: Removing the `default` property affects command `--help` output. Maybe we can update the description, or somehow take the value from some single source of truth, and also provide them to the `options` variable down below. 🤷

barthap (commented):
Great job! 👏 

I'm wondering about usability of iOS profile fields like `sku`, `language`, `companyName` etc... since they're used once to create App Store Connect app and even if user changes them later in `eas.json`, it will have no impact on future submissions (unless BaconLane supports updating app entry in [`ensureAppExists()`](https://github.com/expo/eas-cli/blob/5c7862a685e550bffd5a6fa1b914774fc8143d59/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts#L148)).

packages/eas-cli/src/commands/submit.ts
```diff
@@ -141,25 +150,7 @@ export default class BuildSubmit extends EasCommand {
...
-        'app-name': appName,
-        'bundle-identifier': bundleIdentifier,
-        'company-name': companyName,
-
-        // common
-        ...flags
-      },
-    } = this.parse(BuildSubmit);
-
+    const flags = this.paseFlags();
```
brentvatne, line 153:
```suggestion
    const flags = this.parseFlags();
```

packages/eas-cli/src/commands/submit.ts
```diff
@@ -207,6 +204,53 @@ export default class BuildSubmit extends EasCommand {
       throw new Error(`Unsupported platform: ${platform}!`);
     }
   }
+
+  private paseFlags(): Flags {
```
brentvatne, line 208:
```suggestion
  private parseFlags(): Flags {
```

packages/eas-cli/src/commands/submit.ts
brentvatne: another possible way to handle this is to make `--profile` and other submission options (save for url, etc, which are not in profile) mutually exclusive. so if you profile `--profile` and `--track` then we error. this might be a little bit annoying for users though...

wkozyra95:
> I'm wondering about usability of iOS profile fields like sku, language, companyName etc... since they're used once to create App Store Connect app and even if user changes them later in eas.json, it will have no impact on future submissions

Yes I agree, I didn't want to change behavior for submit in this PR so currently it behaves exactly as flags would, and when running build command with automatic submit we should handle that separately either by aborting or just warning

barthap (approved):

PR #563, opened 2021-08-17
[eas-cli] Adds RudderStack

kgc00:
# Checklist

# Why

We'd like to send events from the EAS CLI to other destinations besides Amplitude. RudderStack is introduced for this purpose, and to keep consistent with how analytics are implemented across the Expo platform.

# How

- adds RudderStack to the analytics file
- updates analytics file to include proper write keys 
- adds some slight changes to how amplitude sent events previously to address an issue where some properties were not getting sent along with events

All analytics calls should now be sent to both Amplitude and RudderStack. This way we can compare volumes. If everything looks alright after some monitoring, I'll submit another PR to remove Amplitude.

# Test Plan

I tested by locally running `eas build` and ensuring that events were sent to the proper destinations.

### DIFF
--- a/packages/eas-cli/src/analytics.ts
+++ b/packages/eas-cli/src/analytics.ts
@@ -1,5 +1,6 @@
 import { Identify } from '@amplitude/identify';
 import * as Amplitude from '@amplitude/node';
+import RudderAnalytics from '@expo/rudder-sdk-node';
 import os from 'os';
 import { v4 as uuidv4 } from 'uuid';
 
@@ -18,8 +19,9 @@ const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
   linux: 'Linux',
 };
 
-let client: Amplitude.NodeClient | null = null;
-let userIdentifyCalled = false;
+let amplitudeClient: Amplitude.NodeClient | null = null;
+let rudderstackClient: RudderAnalytics | null = null;
+let userIdentified = false;
 let identifyData: {
   userId: string;
   deviceId: string;
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
   }
   const amplitudeEnabled = await UserSettings.getAsync('amplitudeEnabled', true);
   if (amplitudeEnabled) {
-    const apiKey =
+    const config =
       process.env.EXPO_STAGING || process.env.EXPO_LOCAL
-        ? 'cdebbc678931403439486c4750781544'
-        : '4ac443afd5073c0df6169291db1d3495';
-    client = Amplitude.init(apiKey, { retryClass: new Amplitude.OfflineRetryHandler(apiKey) });
+        ? {
+            // staging environment
+            amplitudeWriteKey: 'cdebbc678931403439486c4750781544',
+            rudderstackWriteKey: '1wpX20Da4ltFGSXbPFYUL00Chb7',
+            rudderstackDataPlaneURL: 'https://cdp.expo.dev',
+          }
+        : {
+            // prod environment
+            amplitudeWriteKey: '4ac443afd5073c0df6169291db1d3495',
+            rudderstackWriteKey: '1wpXLFxmujq86etH6G6cc90hPcC',
+            rudderstackDataPlaneURL: 'https://cdp.expo.dev',
+          };
+
+    amplitudeClient = Amplitude.init(config.amplitudeWriteKey, {
+      retryClass: new Amplitude.OfflineRetryHandler(config.amplitudeWriteKey),
+    });
+
+    rudderstackClient = new RudderAnalytics(
+      config.rudderstackWriteKey,
+      new URL('/v1/batch', config.rudderstackDataPlaneURL).toString(),
+      {
+        flushInterval: 300,
+      }
+    );
+    rudderstackClient.logger.silent = true;
   }
 }
 
@@ -53,6 +77,7 @@ export async function setUserDataAsync(
   if (!savedDeviceId) {
     await UserSettings.setAsync('amplitudeDeviceId', deviceId);
   }
+
   identifyData = {
     userId,
     deviceId,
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/analytics.ts
```diff
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
   }
   const amplitudeEnabled = await UserSettings.getAsync('amplitudeEnabled', true);
```
kgc00, line 35: I think we will need to rename this to 'analyticsEnabled'. I left it out of this PR initially because it introduces unnecessary complexity when the goal for this PR is just to compare event volumes. I'm planning to introduce this change in the followup PR to remove amplitude.

Also, there is the scenario where this setting gets renamed and a user had previously opted out. In this case, the user will be opted back into enabling analytics. And I'm not quite sure how we want to handle that.

packages/eas-cli/src/analytics.ts
```diff
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
...
   if (amplitudeEnabled) {
-    const apiKey =
+    const analyticsConfig =
       process.env.EXPO_STAGING || process.env.EXPO_LOCAL
```
kgc00, line 38: When testing locally, if `EXPO_LOCAL` or `EXPO_STAGING` is enabled the build command does not succeed. The CLI asks me to log in, then it throws an invalid username/pass error. Does this env variable cause all authenticated commands to fail? is this intentional?

packages/eas-cli/src/analytics.ts
```diff
@@ -63,30 +87,62 @@ export async function setUserDataAsync(
...
   }
+
+  if (rudderstackClient) {
+    const { userId, deviceId } = identifyData ?? {};
+
+    const identity = userId
+      ? { userId, anonymousId: undefined }
+      : {
+          userId: undefined,
+          anonymousId: deviceId ?? uuidv4(),
```
kgc00, line 122: image

packages/eas-cli/src/analytics.ts
```diff
@@ -63,30 +87,62 @@ export async function setUserDataAsync(
...
+      ...getContext(),
+    });
+  }
 }
 
 function ensureUserIdentified() {
-  if (client && !userIdentifyCalled && identifyData) {
-    client.identify(identifyData.userId, identifyData.deviceId, identifyData.identify);
-    userIdentifyCalled = true;
+  if (!(rudderstackClient && amplitudeClient && !userIdentifyCalled && identifyData)) {
```
kgc00, line 136: returning here if both clients are not initialized. this should not, but could break the amplitude identify flow. any pref for the implementation here? i think the other option is to rewrite it to continue if any client is initialized. i think i prefer how it is here because we do want both clients to be initialized.

packages/eas-cli/src/analytics.ts
```diff
@@ -63,41 +88,73 @@ export async function setUserDataAsync(
...
-function getContext() {
+function getAmplitudeContext() {
   const platform = PLATFORM_TO_ANALYTICS_PLATFORM[os.platform()] || os.platform();
   return {
     os_name: platform,
     os_version: os.release(),
     device_brand: platform,
     device_model: platform,
-    source_version: packageJSON?.version,
-    source: 'eas cli',
```
kgc00, line 100: these properties were not reaching amplitude because they [do not exist](https://developers.amplitude.com/docs/http-api-v2#properties-1) on the API. I moved them into the `properties` object so they will be transmitted to the destinations.

packages/eas-cli/src/analytics.ts
```diff
@@ -63,41 +88,73 @@ export async function setUserDataAsync(
...
+  };
+}
+
+function getRudderStackContext() {
```
kgc00, line 152: RudderStack and Segment have a [different API](https://docs.rudderstack.com/rudderstack-api/rudderstack-spec/common-fields) than Amplitude. I needed to update the context object to match this.

packages/eas-cli/src/analytics.ts
dsokal: Setting `EXPO_LOCAL` or `EXPO_STAGING` makes EAS CLI use another API endpoint than the production one.
If you want to use the local API, you need to have the local www running on your machine. Staging should work out of the box. 
Maybe you got the invalid username/password error because you used your production credentials? Staging doesn't use the same database as production, so if you want to use staging you should register another account at staging.expo.dev

packages/eas-cli/src/analytics.ts
dsokal: We could write logic that renames the `amplitudeEnabled` to something else (e.g. `analyticsEnabled`) and remove this after few months (when the majority of users upgrades to a new EAS CLI).

packages/eas-cli/src/analytics.ts
```diff
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
...
   const amplitudeEnabled = await UserSettings.getAsync('amplitudeEnabled', true);
   if (amplitudeEnabled) {
-    const apiKey =
+    const config =
       process.env.EXPO_STAGING || process.env.EXPO_LOCAL
-        ? 'cdebbc678931403439486c4750781544'
-        : '4ac443afd5073c0df6169291db1d3495';
-    client = Amplitude.init(apiKey, { retryClass: new Amplitude.OfflineRetryHandler(apiKey) });
+        ? {
+            // dev environment
```
dsokal, line 40:
```suggestion
            // staging environment
```

packages/eas-cli/src/analytics.ts
```diff
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
...
+            rudderstackDataPlaneURL: 'https://cdp.expo.dev',
+          };
+
+    amplitudeClient = Amplitude.init(config.amplitudeWriteKey, {
+      retryClass: new Amplitude.OfflineRetryHandler(config.amplitudeWriteKey),
+    });
+
+    rudderstackClient = new RudderAnalytics(
+      config.rudderstackWriteKey,
+      `${config.rudderstackDataPlaneURL}/v1/batch`,
```
dsokal, line 58:
```suggestion
      new URL('/v1/batch', config.rudderstackDataPlaneURL).toString(),
```

packages/eas-cli/src/analytics.ts
```diff
@@ -63,41 +88,73 @@ export async function setUserDataAsync(
...
-    userIdentifyCalled = true;
+  if (!(rudderstackClient && amplitudeClient) || userIdentified || !identifyData) {
+    return;
   }
+
+  amplitudeClient.identify(identifyData.userId, identifyData.deviceId, identifyData.identify);
+  rudderstackClient.identify({
+    userId: identifyData.userId,
+    anonymousId: identifyData.deviceId,
+  });
```
dsokal, line 138:
```suggestion
  if (!(rudderstackClient || amplitudeClient) || userIdentified || !identifyData) {
    return;
  }

  if (amplitudeClient) {
    amplitudeClient.identify(identifyData.userId, identifyData.deviceId, identifyData.identify);
  }
  
  if (rudderstackClient) {
    rudderstackClient.identify({
      userId: identifyData.userId,
      anonymousId: identifyData.deviceId,
    });
  }
```

I think this would match the logic in other functions better.

dsokal (approved):

packages/eas-cli/src/analytics.ts
```diff
@@ -32,11 +34,33 @@ export async function initAsync(): Promise<void> {
...
+    const config =
       process.env.EXPO_STAGING || process.env.EXPO_LOCAL
-        ? 'cdebbc678931403439486c4750781544'
-        : '4ac443afd5073c0df6169291db1d3495';
-    client = Amplitude.init(apiKey, { retryClass: new Amplitude.OfflineRetryHandler(apiKey) });
+        ? {
+            // dev environment
+            amplitudeWriteKey: 'cdebbc678931403439486c4750781544',
+            rudderstackWriteKey: '1wpX20Da4ltFGSXbPFYUL00Chb7',
+            rudderstackDataPlaneURL: 'https://cdp.expo.dev',
```
wkozyra95, line 43: if there is staging env this url should point to it, if not we probably do not want to send those events at all

wkozyra95 (approved):

packages/eas-cli/src/analytics.ts
kgc00: gotcha! thank you for the explanation. yeah, everything seems good to me then.
> Maybe you got the invalid username/password error because you used your production credentials
yup, this is definitely what happened

packages/eas-cli/src/analytics.ts
kgc00: There is no staging URL for RudderStack, but I'd like to send development events to:

- match how other codebases are set up
- enable testing of analytics pipelines and tasks related to analytics events

PR #566, opened 2021-08-18
[eas-cli] add ability to double publish

jkhales:
# Checklist

# Why

If a user specifies platform specific runtimes, we want to automatically publish multiple groups, one for each runtime.

# How

Updated graphQL to take an array here: https://github.com/expo/universe/pull/8036

The graphQL call is slightly more general than needed, for now we automatically create updateGroups based on the rtv's specified in the app.json.

# Test Plan

Tested in a demo repo pointed at the a local server running: https://github.com/expo/universe/pull/8036

Double publish output (Note added a message emphasizing that two groups are created as in discussions we agreed that this maybe a bit surprising if we were not clear about it):

Screen Shot 2021-08-18 at 8 43 32 AM

Confirmed that two update groups are created:

Screen Shot 2021-08-18 at 8 43 45 AM

Single publish still works 
Screen Shot 2021-08-18 at 8 44 20 AM
(added a `platforms` this PR as well):

Confirmed only 1 group created by single publish:

Screen Shot 2021-08-18 at 8 44 41 AM

### DIFF
--- a/packages/eas-cli/graphql-codegen.yml
+++ b/packages/eas-cli/graphql-codegen.yml
@@ -1,5 +1,5 @@
 overwrite: true
-schema: 'https://staging.exp.host/--/graphql'
+schema: 'http://localhost:3000/--/graphql'
 documents:
   - 'src/graphql/**/!(*.d).{ts,tsx}'
   - 'src/credentials/ios/api/graphql/**/!(*.d).{ts,tsx}'
--- a/packages/eas-cli/graphql-codegen.yml
+++ b/packages/eas-cli/graphql-codegen.yml
@@ -1,5 +1,5 @@
 overwrite: true
-schema: 'http://localhost:3000/--/graphql'
+schema: 'https://staging.exp.host/--/graphql'
 documents:
   - 'src/graphql/**/!(*.d).{ts,tsx}'
   - 'src/credentials/ios/api/graphql/**/!(*.d).{ts,tsx}'
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -181,7 +181,7 @@ export default class BranchPublish extends Command {
       runtimeVersion = getRuntimeVersionForSDKVersion(sdkVersion);
     }
 
-    let runtimeVersions: { [keyof: string]: string };
+    let runtimeVersions: Record<string, string>;
     const iOSRuntimeVersion = (exp as any).ios?.runtimeVersion; // TODO-JJ remove cast to any
     const androidRuntimeVersion = (exp as any).android?.runtimeVersion; // TODO-JJ remove cast to any
     switch (platformFlag) {
@@ -365,7 +365,7 @@ export default class BranchPublish extends Command {
       }));
     }
 
-    const runtimeToPlatformMapping: { [keyof: string]: string[] } = {};
+    const runtimeToPlatformMapping: Record<string, string[]> = {};
     for (const runtime of new Set(Object.values(runtimeVersions))) {
       runtimeToPlatformMapping[runtime] = Object.entries(runtimeVersions)
         .filter(pair => pair[1] === runtime)
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -1,4 +1,4 @@
-import { getConfig, getDefaultTarget } from '@expo/config';
+import { ExpoConfig, getConfig, getDefaultTarget } from '@expo/config';
 import { getRuntimeVersionForSDKVersion } from '@expo/sdk-runtime-versions';
 import { Command, flags } from '@oclif/command';
 import assert from 'assert';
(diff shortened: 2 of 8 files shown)


packages/eas-cli/graphql-codegen.yml
```diff
@@ -1,5 +1,5 @@
 overwrite: true
-schema: 'https://staging.exp.host/--/graphql'
+schema: 'http://localhost:3000/--/graphql'
```
jkhales, line 2: revert

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -5,7 +5,7 @@ import assert from 'assert';
 import chalk from 'chalk';
 import dateFormat from 'dateformat';
 import gql from 'graphql-tag';
-import { uniqBy } from 'lodash';
+import { groupBy, uniqBy, update } from 'lodash';
```
jkhales, line 8: revert

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -332,32 +366,64 @@ export default class BranchPublish extends Command {
...
-    try {
-      newUpdateGroup = await PublishMutation.publishUpdateGroupAsync({
+    const runtimeToPlatformMapping: { [keyof: string]: string[] } = {};
+    for (const runtime of new Set(Object.values(runtimeVersions))) {
+      runtimeToPlatformMapping[runtime] = Object.entries(runtimeVersions)
+        .filter(pair => pair[1] === runtime)
+        .map(pair => pair[0]);
+    }
+
+    const updateGroups = Object.entries(runtimeToPlatformMapping).map(([runtime, platforms]) => {
```
jkhales, line 376:
```suggestion
    // Split the updates into different updateGroups based on the platform specific runtime versions
    const updateGroups = Object.entries(runtimeToPlatformMapping).map(([runtime, platforms]) => {
```

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -282,6 +315,7 @@ export default class BranchPublish extends Command {
       }
       Log.withTick(publicationPlatformMessage);
 
+      const updateInfoGroup: UpdateInfoGroup = {};
```
jkhales, line 318:
```suggestion
      const unsortedUpdateInfoGroups: UpdateInfoGroup = {};
```

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -332,32 +365,74 @@ export default class BranchPublish extends Command {
...
     } else {
-      Log.log(
-        formatFields([
-          { label: 'branch', value: branchName },
-          { label: 'runtime version', value: runtimeVersion },
-          { label: 'update group ID', value: newUpdateGroup.group },
-          { label: 'message', value: message },
-        ])
-      );
+      if (new Set(newUpdates.map(update => update.group)).size > 1) {
```
wkozyra95, line 409: I understand that you need Set to remove duplicates, but if just checking if it at least one you can just check newUpdates length

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -332,32 +365,74 @@ export default class BranchPublish extends Command {
...
+      newUpdates = await PublishMutation.publishUpdateGroupAsync(updateGroups);
       publishSpinner.succeed('Published!');
     } catch (e) {
       publishSpinner.fail('Failed to published updates');
       throw e;
     }
 
     if (jsonFlag) {
-      Log.log(JSON.stringify(newUpdateGroup));
+      Log.log(JSON.stringify(newUpdates));
```
wkozyra95, line 407: unrelated, I added some utils to better support json flags https://github.com/expo/eas-cli/commit/094c04c7f07ba1f1ee799fafb1acbdeb4e22599a

enableJsonOutput - redirects everything to stderr
printJsonOnlyOutput - prints to real stdout

this allows someone to pipe just JSON to stdout without parsing any other non-json input
it does not work with prompts

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -181,11 +181,44 @@ export default class BranchPublish extends Command {
       runtimeVersion = getRuntimeVersionForSDKVersion(sdkVersion);
     }
 
-    if (!runtimeVersion) {
+    let runtimeVersions: { [keyof: string]: string };
```
wkozyra95, line 184: this function is getting quite big, consider splitting it/extracting some parts

packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -332,32 +365,74 @@ export default class BranchPublish extends Command {
       }));
     }
 
-    let newUpdateGroup;
-    const publishSpinner = ora('Publishing...').start();
-    try {
-      newUpdateGroup = await PublishMutation.publishUpdateGroupAsync({
+    const runtimeToPlatformMapping: { [keyof: string]: string[] } = {};
```
wkozyra95, line 368:
```suggestion
    const runtimeToPlatformMapping: Record<string, string[]> = {};
```

packages/eas-cli/src/commands/branch/publish.ts
wkozyra95:
```suggestion
    let runtimeVersions: Record<string, string>;
```

wkozyra95 (approved):

packages/eas-cli/src/commands/branch/publish.ts
jkhales: We may have a *single* update group with multiple updates. In that case we don't want to print this message.

Thus checking `newUpdates.length` won't work.

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Nice, this seems generally useful.

packages/eas-cli/src/commands/branch/publish.ts
wkozyra95: Ahh sorry, I read that incorrectly

packages/eas-cli/src/commands/branch/publish.ts
jkhales: Good point, extracted a bunch of this PR into a function called `getRuntimeVersionObject`.

PR #606, opened 2021-09-10
[eas-cli] fix secrets mutations

FiberJW:
# Checklist

# Why

We got a report that force-creation of secrets wasn't working. https://forums.expo.dev/t/cant-delete-or-force-update-a-secret/56640 

This report also included other mutation bugs.

# How

- Supplied the correct parameter to a secrets lookup call
- Fixed mutation helpers (it was previously calling a query with a mutation string rather than a proper mutation).

# Test Plan

The following reproduction plan should no longer result in errors:

```
eas secret:create --scope=account --name=WHATEVER --value=a_secret

eas secret:create --scope=account --name=WHATEVER --value=a_new_secret --force

eas secret:delete --id=<my-secret-id>

eas secret:create --scope=project --name=WHATEVER_PROJECT_SECRET --value=a_secret
```

### DIFF
--- a/packages/eas-cli/src/commands/secret/create.ts
+++ b/packages/eas-cli/src/commands/secret/create.ts
@@ -160,9 +160,7 @@ export default class EnvironmentSecretCreate extends Command {
       }
 
       if (force) {
-        const existingSecrets = await EnvironmentSecretsQuery.byAcccountNameAsync(
-          ownerAccount.name
-        );
+        const existingSecrets = await EnvironmentSecretsQuery.byAccountNameAsync(ownerAccount.name);
         const existingSecret = existingSecrets.find(secret => secret.name === name);
 
         if (existingSecret) {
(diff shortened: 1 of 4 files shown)


brentvatne (approved):
looks good to fix this for now -- how do we prevent it from happening again in the future?

packages/eas-cli/src/commands/secret/create.ts
```diff
@@ -159,11 +160,14 @@ export default class EnvironmentSecretCreate extends Command {
       }
 
       if (force) {
-        const existingSecrets = await EnvironmentSecretsQuery.byAcccountNameAsync(projectId);
+        const existingSecrets = await EnvironmentSecretsQuery.byAcccountNameAsync(
```
ide, line 163:
```suggestion
        const existingSecrets = await EnvironmentSecretsQuery.byAccountNameAsync(
```

ide:
What caused this issue? Was it a breaking change between the EAS server API and EAS CLI, or was this code broken to begin with?

FiberJW:
It looks like this code was broken to begin with 😬

ide:
Let's get this fix out because it's affecting developers and is easy to publish, and then follow up with regression tests.

PR #641, opened 2021-09-24
[eas-cli] creds manager: stop prompting every iteration

quinlanj:
# Checklist

# Why
Stop asking user for android build profile every time they choose an action. Instead, prompt the user once and then use the chosen profile for the rest of the actions.

Screen Shot 2021-09-23 at 6 07 49 PM

# Test Plan

- [x] manually tested
- [x] current tests still pass

### DIFF
--- a/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
+++ b/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
@@ -145,17 +145,17 @@ export class ManageAndroid {
   constructor(private callingAction: Action) {}
 
   async runAsync(ctx: Context, currentActions: ActionInfo[] = highLevelActions): Promise<void> {
+    const accountName = ctx.hasProjectContext
+      ? getProjectAccountName(ctx.exp, ctx.user)
+      : ensureActorHasUsername(ctx.user);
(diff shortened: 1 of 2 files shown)


dsokal (commented):
LGTM

packages/eas-cli/src/credentials/manager/ManageAndroid.ts
```diff
@@ -145,17 +145,17 @@ export class ManageAndroid {
   constructor(private callingAction: Action) {}
 
   async runAsync(ctx: Context, currentActions: ActionInfo[] = highLevelActions): Promise<void> {
+    const accountName = ctx.hasProjectContext
```
wkozyra95, line 148: Not sure if it's not done already, but if we don't prompt for every action maybe we should print relevant info e.g package name/bundle identifier so the users know which profile/app they are configuring?

wkozyra95 (approved):

packages/eas-cli/src/credentials/manager/ManageAndroid.ts
quinlanj: Yep, we do this already :)

PR #647, opened 2021-09-24
[eas-cli] change export style

quinlanj:
# Checklist

# famous quote
'We're trying to move away from this weird convention of exports' - dominik

# Why

Followup PR from https://github.com/expo/eas-cli/pull/642#discussion_r715401434 . Made changes except for:
- default exports 
-  files that import a resource and immediately export it

# How

Change export style from 
```
class Foo {...}
export { Foo };
```
to
```
export class Foo {...}
```

# Test Plan


dsokal (approved):
That's great!

PR #655, opened 2021-09-28
[eas-cli] Patch @expo/rudder-sdk-node

kgc00:
# Checklist

# Why

Updates the analytics client to the latest version. Will reduce the size and remove unneeded dependencies.

# How

Patched the npm package. Made small adjustments to code around the new async flush.

# Test Plan

Ran some commands, made sure events were showing up in the backend

### DIFF
--- a/packages/eas-cli/src/analytics.ts
+++ b/packages/eas-cli/src/analytics.ts
@@ -65,7 +65,6 @@ export async function initAsync(): Promise<void> {
         flushInterval: 300,
       }
     );
-    rudderstackClient.logger.silent = true;
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/analytics.ts
```diff
@@ -114,13 +113,11 @@ function ensureUserIdentified(): void {
...
   }
 
-  if (rudderstackClient) {
-    rudderstackClient.identify({
-      userId: identifyData.userId,
-      anonymousId: identifyData.deviceId,
-      traits: identifyData.traits,
-    });
-  }
+  rudderstackClient.identify({
```
kgc00, line 116: line ~110 is `if (!rudderclient) return;`

packages/eas-cli/src/analytics.ts
```diff
@@ -65,7 +65,6 @@ export async function initAsync(): Promise<void> {
         flushInterval: 300,
       }
     );
-    rudderstackClient.logger.silent = true;
```
kgc00, line 68: package is silent by default

dsokal (approved):
Thanks!

PR #657, opened 2021-09-29
[eas-cli] fix ios print formats

quinlanj:
# Checklist

# Why

Fix the way we print iOS credentials, followup from https://github.com/expo/eas-cli/pull/649#discussion_r717252597 
Screen Shot 2021-09-28 at 4 22 45 PM

# Test Plan

- [x] updated existing snapshots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Always use async `fs` functions. ([#652](https://github.com/expo/eas-cli/pull/652) by [@dsokal](https://github.com/dsokal))
-- Improve iOS credentials printing ([#657](https://github.com/expo/eas-cli/pull/657) by [@quinlanj](https://github.com/quinlanj))
+- Improve iOS credentials printing. ([#657](https://github.com/expo/eas-cli/pull/657) by [@quinlanj](https://github.com/quinlanj))
 
 ## [0.29.0](https://github.com/expo/eas-cli/releases/tag/v0.29.0) - 2021-09-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Always use async `fs` functions. ([#652](https://github.com/expo/eas-cli/pull/652) by [@dsokal](https://github.com/dsokal))
+- Improve iOS credentials printing ([#657](https://github.com/expo/eas-cli/pull/657) by [@quinlanj](https://github.com/quinlanj))
 
 ## [0.29.0](https://github.com/expo/eas-cli/releases/tag/v0.29.0) - 2021-09-28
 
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/credentials/ios/utils/printCredentials.ts
```diff
@@ -66,28 +71,46 @@ export function displayIosCredentials(
...
+      Log.log(
+        formatFields([{ label: 'Target', value: targetName }], { labelFormat: chalk.cyan.bold })
+      );
     }
-    Log.log(`  Bundle Identifier: ${chalk.bold(bundleIdentifier)}`);
+    Log.log(
+      formatFields([{ label: 'Bundle Identifier', value: bundleIdentifier }], {
+        labelFormat: chalk.cyan.bold,
+      })
+    );
```
barthap, line 91: `formatFields()` should be called once for each "field group", because it aligns values to be in the same column, basing on the longest label.

So if should be sth like:
```ts
const fields = [];
if (isMultitarget) {
  fields.push({ label: 'Target', value: targetName });
}
fields.push({ label: 'Bundle Identifier', value: bundleIdentifier });
// ... rest of fields from this group
formatFields(fields, { labelFormat: chalk.cyan.bold });
```
Probably that's why the first three fields (project, Bundle identifier, Apple team) are misaligned on the screenshot. 😉

barthap (changes requested):
Just one note 😉 

_I approved the android part, I haven't found this issue there, but if there is, please fix it too._

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Always use async `fs` functions. ([#652](https://github.com/expo/eas-cli/pull/652) by [@dsokal](https://github.com/dsokal))
+- Improve iOS credentials printing ([#657](https://github.com/expo/eas-cli/pull/657) by [@quinlanj](https://github.com/quinlanj))
```
dsokal, line 19:
```suggestion
- Improve iOS credentials printing. ([#657](https://github.com/expo/eas-cli/pull/657) by [@quinlanj](https://github.com/quinlanj))
```
Sorry 😅

packages/eas-cli/src/credentials/ios/utils/printCredentials.ts
quinlanj: Screen Shot 2021-09-29 at 2 01 41 PM

I've updated the PR to push everything to the `fields` array, and do `Log.log(formatFields(fields, { labelFormat: chalk.cyan.bold }));` at the very end of the function. One challenge is handling the logs of new lines -- we originally did this with `Log.newLine()`. But now, we have to track this in `fields`, otherwise all the new lines will get printed first. Kind of hacky, but I've just replaced them all with `fields.push({label:'', value: ''})`

barthap: It looks like the best solution is something in the middle:
```ts
// pseudo-code
Log.log('iOS credentials');
iosFields.push(project);
iosFields.push(bundleId);
iosFields.push(appleTeam);
Log.log(formatFields(iosFields);

Log.newLine();
Log.log('Push key');
pushFields.push(devPortalId);
// ...
Log.log(formatFields(pushFields);

Log.newLine();
// the same for each "field group"
```

If any field or title is too long, I liked your hack from Android:
```ts
fields.push({ label: 'some very long long label', value: '' });
fields.push({ label: '', value: 'a longish value here' });
```

barthap: Ah, I'm sorry, now I see what you mean 😅 I shouldn't have review PRs right after I wake up.

so the problem was with multitarget project? then we have 2 options:
- try with separate group for each target. Target name can (but does not have to) be then displayed as normal `Log.log()`
- leave your newline hack, I don't mind leaving it as is

PR #658, opened 2021-09-29
[eas-cli] remove google service account key

quinlanj:
# Checklist

# Why

This PR:

- allows the user to remove a GoogleServiceAccountKey from their account
- adds this action to the credentials management workflow

# Test Plan

- [x] new tests pass

### DIFF
--- a/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
+++ b/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
@@ -33,7 +33,7 @@ export class SelectAndRemoveGoogleServiceAccountKey {
 export class RemoveGoogleServiceAccountKey {
   constructor(private googleServiceAccountKey: GoogleServiceAccountKeyFragment) {}
 
-  public async runAsync(ctx: Context): Promise<void> {
+  public async runAsync(ctx: CredentialsContext): Promise<void> {
     if (ctx.nonInteractive) {
       throw new Error(`Cannot remove Google Service Account Keys in non-interactive mode`);
     }
--- a/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
+++ b/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
@@ -13,7 +13,7 @@ export class SelectAndRemoveGoogleServiceAccountKey {
       this.account
     );
     if (gsaKeyFragments.length === 0) {
-      Log.error("There aren't any Google Service Account Keys associated with your account.");
+      Log.warn("There aren't any Google Service Account Keys associated with your account.");
       return;
     }
 
--- a/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
+++ b/packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
@@ -1,0 +1,47 @@
+import { GoogleServiceAccountKeyFragment } from '../../../graphql/generated';
+import Log from '../../../log';
+import { confirmAsync } from '../../../prompts';
+import { Account } from '../../../user/Account';
+import { Context } from '../../context';
+import { selectGoogleServiceAccountKeyAsync } from '../utils/googleServiceAccountKey';
+
+export class SelectAndRemoveGoogleServiceAccountKey {
+  constructor(private account: Account) {}
+
+  async runAsync(ctx: Context): Promise<void> {
+    const gsaKeyFragments = await ctx.android.getGoogleServiceAccountKeysForAccountAsync(
+      this.account
+    );
+    if (gsaKeyFragments.length === 0) {
+      Log.error("There aren't any Google Service Account Keys associated with your account.");
+      return;
+    }
+
+    const selected = await selectGoogleServiceAccountKeyAsync(gsaKeyFragments);
+    await new RemoveGoogleServiceAccountKey(selected).runAsync(ctx);
+    Log.succeed('Removed Google Service Account Key');
+    Log.newLine();
+  }
+}
+
+export class RemoveGoogleServiceAccountKey {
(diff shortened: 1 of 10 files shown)


packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export class SelectAndRemoveGoogleServiceAccountKey {
+  constructor(private account: Account) {}
+
+  async runAsync(ctx: Context): Promise<void> {
+    const gsaKeyFragments = await ctx.android.getGoogleServiceAccountKeysForAccountAsync(
+      this.account
+    );
+    if (gsaKeyFragments.length === 0) {
+      Log.error("There aren't any Google Service Account Keys associated with your account.");
```
dsokal, line 16: Maybe this should be a warning instead? 🤔 
```suggestion
      Log.warn("There aren't any Google Service Account Keys associated with your account.");
```

packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
```diff
@@ -0,0 +1,47 @@
...
+import Log from '../../../log';
+import { confirmAsync } from '../../../prompts';
+import { Account } from '../../../user/Account';
+import { Context } from '../../context';
+import { selectGoogleServiceAccountKeyAsync } from '../utils/googleServiceAccountKey';
+
+export class SelectAndRemoveGoogleServiceAccountKey {
+  constructor(private account: Account) {}
+
+  async runAsync(ctx: Context): Promise<void> {
```
dsokal, line 11: Does running this action make sense in non-interactive mode? 
Maybe we should add a guard like this https://github.com/expo/eas-cli/pull/658/files#diff-4b3d674e01626e2dd920fba191efeec8e73653b2576b86ba96d5ac4f02393ffaR31

packages/eas-cli/src/credentials/android/utils/googleServiceAccountKey.ts
```diff
@@ -34,3 +38,44 @@ export function readAndValidateServiceAccountKey(keyJsonPath: string): GoogleSer
...
+
+function formatGoogleServiceAccountKey({
+  projectIdentifier,
+  privateKeyIdentifier,
+  clientEmail,
+  clientIdentifier,
+  updatedAt,
+}: GoogleServiceAccountKeyFragment): string {
+  let line: string = '';
+  line += `Client Email: ${clientEmail}, Project Id: ${projectIdentifier}`;
```
dsokal, line 75:
```suggestion
  let line = `Client Email: ${clientEmail}, Project Id: ${projectIdentifier}`;
```

packages/eas-cli/src/credentials/android/utils/googleServiceAccountKey.ts
```diff
@@ -34,3 +38,44 @@ export function readAndValidateServiceAccountKey(keyJsonPath: string): GoogleSer
...
+  });
+  return chosenKey;
+}
+
+function sortGoogleServiceAccountKeysByUpdatedAtDesc(
+  keys: GoogleServiceAccountKeyFragment[]
+): GoogleServiceAccountKeyFragment[] {
+  return keys.sort(
+    (keyA, keyB) =>
+      new Date(keyB.updatedAt).getMilliseconds() - new Date(keyA.updatedAt).getMilliseconds()
```
dsokal, line 63:
```suggestion
      new Date(keyB.updatedAt) - new Date(keyA.updatedAt)
```

Screenshot 2021-09-29 at 10 21 27

dsokal: ;)

dsokal (approved):

packages/eas-cli/src/credentials/android/utils/googleServiceAccountKey.ts
wkozyra95: I think that using getTime is cleaner than relying on js to auto-convert date to number

wkozyra95: also the same mistake was already merged in https://github.com/expo/eas-cli/commit/4674cbebc5b99c38e6bd14e11b953e57e96e58fd

packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
wkozyra95: guards like that should be needed only if action can be run in a context other than credentials manager, `eas credentials` is always interactive.

wkozyra95 (approved):

packages/eas-cli/src/credentials/android/actions/RemoveGoogleServiceAccountKey.ts
dsokal: @wkozyra95 How can we guarantee that no one will ever use this in another context?

quinlanj: I've updated this to add the `non-interactive` guard at the beginning for fast failure, in case this is called in the wrong context

packages/eas-cli/src/credentials/android/utils/googleServiceAccountKey.ts
quinlanj: Good catch! I've changed this to `getTime` because typescript errors when you perform arithmetic on Date objects. Though it is worth noting vanilla javascript auto-converts dates to a number in this case.

quinlanj: This PR refactors https://github.com/expo/eas-cli/commit/4674cbebc5b99c38e6bd14e11b953e57e96e58fd as well, so fixing the mistake here is sufficient :)

PR #688, opened 2021-10-12
[eas-cli][submit] Ability to select a build from the list

barthap:
# Why

Closes ENG-1952

# How

- Command-line option `--latest` works with no changes.
- In interactive mode, the prompt no longer have ~~`Latest finished build from EAS`~~ option, but `A selected build from EAS` option which displays a choice of _4 latest, finished builds, with `Store` distribution and matching platform._ A `none of the above` option is available, which falls back to the previous prompt.
- Expired builds are displayed on the list, but are marked as ~~`Disabled`~~ - they cannot be chosen. If all artifacts are expired, the list is not shown at all - a proper message is displayed instead.
- Displayed info:
   ```
   - Build ID, Finished at
      App version, Build/Version Number, Runtime Version (if available), SDK version (if available)
      Build Profile, Release Channel
   ```

# Test Plan

Manually + added test cases

---

Screenshot 2021-10-12 at 19 28 23
Screenshot 2021-10-12 at 19 30 15
Screenshot 2021-10-12 at 19 30 52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - Add App Store Connect Api Key support to iOS submissions. ([#678](https://github.com/expo/eas-cli/pull/678) by [@quinlanj](https://github.com/quinlanj))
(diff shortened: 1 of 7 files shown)


barthap:
/changelog-entry new-feature Added ability to select a build from a list in `eas submit` interactive mode.

PR #727, opened 2021-11-03
Replace eas:secrets with eas:secret

ryan-blunden:
# Why

`eas secrets` usage wasn't updated to `eas secret`.

# How

Replaced instances with `eas:secret`.


dsokal (approved):
Thank you!

PR #745, opened 2021-11-06
[eas-cli] only prompt for appleIdUsername if authenticating via asp

quinlanj:
# Checklist

# Why

Only prompt for Apple Id Username (ie) `quin@expo.io` if we are authenticating via App Specific Password. We don't need this information if we are using ASC Api Keys.

# TODO

- [x] Land and deploy https://github.com/expo/universe/pull/8518

# Test Plan

- [x] tests still pass
- [x] manual sanity check

### DIFF
--- a/packages/eas-cli/src/submit/ios/AppSpecificPasswordSource.ts
+++ b/packages/eas-cli/src/submit/ios/AppSpecificPasswordSource.ts
@@ -1,3 +1,9 @@
+import { Platform } from '@expo/eas-build-job';
+import getenv from 'getenv';
+
+import { promptAsync } from '../../prompts';
+import UserSettings from '../../user/UserSettings';
+import { SubmissionContext } from '../context';
 export enum AppSpecificPasswordSourceType {
   userDefined,
 }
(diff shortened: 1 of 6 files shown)


dsokal (approved):
👍

packages/eas-cli/src/submit/ios/AppSpecificPasswordSource.ts
```diff
@@ -11,15 +17,62 @@ interface AppSpecificPasswordUserDefinedSource extends AppSpecificPasswordSource
   appSpecificPassword: string;
 }
 
+export type AppSpecificPassword = {
```
barthap, line 20:
```suggestion
export interface AppSpecificPasswordCredentials {
```
- I think we prefer interfaces over types
- IMO the `Credentials` suffix will make this name more accurate

barthap (approved):

quinlanj:
/changelog-entry bug-fix Only prompt for Apple Id username if authenticating with an App Specific Password.

PR #750, opened 2021-11-09
Revert "fix working with local api server (#747)"

ide:
Why
---
Make local development behave more closely to production and use the same API route paths. I made 127.0.0.1 behave the same way as api.expo.test and localhost. Another fix would have been to change 127.0.0.1 to localhost in EAS CLI, but it's a good idea to make the API server treat localhost and 127.0.0.1 similarly (look up PR 8542).

How
---
This reverts commit 5ba3cc66237d8ca8984d1fbd1efff65097f5feaf.

Test Plan
---------
Started the API server locally, then started EAS CLI locally with `EXPO_LOCAL=1 yarn eas login`. Logged in with my staging credentials and saw the API server locally print out that it had made new database connections in the log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Fix auto-submitting with `--auto-submit-with-profile`. ([#748](https://github.com/expo/eas-cli/pull/748) by [@dsokal](https://github.com/dsokal))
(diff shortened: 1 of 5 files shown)


ide:
/changelog-entry bug-fix Use same URL paths for local development of EAS CLI as production

ide:
/changelog-entry chore Use same URL paths for local development of EAS CLI as production

dsokal (approved):

PR #789, opened 2021-11-24
[eas-cli] check EAS Update URL is correctly configured before publish

jkhales:
# Checklist

# Why

Prevent users from publishing without having configured their updates URL. 

This will prevent frustrating foot guns.

# How

Check app.json and throw if incorrectly configured.

Note: the question of how to keep the url in app.json and native config files is already managed here: https://github.com/expo/expo-cli/blob/master/packages/config-plugins/src/android/Updates.ts#L199

# Test Plan

Tested in demo repo

### DIFF
--- a/packages/eas-cli/src/commands/branch/publish.ts
+++ b/packages/eas-cli/src/commands/branch/publish.ts
@@ -40,6 +40,8 @@ import { createUpdateBranchOnAppAsync } from './create';
 import { listBranchesAsync } from './list';
 import { viewUpdateBranchAsync } from './view';
 
+const EAS_UPDATE_URL = 'https://u.expo.dev';
+
 export const defaultPublishPlatforms: PublishPlatform[] = ['android', 'ios'];
 type PlatformFlag = PublishPlatform | 'all';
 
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/commands/branch/publish.ts
```diff
@@ -473,3 +475,19 @@ function formatUpdateTitle(
...
+}
+
+async function checkEASUpdateURLIsSetAsync(exp: ExpoConfig): Promise<void> {
+  const configuredURL = exp.updates?.url;
+  const expectedURL = await getEASUpdateURLAsync(exp);
+
+  if (configuredURL !== expectedURL) {
+    throw new Error(
+      `The update URL is incorrectly configured for EAS Update. Please set updates.url to ${expectedURL} in your app.json.`
+    );
```
jkhales, line 491: One DX question is whether a throw is too aggressive here. I considered just logging a warning, but those are easier to miss and I can't think of a situation in which a user would want to publish without a correctly configured URL.

wkozyra95: what is the point of that field if we can get that value using other means and overriding it is not allowed?

wkozyra95 (approved):

packages/eas-cli/src/commands/branch/publish.ts
jkhales: I'm planning to allow overwriting it here: https://github.com/expo/eas-cli/pull/800/files

I didn't want to automatically overwrite it here as the developer maybe wants to use classic updates and the mistake was  note the URL, but using `eas branch:publish` instead of `expo publish`

PR #791, opened 2021-11-24
docs: fix broken links to all command source files in readme

kbrandwijk:
# Checklist

# Why

They pointed to .js, but they're all .ts.

# How

Manually.

# Test Plan

I clicked them all.


brentvatne:
the readme is generated using `oclif-dev readme` ([source](https://github.com/oclif/dev-cli/blob/master/src/commands/readme.ts)) - we need to check why oclif is not properly generating paths to ts files

kbrandwijk:
image
Let's just say I had a premonition...

dsokal:
@kbrandwijk thanks for the PR. As Brent mentioned, the readme is auto-generated so this is a one-time fix.

Fortunately, there is a way to fix it. We have a script that we run to start the release process - https://github.com/expo/eas-cli/blob/main/packages/eas-cli/package.json#L178 A part of the script is responsible for patching the readme. The script is here https://github.com/expo/eas-cli/blob/main/packages/eas-cli/patch-readme.js I think the proper fix is just a matter of adding a new `.replace` call. It's not ideal but it does not require fixing `oclif-dev readme`.

kbrandwijk:
> @kbrandwijk thanks for the PR. As Brent mentioned, the readme is auto-generated so this is a one-time fix.
> 
> Fortunately, there is a way to fix it. We have a script that we run to start the release process - https://github.com/expo/eas-cli/blob/main/packages/eas-cli/package.json#L178 A part of the script is responsible for patching the readme. The script is here https://github.com/expo/eas-cli/blob/main/packages/eas-cli/patch-readme.js I think the proper fix is just a matter of adding a new `.replace` call. It's not ideal but it does not require fixing `oclif-dev readme`.

@dsokal Thanks for the pointers. I'll follow up with a commit with the fix to the script.

kbrandwijk:
Apparently, according to https://github.com/oclif/dev-cli/blob/master/src/commands/readme.ts#L210, oclif-dev looks for the project devdepencies to see if typescript is present, and if so, generates links to the .ts instead. 

So instead of adding another manual replace patch to the existing ones, I went that route, and verified that the generated commands block does indeed link to the .ts source now.

dsokal (commented):
Looks good!

Currently, `typescript` is installed only in the repo root. With this PR, we'd be a little inconsistent with the configuration across other workspaces. Could you please also add `typescript` to package.json in `packages/eas-json`?

PR #830, opened 2021-12-08
[eas-cli] change eas update:publish to eas update

jkhales:
# Checklist

# Why

Make EAS Update adhere to the cli format of EAS Build and EAS Submit:

currently `eas build` and `eas submit` trigger actions.

This PR makes `eas update` trigger a publish, instead of `eas update:publish`
# How

Move file.
# Test Plan

Tested in demo repo. Help menu is still accessible via `eas update --help`


jkhales:
/changelog-entry breaking-change [eas-cli] change eas update:publish to eas update

jonsamp (approved):

wkozyra95 (approved):
I don't have strong opinion that, but not that you can alias  one command to another. For example `eas submit` is also aliased to `eas build:submit`

https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commands/submit.ts#L47

jkhales:
> I don't have strong opinion that, but note that you can alias one command to another. For example `eas submit` is also aliased to `eas build:submit`
> 
> https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commands/submit.ts#L47

That makes sense, not sure an alias here would provide more clarity and since it is not yet deployed we don't have to worry about legacy users other than @jonsamp (sorry jon. 😆 )

Issue #824, opened 2021-12-06
Error ITMS-4088 when submitting iOS builds from CI


pothos-dev:
### Summary

In our CI build, we're building and submitting an Expo App via CLI:
```
eas build --non-interactive --platform=all --profile=preview --auto-submit
```

This works for Android, but iOS submission fails with:
```
ERROR ITMS-4088: "The organization ((our company)) does not have permission to modify the application with the Apple ID 123456789. Please try again, using an account for an organization with access to this application." at SoftwareAssets
Invalid AdamId: 123456789.
```
Same same issue occurs if I split the `eas build` and `eas submit` commands, so it does not seem to be related to `--auto-submit`.

The same exact command works when run from my dev machine.

We're developing an app for another company ((their company)), so there are multiple Developer Teams involved.
The Distribution Certificate and Provisioning Profile assigned to the app both relate to ((their company)) in the expo.dev Web UI.
Our `eas.json` specifies the `ascAppId`, but no apple team ids.

I noticed that the output of the `submit` command is different on CI:
On my machine:
```
- Linking to project @the/project
✔ Linked to project @the/project(​https://expo.dev/accounts/the/projects/project)

ASC App ID:  1234567890
Project ID:  ...
Apple ID:   our.apple.id
```

On the CI:
```
- Linking to project @the/project
✔ Linked to project @the/project(​https://expo.dev/accounts/the/projects/project)

Looking up credentials configuration for com.bundle.identifier...
✔ App Store Connect API Key already set up.
Using API Key ID: 7A8W74KWY2 ([Expo] EAS Submit p548nhOiPM)     

ASC App ID:                 1234567890
Project ID:                 ....
App Store Connect API Key:  
    Key Name  :  [Expo] EAS Submit p548nhOiPM
    Key ID    :  7A8W74KWY2
    Key Source:  EAS servers
```

So this looks like EAS is picking up some environmental configuration on my dev machine that is not available on CI.

### Managed or bare?

Managed

### Environment

Environment on our CI container:
```
  EAS CLI 0.39.0 environment info:
    System:
      OS: Linux 4.15 Alpine Linux
      Shell: 1.33.1 - /bin/ash
    Binaries:
      Node: 14.17.6 - /usr/bin/node
      npm: 7.17.0 - /usr/bin/npm
    Utilities:
      Git: 2.32.0 - /usr/bin/git
    npmPackages:
      expo: ^43.0.3 => 43.0.3
[...]

PR #841, opened 2021-12-11
[eas-cli] Fix ASC prompt displayed in --non-interactive

barthap:
# Checklist

# Why

When testing my answer https://github.com/expo/eas-cli/issues/824#issuecomment-991584905 I discovered that a prompt is displayed in non-interactive mode, when user passed _some, but not all_ ASC API credentials in `eas.json`

# How

Instead of failing right away, I've decided to skip the `eas.json` resolution with a warning, so the CLI can try to resolve them from EAS servers.

# Test Plan

### Before

Screenshot 2021-12-11 at 11 43 58

### After

Screenshot 2021-12-11 at 11 43 02

### DIFF
--- a/packages/eas-cli/src/submit/ios/IosSubmitCommand.ts
+++ b/packages/eas-cli/src/submit/ios/IosSubmitCommand.ts
@@ -172,12 +172,18 @@ export default class IosSubmitCommand {
       });
     }
 
-    // interpret this to mean the user had some intention of passing in ASC Api key
     if (ascApiKeyPath || ascApiKeyIssuerId || ascApiKeyId) {
       Log.warn(`ascApiKeyPath, ascApiKeyIssuerId and ascApiKeyId must all be defined in eas.json`);
-      return result({
-        sourceType: AscApiKeySourceType.prompt,
-      });
+
+      // in interactive mode, interpret this to mean the user had some intention of passing in ASC Api key
+      // skip in non-interactive mode with warning
+      if (!this.ctx.nonInteractive) {
+        return result({
+          sourceType: AscApiKeySourceType.prompt,
+        });
+      } else {
+        Log.warn("Proceeding because you've run this command in non-interactive mode.");
+      }
     }
 
     return result(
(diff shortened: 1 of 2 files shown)


barthap:
/changelog-entry bug-fix Fix `eas submit` to display prompt in non-interactive mode when some ASC API credentials are missing.

quinlanj (commented):
I'm worried this will cause confusion. 

- With the current behaviour, if someone has partial `eas.json` credentials, we error to let them know they need to update it. 
- With the new behaviour, if someone has partial `eas.json` credentials, we log a warning, then use their remote credentials if it exists. 

If someone wants to use their credentials in `eas.json` and we use their remote credentials instead, this may be unexpected.

barthap:
> With the current behaviour, if someone has partial eas.json credentials, we error to let them know they need to update it.

Actually we warn them, then try to display prompt in non-interactive 😅  That's why I opened this pr. It will most likely fail anyway, but in less obvious way.

But explicitly erroring, in this case, was the first idea I came up with and was the most obvious to me. However, then I thought it could check other options before eventually failing, to be more errr foolproof?
But you are right, this might be confusing. I can revert it then to my original idea (that one you defined as "current behavior" 😉 ).

dsokal (approved):
LGTM. I'm also fine with throwing an error.

wkozyra95 (approved):

quinlanj:
@barthap oops my bad, you are correct. 'current behaviour' is not actually current behaviour 😓  sorry for the confusion.

I prefer throwing an error instead of looking for server credentials. I agree that we will have lower rates of failure if we always fall back to server credentials. But as a user, I'd be pretty confused if I expected to use the credentials in `eas.json` only to have it fall back to server because I forgot a field.

packages/eas-cli/src/submit/ios/IosSubmitCommand.ts
```diff
@@ -172,12 +172,18 @@ export default class IosSubmitCommand {
...
-      });
+
+      // in interactive mode, interpret this to mean the user had some intention of passing in ASC Api key
+      // skip in non-interactive mode with warning
+      if (!this.ctx.nonInteractive) {
+        return result({
+          sourceType: AscApiKeySourceType.prompt,
+        });
+      } else {
+        Log.warn("Proceeding because you've run this command in non-interactive mode.");
```
quinlanj, line 185: this is better than the actual current behaviour, but i'd prefer if we throw an error instead

quinlanj (approved):

PR #846, opened 2021-12-13
[eas-cli] add channel:delete

jkhales:
# Checklist

# Why

Add eas channel:delete

# How

copied eas branch:delete

# Test Plan

tested in demo repo

### DIFF
--- a/packages/eas-cli/src/commands/channel/delete.ts
+++ b/packages/eas-cli/src/commands/channel/delete.ts
@@ -20,61 +20,6 @@ import {
 import { promptAsync, toggleConfirmAsync } from '../../prompts';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
-async function getChannelInfoAsync({
-  appId,
-  name,
-}: GetChannelInfoQueryVariables): Promise<GetChannelInfoQuery> {
-  const data = await withErrorHandlingAsync(
-    graphqlClient
-      .query<GetChannelInfoQuery, GetChannelInfoQueryVariables>(
-        gql`
-          query GetChannelInfo($appId: String!, $name: String!) {
-            app {
-              byId(appId: $appId) {
-                id
-                updateChannelByName(name: $name) {
-                  id
-                  name
-                }
-              }
-            }
-          }
-        `,
-        {
-          appId,
-          name,
-        },
-        { additionalTypenames: ['UpdateChannel'] }
-      )
-      .toPromise()
-  );
-  return data;
-}
-
-async function deleteChannelOnAppAsync({
-  channelId,
-}: DeleteUpdateChannelMutationVariables): Promise<DeleteUpdateChannelResult> {
-  const data = await withErrorHandlingAsync(
-    graphqlClient
-      .mutation<DeleteUpdateChannelMutation, DeleteUpdateChannelMutationVariables>(
-        gql`
-          mutation DeleteUpdateChannel($channelId: ID!) {
-            updateChannel {
-              deleteUpdateChannel(channelId: $channelId) {
-                id
-              }
-            }
-          }
-        `,
-        {
-          channelId,
-        }
-      )
-      .toPromise()
-  );
-  return data.updateChannel.deleteUpdateChannel;
-}
-
 export default class ChannelDelete extends EasCommand {
   static hidden = true;
   static description = 'Delete a channel on the current project';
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,148 @@
...
+      description: `return JSON with the edited branch's ID and name.`,
+      default: false,
+    }),
+  };
+
+  async runAsync(): Promise<void> {
+    let {
+      args: { name },
+      flags: { json: jsonFlag },
+    } = this.parse(ChannelDelete);
```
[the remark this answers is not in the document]
jkhales: Ah I see, OK. Going to put us a another PR to correct this in a bunch of other commands as well.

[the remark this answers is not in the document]
jkhales: https://github.com/expo/eas-cli/pull/848

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+  static args = [
+    {
+      name: 'name',
+      required: false,
+      description: 'Name of the channel to delete',
+    },
+  ];
+  static flags = {
+    json: flags.boolean({
+      description: `return JSON with the edited branch's ID and name.`,
```
dsokal, line 91: edited?

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+      default: false,
+    }),
+  };
+
+  async runAsync(): Promise<void> {
+    let {
```
dsokal, line 97:
```suggestion
    const {
```
Don't use `let` if you don't need it.

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+    const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
+    const fullName = await getProjectFullNameAsync(exp);
+    const projectId = await getProjectIdAsync(exp);
+
+    if (!name) {
+      const validationMessage = 'Channel name may not be empty.';
+      if (jsonFlag) {
+        throw new Error(validationMessage);
+      }
+      ({ name } = await promptAsync({
```
dsokal, line 115: Do not reassign variables. IMO it's an anti-pattern and makes refactoring a nightmare.

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    const projectDir = await findProjectRootAsync();
+    const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
+    const fullName = await getProjectFullNameAsync(exp);
+    const projectId = await getProjectIdAsync(exp);
+
+    if (!name) {
+      const validationMessage = 'Channel name may not be empty.';
+      if (jsonFlag) {
```
dsokal, line 112: Hmmm, I'm not sure if we want to throw an error if someone passes the `--json` flag. I'd prefer to add another flag - `--non-interactive` - and throw only then.

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+      }));
+    }
+
+    const data = await getChannelInfoAsync({ appId: projectId, name });
+    const channelId = data.app?.byId.updateChannelByName?.id;
+    if (!channelId) {
+      throw new Error(`Could not find channel ${name} on ${fullName}`);
+    }
+
+    if (!jsonFlag) {
```
dsokal, line 129: Same here, this looks weird. I'm 100% sure the logic shouldn't be based on the `--json` flag.
We should either don't let users run this in non-interactive mode or handle the `--non-interactive` flag. Is there a use case for running this on CI?

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+      Log.addNewLineIfNone();
+      Log.warn(
+        `You are about to permamently delete channel: "${name}".\nThis action is irreversible.`
+      );
+      Log.newLine();
+      const confirmed = await toggleConfirmAsync({ message: 'Are you sure you wish to proceed?' });
+      if (!confirmed) {
+        Log.error(`Cancelled deletion of channel: "${name}".`);
```
dsokal, line 137:
```suggestion
        Log.error(`Canceled deletion of channel: "${name}".`);
```
🇺🇸

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const deletionResult = await deleteChannelOnAppAsync({
+      channelId,
+    });
+
+    if (jsonFlag) {
+      printJsonOnlyOutput(deletionResult);
+    }
+
+    Log.withTick(`️Deleted channel "${name}".`);
```
dsokal, line 150: Do we want to print this with `--json` flag?

dsokal (changes requested):

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,152 @@
...
+
+async function getChannelInfoAsync({
+  appId,
+  name,
+}: GetChannelInfoQueryVariables): Promise<GetChannelInfoQuery> {
+  const data = await withErrorHandlingAsync(
+    graphqlClient
+      .query<GetChannelInfoQuery, GetChannelInfoQueryVariables>(
+        gql`
+          query GetChannelInfo($appId: String!, $name: String!) {
```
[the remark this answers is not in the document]
jkhales: There is an extra database query on the most similar one I could find:
Screen Shot 2021-12-16 at 3 39 08 PM

But I guess this is an infrequent command and it wouldn't be so bad to combine these commands in the name of deduplication... on the other hand the future of these two functions are not inherently tied so there isn't a great reason to unify them.

Do you feel strongly about combining them?

wkozyra95 (approved):

dsokal (approved):
I don't like the behavior change with `--json` flag but 🤷

jkhales:
> I don't like the behavior change with `--json` flag but 🤷

I think you are right.

I didn't get around to addressing those comments yesterday, sorry if it came off as me ignoring them!

I'm going to push some changes when I get back from lunch and would greatly appreciate you taking a look at them and also making sure I am using the correct english 🇺🇸 😄

packages/eas-cli/src/commands/channel/delete.ts
jkhales: moved behind an if/else

packages/eas-cli/src/commands/channel/delete.ts
jkhales: switched to `non-interactive`

packages/eas-cli/src/commands/channel/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+  async runAsync(): Promise<void> {
+    const {
+      args: { name: nameArg },
+      flags: { json: jsonFlag, 'non-interactive': nonInteractiveFlag },
+    } = this.parse(ChannelDelete);
+    if (jsonFlag && !nonInteractiveFlag) {
+      error('--json is allowed only in non-interactive mode', { exit: 1 });
+    }
```
jkhales, line 53: There is an argument that setting `--json` should automatically set `nonInteractiveFlag = true`, but for now am following the convention set here: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commands/build/index.ts#L107

dsokal (approved):
👏

PR #848, opened 2021-12-14
[eas-cli] fix json output in EAS Update commands

jkhales:
# Checklist

# Why

Applying suggestions from this comment more generally: https://github.com/expo/eas-cli/pull/846#discussion_r768701581

### DIFF
--- a/packages/eas-cli/src/commands/branch/view.ts
+++ b/packages/eas-cli/src/commands/branch/view.ts
@@ -126,35 +126,35 @@ export default class BranchView extends EasCommand {
 
     if (jsonFlag) {
       printJsonOnlyOutput({ ...UpdateBranch, updates });
+    } else {
+      const groupTable = new Table({
+        head: UPDATE_COLUMNS,
+        wordWrap: true,
+      });
+
+      for (const update of updates) {
+        groupTable.push([
+          formatUpdate(update),
+          update.runtimeVersion,
+          update.group,
+          getPlatformsForGroup({
+            updates: UpdateBranch.updates,
+            group: update.group,
+          }),
+        ]);
+      }
+
+      Log.addNewLineIfNone();
+      Log.log(chalk.bold('Branch:'));
+      Log.log(
+        formatFields([
+          { label: 'Name', value: UpdateBranch.name },
+          { label: 'ID', value: UpdateBranch.id },
+        ])
+      );
+      Log.addNewLineIfNone();
+      Log.log(chalk.bold('Recently published update groups:'));
+      Log.log(groupTable.toString());
     }
-
-    const groupTable = new Table({
-      head: UPDATE_COLUMNS,
-      wordWrap: true,
-    });
-
-    for (const update of updates) {
-      groupTable.push([
-        formatUpdate(update),
-        update.runtimeVersion,
-        update.group,
-        getPlatformsForGroup({
-          updates: UpdateBranch.updates,
-          group: update.group,
-        }),
-      ]);
-    }
-
-    Log.addNewLineIfNone();
-    Log.log(chalk.bold('Branch:'));
-    Log.log(
-      formatFields([
-        { label: 'Name', value: UpdateBranch.name },
-        { label: 'ID', value: UpdateBranch.id },
-      ])
-    );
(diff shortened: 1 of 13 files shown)


packages/eas-cli/src/commands/update/delete.ts
```diff
@@ -80,7 +83,7 @@ export default class UpdateDelete extends EasCommand {
     await deleteUpdateGroupAsync({ group });
 
     if (jsonFlag) {
-      Log.log(JSON.stringify({ group }));
+      printJsonOnlyOutput({ group });
       return;
```
jkhales, line 87:
```suggestion
```

packages/eas-cli/src/commands/update/view.ts
```diff
@@ -73,11 +74,14 @@ export default class UpdateView extends EasCommand {
...
+    if (jsonFlag) {
+      enableJsonOutput();
+    }
 
     const { updatesByGroup } = await viewUpdateAsync({ groupId });
 
     if (jsonFlag) {
-      Log.log(JSON.stringify(updatesByGroup));
+      printJsonOnlyOutput(updatesByGroup);
       return;
```
jkhales, line 85:
```suggestion
```

dsokal (approved):
LGTM but leaving the final decision to @wkozyra95

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -94,8 +98,7 @@ export default class ChannelList extends EasCommand {
     }
 
     if (jsonFlag) {
-      Log.log(JSON.stringify(channels));
-      return;
+      printJsonOnlyOutput(channels);
```
wkozyra95, line 101: duplicated info, the same as bove

packages/eas-cli/src/commands/channel/view.ts
```diff
@@ -243,8 +247,7 @@ export default class ChannelView extends EasCommand {
     }
 
     if (jsonFlag) {
-      Log.log(JSON.stringify(channel));
-      return;
+      printJsonOnlyOutput(channel);
```
wkozyra95, line 250: duplicated info, the same as bove

packages/eas-cli/src/commands/update/view.ts
```diff
@@ -73,12 +74,14 @@ export default class UpdateView extends EasCommand {
...
+    if (jsonFlag) {
+      enableJsonOutput();
+    }
 
     const { updatesByGroup } = await viewUpdateAsync({ groupId });
 
     if (jsonFlag) {
-      Log.log(JSON.stringify(updatesByGroup));
-      return;
+      printJsonOnlyOutput(updatesByGroup);
```
wkozyra95, line 84: duplicated info, the same as bove

wkozyra95 (changes requested):

packages/eas-cli/src/commands/branch/view.ts
```diff
@@ -122,8 +126,7 @@ export default class BranchView extends EasCommand {
     );
 
     if (jsonFlag) {
-      Log.log(JSON.stringify({ ...UpdateBranch, updates }));
-      return;
+      printJsonOnlyOutput({ ...UpdateBranch, updates });
```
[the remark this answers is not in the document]
jkhales: Isn't the table below logged to `stderr` when `--json` is enabled?

I thought duplication is the right approach in the sense that one is for humans, the other for CI (or whatever).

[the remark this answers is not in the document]
wkozyra95: yes it will go to stderr, and it will work correctly if you would do `eas command_Name | jq`, but I have never seen json flag used this way in cli tools, it's always replacing human-readable content.

If you feel strongly about it, I'm not a fan of this approach, but I won't object if you decide to go with this solution.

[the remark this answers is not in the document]
jkhales: Moved everything behind an else statement as you are suggesting. Thanks for explaining you thinking in slack DM's! 🙏

wkozyra95 (approved):

PR #852, opened 2021-12-15
[eas-cli] add support for autocompletion

wkozyra95:
# Checklist

# Why

Ads autocompletion support

# How

`eas autocomplete` will print the command you need to execute to configure your shell

For local eas-cli development
if you have alias like `alias eas_dev=$HOME/expo/eas-cli/bin/run` for local devlopment
you can add completion to it by adding `compdef $HOME/expo/eas-cli/bin/run='eas'` in you zshrc after `eas` completion is set up

# Test Plan

tested with eas if completion shows up


wkozyra95:
/changelog-entry new-feature Add shell completion for bash and zsh.

kbrandwijk:
Will that include `fig` support like expo-cli?

wkozyra95:
> Will that include `fig` support like expo-cli?

This is standard zsh/bash completion. It might work if fig integrates with shell completion, but I don't know if it does.

kbrandwijk:
No, it needs it's own command config file, there's a whole script for it in expo-cli, and seeing this was still in draft, I asked in case you were considering to add it.

wkozyra95:
> No, it needs it's own command config file, there's a whole script for it in expo-cli (https://github.com/expo/expo-cli/blob/main/packages/expo-cli/scripts/introspect.ts), and seeing this was still in draft, I asked in case you were considering to add it.

This is a draft because updating oclif packages caused 2-3 MB increase in size for eas-cli, we will need to resolve that before merging, I'm not saying we will never support fig, but it's quite a niche solution and bash/zsh completion works everywhere.

PR #890, opened 2022-01-05
Delete device functionality

### LINKED ISSUE #287: There is no way to remove old devices
Currently, we can add new devices via `eas device:create`. However, there is no way to remove devices.

kbrandwijk:
# Checklist

# Why

Resolves #287 

# How

- [X] Add `device:delete` command for device selection
- [X] Delete device using existing gql api
- [X] Update adhoc build provision logic to take deleted devices into account - not needed

# Test Plan

- [X] Manually tested deleting, and subsequent `device:list` not displaying the device anymore. Also tested how this affects the adhoc provisioning after removing a device that was previously provisionined for adhoc deployment on a project.


packages/eas-cli/src/credentials/ios/actions/DeviceUtils.ts
```diff
@@ -26,6 +26,23 @@ export async function chooseDevicesAsync(
   return devices;
 }
 
+export async function chooseSingleDeviceToDeleteAsync(
+  allDevices: AppleDeviceFragment[]
+): Promise<AppleDevice> {
+  const { device } = await promptAsync({
+    type: 'select',
+    name: 'device',
+    message: 'Which devices do you want to remove?',
```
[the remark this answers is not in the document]
dsokal: Generally, I think it'd be better to let users delete multiple devices at once.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    let spinner: Ora;
+
+    if (!appleTeamIdentifier) {
+      spinner = ora().start('Fetching the list of teams for the project…');
+
+      try {
+        const teams = await AppleTeamQuery.getAllForAccountAsync(accountName);
+
+        if (teams.length) {
```
dsokal, line 41:
```suggestion
        if (teams.length > 0) {
```
- It's always better to be explicit instead of depending on implicit conversion.
- Yes, I know this is copy-pasted.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,125 @@
...
+import Log from '../../log';
+import { Ora, ora } from '../../ora';
+import { findProjectRootAsync, getProjectAccountNameAsync } from '../../project/projectUtils';
+import { promptAsync, toggleConfirmAsync } from '../../prompts';
+
+export default class DeviceDelete extends EasCommand {
+  static description = 'remove a registered device from your account';
+
+  static flags = {
+    'apple-team-id': Flags.string(),
```
dsokal, line 23: I'd be nice if this command accepted the device udid, so you could remove a device like this:
`eas device:delete --apple-team-id TEAM_ID --udid UDID`

packages/eas-cli/src/commands/device/delete.ts
kbrandwijk: I'll adjust it in the original location too.

packages/eas-cli/src/credentials/ios/actions/DeviceUtils.ts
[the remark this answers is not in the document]
kbrandwijk: Just to be clear, do you want the behavior adjusted or the question adjusted?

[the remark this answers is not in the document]
dsokal: I'd prefer to change the behavior (so it's possible to delete multiple devices with a single `eas device:delete` run). If you think this doesn't make sense then we need to change the question.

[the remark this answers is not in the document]
kbrandwijk: Multi-select implemented

packages/eas-cli/src/commands/device/delete.ts
kbrandwijk: Parameter implemented, including multi-select

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,155 @@
...
+    udid: Flags.string({ multiple: true }),
+  };
+
+  async runAsync(): Promise<void> {
+    let appleTeamIdentifier = (await this.parse(DeviceDelete)).flags['apple-team-id'];
+    const udids = (await this.parse(DeviceDelete)).flags.udid;
```
wkozyra95, line 30: run parse once

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,155 @@
...
+            }
+          });
+        }
+
+        if (chosenDevices.length === 0) {
+          Log.addNewLineIfNone();
+          chosenDevices = await chooseDevicesToDeleteAsync(appleDevices);
+        }
+
+        chosenDevices.map(device => {
```
wkozyra95, line 109: you are not returning anything, use forEach or for loop

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,155 @@
...
+          Log.log(
+            `\n${formatDevice(device, {
```
wkozyra95, line 111: use Log.newline instead of `\n`

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,155 @@
...
+              appleTeamName,
+              appleTeamIdentifier: appleTeamIdentifier!,
+            })}`
+          );
+          Log.newLine();
+        });
+
+        if (chosenDevices.length > 0) {
+          Log.warn(
+            `You are about to remove the Apple device${
```
wkozyra95, line 121:
```suggestion
            `You are about to remove the Apple device ${
```

kbrandwijk: No, it adds an -s for multiple.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,184 @@
...
+          const confirmed = await toggleConfirmAsync({
+            message: 'Are you sure you wish to proceed?',
+          });
+
+          if (confirmed) {
+            const removalSpinner = ora(`Removing Apple devices on Expo`).start();
+            try {
+              chosenDevices.forEach(async chosenDevice => {
+                await AppleDeviceMutation.deleteAppleDeviceAsync(chosenDevice.id);
+              });
```
wkozyra95, line 139:
```suggestion
              await Promise.all(chosenDevices.map(chosenDevice => {
                 return AppleDeviceMutation.deleteAppleDeviceAsync(chosenDevice.id);
              }));
```
you are not awaiting for those values, forEach works only with synchronous methods (alternatively you could use for loop if it has to be called one by one)

plus nit pick: Irecomend moving `removalSpinner.succeed` in the try  block at the end

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,184 @@
...
+              const context: RequestContext = getRequestContext(ctx);
+
+              Log.addNewLineIfNone();
+              const removeAppleSpinner = ora('Removing devices from Apple').start();
+              try {
+                const realDevices = await Device.getAllIOSProfileDevicesAsync(context);
+                realDevices
+                  .filter(d => chosenDevices.map(cd => cd.identifier).includes(d.attributes.udid))
+                  .forEach(device => {
+                    device.updateAsync({ status: DeviceStatus.DISABLED });
```
wkozyra95, line 163: the same as above
- forEach is not awaiting correctlly
- spinner.succeed in try block

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export default class DeviceDelete extends EasCommand {
+  static description = 'remove a registered device from your account';
+
+  static flags = {
+    'apple-team-id': Flags.string(),
+    udid: Flags.string({ multiple: true }),
+  };
+
+  async runAsync(): Promise<void> {
```
dsokal, line 30: This function is 130 LOC. That's a lot. Please break it down.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,184 @@
...
+
+            Log.newLine();
+            const deleteOnApple = await toggleConfirmAsync({
+              message: 'Do you want to remove these devices from your Apple account as well?',
+            });
+
+            if (deleteOnApple) {
+              // Delete on Apple?
+              const ctx = await authenticateAsync({ teamId: appleTeamIdentifier });
+              const context: RequestContext = getRequestContext(ctx);
```
dsokal, line 154: I think `: RequestContext` is not necessary.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,184 @@
...
+                realDevices
+                  .filter(d => chosenDevices.map(cd => cd.identifier).includes(d.attributes.udid))
+                  .forEach(device => {
+                    device.updateAsync({ status: DeviceStatus.DISABLED });
+                  });
+              } catch (err) {
+                removeAppleSpinner.fail();
+                throw err;
+              }
+              removeAppleSpinner.succeed('Removed devices from Apple');
```
dsokal, line 169: Maybe we should be explicit here. We actually don't **remove** the devices, we **disable** them.

dsokal: Change that in other logs as well.

packages/eas-cli/src/commands/device/delete.ts
kbrandwijk: Reading up on this, I will actually turn this into a normal for loop, as I cannot control the degree of parallelism and I don't want to run into weird errors if a user decides to remove 20 devices at the same time.

packages/eas-cli/src/commands/device/delete.ts
kbrandwijk: `list` is 100 LOC. What's the target value?

dsokal: Basically, I think we should have a lint rule for that. It's hard to add that in right now as it'd require a lot of refactoring. I'd say a rule of thumb is that a function should fit a reasonable display (without scrolling).

kbrandwijk: Works for me, I'll make a note to include that rule for myself whenever I need to 'touch' existing code as well.

dsokal (approved):
LGTM, let's wait for a review from @wkozyra95

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -0,0 +1,246 @@
...
+    appleTeamName: Maybe<string> | undefined,
+    appleTeamIdentifier: string
+  ): void {
+    Log.addNewLineIfNone();
+    chosenDevices.forEach(device => {
+      Log.log(
+        `${formatDevice(device, {
+          appleTeamName,
+          appleTeamIdentifier: appleTeamIdentifier!,
+        })}`
```
wkozyra95, line 149:
```suggestion
        formatDevice(device, {
          appleTeamName,
          appleTeamIdentifier: appleTeamIdentifier!,
        })
```

wkozyra95 (approved):

PR #904, opened 2022-01-11
[eas-cli] Only configure updates in generic projects

brentvatne:
# Checklist

# Why

In #888 we refactored the `build:configure` command and as a result of that change we ended up attempting to configure native iOS and Android projects for expo-updates even if they don't exist because the project is managed. It's possible to repro this with:

```
expo init project # blank app
cd project && expo install expo-updates
eas build:configure
# Error: Android project folder is missing in project ...
# Same sort of thing for iOS
``` 

# How

I used the same logic that we use in `syncProjectConfiguration` to detect if we are in a generic project or not, then skip running updates configuration in managed projects:

https://github.com/expo/eas-cli/blob/dd2303dbb2b01dcd941591b5ccba01bbd04b29a7/packages/eas-cli/src/build/ios/syncProjectConfiguration.ts#L26-L29

# Test Plan

- Run `easd build:configure` in managed app with `expo-updates` installed, it should just create **eas.json** and nothing else
- Run `easd build:configure` in bare app with `expo-updates` installed, it should sync config from **app.json** to native projects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fix build:configure expo-updates configuration to only run on generic projects. ([#904](https://github.com/expo/eas-cli/pull/904) by [@brentvatne](https://github.com/brentvatne))
+
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


brentvatne:
/changelog-entry bug-fix Fix build:configure expo-updates configuration to only run on generic projects

dsokal (approved):

PR #964, opened 2022-02-08
[eas-cli] Add code signing

wschurman:
# Checklist

# Why

This implements code signing in the EAS CLI which was implemented on the server in https://github.com/expo/universe/pull/8825.

# How

The flow is roughly:
- Create project, install expo-updates.
- `yarn expo-updates codesigning:generate` - generates a codesigning keypair and certificate (or the user can bring their own)
- `yarn expo-updates codesigning:configure` - configures app.json expo-updates config section for code signing.
- Build a binary. The expo-updates config plugin will transfer the code signing config from app.json to the native configuration (AndroidManifest.xml or Expo.plist). This built app will now expect signed OTA updates in the expo-update module.
- Publish an update via `eas update` (this PR automatically does the following)
    1. publish the update with the awaitingCodeSigningInfo flag (so that it doesn't get served to users before code signing info is uploaded)
    2. download that update's manifest from the server as if it were being served to a client
    3. verify the integrity of the manifest on the server (verify it matches the info we just uploaded)
    4. sign the server's manifest with our private key
    5. Upload that signature to the server, indicating that the update is ready to serve to clients that need signed manifests.

# Test Plan

1. Follow above steps, then run `DEBUG=true ~/expo/eas-cli/packages/eas-cli/bin/run update --branch preview --message "Updating the app 2"`
2. In curl/insomnia, load update URL for app (from app.json) with header: `expo-expect-signature` = `sig, keyid="main", alg="rsa-v1_5-sha256"`.
3. Ensure response from EAS has the `expo-signature` header.


packages/eas-cli/package.json
```diff
@@ -30,8 +31,10 @@
     "chalk": "4.1.2",
     "cli-table3": "0.6.1",
     "dateformat": "4.6.3",
+    "dicer": "0.3.1",
```
ide, line 34: Sizes look OK to me:

- dicer: 2.4KB min+gz (incl. deps)
- fast-deep-equal: 414B min+gz
- structured-headers: 2.7KB min+gz

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+    certificate,
+  };
+}
+
+export type MultipartPart = { headers: Map<string, string>; body: string };
+
+export async function parseMultipartMixedResponseAsync(res: Response): Promise<MultipartPart[]> {
+  const contentType = res.headers['content-type'];
+  if (!contentType) {
+    throw new Error('Missing content-type in multipart response');
```
ide, line 48: Since these errors might show up to developers, making them a little more human-readable will make error messages sent our way easier to understand:
```suggestion
    throw new Error('The multipart response is missing the content-type header');
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  const boundaryRegex = /^multipart\/.+?; boundary=(?:"(.+)"|([^\s]+))$/i;
```
ide, line 51: Might want to remove the $ to leave open the possibility for extra parameters without needing to upgrade the CLI:
```suggestion
  const boundaryRegex = /^multipart\/.+?; boundary=(?:"([^"]+)"|([^\s;]+))/i;
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+  const matches = boundaryRegex.exec(contentType);
+  if (!matches) {
+    throw new Error('content-type header in response not multipart');
```
ide, line 54:
```suggestion
    throw new Error('The content-type header in the HTTP response is not a multipart media type');
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+            headers: new Map(),
+          };
+
+          p.on('header', headers => {
+            for (const h in headers) {
+              part.headers.set(h, (headers as { [key: string]: string[] })[h][0]);
+            }
+          });
+          p.on('data', data => {
+            part.body = data.toString();
```
ide, line 78: Can't `data` get emitted multiple times?
```suggestion
            part.body += data.toString();
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+          });
+          p.on('end', () => {
+            parts.push(part);
+          });
+        })
+        .on('finish', () => {
+          resolve(parts);
+        })
+        .on('error', err => reject(err))
```
ide, line 87: match style above:
```suggestion
        .on('error', (error) => {
          reject(error);
        })
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+    );
+  });
+}
+
+function isManifestMultipartPart(multipartPart: MultipartPart): boolean {
+  const partName = parseItem(nullthrows(multipartPart.headers.get('content-disposition')))[1].get(
+    'name'
+  );
```
ide, line 95: IMO I'd find it clearer to define the header parameters separately because I was unfamiliar with how the structured-headers library returns a (value, parameters) tuple.
```suggestion
  const [, parameters] = parseItem(nullthrows(multipartPart.headers.get('content-disposition')));
  const partName = parameters.get('name');
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+  partialManifestAsset: PartialManifestAsset
+): void {
+  if (
+    manifestResponseBodyAssetJSON.hash !== partialManifestAsset.fileSHA256 ||
+    manifestResponseBodyAssetJSON.contentType !== partialManifestAsset.contentType ||
+    manifestResponseBodyAssetJSON.key !== partialManifestAsset.bundleKey
+  ) {
+    throw new Error(
+      `Code signing manifest integrity error: Manifest asset tamper detected for asset: ${partialManifestAsset.bundleKey}`
+    );
```
ide, line 124: If this error occurs and a developer sends the message to us, it will be much more helpful to see which field(s) didn't match and what the expected and actual values were. (Make this message easy for future-us to diagnose.)

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+  // server has compromised the integrity of the manifest.
+  // JSON stringify and unstringify to remove any undefined values and bring it as close
+  // to the server sanitized value as possible
+  const isExtraEqual = isDeepEqual(
+    JSON.parse(JSON.stringify(partialManifest.extra?.expoClient)),
+    manifestResponseBodyJSON.extra?.expoClient
+  );
+  if (!isExtraEqual) {
+    throw new Error(
+      'Code signing manifest integrity error: Manifest extra.expoClient does not match uploaded manifest extra.expoClient'
```
ide, line 144:
```suggestion
      `Code signing manifest integrity error: The manifest to code sign contains an extra.expoClient field that does not match the initially uploaded manifest's extra.expoClient field`
```
or something like this

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  for (const partialManifestAsset of partialManifest.assets) {
+    const partialManifestAssetBundleKey = nullthrows(partialManifestAsset).bundleKey;
+    const correspondingManifestResponseBodyAssetJSON = manifestResponseBodyJSON.assets.find(
+      (manifestResponseBodyAssetJSON: { [key: string]: any }) =>
+        manifestResponseBodyAssetJSON.key === partialManifestAssetBundleKey
+    );
+    if (!correspondingManifestResponseBodyAssetJSON) {
+      throw new Error(
+        `Code signing manifest integrity error: Manifest asset not found in uploaded manifest: ${partialManifestAssetBundleKey}`
```
ide, line 164: This could be more human-friendly / easier to debug as well

ide (changes requested):

ide:
Most of the requested changes are around the readability of error messages and making easier for both developers and our team to diagnose issues.

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (!contentType) {
+    throw new Error('Missing content-type in multipart response');
+  }
+
+  const boundaryRegex = /^multipart\/.+?; boundary=(?:"(.+)"|([^\s]+))$/i;
+  const matches = boundaryRegex.exec(contentType);
+  if (!matches) {
+    throw new Error('content-type header in response not multipart');
+  }
+  const boundary = matches[1] ?? matches[2];
```
jkhales, line 56: what is the situation in which `matches[1]` would be undefined, but `matches[2]` would be the correct?

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -413,30 +443,73 @@ export default class UpdatePublish extends EasCommand {
...
-    let newUpdates;
+    );
+    let newUpdates: UpdatePublishMutation['updateBranch']['publishUpdateGroups'];
     const publishSpinner = ora('Publishing...').start();
     try {
       newUpdates = await PublishMutation.publishUpdateGroupAsync(updateGroups);
+
+      const updatesTemp = [...newUpdates];
+      const updateGroupsAndTheirUpdates = updateGroups.map((updateGroup, i) => {
+        const newUpdates = updatesTemp.splice(0, Object.keys(updateGroup.updateInfoGroup).length);
```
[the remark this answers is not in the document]
ide: splice mutates the array in place. this is like a big array.shift.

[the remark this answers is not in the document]
jkhales: ah thanks

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -413,30 +443,73 @@ export default class UpdatePublish extends EasCommand {
...
+        return {
+          updateGroup,
+          newUpdates,
+        };
+      });
+
+      if (certificate && privateKey) {
+        await Promise.all(
+          updateGroupsAndTheirUpdates.map(async ({ updateGroup, newUpdates }) => {
+            await Promise.all(
```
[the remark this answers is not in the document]
wschurman: Since this is a void function (or `Promise<void>`), I generally opt to not return results. Even though they're technically equivalent.

packages/eas-cli/src/utils/code-signing.ts
wschurman: The boundary is not necessarily always in quotes, so `matches[1]` will be the case when it is quoted (probably contains whitespace) and `matches[2]` will be the case otherwise (can't contain whitespace). Now that I think about it, we may be able to use structured-headers to parse this?

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -4,17 +4,59 @@ import {
...
+
+export async function getCodeSigningInfoAsync(
+  config: ExpoConfig,
+  privateKeyPath: string | undefined
+): Promise<CodeSigningInfo | undefined> {
+  const codeSigningCertificatePath = (config.updates as any)?.codeSigningCertificate;
+  const codeSigningMetadata = (config.updates as any)?.codeSigningMetadata;
+
+  if (codeSigningCertificatePath && !privateKeyPath) {
+    privateKeyPath = path.join(path.dirname(codeSigningCertificatePath));
```
ide, line 34:
```suggestion
    privateKeyPath = path.join(path.dirname(codeSigningCertificatePath), 'private-key.pem');
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -4,17 +4,59 @@ import {
...
+  }
+
+  if (!codeSigningMetadata || !codeSigningMetadata.alg || !codeSigningMetadata.keyid) {
+    throw new Error('Must specify codeSigningMetadata in config.updates for EAS code signing');
```
ide, line 38: This sounds like there is a field called "config":
```suggestion
    throw new Error('Must specify codeSigningMetadata under the "updates" field of your app config file to use EAS code signing');
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -23,8 +65,14 @@ export async function getKeyAndCertificateFromPathsAsync({
   privateKeyPath: string;
 }): Promise<{ privateKey: PKI.rsa.PrivateKey; certificate: PKI.Certificate }> {
   const [codeSigningCertificatePEM, privateKeyPEM] = await Promise.all([
-    fs.readFile(codeSigningCertificatePath, 'utf8'),
-    fs.readFile(privateKeyPath, 'utf8'),
+    readFileAsync(
+      codeSigningCertificatePath,
+      `Code signing certificate can not be read from path: ${codeSigningCertificatePath}`
```
ide, line 70:
```suggestion
      `Code signing certificate cannot be read from path: ${codeSigningCertificatePath}`
```

packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -23,8 +65,14 @@ export async function getKeyAndCertificateFromPathsAsync({
...
+    ),
+    readFileAsync(
+      privateKeyPath,
+      `Code signing private key can not be read from path: ${privateKeyPath}`
```
ide, line 74:
```suggestion
      `Code signing private key cannot be read from path: ${privateKeyPath}`
```

packages/eas-cli/src/utils/code-signing.ts
ide: Looks parseable https://runkit.com/ide/structured-headers-test

ide (approved):

jkhales (approved):
🥳

packages/eas-cli/src/utils/code-signing.ts
wschurman: Just tried, seems like our parser can't parse it. Will create a task for a follow-up investigation.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -208,6 +225,19 @@ export default class UpdatePublish extends EasCommand {
       skipSDKVersionRequirement: true,
       isPublicConfig: true,
     });
+    const codeSigningCertificatePath = (exp.updates as any)?.codeSigningCertificate;
+    const codeSigningMetadata = (exp.updates as any)?.codeSigningMetadata;
```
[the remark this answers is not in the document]
wschurman: Talked to CLI people about this. Will do in a follow-up.

PR #1018, opened 2022-03-15
[eas-cli] Fix code signing error when not yet configured

wschurman:
# Checklist

# Why

The logic here was incorrect as it was requiring code signing metadata even if there was not code signing certificate.

# How

Fix logic.

# Test Plan

We should really add some tests. Will work on that.


wschurman:
/changelog-entry bug-fix Fix code signing error when not yet configured

jonsamp (approved):

PR #1039, opened 2022-03-29
[eas-cli] generate credentilas for iOS app extensions

wkozyra95:
# Checklist

# Why

support generating credentials for iOS ap extension

# How

allow specifying those values via extra field

# Test Plan

run build with onesignal config

```
    "extra": {
      "eas": {
        "build": {
          "experimental": {
            "ios": {
              "appExtensions": [
                {
                  "targetName": "OneSignalNotificationServiceExtension",
                  "bundleIdentifier": "com.wkozyra95.sdk44.OneSignalNotificationServiceExtension"
                }
              ]
            }
          }
        }
      }
    },
    "plugins": [
      [
        "onesignal-expo-plugin",
        {
          "mode": "development"
        }
      ]
    ]
  }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Experimental support for iOS App Extensions in managed projects. ([#1039](https://github.com/expo/eas-cli/pull/1039) by [@wkozyra95](https://github.com/wkozyra95))
+
 ### 🐛 Bug fixes
 
 - Fix proxy support. ([#1032](https://github.com/expo/eas-cli/pull/1032) by [@wkozyra95](https://github.com/wkozyra95))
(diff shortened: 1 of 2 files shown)


dsokal (approved):

kbrandwijk:
I'm sad that this didn't get implemented as a 'first class citizen' in app.json to just generate the targets. Mostly because I think having to define this here _just_ for credential management is not a good DX, especially not in this 6-levels deep path in app.json if it's only used for EAS credential management. I feel this implementation doesn't address any of the concerns that were raised in the proposal stage. And I wonder how this differs from using managed credentials for additional targets in the bare workflow.

wkozyra95:
>  I feel this implementation doesn't address any of the concerns that were raised in the proposal stage

@kbrandwijk If you referring to the structure of the object then I decided to keep it as simple as possible from a code perspective, because it's sth that plugins will add not the user. If you are referring to sth else then please be specific.

> And I wonder how this differs from using managed credentials for additional targets in the bare workflow.

It doesn't, that's the point, now managed workflow is on par with bare in this respect.

Also, note that this is not full support, it's on par with the current workaround that uses credentials.json, but e.g. we do not correctly resolve sentiments for targets so it might be necessary to build with EXPO_NO_CAPABILITY_SYNC enabled

brentvatne (approved):

wkozyra95:
/changelog-entry new-feature Experimental support for iOS App Extensions in managed projects.

PR #1051, opened 2022-04-06
[ENG-4419][eas-cli] use asc api token to authenticate for EAS build

quinlanj:
# Checklist

# Why

- Authenticate locally with a ASC Api Key by specifying the following environment variables: `EXPO_ASC_API_KEY_PATH, EXPO_ASC_KEY_ID, EXPO_ASC_ISSUER_ID, EXPO_APPLE_TEAM_ID, EXPO_APPLE_TEAM_TYPE`. You can create distribution certs, provisioning profiles and bundle identifiers with the api key. But you cannot create a push key and modify capabilities. 
- There isn't a good way to determine the Apple Team with ASC Api Key (based on my research, it seems that Apple has intentionally done this for privacy reasons). Because we are resigning/creating provisioning profiles, we need to know information about the apple team, like if it is inHouse/Enterprise. For this reason, I also require apple team info be passed in as well.

# how you can help (dominik, wojtek)
- is the build profile the best place to pass in a local asc api key? i see there's also `credentials.json` but that seems target based and only for build creds like dist cert and provisioning profile. 

# how you can help (bacon)
- There isn't a good way to type a jwt `AuthState` right now. The object I return is missing a lot of things that are expected like `session` or `appleId/applePassword/username`. I think this may lead to undefined behaviour in the third party package if it is expecting those fields. See review comments below.
- I'd suggest making upstream changes in Thirdparty so we can differentiate between different types of AuthState (ie) UserAuthState or JwtAuthState. 

# Test Plan

- [x] new tests pass

### DIFF
--- a/packages/eas-cli/src/build/createContext.ts
+++ b/packages/eas-cli/src/build/createContext.ts
@@ -1,12 +1,14 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile } from '@expo/eas-json';
 import JsonFile from '@expo/json-file';
+import fs from 'fs-extra';
 import resolveFrom from 'resolve-from';
 import { v4 as uuidv4 } from 'uuid';
 
 import { TrackingContext } from '../analytics/common';
 import { Analytics, BuildEvent } from '../analytics/events';
-import { CredentialsContext } from '../credentials/context';
+import { CredentialsContext, CredentialsContextOptions } from '../credentials/context';
+import { AuthenticationMode } from '../credentials/ios/appstore/authenticate';
 import { getExpoConfig } from '../project/expoConfig';
 import { getProjectAccountName, getProjectIdAsync } from '../project/projectUtils';
 import { resolveWorkflowAsync } from '../project/workflow';
@@ -43,13 +45,48 @@ export async function createBuildContextAsync<T extends Platform>({
   const workflow = await resolveWorkflowAsync(projectDir, platform);
   const accountId = findAccountByName(user.accounts, accountName)?.id;
 
-  const credentialsCtx = new CredentialsContext({
+  let credentialsCtxOptions = {
     exp,
     nonInteractive,
     projectDir,
     user,
     env: buildProfile.env,
-  });
+  } as CredentialsContextOptions;
+  if (platform === Platform.IOS) {
+    const iosBuildProfile = buildProfile as BuildProfile<Platform.IOS>;
+    const isPartialLocalAscApiKeyProvided =
+      iosBuildProfile.ascApiKeyPath ||
+      iosBuildProfile.ascApiKeyIssuerId ||
+      iosBuildProfile.ascApiKeyId;
+    const isEntireLocalAscApiKeyProvided =
+      iosBuildProfile.ascApiKeyPath &&
+      iosBuildProfile.ascApiKeyIssuerId &&
+      iosBuildProfile.ascApiKeyId &&
+      iosBuildProfile.appleTeamId &&
+      iosBuildProfile.appleTeamType;
+
+    if (isPartialLocalAscApiKeyProvided && !isEntireLocalAscApiKeyProvided) {
+      throw new Error(
+        'To authenticate with an App Store Connect API Key, you must provide the following fields in your build profile: ascApiKeyPath, ascApiKeyIssuerId, ascApiKeyId, appleTeamId, appleTeamType.'
+      );
+    } else if (isEntireLocalAscApiKeyProvided) {
+      const keyP8 = await fs.readFile(iosBuildProfile.ascApiKeyPath!, 'utf-8');
+      credentialsCtxOptions = {
+        ...credentialsCtxOptions,
+        defaultAppStoreAuthentication: {
+          mode: AuthenticationMode.API_KEY,
+          ascApiKey: {
+            keyP8,
+            keyId: iosBuildProfile.ascApiKeyId!,
+            issuerId: iosBuildProfile.ascApiKeyIssuerId!,
+          },
+          teamId: iosBuildProfile.appleTeamId!,
+          teamType: iosBuildProfile.appleTeamType!,
+        },
+      };
+    }
+  }
+  const credentialsCtx = new CredentialsContext(credentialsCtxOptions);
 
   const devClientProperties = getDevClientEventProperties({
     platform,
--- a/packages/eas-cli/src/build/createContext.ts
+++ b/packages/eas-cli/src/build/createContext.ts
@@ -1,14 +1,12 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile } from '@expo/eas-json';
 import JsonFile from '@expo/json-file';
-import fs from 'fs/promises';
 import resolveFrom from 'resolve-from';
 import { v4 as uuidv4 } from 'uuid';
 
 import { TrackingContext } from '../analytics/common';
 import { Analytics, BuildEvent } from '../analytics/events';
-import { CredentialsContext, CredentialsContextOptions } from '../credentials/context';
-import { AuthenticationMode } from '../credentials/ios/appstore/authenticate';
+import { CredentialsContext } from '../credentials/context';
 import { getExpoConfig } from '../project/expoConfig';
 import { getProjectAccountName, getProjectIdAsync } from '../project/projectUtils';
 import { resolveWorkflowAsync } from '../project/workflow';
@@ -45,48 +43,13 @@ export async function createBuildContextAsync<T extends Platform>({
   const workflow = await resolveWorkflowAsync(projectDir, platform);
   const accountId = findAccountByName(user.accounts, accountName)?.id;
 
-  let credentialsCtxOptions = {
+  const credentialsCtx = new CredentialsContext({
     exp,
     nonInteractive,
     projectDir,
     user,
     env: buildProfile.env,
-  } as CredentialsContextOptions;
-  if (platform === Platform.IOS) {
-    const iosBuildProfile = buildProfile as BuildProfile<Platform.IOS>;
-    const isPartialLocalAscApiKeyProvided =
-      iosBuildProfile.ascApiKeyPath ||
-      iosBuildProfile.ascApiKeyIssuerId ||
-      iosBuildProfile.ascApiKeyId;
-    const isEntireLocalAscApiKeyProvided =
-      iosBuildProfile.ascApiKeyPath &&
-      iosBuildProfile.ascApiKeyIssuerId &&
-      iosBuildProfile.ascApiKeyId &&
-      iosBuildProfile.appleTeamId &&
-      iosBuildProfile.appleTeamType;
-
-    if (isPartialLocalAscApiKeyProvided && !isEntireLocalAscApiKeyProvided) {
-      throw new Error(
-        'To authenticate with an App Store Connect API Key, you must provide the following fields in your build profile: ascApiKeyPath, ascApiKeyIssuerId, ascApiKeyId, appleTeamId, appleTeamType.'
-      );
-    } else if (isEntireLocalAscApiKeyProvided) {
-      const keyP8 = await fs.readFile(iosBuildProfile.ascApiKeyPath!, 'utf-8');
-      credentialsCtxOptions = {
-        ...credentialsCtxOptions,
-        defaultAppStoreAuthentication: {
-          mode: AuthenticationMode.API_KEY,
-          ascApiKey: {
-            keyP8,
-            keyId: iosBuildProfile.ascApiKeyId!,
-            issuerId: iosBuildProfile.ascApiKeyIssuerId!,
-          },
-          teamId: iosBuildProfile.appleTeamId!,
-          teamType: iosBuildProfile.appleTeamType!,
-        },
-      };
-    }
-  }
-  const credentialsCtx = new CredentialsContext(credentialsCtxOptions);
+  });
 
   const devClientProperties = getDevClientEventProperties({
     platform,
--- a/packages/eas-cli/src/build/createContext.ts
+++ b/packages/eas-cli/src/build/createContext.ts
@@ -1,7 +1,7 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile } from '@expo/eas-json';
 import JsonFile from '@expo/json-file';
-import fs from 'fs-extra';
+import fs from 'fs/promises';
 import resolveFrom from 'resolve-from';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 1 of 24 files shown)


packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -144,6 +178,40 @@ async function loginWithUserCredentialsAsync({
...
+  }
+}
+
+// TODO: this may cause undefined behaviour in third party code
+async function authenticateWithApiKeyAsync(options: Options = {}): Promise<ApiKeyAuthCtx> {
+  const { ascApiKey, teamId, teamName, teamType } = options;
+  assert(ascApiKey, 'ascApiKey must be defined');
+  assert(teamId && teamType !== undefined, 'teamId and teamType must be defined');
+  const isInHouse = teamType === 'inHouse';
+  return {
```
quinlanj, line 194: @EvanBacon this object is missing a lot of expected fields like  username, cookies, session. Everything works fine when i test it out manually, but it may lead to some unexpected behaviour down the road if third party is expecting those fields to be present

packages/eas-cli/src/build/createContext.ts
```diff
@@ -1,12 +1,14 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile } from '@expo/eas-json';
 import JsonFile from '@expo/json-file';
+import fs from 'fs-extra';
```
EvanBacon, line 4:
```suggestion
import fs from 'fs/promises';
```

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -18,22 +21,31 @@ import {
 
 const APPLE_IN_HOUSE_TEAM_TYPE = 'in-house';
 
+export enum AuthenticationMode {
+  API_KEY,
```
EvanBacon, line 25:
```suggestion
  /** App Store API requests will be made using the official API via an API key, used for CI environments where 2FA cannot be performed. */
  API_KEY,
```

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -18,22 +21,31 @@ import {
...
+  USER,
```
EvanBacon, line 26:
```suggestion
  /** Uses cookies based authentication and the unofficial App Store web API, this provides more functionality than the official API but cannot be reliably used in CI because it requires 2FA. */
  USER,
```

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -18,22 +21,31 @@ import {
...
   appleId?: string;
   teamId?: string;
+  teamName?: string;
+  teamType?: AppleTeamType;
+  ascApiKey?: MinimalAscApiKey;
   /**
    * Can be used to restore the Apple auth state via apple-utils.
    */
   cookies?: Session.AuthState['cookies'];
+  mode?: AuthenticationMode;
```
EvanBacon, line 39:
```suggestion
  /** Indicates how Apple network requests will be made. */
  mode?: AuthenticationMode;
```

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -18,22 +21,31 @@ import {
...
 };
 
 export type Team = {
   id: string;
-  name: string;
+  name?: string;
```
EvanBacon, line 44:
```suggestion
    /** Name of the development team, this is undefined when ASC API keys are used instead of cookies for authentication. */
  name?: string;
```

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
EvanBacon: Makes sense, originally we moved a lot of the authentication handling into EAS CLI so it was clearer to users how we handled/stored their credentials.

packages/eas-cli/src/credentials/ios/utils/authType.ts
```diff
@@ -1,6 +1,7 @@
 import { RequestContext } from '@expo/apple-utils';
 
 /** Is the request context App Store Connect only with no access to cookies authentication. */
+// todo: perhaps good to differentiate jwt vs cookies authentication upstream?
 export function isAppStoreConnectTokenOnlyContext(authContext: RequestContext): boolean {
```
EvanBacon, line 5: Internally we allow for both types and defer to using App Store Connect API when available [ref](https://github.com/expo/third-party/blob/3ac7f272d26bfc47e7660c0bb83b4bdf7cf6520d/packages/app-store/src/network/Request.ts#L107-L119). I added this extra step so `eas-cli` had explicit control on how we use the API. This is largely due to the fact that `apple-utils` doesn't have a proper governing body (or even CHANGELOG) to prevent logic changes like flipping the authentication priority to favor cookies first.

The purpose of this specific function is currently to bail out of API calls that we know do not work (like push keys, capability ids, provisioning profile repair).

packages/eas-json/src/build/types.ts
```diff
@@ -52,6 +53,12 @@ export interface IosBuildProfile extends CommonBuildProfile {
   artifactPath?: string;
   scheme?: string;
   buildConfiguration?: string;
+
+  ascApiKeyPath?: string;
```
EvanBacon, line 57:
```suggestion
  /** File path to the App Store Connect API key. */
  ascApiKeyPath?: string;
```

packages/eas-json/src/build/types.ts
```diff
@@ -52,6 +53,12 @@ export interface IosBuildProfile extends CommonBuildProfile {
...
+  ascApiKeyIssuerId?: string;
+  ascApiKeyId?: string;
+  appleTeamId?: string;
+  appleTeamType?: AppleTeamType;
```
EvanBacon, line 61:
```suggestion
  /** App Store Connect API issuer ID. This is a UUID v4 string (reserved variant). Example: `00x0xx00-0000-47e3-e053-0x0x0x00x0x0` */
  ascApiKeyIssuerId?: string;
  /** Opaque identifier for the App Store Connect API key. Example: `ZJZAAA11AA`*/
  ascApiKeyId?: string;
  appleTeamId?: string;
  appleTeamType?: AppleTeamType;
```

wkozyra95:
> is the build profile the best place to pass in a local asc api key? i see there's also credentials.json but that seems target based and only for build creds like dist cert and provisioning profile.

values in credentials.json are intended for unmanaged credentials so if credentials.json is present asc api key is unnecessary, I would prefer not to mix those 2 methods. We technically could put that in eas.json like we already do in submit a profile, but that case is a bit for legacy reasons. 

My preference would be to prompt for the key the same way we are already prompting for username and password, but also provide a way to configure that via environment variable(can we just pass a path to the file, or are those 3 values necessary like for submit profile?)

packages/eas-json/src/build/types.ts
wkozyra95, line 61: I would prefer not to configure that via eas.json, but I don't have strong opinion on that. If you decide to stay with this approach wrap all those fields in some object, so we are clear that e.g appleTeamId is required for asc API key not for other build stuff

packages/eas-cli/src/build/createContext.ts
```diff
@@ -43,13 +45,48 @@ export async function createBuildContextAsync<T extends Platform>({
   const workflow = await resolveWorkflowAsync(projectDir, platform);
   const accountId = findAccountByName(user.accounts, accountName)?.id;
 
-  const credentialsCtx = new CredentialsContext({
+  let credentialsCtxOptions = {
```
dsokal, line 48: Try avoiding using `let`. It's an easy refactor here.

packages/eas-cli/src/build/createContext.ts
```diff
@@ -43,13 +45,48 @@ export async function createBuildContextAsync<T extends Platform>({
...
+      iosBuildProfile.ascApiKeyId &&
+      iosBuildProfile.appleTeamId &&
+      iosBuildProfile.appleTeamType;
+
+    if (isPartialLocalAscApiKeyProvided && !isEntireLocalAscApiKeyProvided) {
+      throw new Error(
+        'To authenticate with an App Store Connect API Key, you must provide the following fields in your build profile: ascApiKeyPath, ascApiKeyIssuerId, ascApiKeyId, appleTeamId, appleTeamType.'
+      );
+    } else if (isEntireLocalAscApiKeyProvided) {
+      const keyP8 = await fs.readFile(iosBuildProfile.ascApiKeyPath!, 'utf-8');
```
dsokal, line 73: I'm not a fan of using `!`. The reason the variable is not truthy in this branch is that you can't guarantee that e.g. `.ascApiKeyPath` wasn't changed to a falsy value after the `isEntireLocalAscApiKeyProvided` was evaluated. This seems reasonable. In this case, I'd much rather prefer using nullthrows/assert

packages/eas-cli/src/build/createContext.ts
```diff
@@ -43,13 +45,48 @@ export async function createBuildContextAsync<T extends Platform>({
...
+      credentialsCtxOptions = {
+        ...credentialsCtxOptions,
+        defaultAppStoreAuthentication: {
+          mode: AuthenticationMode.API_KEY,
+          ascApiKey: {
+            keyP8,
+            keyId: iosBuildProfile.ascApiKeyId!,
+            issuerId: iosBuildProfile.ascApiKeyIssuerId!,
+          },
+          teamId: iosBuildProfile.appleTeamId!,
```
dsokal, line 83: Same comment about other `!` here.

packages/eas-cli/src/credentials/ios/actions/AscApiKeyUtils.ts
```diff
@@ -94,7 +94,6 @@ async function generateAscApiKeyAsync(
   ctx: CredentialsContext,
   purpose: AppStoreApiKeyPurpose
 ): Promise<MinimalAscApiKey> {
-  //console.log('debug');
```
dsokal, line 97: 👀 Lame, polish devs use `console.log('dupa');` 🤣  (this is true)

barthap: ^ That's correct, this methodology is officially called DDD (Dupa-driven debugging) 😆

dsokal:
I'm okay with the eas.json approach. However, as Wojtek suggested, I'd like to have support for env variables there. Similarly to what we have for submit profiles - see https://github.com/expo/eas-cli/blob/main/packages/eas-json/src/submit/resolver.ts#L28
This is important so we don't force users to store potentially sensitive data in files committed to the repo.

packages/eas-cli/src/credentials/ios/actions/AscApiKeyUtils.ts
quinlanj: 😠 😠 😠 😠 😠 😠 😠 😠 😠 
Screen Shot 2022-04-06 at 11 46 21 AM

quinlanj:
I've updated the PR with wojtek's suggestion where we authenticate with the api key the same way we do with username/password. You can supply the following env vars: `EXPO_ASC_API_KEY_PATH, EXPO_ASC_KEY_ID, EXPO_ASC_ISSUER_ID, EXPO_APPLE_TEAM_ID, EXPO_APPLE_TEAM_TYPE`. If there are any missing, it will prompt the user. 

If none of the ASC env vars are present, it will assume we are doing username/password authentication.

packages/eas-cli/src/build/createContext.ts
quinlanj: i've gotten rid of this code since we authenticate using env vars/ prompts now.

packages/eas-cli/src/build/createContext.ts
quinlanj: ditto

packages/eas-cli/src/submit/ios/AppProduce.ts
```diff
@@ -81,9 +80,11 @@ async function createAppStoreConnectAppAsync(
       bundleIdentifier: bundleId,
     });
   } else {
-    Log.warn(
-      `Provisioning is not available for user "${authCtx.appleId}", skipping bundle identifier check.`
-    );
+    const entity = isUserAuthCtx(authCtx)
```
wkozyra95, line 83: If I understand correctly asc api key does not support everything for submit, I might be wrong, but currently only creating new app (or maybe querying) on appstore connect requires a user login. We should probably handle that case before it errors with some cryptic error.

Another issue is that it might be weird if users do not have asc key configured and they have ascAppId in eas.json and will ask to provide asc API key even if they already authenticated with asc API key to login to apple

Both above points do not introduce any regressions on current behavior, so it's probably fine to handle them in separate PR

wkozyra95 (approved):

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -144,6 +152,33 @@ async function loginWithUserCredentialsAsync({
 }
 
 export async function authenticateAsync(options: Options = {}): Promise<AuthCtx> {
+  if (options.mode === AuthenticationMode.API_KEY) {
+    return await authenticateWithApiKeyAsync(options);
+  }
+  return await authenticateAsUserAsync(options);
```
dsokal, line 158:
```suggestion
  } else {
    return await authenticateAsUserAsync(options);
  }
```
It's easier to think about the branches in this case.

packages/eas-cli/src/credentials/ios/appstore/authenticate.ts
```diff
@@ -144,6 +152,33 @@ async function loginWithUserCredentialsAsync({
...
+  const team = await resolveAppleTeamAsync(options);
+  return {
+    team,
+    authState: {
+      context: {
+        token: new Token({
+          key: ascApiKey.keyP8,
+          issuerId: ascApiKey.issuerId,
+          keyId: ascApiKey.keyId,
+          duration: 1200, // 20 minutes
```
dsokal, line 173: I don't like "magic numbers". I'd assign this to a properly-named variable. @wkozyra95 would probably not agree with me though :D

packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -186,9 +187,11 @@ export async function ensureAppExistsAsync(
       });
     } catch (error: any) {
       if (error.message.match(/An App ID with Identifier '(.*)' is not available/)) {
-        const providerName = authCtx.authState?.session.provider.name;
+        const entity = isUserAuthCtx(authCtx)
+          ? `provider "${authCtx.authState?.session.provider.name}"`
+          : `apple team: ${authCtx.team.id}`;
```
dsokal, line 192:
```suggestion
          : `Apple Team: ${authCtx.team.id}`;
```

dsokal (approved):
LGTM

PR #1054, opened 2022-04-07
[eas-cli] Use @expo/multipart-body-parser library

wschurman:
# Why

https://github.com/expo/universe/pull/9396 extracted this code out into a library. This PR substitutes the library in.

# How

Add lib, update code.

# Test Plan

Run tests.


packages/eas-cli/src/utils/code-signing.ts
```diff
@@ -103,66 +104,26 @@ export async function getKeyAndCertificateFromPathsAsync({
...
 }
 
-export type MultipartPart = { headers: Map<string, string>; body: string };
+function convertArrayBufferToBuffer(ab: ArrayBuffer): Buffer {
+  const buf = Buffer.alloc(ab.byteLength);
+  const view = new Uint8Array(ab);
+  for (let i = 0; i < buf.length; ++i) {
+    buf[i] = view[i];
+  }
+  return buf;
```
ide, line 113: I would expect something like this to work:
```suggestion
  return Buffer.from(new Uint8Array(ab));
```
https://nodejs.org/api/buffer.html#static-method-bufferfrombuffer

or just inline the code

ide (approved):

PR #1070, opened 2022-04-15
[eas-cli] Enable update bundle output when debugging

byCedric:
# Checklist

# Why

In rare cases, the bundling could fail with heap out of memory errors in Github Actions. Without these logs, it's super hard to find this cause.

# How

Removed the harcoded `silent: true` and replaced it with `silent: !Log.isDebug`.

# Test Plan

- Linux/MacOS: `$ EXPO_DEBUG=true; eas update`
- Windows: `$ npx cross-env 'EXPO_DEBUG=true' eas update`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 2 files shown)


byCedric:
/changelog-entry bug-fix Enable full bundling logging for eas update when debugging

jonsamp (approved):
Thanks for implementing this debug logging fix

PR #1094, opened 2022-05-07
[eas-cli] set default build platform to all

giautm:
# Checklist

# Why

Resolve ENG-4624

# How

# Test Plan

### DIFF
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -37,6 +37,7 @@ export default class Build extends EasCommand {
 
   static flags = {
     platform: Flags.enum({
+      default: 'all',
       char: 'p',
       options: ['android', 'ios', 'all'],
     }),


kbrandwijk (changes requested):
Not a fan of adding this default, especially when usage based limits will be a thing soon, and forgetting to specify kicks of 2 builds. Especially on CI, it's very easy to be explicit about this, and locally you used to get an interactive question.

If we _really_ want to move forward with this change, I would do it for non-interactive mode only, but I'd rather not.

And if this somehow lands anyway, the code for the interactive prompt should be cleaned up.

wkozyra95 (changes requested):
Agree with @kbrandwijk , also there few other problems:
- it would be bad experience for free users because it would block builds for a bit for them if they can run just one concurrent build 
- a problem for us because we would run some unnecessary builds
- some options e.g. `--local` do not allow running builds with option all

dsokal (commented):
I think the same as Kim and Wojtek.

giautm:
Thanks for everyone giving the reviews. I will update the GA to mention that `--platform` is a required flag.

PR #1112, opened 2022-05-20
fix: match capability types more accurately

EvanBacon:
# Checklist

# Why

- Fix bug here https://forums.expo.dev/t/eas-build-failed-on-ios-associated-domains-capability/61662/5

# How

- To support both App Store Connect (`FEAT`) and cookies authentication (`BUNDLEID_FEAT`) we were matching capability types with `endsWith`. This caused and issue where projects with `MDM_MANAGED_ASSOCIATED_DOMAINS` enabled and `ASSOCIATED_DOMAINS` in use locally would be corrupted because `MDM_MANAGED_ASSOCIATED_DOMAINS` would match `ASSOCIATED_DOMAINS` and `ASSOCIATED_DOMAINS` would not be enabled remotely. We now have a more complex regex https://github.com/expo/third-party/pull/81 which has a higher accuracy rate.

# Test Plan

- Added tests here and in https://github.com/expo/third-party/pull/81

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -8,7 +8,7 @@
   },
   "bugs": "https://github.com/expo/eas-cli/issues",
   "dependencies": {
-    "@expo/apple-utils": "0.0.0-alpha.31",
+    "@expo/apple-utils": "0.0.0-alpha.33",
     "@expo/code-signing-certificates": "0.0.2",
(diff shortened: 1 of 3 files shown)


EvanBacon:
/changelog-entry bug-fix match bundle identifier capabilities more accurately.

byCedric (approved):

PR #1116, opened 2022-05-23
[eas-cli] Add metadata telemetry and error reporting

byCedric:
# Checklist

# Why

This adds extensive error reporting whenever a request to ASC fails. It includes a couple of things:
- Set the error code on the command in case of failures
- In case of failures, send any failures as error reports to us
- Remove sensitive data like App IDs and/or keys
- Group the events by a unique execution ID, and give the user an opt-in to share that with us

# How

Because of our history with the ASC API, we need extensive visibility of any error responses to keep our validation and scripts in sync with future changes to the Apple ASC API. By sending us error reports, we should have that visibility without sending us sensitive data. This allows us to anticipate these future changes.

# Test Plan

TBD


packages/eas-cli/src/metadata/errors.ts
```diff
@@ -0,0 +1,17 @@
+import type { ErrorObject } from 'ajv';
+
+export class MetadataValidationError extends Error {
+  constructor(message?: string, public readonly errors?: ErrorObject[]) {
+    super(message ?? 'Metadata validation failed');
+  }
+}
+
+export class MetadataUploadError extends Error {
```
EvanBacon, line 9: Comment why this would be thrown

PR #1119, opened 2022-05-24
[eas-cli] lower page limit for UpdateQueries

kgc00:
# Checklist

# Why

A user has [reported](https://forums.expo.dev/t/eas-update-gateway-timeout/63783) that requests to publish updates started to time out a few weeks ago. 

# How

I inspected the screenshot they provided and saw that they time out after the update has been linked to the project. I then inspected the code in that commands execution path after an update gets linked for any web requests. I found `UpdateQuery.viewBranchAsync` on `.../commands/update/index.ts` ln ~132 and realized that this query times out consistently when i run it for their particular branch. the fix:
 
* lower page limit from 10_000 to 300 in UpdateQuery.ts

# Test Plan

I noticed that the `app.byId.updateBranch*.updates` query started to time out after the branch contained > ~600 updates. I ran queries against production using the website and an API client for a branch that contained ~2k updates:

```
query branchDetail($fullName: String!, $name: String!, $limit: Int!) {
  app {
      updateBranchByName(name: $name) {
        updates(offset: 0, limit: $limit) {
          id
          group
          message
          createdAt
          runtimeVersion
          platform
        }
      }
    }
  }
}
```

I started small and increased the limit until the requests started to time out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Match bundle identifier capabilities more accurately. ([#1112](https://github.com/expo/eas-cli/pull/1112) by [@EvanBacon](https://github.com/EvanBacon))
+- No longer timeout on publishing updates for branches with many updates. ([#1119](https://github.com/expo/eas-cli/pull/1119)) by [@kgc00](https://github.com/kgc00/)
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


kgc00:
@ide I was initially going to do a more thorough re-write of `eas update` that included multiple queries to get all the updates a user has published. However, I think it's worth splitting that out into a follow-up PR as it will take time. In the meantime, just this should unblock our users that work on branches with thousands of updates in regards to using eas cli.

Work that should still be done in a follow up:
- `eas update` when used with the `republish` or `group` options, uses the updates queried to figure out which group to republish... 
  - IMO we should expose some gql endpoint we can query that would provide update groups for an account and use that instead. 
- `eas update:list` renders a list of updates to the user. 
  - I'd probably want to implement some 'fetch 50 updates, if the user wants the next page of updates, they can hit a button and we will make another request'. 

Alternatively, I can just re-query with offset + limit until we have all the updates for both cases (simulating our current behavior). But it seems unnecessary and slow to fetch all the updates you've ever published every time you run a cli command.

ide (approved):

PR #1122, opened 2022-05-25
[eas-cli] use getExpoConfig wrapper

wkozyra95:
# Checklist

# Why

Use wrapper around config everywhere.

Most likely we will need to override some values returned in config to support https://linear.app/expo/issue/ENG-3863/version-bumping so having one wrapper would enable us to override config everywhwere

# How

replace all uses of getConfig

# Test Plan

TODO


dsokal (approved):

wkozyra95:
/changelog-entry chore Use getExpoConfig to access config

PR #1126, opened 2022-05-30
[eas-cli] Export ora as default from mock

byCedric:
# Checklist

This is a mock-fix only, so I don't think it should be listed in the changelogs. It's just an internal thing.

# Why

We import `ora` as default import in **[src/ora.ts](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/ora.ts#L2)**. Whenever we test anything that includes Ora code, it would fail with this error ([source](https://github.com/expo/eas-cli/runs/6655213156?check_suite_focus=true#step:6:197)):

```
TypeError: (0 , ora_1.default) is not a function
  24 |   const inputOptions = typeof options === 'string' ? { text: options } : options ?? {};
  25 |   const disabled = Log.isDebug || !process.stdin.isTTY;
> 26 |   const spinner = oraReal({
     |                          ^
  27 |     // Ensure our non-interactive mode emulates CI mode.
  28 |     isEnabled: !disabled,
  29 |     // In non-interactive mode, send the stream to stdout so it prevents looking like an error.
```

# How

Export the ora stub as default from the mock.

# Test Plan

Add this export as default, and try to test code using ora (e.g. see PR #1116)

### DIFF
--- a/packages/eas-cli/src/__mocks__/ora.ts
+++ b/packages/eas-cli/src/__mocks__/ora.ts
@@ -7,3 +7,6 @@ export const ora = jest.fn().mockReturnValue({
   stopAndPersist: jest.fn(),
   succeed: jest.fn(),
 });
+
+// Ora is imported as default import
+export default ora;


byCedric:
Ah I think you are right, I mistakenly took this for a node module mock while it's positioned as a user module mock.

PR #1083, opened 2022-04-25
[metadata] Add first draft of app store connect metadata sync

byCedric:
# Checklist

- [ ] I've added an entry to [CHANGELOG.md](https://github.com/expo/eas-cli/blob/main/CHANGELOG.md#main) if necessary. You can comment this pull request with `/changelog-entry [breaking-change|new-feature|bug-fix|chore] [message]` and CHANGELOG.md will be updated automatically.

# This PR has been split

In order to do proper review, I separated this PR into a set of other PRs.

- ~~https://github.com/expo/eas-cli/pull/1130~~
  - ~~https://github.com/expo/eas-cli/pull/1131~~
  - ~~https://github.com/expo/eas-cli/pull/1132~~
  - ~~https://github.com/expo/eas-cli/pull/1133~~
- ~~https://github.com/expo/eas-cli/pull/1134~~
- ~~https://github.com/expo/eas-cli/pull/1135~~
  - https://github.com/expo/eas-cli/pull/1136

# Why

First draft of the EAS submit metadata, running everything locally.

> Note, we might need to move this over to EAS submit service instead.

# How

This adds basic App Store Connect metadata configuration syncs using a couple of concepts.

1. **config schema** → JSON Schema that defines the available configuration options available. (Using JSON Schemas allows us to create autocomplete in editors like vscode)
2. **config reader** → Deserializer object to convert the JSON data to ASC model attributes. (Could change per config schema version)
3. **context** → A set of ASC model instances to communicate with the ASC API.
4. **sync tasks** → A simple task runner system using `preupload` and `upload` stages, more info in the proposal.

# Test Plan

For all apple metadata tasks, there are unit tests with mocked ASC responses.

> The underlying HTTP library needs some patches for nock, the tests might fail until that's available.

PR #1130, opened 2022-05-31
[eas-cli] Add metadata store config serializers

byCedric:
# Checklist

# Why

Part of PR #1083

This adds the basic store config serializer to read and write the JSON file. It also includes [some utility methods](https://github.com/expo/eas-cli/tree/%40bycedric/metadata/draft/packages/eas-cli/src/metadata/utils) that we use throughout the metadata part.

# How

- The config reader should support multiple versions of the store config, mostly for when we add or change functionality in the future.
- The config writer can always be the latest version, we don't need to support older versions as we generate them on the fly.

# Test Plan

The config serializers are used before we download or upload the store configuration. The writer has some unit tests, I omitted the reader unit tests to focus on the functionality. Can add that later though!

### DIFF
--- a/packages/eas-cli/src/metadata/apple/config/__tests__/fixtures/appStoreVersionLocalization.ts
+++ b/packages/eas-cli/src/metadata/apple/config/__tests__/fixtures/appStoreVersionLocalization.ts
@@ -1,0 +1,23 @@
+import { AppStoreVersionLocalization } from '@expo/apple-utils';
+
+import { AttributesOf } from '../../../../utils/asc';
+
+export const englishVersion: AttributesOf<AppStoreVersionLocalization> = {
+  locale: 'en-US',
+  description: 'This is a description of this version',
+  keywords: 'some, description',
+  marketingUrl: 'https://example.com/en/marketing',
+  promotionalText: 'This is some promotional text',
+  supportUrl: 'https://example.com/en/support',
+  whatsNew: 'Bugfixes and improvements',
+};
+
+export const dutchVersion: AttributesOf<AppStoreVersionLocalization> = {
+  locale: 'nl-NL',
+  description: 'Dit is een beschrijving van deze versie',
+  keywords: 'een, beschrijving',
+  marketingUrl: 'https://example.com/nl/marketing',
+  promotionalText: 'Dit is wat promotie tekst',
+  supportUrl: 'https://example.com/nl/support',
+  whatsNew: 'Beestreparaties en verbeteringen',
+};
--- a/packages/eas-cli/src/metadata/apple/config/reader.ts
+++ b/packages/eas-cli/src/metadata/apple/config/reader.ts
@@ -1,0 +1,121 @@
+import {
+  AgeRatingDeclaration,
+  AppInfoLocalization,
+  AppStoreVersion,
+  AppStoreVersionLocalization,
+  CategoryIds,
+  ReleaseType,
+} from '@expo/apple-utils';
+
+import { unique } from '../../utils/array';
+import { AttributesOf } from '../../utils/asc';
+import { removeDatePrecision } from '../../utils/date';
+import { AppleMetadata } from '../types';
+
+type PartialExcept<T, K extends keyof T> = Pick<T, K> & Partial<Omit<T, K>>;
+
+// TODO: find out if we can move this to default JSON schema normalization
+export const DEFAULT_WHATSNEW = 'Bug fixes and improved stability';
+
+/**
+ * Deserializes the metadata config schema into attributes for different models.
+ * This uses version 0 of the config schema.
+ */
+export class AppleConfigReader {
+  constructor(public readonly schema: AppleMetadata) {}
+
+  getAgeRating(): Partial<AttributesOf<AgeRatingDeclaration>> | null {
+    return this.schema.advisory || null;
+  }
+
+  getLocales(): string[] {
+    // TODO: filter "default" locales, add option to add non-localized info to the config
+    return unique(Object.keys(this.schema.info || {}));
+  }
+
+  getInfoLocale(
+    locale: string
+  ): PartialExcept<AttributesOf<AppInfoLocalization>, 'locale' | 'name'> | null {
+    const info = this.schema.info?.[locale];
+    if (!info) {
+      return null;
+    }
+
+    return {
+      locale,
+      name: info.title || 'no name provided',
+      subtitle: info.subtitle,
+      privacyChoicesUrl: info.privacyChoicesUrl,
+      privacyPolicyText: info.privacyPolicyText,
+      privacyPolicyUrl: info.privacyPolicyUrl,
+    };
+  }
+
+  getCategories(): CategoryIds | null {
+    if (Array.isArray(this.schema.categories) && this.schema.categories.length > 0) {
+      return {
+        primaryCategory: this.schema.categories[0],
+        secondaryCategory: this.schema.categories[1],
+      };
+    }
+
+    return null;
+  }
+
+  getVersion(): Partial<
+    Omit<AttributesOf<AppStoreVersion>, 'releaseType' | 'earliestReleaseDate'>
+  > | null {
+    return this.schema.copyright ? { copyright: this.schema.copyright } : null;
+  }
+
+  getVersionRelease(): Partial<
+    Pick<AttributesOf<AppStoreVersion>, 'releaseType' | 'earliestReleaseDate'>
+  > | null {
+    const { release } = this.schema;
+
+    if (release?.autoReleaseDate) {
+      return {
+        releaseType: ReleaseType.SCHEDULED,
+        // Convert time format to 2020-06-17T12:00:00-07:00
+        earliestReleaseDate: removeDatePrecision(release.autoReleaseDate)?.toISOString() ?? null,
+      };
+    }
+
+    if (release?.automaticRelease === true) {
+      return {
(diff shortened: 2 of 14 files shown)


packages/eas-cli/src/metadata/apple/config/__tests__/fixtures/appStoreVersionLocalization.ts
```diff
@@ -0,0 +1,23 @@
+import { AppStoreVersionLocalization } from '@expo/apple-utils';
```
EvanBacon, line 1: These are some articulate fixtures...

packages/eas-cli/src/metadata/apple/config/reader.ts
```diff
@@ -0,0 +1,121 @@
...
+      return {
+        primaryCategory: this.schema.categories[0],
+        secondaryCategory: this.schema.categories[1],
+      };
+    }
+
+    return null;
+  }
+
+  getVersion(): Partial<
```
EvanBacon, line 65:
```suggestion
  /** Get the `AppStoreVersion` object. */
  getVersion(): Partial<
```

added a comment cuz usually I think of "1' or "2.0.0" when I read version.

packages/eas-cli/src/metadata/utils/date.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * "status": "409",
+ * "code": "ENTITY_ERROR.ATTRIBUTE.INVALID",
+ * "title": "An attribute value is invalid.",
+ * "detail": "The attribute 'earliestReleaseDate' only allows hour precision",
+ * "source": {
+ *   "pointer": "/data/attributes/earliestReleaseDate"
+ * }
+ */
+export function removeDatePrecision(date: any): null | Date {
```
EvanBacon, line 12: /testme

packages/eas-cli/src/metadata/apple/config/__tests__/fixtures/appStoreVersionLocalization.ts
byCedric: I blame @apple 🙈

packages/eas-cli/src/metadata/apple/config/writer.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { AttributesOf } from '../../utils/asc';
+import { AppleMetadata } from '../types';
+
+/**
+ * Serializes the Apple ASC entities into the metadata configuration schema.
+ * This uses version 0 of the config schema.
+ */
+export class AppleConfigWriter {
+  constructor(public readonly schema: Partial<AppleMetadata> = {}) {}
```
wkozyra95, line 18: use public/private access modifiers explicitly on all methods in class

packages/eas-cli/src/metadata/apple/config/writer.ts
```diff
@@ -0,0 +1,107 @@
...
+      releaseNotes: optional(attributes.whatsNew),
+      marketingUrl: optional(attributes.marketingUrl),
+      promoText: optional(attributes.promotionalText),
+      supportUrl: optional(attributes.supportUrl),
+    };
+  }
+}
+
+/** Helper function to convert `T | null` to `T | undefined`, required for the entity properties */
+function optional<T>(value: T | null): T | undefined {
```
wkozyra95, line 105: name does not tell you much about what it is doing, also you are not saving any space here, because
`optional(x)` is not much shorter than `x ?? undefined`

packages/eas-cli/src/metadata/apple/config/writer.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.schema.info[attributes.locale] = {
+      ...existing,
+      description: optional(attributes.description),
+      keywords: optional(attributes.keywords)
+        ?.split(',')
+        .map(keyword => keyword.trim()),
+      releaseNotes: optional(attributes.whatsNew),
+      marketingUrl: optional(attributes.marketingUrl),
+      promoText: optional(attributes.promotionalText),
+      supportUrl: optional(attributes.supportUrl),
```
wkozyra95, line 99: in most cases, it does not matter, but we usually prefer the syntax
`...(attributes.supportUrl ? { supportUrl: attributes.supportUrl } : {})`
in case a key with undefined value is treated differently than when the key is missing. It would be rare for it to work this way, but with above approach you do not need to think about it.

packages/eas-cli/src/metadata/utils/retry.ts
```diff
@@ -0,0 +1,28 @@
...
+  return new Promise(resolve => setTimeout(resolve, duration));
+}
+
+type WithRetryOptions = {
+  tries?: number;
+  delay?: number;
+  onRetry?: (triesLeft: number) => void;
+};
+
+export async function retryIfNullAsync<T>(
```
wkozyra95, line 11: I would write a generic retry function and move it to common utils for cli
retrying when the result is null is a bit weird, so probably a cleaner version to handle that would be on call site
e.g.
```
await retryAsync(async () => {
   const result = app.getEditAppStoreVersionAsync({ platform });
   if (!resul) {
     throw new Error('some not found message')
   }
}, retryOptions)
```
also you probably want to retry if `method` throws an exception

packages/eas-cli/src/metadata/utils/date.ts
```diff
@@ -0,0 +1,26 @@
...
```
wkozyra95, line 12: This shouldn't need any type

packages/eas-cli/src/metadata/utils/array.ts
```diff
@@ -0,0 +1,4 @@
+export function unique<T = any>(items: T[]): T[] {
```
wkozyra95, line 1: we already have uniqBy https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/utils/expodash/uniqBy.ts

byCedric: `unique` and `uniqBy` seems to be different functions, we don't need the `By` functionality here. We just want whatever is unique, I think we mostly use this to find locales from the config. This `uniqBy` seems to also require a `getKey` method, which would be `v => v` in our case, plus it seems to be slower as it would iterate everything (not sure if that would be noticeable)

Happy to just add a `unique` method to expodash though?

packages/eas-cli/src/metadata/apple/config/writer.ts
byCedric: It's not shorter, I agree with that but makes more sense when reading big chunks of code in the writer class. "optional attribute X"

packages/eas-cli/src/metadata/utils/array.ts
wkozyra95: I would just use uniqBy, but I'm fine with just moving it to expodash(and renaming to be consistent `uniq`)

packages/eas-cli/src/metadata/utils/date.ts
byCedric: Fixed in https://github.com/expo/eas-cli/pull/1147

packages/eas-cli/src/metadata/apple/config/writer.ts
byCedric: Yep, agreed! In this case, we use the data from the writer to serialize everything to JSON. And that is being handled properly:

Screenshot 2022-06-07 at 18 45 19

Not sure if it's worth making the code harder to read/maintain and adding nested object creations + object spreads for each individual property?

byCedric: Also, one of the reasons for this is that the types in apple utils mostly require `null` or a value. E.g.

Screenshot 2022-06-07 at 18 48 05

So, if we want to use these properties for models, we have to use `null` instead. And if we want to omit those "non-values" in JSON, we'd have to use `undefined`.

packages/eas-cli/src/metadata/apple/config/writer.ts
wkozyra95: using wrapper function might be cleaner, but in this optional is not a good name, in most languages function optional would return abstraction that provide interface e.g. `hasValue(): bool`, `getValue(): T` where T is not nullable

in this case, your function is replacing nulls with undefined

packages/eas-cli/src/metadata/utils/array.ts
byCedric: Done in PR https://github.com/expo/eas-cli/pull/1147

packages/eas-cli/src/metadata/apple/config/writer.ts
byCedric: > in this case, your function is replacing nulls with undefined

Yes, that's what I meant to make clear here. I also tried to explain why in the comment below, but basically we have this:

- For apple utils, we need to use `null` values for non-values.
- For the `store.config.json` we need to omit writing these values so users won't have configs with all available properties written out (we'll default to sensible values when users don't write it)
- Some attributes can actually be `null`, so we have to do the `null -> undefined` conversion per attribute.

What would be your suggestion to name this? I think optional makes sense in this case, but I'm open to improvements :)

wkozyra95: > What would be your suggestion to name this

that's the point, I think doing that inline is cleaner because it's hard to name

PR #1131, opened 2022-05-31
[eas-cli] Add metadata config validation using json schema

byCedric:
# Checklist

# Why

- Part of PR #1083
- [x] ~~Requires PR #1130~~

This adds basic JSON validation for the store configuration, using AJV and a custom JSON schema.

# How

- This contains well-known issues/limitations of the stores, to help users inform of configurational errors.
- In cases of errors, it fails before we communicate with the stores to keep the feedback loop as fast as we can.

# Test Plan

- If you go to PR #1083, generate a store config, and mess up some configuration, you can see the errors.
- Alternatively, you could add the custom JSON schema to vscode to see live validation as you code. Will probably add this to Expo Tools later as well.

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -30,6 +30,7 @@
     "@oclif/core": "1.9.0",
     "@urql/core": "2.4.4",
     "@urql/exchange-retry": "0.3.3",
+    "ajv": "^6.12.6",
     "chalk": "4.1.2",
     "cli-progress": "3.11.0",
     "cli-table3": "0.6.2",
--- a/packages/eas-cli/schema/metadata-0.json
+++ b/packages/eas-cli/schema/metadata-0.json
@@ -1,0 +1,928 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema",
+  "type": "object",
+  "definitions": {
+    "apple": {
+      "AppleLocalizedPreviewFolder": {
+        "anyOf": [
+          {
+            "description": "Path to folder containing preview files",
+            "type": "string"
+          },
+          {
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+              "type": "string"
+            }
+          }
+        ]
+      },
+      "AppleKidsAge": {
+        "enum": ["FIVE_AND_UNDER", "SIX_TO_EIGHT", "NINE_TO_ELEVEN"]
+      },
+      "AppleRating": {
+        "enum": ["NONE", "INFREQUENT_OR_MILD", "FREQUENT_OR_INTENSE"]
+      },
+      "AppleCategory": {
+        "enum": [
+          "FOOD_AND_DRINK",
+          "BUSINESS",
+          "EDUCATION",
+          "SOCIAL_NETWORKING",
+          "BOOKS",
+          "SPORTS",
+          "FINANCE",
+          "REFERENCE",
+          "GRAPHICS_AND_DESIGN",
+          "DEVELOPER_TOOLS",
+          "HEALTH_AND_FITNESS",
+          "MUSIC",
+          "WEATHER",
+          "TRAVEL",
+          "ENTERTAINMENT",
+          "STICKERS",
+          "GAMES",
+          "LIFESTYLE",
+          "MEDICAL",
+          "MAGAZINES_AND_NEWSPAPERS",
+          "UTILITIES",
+          "SHOPPING",
+          "PRODUCTIVITY",
+          "NEWS",
+          "PHOTO_AND_VIDEO",
+          "NAVIGATION"
+        ]
+      },
+      "AppleSubcategory": {
+        "enum": [
+          "STICKERS_PLACES_AND_OBJECTS",
+          "STICKERS_EMOJI_AND_EXPRESSIONS",
+          "STICKERS_CELEBRATIONS",
+          "STICKERS_CELEBRITIES",
+          "STICKERS_MOVIES_AND_TV",
+          "STICKERS_SPORTS_AND_ACTIVITIES",
+          "STICKERS_EATING_AND_DRINKING",
+          "STICKERS_CHARACTERS",
+          "STICKERS_ANIMALS",
+          "STICKERS_FASHION",
+          "STICKERS_ART",
+          "STICKERS_GAMING",
+          "STICKERS_KIDS_AND_FAMILY",
+          "STICKERS_PEOPLE",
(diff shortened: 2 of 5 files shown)


packages/eas-cli/schema/metadata-0.json
```diff
@@ -0,0 +1,928 @@
...
+            "maxLength": 255,
+            "description": "A URL where users can modify and delete the data collected from the app, or decide how their data is used and shared.",
+            "type": "string"
+          }
+        }
+      },
+      "AppleAdvisory": {
+        "type": "object",
+        "additionalProperties": false,
+        "defaultSnippets": [
```
EvanBacon, line 254: Probably safe to delete these default snippets for now, can add back later when we update to match the latest spec.

packages/eas-cli/schema/metadata-0.json
```diff
@@ -0,0 +1,928 @@
...
+          "privacyPolicyUrl": {
+            "meta": {
+              "versioned": true,
+              "liveEdits": true
+            },
+            "maxLength": 255,
+            "description": "A URL that links to your privacy policy. A privacy policy is required for all apps.",
+            "type": "string"
+          },
+          "privacyChoiceUrl": {
```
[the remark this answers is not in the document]
byCedric: Fixed

EvanBacon (approved):

packages/eas-cli/package.json
```diff
@@ -30,6 +30,7 @@
     "@oclif/core": "1.9.0",
     "@urql/core": "2.4.4",
     "@urql/exchange-retry": "0.3.3",
+    "ajv": "^6.12.6",
```
wkozyra95, line 33: we already use joi for validation, do we need a separate library?
also if it's needed use strict dependency

packages/eas-cli/src/metadata/config.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Run the JSON Schema validation to normalize defaults and flag early config errors.
+ * This includes validating the known store limitations for every configurable property.
+ */
+export function validateConfig(config: unknown): {
+  valid: boolean;
+  errors: Ajv.ErrorObject[];
+} {
+  const validator = new Ajv({ allErrors: true, useDefaults: true })
+    .addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'))
```
wkozyra95, line 24: Is there any reason you are using require instead of import?

packages/eas-cli/src/metadata/config.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Run the JSON Schema validation to normalize defaults and flag early config errors.
+ * This includes validating the known store limitations for every configurable property.
+ */
+export function validateConfig(config: unknown): {
+  valid: boolean;
+  errors: Ajv.ErrorObject[];
+} {
```
wkozyra95, line 22: missing return type

packages/eas-cli/src/metadata/config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const valid = validator(config) as boolean;
+
+  return { valid, errors: validator.errors || [] };
+}
+
+/** Create a versioned deserializer to fetch App Store data from the store configuration. */
+export function createAppleReader(config: MetadataConfig): AppleConfigReader {
+  assert(config.configVersion === 0, 'Unsupported store configuration version');
+  assert(config.apple, 'No apple configuration found');
```
wkozyra95, line 35: assert should be used to mitigate programmer's error, if you want to handle the error that might happen because the user e.g. select  incorrect option then it should throw an exception

packages/eas-cli/package.json
byCedric: We need JSON schema validation, not Joi. The reason for that is that we want to integrate this with intellisense in vscode, which only supports JSON schema.

I've tried to look for Joi -> JSON Schema libraries to integrate the `eas.json` validation in https://github.com/expo/vscode-expo, but none would also take a description and other documentation features into account. That's why we opted to go for JSON schema directly.

packages/eas-cli/src/metadata/config.ts
byCedric: will fix!

packages/eas-cli/src/metadata/config.ts
byCedric: I'm not sure if I understand this. You are right that these would be user-configuration errors, but I'm not sure why those should be thrown errors.

From [the docs](https://nodejs.org/api/assert.html):
> The `node:assert` module provides a set of assertion functions for verifying invariants.

With React, for example, the user could be returning the wrong underlying component, causing invariant errors, right?

packages/eas-cli/src/metadata/config.ts
byCedric: I don't think this reason applies to EAS CLI, but in the new ES module syntax you can't just import JSON files like `import('path/to/file.json')` anymore. It requires a [new syntax from ES JSON modules](https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/), so I try to avoid using `import` for plain JSON files in case we want to move over to ES modules.

packages/eas-cli/src/metadata/config.ts
wkozyra95: assert should not be used for error handling, few examples

bad use cases:
 - user-specified an invalid value
 - network request failed
 - API returned an error

 good use cases:
 - if you know that you validated sth in the past but type does not reflect that (e.g. after refactor assert might fail, which shows a bug in a code)
 - if API response returns sth in a wrong format and I'm not talking (already a bit questionable)
 - vary rare errors e.g. out of disk out of memory (also questionable)
 
 https://en.wikipedia.org/wiki/Assertion_(software_development)

> an assertion is a [predicate](https://en.wikipedia.org/wiki/Predicate_(mathematical_logic)) (a [Boolean-valued function](https://en.wikipedia.org/wiki/Boolean-valued_function) over the [state space](https://en.wikipedia.org/wiki/State_space), usually expressed as a [logical proposition](https://en.wikipedia.org/wiki/Logical_proposition) using the [variables](https://en.wikipedia.org/wiki/Variable_(programming)) of a program) connected to a point in the program, that always should evaluate to true at that point in code execution. Assertions can help a programmer read the code, help a [compiler](https://en.wikipedia.org/wiki/Compiler) compile it, or help the program detect its own defects.

PR #1133, opened 2022-05-31
[eas-cli] Add metadata app store task runner

byCedric:
# Checklist

# Why

- Part of PR #1083
- [x] ~~Requires PR #1130~~
- [x] ~~Requires PR #1134~~

This adds the task-runner based entity actions to perform the download/upload actions, see Notion for more info.

# How

- Every task runner works independently from another task
- Every task can have data-dependencies that have to be asserted
- When a task fails, the other tasks should still be executed to try and send/get as much data as possible
- When the command is finished, the exit status is determined by "encountered any issue or not" and should log helpful context to the user

# Test Plan

See tests with fixtures from the ASC

### DIFF
--- a/packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
+++ b/packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
@@ -1,0 +1,42 @@
+import { AgeRatingDeclaration } from '@expo/apple-utils';
+import assert from 'assert';
+import chalk from 'chalk';
+
+import Log from '../../../log';
+import { logAsync } from '../../utils/log';
+import { AppleTask, TaskDownloadOptions, TaskPrepareOptions, TaskUploadOptions } from '../task';
+
+export type AgeRatingData = {
+  /** The app age rating declaration for the app version */
+  ageRating: AgeRatingDeclaration;
+};
+
+export class AgeRatingTask extends AppleTask {
+  name = (): string => 'age rating declarations';
+
+  async prepareAsync({ context }: TaskPrepareOptions): Promise<void> {
+    assert(context.version, `App version information is not prepared, can't update age rating`);
+    context.ageRating = (await context.version.getAgeRatingDeclarationAsync()) || undefined;
+  }
+
+  async downloadAsync({ config, context }: TaskDownloadOptions): Promise<void> {
+    if (context.ageRating) {
+      config.setAgeRating(context.ageRating.attributes);
+    }
+  }
+
+  async uploadAsync({ config, context }: TaskUploadOptions): Promise<void> {
+    assert(context.ageRating, `Age rating not initialized, can't update age rating`);
+
+    const ageRating = config.getAgeRating();
+    if (!ageRating) {
+      Log.log(chalk`{dim - Skipped age rating update, no advisory configured}`);
+    } else {
+      context.ageRating = await logAsync(() => context.ageRating.updateAsync(ageRating), {
+        pending: 'Updating age rating declaration...',
+        success: 'Updated age rating declaration',
+        failure: 'Failed to update age rating declaration',
+      });
+    }
+  }
+}
--- a/packages/eas-cli/src/metadata/apple/tasks/app-info.ts
+++ b/packages/eas-cli/src/metadata/apple/tasks/app-info.ts
@@ -1,0 +1,88 @@
+import { AppInfo, AppInfoLocalization } from '@expo/apple-utils';
+import assert from 'assert';
+import chalk from 'chalk';
+
+import Log from '../../../log';
+import { logAsync } from '../../utils/log';
+import { retryIfNullAsync } from '../../utils/retry';
+import { AppleTask, TaskDownloadOptions, TaskPrepareOptions, TaskUploadOptions } from '../task';
+
+export type AppInfoData = {
+  /** The current app info that should be edited */
+  info: AppInfo;
+  /** All info locales that are enabled */
+  infoLocales: AppInfoLocalization[];
+};
+
+export class AppInfoTask extends AppleTask {
+  name = (): string => 'app information';
+
+  async prepareAsync({ context }: TaskPrepareOptions): Promise<void> {
+    const info = await retryIfNullAsync(() => context.app.getEditAppInfoAsync());
+    assert(info, 'Could not resolve the editable app info to update');
+
+    context.info = info;
+    context.infoLocales = await info.getLocalizationsAsync();
+  }
+
+  async downloadAsync({ config, context }: TaskDownloadOptions): Promise<void> {
+    assert(context.info, `App info not initialized, can't download info`);
+
+    // TODO: see why this type mismatch occurs
+    // @ts-ignore
+    config.setCategories(context.info.attributes);
+
+    for (const locale of context.infoLocales) {
+      config.setInfoLocale(locale.attributes);
+    }
+  }
+
+  async uploadAsync({ config, context }: TaskUploadOptions): Promise<void> {
+    assert(context.info, `App info not initialized, can't update info`);
+
+    const categories = config.getCategories();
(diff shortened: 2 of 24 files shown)


packages/eas-cli/src/metadata/apple/tasks/app-info.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    context.info = info;
+    context.infoLocales = await info.getLocalizationsAsync();
+  }
+
+  async downloadAsync({ config, context }: TaskDownloadOptions): Promise<void> {
+    assert(context.info, `App info not initialized, can't download info`);
+
+    // TODO: see why this type mismatch occurs
+    // @ts-ignore
```
EvanBacon, line 32:
```suggestion
    // @ts-expect-error
```

so you're reminded to remove this

packages/eas-cli/src/metadata/apple/tasks/app-info.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    context.info = info;
+    context.infoLocales = await info.getLocalizationsAsync();
+  }
+
+  async downloadAsync({ config, context }: TaskDownloadOptions): Promise<void> {
+    assert(context.info, `App info not initialized, can't download info`);
+
+    // TODO: see why this type mismatch occurs
+    // @ts-expect-error
```
EvanBacon, line 32: This appears to be unused in the tests? If this is for sure a type error locally then we should switch to `ts-ignore` and fix the tests later.

byCedric: ah, I think I fixed this issue just before opening the PRs. (during polishing) So here we can see the value of `ts-expect-error` over `ts-ignore` 😂

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
```diff
@@ -0,0 +1,42 @@
...
+export type AgeRatingData = {
+  /** The app age rating declaration for the app version */
+  ageRating: AgeRatingDeclaration;
+};
+
+export class AgeRatingTask extends AppleTask {
+  name = (): string => 'age rating declarations';
+
+  async prepareAsync({ context }: TaskPrepareOptions): Promise<void> {
+    assert(context.version, `App version information is not prepared, can't update age rating`);
```
wkozyra95, line 18: the same as in other PR, if this is error handling it shouldn't use asserts(the same applies to other asserts)

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
```diff
@@ -0,0 +1,42 @@
...
+    context.ageRating = (await context.version.getAgeRatingDeclarationAsync()) || undefined;
+  }
+
+  async downloadAsync({ config, context }: TaskDownloadOptions): Promise<void> {
```
wkozyra95, line 22: It does not look like this method is doing what the description in abstract class is claiming (the same applies to other tasks)

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
```diff
@@ -0,0 +1,42 @@
...
+import { logAsync } from '../../utils/log';
+import { AppleTask, TaskDownloadOptions, TaskPrepareOptions, TaskUploadOptions } from '../task';
+
+export type AgeRatingData = {
+  /** The app age rating declaration for the app version */
+  ageRating: AgeRatingDeclaration;
+};
+
+export class AgeRatingTask extends AppleTask {
+  name = (): string => 'age rating declarations';
```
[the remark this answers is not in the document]
byCedric: With methods, we are able to take dynamic properties into account. For the current task runners, we don't need that specifically. But I'm also working on adding screenshot tasks. With that we would be able to mention the path of the screenshots or the amount of screenshots we are uploading.

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  async prepareAsync({ context }: TaskPrepareOptions): Promise<void> {
```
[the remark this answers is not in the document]
byCedric: Will fix

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
byCedric: Are you referring to this?
> Download all information from the App Store to generate the store configuration

If so, I think it does. The only thing `downloadAsync` is doing is pulling information from the stores. That would be done by loading each individual models or entities (Download all information from the App Store), and populating the `AppleConfigWriter` (to generate the store configuration).

It does not generate the store config itself, that's what the `AppleConfigWriter` is doing. But it is required to be able to generate that config.

packages/eas-cli/src/metadata/apple/tasks/index.ts
```diff
@@ -0,0 +1,12 @@
+import { MetadataContext } from '../../context';
+import { AppleTask } from '../task';
+import { AgeRatingTask } from './age-rating';
+import { AppInfoTask } from './app-info';
+import { AppVersionTask } from './app-version';
+
+/**
+ * List of all eligible tasks to sync local store configuration to the App store.
+ */
+export function createAppleTasks(_ctx: MetadataContext): AppleTask[] {
```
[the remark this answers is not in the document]
byCedric: It's for the screenshots that I'm working on. For the screenshots, we need to know a couple of things:
1. What is the project root? - We are going to interact with the file system to validate and prep the images for upload to ASC API
2. What would be the preferred path of storing these screenshots - Defined in the `store.config.json`, but overwritable, mostly to set a path to where we would be downloading the files to.

packages/eas-cli/src/metadata/apple/tasks/app-version.ts
```diff
@@ -0,0 +1,150 @@
...
+        if (!attributes) {
+          continue;
+        }
+
+        const oldModel = context.versionLocales.find(model => model.attributes.locale === locale);
+        await logAsync(
+          () =>
+            oldModel
+              ? oldModel.updateAsync(attributes)
+              : context.version.createLocalizationAsync({ ...attributes, locale }),
```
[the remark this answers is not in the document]
wkozyra95: logAsync expects an async method, but you are passing a sync one that just returns a promise, from logic like that it's easy to just add braces and lost that promise altogether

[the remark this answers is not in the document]
byCedric: I'm not sure if this makes it better, in those cases I have to add `return` to the `logAsync` action method. But that return value isn't used at all. IMHO it just creates another wrapped promise which might be worse for debugging than returning the original promise.

Screenshot 2022-06-07 at 19 11 11

Either way, I added `return` and `await` in https://github.com/expo/eas-cli/pull/1147

packages/eas-cli/src/metadata/apple/tasks/age-rating.ts
byCedric: Alright, so I think these asserts are right:

- If `Task2` is being executed before `Task1`, and `Task2` requires the data from `Task1` -> that would be a programmer error
- If `Task1` does not set the data required for `Task2` -> that would be a wrong permissive `null` value that would always fail `Task2` (either programmer error, or wrongfully handled exception from ASC)

I'll let this be like this for now

packages/eas-cli/src/metadata/apple/tasks/app-version.ts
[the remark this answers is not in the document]
wkozyra95: If didn't realize that those function did not return anything, in that case it should be if-else instead of ternarry

PR #1134, opened 2022-05-31
[eas-cli] Add metadata context based on submission contexts

byCedric:
# Checklist

# Why

Part of PR #1083

This adds a new context around the metadata, based on the submission contexts. It is used to initialize and authenticate the user to interact with the metadata stuff.

# How

- Metadata is configured through `eas.json`'s submission profiles, that's why we read those profiles in the context.
- If it's not configured, we assume the store configuration file.
- When the store configuration file is set in the profile, we use that to read/write the data.

# TODO

- Add the submission property `meta` with default value whenever users configure it.

# Test Plan

I need to add some tests for this, can add them later.

### DIFF
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -1,7 +1,7 @@
 import { App, Session } from '@expo/apple-utils';
 import { ExpoConfig } from '@expo/config';
 import { Platform } from '@expo/eas-build-job';
-import { IosSubmitProfile } from '@expo/eas-json/build/submit/types';
+import { SubmitProfile } from '@expo/eas-json';
 import assert from 'assert';
 
 import { CredentialsContext } from '../credentials/context';
@@ -16,7 +16,7 @@ export type MetadataContext = {
   /** Submission profile platform to use */
   platform: Platform.IOS;
   /** Resolved submission profile configuration */
-  profile: IosSubmitProfile;
+  profile: SubmitProfile<Platform.IOS>;
   /** Configured store configuration file name (defaults to store.config.json) */
   metadataFile: string;
   /** Authenticated Expo account */
@@ -65,7 +65,7 @@ export async function createMetadataContextAsync(params: {
     submissionProfile,
     'Could not resolve the iOS submission profile, only iOS is supported for metadata'
   );
-  const iosSubmissionProfile = submissionProfile.profile as IosSubmitProfile;
+  const iosSubmissionProfile = submissionProfile.profile as SubmitProfile<Platform.IOS>;
 
   const exp = params.exp ?? getExpoConfig(params.projectDir);
   const user = await ensureLoggedInAsync({ nonInteractive: params.nonInteractive });
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -86,7 +86,7 @@ export async function createMetadataContextAsync(params: {
     credentialsCtx,
     bundleIdentifier,
     projectDir: params.projectDir,
-    exp: params.exp,
+    exp,
     nonInteractive: params.nonInteractive ?? false,
   };
 }
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -23,7 +23,7 @@ export type MetadataContext = {
   user: Actor;
   /** The store credentials manager */
   credentialsCtx: CredentialsContext;
-  /** The app bundle identifier to use for the stores */
+  /** The app bundle identifier to use for the store */
   bundleIdentifier: string;
   /** Root of the Expo project directory */
   projectDir: string;
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -29,11 +29,6 @@ export type MetadataContext = {
   projectDir: string;
   /** Resolved Expo app configuration */
   exp: ExpoConfig;
-  /**
-   * If we should run in non-interactive mode, as much as we can.
-   * The store authentication requires a fully configured ASC environment, or interactive mode.
-   */
-  nonInteractive: boolean;
 };
 
 export type MetadataAppStoreAuthentication = {
@@ -52,7 +47,6 @@ export async function createMetadataContextAsync(params: {
   credentialsCtx: CredentialsContext;
   exp?: ExpoConfig;
   profileName?: string;
-  nonInteractive?: boolean;
 }): Promise<MetadataContext> {
   const submissionProfiles = await getProfilesAsync({
     type: 'submit',
@@ -69,7 +63,7 @@ export async function createMetadataContextAsync(params: {
   const iosSubmissionProfile = submissionProfile.profile as SubmitProfile<Platform.IOS>;
 
   const exp = params.exp ?? getExpoConfig(params.projectDir);
-  const user = await ensureLoggedInAsync({ nonInteractive: params.nonInteractive });
+  const user = await ensureLoggedInAsync();
   const bundleIdentifier = await getBundleIdentifierAsync(params.projectDir, exp);
 
   return {
@@ -81,7 +75,6 @@ export async function createMetadataContextAsync(params: {
     bundleIdentifier,
     projectDir: params.projectDir,
     exp,
-    nonInteractive: params.nonInteractive ?? false,
   };
 }
 
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -18,7 +18,7 @@ export type MetadataContext = {
   /** Resolved submission profile configuration */
   profile: SubmitProfile<Platform.IOS>;
   /** Configured store configuration file name (defaults to store.config.json) */
-  metadataFile: string;
+  metadataFilename: string;
   /** Authenticated Expo account */
   user: Actor;
   /** The store credentials manager */
@@ -81,7 +81,7 @@ export async function createMetadataContextAsync(params: {
   return {
     platform: Platform.IOS,
     profile: iosSubmissionProfile,
-    metadataFile: iosSubmissionProfile.meta ?? 'store.config.json',
+    metadataFilename: iosSubmissionProfile.meta ?? 'store.config.json',
     user,
     credentialsCtx,
     bundleIdentifier,
--- a/packages/eas-cli/src/metadata/context.ts
+++ b/packages/eas-cli/src/metadata/context.ts
@@ -1,0 +1,110 @@
+import { App, Session } from '@expo/apple-utils';
+import { ExpoConfig } from '@expo/config';
+import { Platform } from '@expo/eas-build-job';
+import { IosSubmitProfile } from '@expo/eas-json/build/submit/types';
+import assert from 'assert';
+
+import { CredentialsContext } from '../credentials/context';
+import { getRequestContext } from '../credentials/ios/appstore/authenticate';
+import { getExpoConfig } from '../project/expoConfig';
+import { getBundleIdentifierAsync } from '../project/ios/bundleIdentifier';
+import { Actor } from '../user/User';
+import { ensureLoggedInAsync } from '../user/actions';
+import { getProfilesAsync } from '../utils/profiles';
+
+export type MetadataContext = {
+  /** Submission profile platform to use */
+  platform: Platform.IOS;
+  /** Submission profile name to use */
+  profile: IosSubmitProfile;
+  /** Configured store configuration file name (defaults to store.config.json) */
+  metadataFile: string;
+  /** Authenticated Expo account */
+  user: Actor;
+  /** The store credentials manager */
+  credentialsCtx: CredentialsContext;
+  /** The app bundle identifier to use for the stores */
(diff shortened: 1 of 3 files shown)


EvanBacon (approved):
Looks fine, mostly composing existing modules so not much to add.

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+  });
+
+  const submissionProfile = submissionProfiles.find(profile => profile.platform === Platform.IOS);
+  assert(
+    submissionProfile,
+    'Could not resolve the iOS submission profile, only iOS is supported for metadata'
+  );
+  const iosSubmissionProfile = submissionProfile.profile as IosSubmitProfile;
+
+  const exp = params.exp ?? getExpoConfig(params.projectDir);
```
wkozyra95, line 70: exp is not optional in the params, so either this line or type definition is incorrect

packages/eas-json/src/submit/schema.ts
```diff
@@ -26,6 +26,7 @@ export const IosSubmitProfileSchema = Joi.object({
   companyName: Joi.string(),
   appName: Joi.string(),
   bundleIdentifier: Joi.string(),
+  meta: Joi.string(),
```
wkozyra95, line 29: what is meta?
`metadata` or maybe `storeMetadata` if we want to keep consistent naming for android too, or `appstoreMetadata` if we don't
if it's a path maybe also add `File` or `Path` suffix to the name

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * To start syncing ASC entities, we need access to the apple utils App instance.
+ * This ensures we are authenticated and have the app instance ready to use.
+ */
+export async function ensureMetadataAppStoreAuthenticatedAsync({
+  credentialsCtx,
+  bundleIdentifier,
+}: MetadataContext): Promise<MetadataAppStoreAuthentication> {
+  const authCtx = await credentialsCtx.appStore.ensureAuthenticatedAsync();
```
wkozyra95, line 102: how are you handling nonInteractive option here? ensureAuthenticatedAsync is interactive

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+    submissionProfile,
+    'Could not resolve the iOS submission profile, only iOS is supported for metadata'
+  );
+  const iosSubmissionProfile = submissionProfile.profile as IosSubmitProfile;
+
+  const exp = params.exp ?? getExpoConfig(params.projectDir);
+  const user = await ensureLoggedInAsync({ nonInteractive: params.nonInteractive });
+  const bundleIdentifier = await getBundleIdentifierAsync(params.projectDir, exp);
+
+  const credentialsCtx = new CredentialsContext({
```
wkozyra95, line 74: CredentialsContext should not be created in util functions, it should be done in some top-level code e.g. in command, one constraint we want to ensure is that this is not called more than once per command, if this is true then it's fine to leave it here

The other point is it looks like you only need access to apple API, if that is the case you do not need credentials context.

wkozyra95 (changes requested):

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+import { getBundleIdentifierAsync } from '../project/ios/bundleIdentifier';
+import { Actor } from '../user/User';
+import { ensureLoggedInAsync } from '../user/actions';
+import { getProfilesAsync } from '../utils/profiles';
+
+export type MetadataContext = {
+  /** Submission profile platform to use */
+  platform: Platform.IOS;
+  /** Submission profile name to use */
+  profile: IosSubmitProfile;
```
dsokal, line 19: The comment is misleading. `IosSubmitProfile` is not the profile name.

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** Configured store configuration file name (defaults to store.config.json) */
+  metadataFile: string;
```
dsokal, line 21:
```suggestion
  metadataFilename: string;
```

dsokal: "File" sounds like a file descriptor to me.

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** Authenticated Expo account */
```
dsokal, line 22:
```suggestion
  /** Authenticated Expo user */
```

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+  user: Actor;
+  /** The store credentials manager */
+  credentialsCtx: CredentialsContext;
+  /** The app bundle identifier to use for the stores */
```
dsokal, line 26: Why store**s**?

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+   * The store authentication requires a fully configured ASC environment, or interactive mode.
+   */
+  nonInteractive: boolean;
+};
+
+export type MetadataAppStoreAuthentication = {
+  /** The root entity of the App store  */
+  app: App;
+  /** The authentication state we used to fetch the root entity */
+  auth: Partial<Session.AuthState>;
```
dsokal, line 43: Why is it `Partial<..>`?

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  return {
+    platform: Platform.IOS,
+    profile: iosSubmissionProfile,
+    metadataFile: iosSubmissionProfile.meta ?? 'store.config.json',
+    user,
+    credentialsCtx,
+    bundleIdentifier,
+    projectDir: params.projectDir,
+    exp: params.exp,
```
dsokal, line 89: Related to https://github.com/expo/eas-cli/pull/1134/files#r886412962
Shouldn't it be:
```suggestion
    exp,
```

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
+import { App, Session } from '@expo/apple-utils';
+import { ExpoConfig } from '@expo/config';
+import { Platform } from '@expo/eas-build-job';
+import { IosSubmitProfile } from '@expo/eas-json/build/submit/types';
+import assert from 'assert';
```
dsokal, line 5: I'm not sure why it's not enforced by eslint (it should be) but the convention in Expo projects is usally:

```
// node built-in modules imports

// 3rd party modules imports

// local imports
```

There should be an empty line between each group.

packages/eas-json/src/submit/schema.ts
dsokal: Agreed with Wojtek, `meta` is really confusing. I could think it's somehow related to Meta (formerly Facebook).

dsokal (changes requested):

packages/eas-cli/src/metadata/context.ts
```diff
@@ -0,0 +1,110 @@
+import { App, Session } from '@expo/apple-utils';
+import { ExpoConfig } from '@expo/config';
+import { Platform } from '@expo/eas-build-job';
+import { IosSubmitProfile } from '@expo/eas-json/build/submit/types';
```
wkozyra95, line 4: import stuff only via entry point

packages/eas-cli/src/metadata/context.ts
byCedric: Good catch, it should be optional. I noticed we double-load this in the submission context (on the command, to load project id etc, and in the actual context creation). So this should just be "if it's loaded already, reuse it or load it".

packages/eas-json/src/submit/schema.ts
byCedric: Alright, I think both `metadata` or `storeMetadata` (or even `storeConfig`) should be best here. The store config file should be able to support other stores in the future too. So I would avoid `appStoreMetadata`.

packages/eas-cli/src/metadata/context.ts
byCedric: It can't. None of the `AppStoreApi` methods from the `CredentialsContext` can be set to non-interactive, that's why I added a warning here: https://github.com/expo/eas-cli/blob/%40bycedric/metadata/context/packages/eas-cli/src/metadata/context.ts#L32-L35

Preferably, we would just pass through the `nonInteractive` mode set on the `CredentialsContext` here: https://github.com/expo/eas-cli/blob/%40bycedric/metadata/context/packages/eas-cli/src/credentials/context.ts#L19

But, that would change existing functionality. We can still do that, but I guess in a different PR. For now, it would fail in the `prompts` method when authenticating in CI without the proper envvars. If people configured the right envvars, it should not go into an interactive mode (should pick up the right envvars and just authenticate with that directly).

packages/eas-cli/src/metadata/context.ts
byCedric: > CredentialsContext should not be created in util functions, it should be done in some top-level code e.g. in command, one constraint we want to ensure is that this is not called more than once per command, if this is true then it's fine to leave it here

Alright, thanks for the context! Yeah, we only use the `AppStoreApi` instance to authenticate with the ASC. (see below why I used `CredentialContext` instead of the `AppStoreApi` directly)

> The other point is it looks like you only need access to apple API, if that is the case you do not need credentials context.

I can also initialize the `AppStoreApi` directly, instead of the credentials context. I did this to not go too deep into the inner workings of the credentials. But from your comment I assume you'd prefer that instead :) I'll update it!

packages/eas-cli/src/metadata/context.ts
byCedric: It's because the `AppStoreApi`'s [`AuthCtx` is `Partial<Session.AuthState> | Session.AuthState`](https://github.com/expo/eas-cli/blob/632577865c6e966dd7b92d925be6e301d6db7d3c/packages/eas-cli/src/credentials/ios/appstore/authenticateTypes.ts#L36-L45), depending on the type of authentication, e.g. credentials or ASC API keys.

packages/eas-cli/src/metadata/context.ts
byCedric: I think people can also use robot accounts, which are not exactly users. The user prop is identical to the one from [`SubmissionContext`](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/submit/context.ts#L27), but these are actually actors (so either User or Robot, which only have "accounts" in common)

packages/eas-cli/src/metadata/context.ts
byCedric: For the App Store, this is the "link" to the `App` model instance. This property determines the actual app instance in ASC. We [resolve that in this method](https://github.com/expo/eas-cli/blob/%40bycedric/metadata/context/packages/eas-cli/src/metadata/context.ts#L106).

packages/eas-cli/src/metadata/context.ts
byCedric: I thought the convention was:

```
// Node or 3rd party modules

// Local imports
```

Just like the **[project/expoSdk.ts](https://github.com/expo/eas-cli/blob/07d6b5269bc1bf4b18cd2cd5d5a3342603c8218d/packages/eas-cli/src/project/expoSdk.ts#L1-L8)**.

packages/eas-cli/src/metadata/context.ts
wkozyra95: If there is no way to do that in nonInteractive mode then there should be no nonInteractive option

byCedric: @wkozyra95 would you be ok with commenting the `nonInteractive` option out, but still add a `todo` to implement that later in the `AppStoreApi`?

packages/eas-cli/src/metadata/context.ts
byCedric: Done, would need to update the command PR, but I'll do that after merging this and bringing the stacked PRs up-to-date.

packages/eas-cli/src/metadata/context.ts
dsokal: Ok, I thought we follow this convention in EAS CLI too. 

In turtle-v2 repo we certainly have this rule enforced. See e.g. https://github.com/expo/turtle-v2/blob/main/src/services/submission-service/src/tasks/iosSubmit/ascApiKey.ts#L1 

When I change the order I get an error from linter
Screenshot 2022-06-02 at 15 58 21

packages/eas-cli/src/metadata/context.ts
dsokal: I only meant why it's plural :P

dsokal (approved):

packages/eas-cli/src/metadata/context.ts
byCedric: Ah, I guess that should be singular here :) If we support multiple stores, every store should get it's own metadata context instance, just like submission + platform.

PR #1136, opened 2022-05-31
[eas-cli] Add metadata commands

byCedric:
# Checklist

# Why

**Part of PR #1083**
- [x] ~~Requires PR #1130~~
- [x] ~~Requires PR #1131~~
- [x] ~~Requires PR #1132~~
- [x] ~~Requires PR #1133~~
- [x] ~~Requires PR #1134~~
- [x] ~~Requires PR #1135~~

**Additional PRs merged**
- [x] PR #1159
- [x] PR #1156

This adds the EAS CLI commands to let users interact with the metadata features.

# How

Originally we wanted to include it in the submission commands, but that workflow does not work well.
- We have to wait until the binary is shipped and processed, else we might be editing a wrong version
- There is no good way to wait until the binary is processed
- It's not that great to ask the user to wait until that's all done

As a temporary solution, the metadata is split into its own commands.

# Test Plan

After all PRs are merged and ready:
- `$ eas metadata`
- `$ eas metadata:pull`
- _Edit the store configuration_
- `$ eas metadata:push`

### DIFF
--- a/packages/eas-cli/src/commands/metadata/configure.ts
+++ b/packages/eas-cli/src/commands/metadata/configure.ts
@@ -2,12 +2,14 @@ import { getConfig } from '@expo/config';
 import { Flags } from '@oclif/core';
 
 import EasCommand from '../../commandUtils/EasCommand';
+import { CredentialsContext } from '../../credentials/context';
 import Log, { learnMore } from '../../log';
 import { createMetadataContextAsync } from '../../metadata/context';
 import { downloadMetadataAsync } from '../../metadata/download';
 import { handleMetadataError } from '../../metadata/errors';
 import { RequestedPlatform } from '../../platform';
 import { findProjectRootAsync, getProjectIdAsync } from '../../project/projectUtils';
+import { ensureLoggedInAsync } from '../../user/actions';
 
 type RawCommandFlags = {
   platform?: string;
@@ -44,7 +46,15 @@ export default class MetadataConfigure extends EasCommand {
     const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
     await getProjectIdAsync(exp);
 
-    const metadataContext = await createMetadataContextAsync({
+    const credentialsCtx = new CredentialsContext({
+      exp,
+      projectDir,
+      user: await ensureLoggedInAsync(),
+      nonInteractive: false,
+    });
+
+    const metadataCtx = await createMetadataContextAsync({
+      credentialsCtx,
       projectDir,
       exp,
       profileName: flags.profile,
@@ -52,7 +62,7 @@ export default class MetadataConfigure extends EasCommand {
 
     try {
       Log.addNewLineIfNone();
-      const filePath = await downloadMetadataAsync(metadataContext);
+      const filePath = await downloadMetadataAsync(metadataCtx);
       Log.addNewLineIfNone();
 
       Log.succeed('Your store configuration has been generated!');
--- a/packages/eas-cli/src/commands/metadata/configure.ts
+++ b/packages/eas-cli/src/commands/metadata/configure.ts
@@ -36,8 +36,6 @@ export default class MetadataConfigure extends EasCommand {
     }),
   };
 
-  static args = [];
-
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(MetadataConfigure);
     const flags = await this.sanitizeFlagsAsync(rawFlags);
--- a/packages/eas-cli/src/commands/metadata/configure.ts
+++ b/packages/eas-cli/src/commands/metadata/configure.ts
@@ -23,7 +23,7 @@ type CommandFlags = {
 
 export default class MetadataConfigure extends EasCommand {
   static hidden = true;
-  static description = 'upload metadata configuration to the app stores';
+  static description = 'configure the store configuration file in your project';
 
   static flags = {
     platform: Flags.enum({
--- a/packages/eas-cli/src/commands/metadata/configure.ts
+++ b/packages/eas-cli/src/commands/metadata/configure.ts
@@ -1,0 +1,86 @@
+import { getConfig } from '@expo/config';
+import { Errors, Flags } from '@oclif/core';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import Log, { learnMore } from '../../log';
+import { createMetadataContextAsync } from '../../metadata/context';
+import { downloadMetadataAsync } from '../../metadata/download';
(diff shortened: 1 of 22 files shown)


packages/eas-cli/src/commands/metadata/configure.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    try {
+      Log.addNewLineIfNone();
+      const filePath = await downloadMetadataAsync(metadataContext);
+      Log.addNewLineIfNone();
+
+      Log.succeed('Your store configuration has been generated!');
+      Log.log(filePath);
+      Log.log(learnMore('https://docs.expo.dev/eas-metadata/introduction/'));
```
byCedric, line 66: This is not available yet, before merging I have to check if we have this ready or if we should omit this link until the docs are ready.

byCedric:
This also applies to this PR https://github.com/expo/eas-cli/pull/1135#pullrequestreview-995166194 (the commands)

packages/eas-cli/src/commands/metadata/configure.ts
```diff
@@ -0,0 +1,85 @@
+import { getConfig } from '@expo/config';
```
wkozyra95, line 1: metadata name is not descriptive, without looking at docs users will have no idea what it means, also metadata world on its own is not clear what it will do( build,submit,update are both verbs and nouns metadata on it's own is not clear)

few suggestions
- `submit:metadata:update` and `submit:metadata:configure`
- `metadata:appstore:update`/`metadata:playstore:update`/`metadata:appstore:configure`/`metadata:playstore:configure` (this might be better alternative to platform flag, because we might want to support multiple stores per platform

wkozyra95 (commented):
What are the prerequisites for those commands to work?
- do both of them work if there is no bundle identifier with that name
- do both of them work if there is no app store connect entry?
- please test f it works with a bare project(and if bundleIdentifier is resolved from native code not app.json)

Note that some of the above might not work if used with app store API key authentication

packages/eas-cli/src/commands/metadata/configure.ts
byCedric: 1. That seems ok to me, would you also be ok to add short aliases with the current ones? e.g. `submit:metadata:update`, which aliases to `submit:metadata` or even `metadata`?
2. Right now, iOS doesn't allow any other app store and Android doesn't have a good official 2nd store without going into shady websites, so I don't think we want to support multiple stores per platform (just the official ones). But let me check with Evan first!

packages/eas-cli/src/commands/metadata/configure.ts
```diff
@@ -0,0 +1,85 @@
...
+  requestedPlatforms: RequestedPlatform;
+  profile?: string;
+};
+
+export default class MetadataConfigure extends EasCommand {
+  static hidden = true;
+  static description = 'configure the store configuration file in your project';
+
+  static flags = {
+    platform: Flags.enum({
```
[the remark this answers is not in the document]
byCedric: We definitely intend to extend this to the play store, but since this only supports ios for now, I think we can leave it out but that might break user's pipelines when we do that.

wkozyra95:
> If we can't find the app entity with this bundle identifier, we will [exit early here](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/metadata/context.ts#L94). We could add some additional help text to guide users to check their config, or to create a new ASC app entry with that bundle identifier though.

we create app store connect entry in submit, so I would expect that those commands should also do that. A lot of people will want to setup metadata before submit

> I've reused the [getBundleIdentifierAsync](https://github.com/expo/eas-cli/blob/940e3f6d40a7e65d449d3d9b8b2ab1bc3fbf25fb/packages/eas-cli/src/project/ios/bundleIdentifier.ts#L37) from the submit parts ([here](https://github.com/expo/eas-cli/blob/91456d1c65a08a7340edf87caaadbc72363e90d8/packages/eas-cli/src/submit/ios/AppProduce.ts), and [there](https://github.com/expo/eas-cli/blob/e2277b948272eddfc8c0d033ec8b1ba767e1252e/packages/eas-cli/src/submit/ios/IosSubmitCommand.ts)). That should take both managed and bare into account, and, if necessary, load it from the native files. This is [done in the metadata context](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/metadata/context.ts#L67), and I think you and Dominik approved that one right?

I double-checked that and it's not correct for multitarget projects because you are not passing the third argument there. We have 2 approaches:
- for build we always know the build profile, so we can identify a specific target.
- for submit we didn't want to define both build and submit profiles, so if bundleIdentifier is ambigous it can be defined in submit profile

given that you already have submit profile the second approach is better, it should be sth like this
```
const bundleIdentifier = iosSubmissionProfiile.bundleIdentifier ?? await getBundleIdentifierAsync(params.projectDir, exp);
```

packages/eas-cli/src/commands/metadata/configure.ts
wkozyra95: Ad:1 I'm not suggesting a change because I think my option is better, but because the shorter version is actively confusing, so adding aliases would defeat that purpose.
Ad: 2 I was thinking about android stores e.g. amazon one

packages/eas-cli/src/commands/metadata/configure.ts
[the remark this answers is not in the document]
wkozyra95: On the other hand, people will complain that they need to pass `-p ios` even though it's always ios

My preference would be to use sth like `metadata:appstore:update` which does not require options or breaking changes if we add new stores, but if you do not want to embed platform in name of the command I think breaking change in the future is better.

byCedric:
> given that you already have submit profile the second approach is better, it should be sth like this

Done in https://github.com/expo/eas-cli/pull/1147

byCedric:
> we create app store connect entry in submit, so I would expect that those commands should also do that. A lot of people will want to setup metadata before submit

I think so too, but one of the issues that I ran into with adding this part in the submit workflow is that we edit everything per version. Once users upload a new binary, a new version is created. We/I have to think and test a bit more for these scenarios, where we intentionally edit the right version of the app (instead of the old one).

wkozyra95 (approved):

dsokal (approved):

packages/eas-cli/src/commands/metadata/configure.ts
byCedric: As discussed, keeping this open for the next store to implement.

PR #1144, opened 2022-06-06
[eas-cli] make viewBranchAsync updates limit configurable

kgc00:
# Checklist

# Why

Follow up to #1119. This will help avoid timeouts and network errors related to publishing updates by avoiding querying for large payloads of updates when it is not necessary.

# How

For the execution flow to publish a new update we just need to check the existence of the branch, and if it is not extant, create it. We can limit the updates returned in the GQL query in this case to 0 because we are not using them.

For the execution flow to republish an old update, we need to both check that a branch exists and get updates associated with it. We can limit the updates returned in the GQL query in this case to however many we need. We lowered the number of updates returned from 10k to 300 previously. We will need to add in some new functionality to paginate these results and re-query with an offset if the user did not find the update they'd like to re-publish on the first page. This will be done in a follow-up PR.

# Test Plan

Added automated tests to ensure the configurable update limit is working. Additionally, one can:
- run `yarn start`
- alias their local `yarn start` output to `edl` and run:
  - `edl update --republish`
  - `edl update`

to ensure that both flows work properly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ packages/eas-cli/tmp
 # Code editors
 .idea
 .history
+.vscode
 
 # macOS
 .DS_Store
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -120,9 +120,11 @@ async function ensureChannelExistsAsync({
 async function ensureBranchExistsAsync({
   appId,
   name: branchName,
+  limit,
 }: {
   appId: string;
   name: string;
+  limit?: number;
 }): Promise<{
   id: string;
   updates: Exclude<
@@ -133,6 +135,7 @@ async function ensureBranchExistsAsync({
   const { app } = await UpdateQuery.viewBranchAsync({
     appId,
     name: branchName,
+    limit,
   });
   const updateBranch = app?.byId.updateBranchByName;
   if (updateBranch) {
@@ -309,11 +312,6 @@ export default class UpdatePublish extends EasCommand {
       assert(branchName, 'Branch name must be specified.');
     }
 
-    const { id: branchId, updates } = await ensureBranchExistsAsync({
-      appId: projectId,
-      name: branchName,
-    });
-
     let unsortedUpdateInfoGroups: UpdateInfoGroup = {};
     let oldMessage: string, oldRuntimeVersion: string;
     if (republish) {
@@ -335,6 +333,10 @@ export default class UpdatePublish extends EasCommand {
           throw new Error('You must specify the update group to republish.');
         }
 
+        const { updates } = await ensureBranchExistsAsync({
+          appId: projectId,
+          name: branchName,
+        });
         const updateGroups = uniqBy(updates, u => u.group)
           .filter(update => {
             // Only show groups that have updates on the specified platform(s).
@@ -464,6 +466,12 @@ export default class UpdatePublish extends EasCommand {
         .filter(pair => pair[1] === runtime)
         .map(pair => pair[0]);
     }
+
+    const { id: branchId } = await ensureBranchExistsAsync({
+      appId: projectId,
+      name: branchName,
+      limit: 0,
+    });
 
     // Sort the updates into different groups based on their platform specific runtime versions
     const updateGroups: PublishUpdateGroupInput[] = Object.entries(runtimeToPlatformMapping).map(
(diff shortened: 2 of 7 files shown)


packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -54,7 +57,7 @@ export const UpdateQuery = {
         .toPromise()
     );
   },
-  async viewBranchAsync({ appId, name }: Pick<ViewBranchUpdatesQueryVariables, 'appId' | 'name'>) {
+  async viewBranchAsync({ appId, name, limit }: ViewBranchUpdatesQueryVariablesWithOptionalLimit) {
```
kgc00, line 60: alternatively, instead of making limit an optional argument, we could just create a second gql query that does not include updates at all:
```
query ViewBranchUpdates($appId: String!, $name: String!) {
              app {
                byId(appId: $appId) {
                  id
                  updateBranchByName(name: $name) {
                    id
                    name
                    }
                  }
                }
              }
```

I chose to set the limit to 0 to make the smallest change on the API. Seemed to get all the benefits of a new query without the boilerplate of making/maintaining a new query.

.gitignore
```diff
@@ -6,6 +6,7 @@ packages/eas-cli/tmp
 # Code editors
 .idea
 .history
+.vscode
```
kgc00, line 9: added .vscode ignore because it was missing 🤷

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -335,6 +333,10 @@ export default class UpdatePublish extends EasCommand {
           throw new Error('You must specify the update group to republish.');
         }
 
+        const { updates } = await ensureBranchExistsAsync({
+          appId: projectId,
+          name: branchName,
+        });
```
ide, line 339: Passing in the limit here would be a good idea because the number of updates to query from the history is part of this command's UX, rather than relying on the default limit `ensureBranchExistsAsync` happens to choose.

Also, on line 351 we maybe should change the error message to say, "There are no [recent] updates" since we now fetch only a subset of the updates.

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -8,7 +8,10 @@ import {
   ViewBranchUpdatesQueryVariables,
 } from '../generated';
 
-const PAGE_LIMIT = 300;
+export const PAGE_LIMIT = 300;
```
ide, line 11: Give this constant a more global, less context-sensitive name since it's now exported (it's not a global page limit we use everywhere). Also, the name should indicate it's a limit on the number of updates per page (vs, say, a limit on the number of pages).

ide (changes requested):
One additional thing we should do is enforce a reasonable limit on the server with something like `Math.min(limit, MAX_UPDATES_LIMIT)`.

packages/eas-cli/src/commands/update/index.ts
kgc00: > Passing in the limit here would be a good idea because the number of updates to query from the history is part of this command's UX, rather than relying on the default limit ensureBranchExistsAsync happens to choose.

This *should* be addressed in the follow up [PR](https://github.com/expo/eas-cli/pull/1148) which deals solely with the republish flow.

kgc00:
> One additional thing we should do (separately) is enforce a reasonable limit on the server with something like `Math.min(limit, MAX_UPDATES_LIMIT)`.

https://linear.app/expo/issue/ENG-5228/add-a-cap-to-number-of-updates-fetched-from-www
will add a PR for this

packages/eas-cli/src/update/__tests__/update-test.ts
```diff
@@ -0,0 +1,85 @@
...
+              },
+            },
+          ],
+        },
+      }),
+    })),
+  },
+}));
+
+describe.only('UpdatePublish', () => {
```
ide, line 56:
```suggestion
describe('UpdatePublish', () => {
```

ide (approved):

packages/eas-cli/src/update/__tests__/update-test.ts
kgc00: thank you! 😳 
i will add a follow up task to add some linter error if .only is present

PR #1148, opened 2022-06-08
[eas-cli] add fetching additional pages for update --republish

kgc00:
# Checklist

# Why

Follow up to #1119. This allows CLI users to `update --republish` any previously published update, even if it is beyond the initial page limit (n=300).

# How

I added the `getUpdatesToRepublishAsync` method which will recurse when the user asks for a new page updates. The updates are accumulated and presented to the user. Once a branch has been fully loaded, the option to fetch another page (`Next page...`) will disappear.

# Test Plan

Run locally (`edl update --republish`) with a low page limit (n=5) on a branch with few updates. Ensure that:
- the option to fetch new pages of updates is present
- updates fetched in both the current network request and previous network requests can be republished
- once all updates are fetched, the option to fetch another page disappears

Also, added some automated tests for above

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -25,7 +25,7 @@ import {
   ViewBranchUpdatesQuery,
 } from '../../graphql/generated';
 import { PublishMutation } from '../../graphql/mutations/PublishMutation';
-import { UpdateQuery } from '../../graphql/queries/UpdateQuery';
+import { UpdateQuery, viewBranchUpdatesQueryUpdateLimit } from '../../graphql/queries/UpdateQuery';
 import Log, { link } from '../../log';
 import { ora } from '../../ora';
 import { getExpoConfig } from '../../project/expoConfig';
@@ -124,10 +124,12 @@ export async function ensureBranchExistsAsync({
   appId,
   name: branchName,
   limit,
+  offset,
 }: {
   appId: string;
   name: string;
   limit?: number;
+  offset?: number;
 }): Promise<{
   id: string;
   updates: Exclude<
@@ -139,6 +141,7 @@ export async function ensureBranchExistsAsync({
     appId,
     name: branchName,
     limit,
+    offset,
   });
   const updateBranch = app?.byId.updateBranchByName;
   if (updateBranch) {
@@ -338,30 +341,11 @@ export default class UpdatePublish extends EasCommand {
           throw new Error('You must specify the update group to republish.');
         }
 
-        const { updates } = await ensureBranchExistsAsync({
-          appId: projectId,
-          name: branchName,
-        });
-        const updateGroups = uniqBy(updates, u => u.group)
-          .filter(update => {
-            // Only show groups that have updates on the specified platform(s).
-            return platformFlag === 'all' || update.platform === platformFlag;
-          })
-          .map(update => ({
-            title: formatUpdateTitle(update),
-            value: update.group,
-          }));
-        if (updateGroups.length === 0) {
-          throw new Error(
-            `There are no updates on branch "${branchName}" published on the platform(s) ${platformFlag}. Did you mean to publish a new update instead?`
-          );
-        }
-
-        const selectedUpdateGroup = await selectAsync<string>(
-          'which update would you like to republish?',
-          updateGroups
+        updatesToRepublish = await getUpdatesToRepublishInteractiveAsync(
+          projectId,
+          branchName,
+          platformFlag
         );
-        updatesToRepublish = updates.filter(update => update.group === selectedUpdateGroup);
       }
       const updatesToRepublishFilteredByPlatform = updatesToRepublish.filter(
         // Only republish to the specified platforms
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/commands/update/index.ts
```diff
@@ -584,6 +571,59 @@ export default class UpdatePublish extends EasCommand {
...
+    appId: projectId,
+    name: branchName,
+    offset,
+  });
+  cumulativeUpdates = [...cumulativeUpdates, ...updates];
+  const updateGroups = uniqBy(cumulativeUpdates, u => u.group)
+    .filter(update => {
+      // Only show groups that have updates on the specified platform(s).
+      return platformFlag === 'all' || update.platform === platformFlag;
+    })
```
ide, line 596: I'm not quite sure how the old logic worked -- it seems like we need to filter by platform first, and then remove duplicates within the group. Also if platformFlag is all, we can skip the filter altogether.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -584,6 +571,59 @@ export default class UpdatePublish extends EasCommand {
...
+    .map(update => ({
+      title: formatUpdateTitle(update),
+      value: update.group,
+    }));
+  if (!updateGroups.length) {
+    throw new Error(
+      `There are no updates on branch "${branchName}" published on the platform(s) ${platformFlag}. Did you mean to publish a new update instead?`
```
ide, line 603:
```suggestion
      `There are no updates on branch "${branchName}" published for the platform(s): ${platformFlag}. Did you mean to publish a new update instead?`
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -584,6 +571,59 @@ export default class UpdatePublish extends EasCommand {
...
+    );
+  }
+
+  if (updates.length && updates.length === getViewBranchUpdatesQueryUpdateLimit()) {
```
ide, line 607: updates.length will always be defined
```suggestion
  if (updates.length === getViewBranchUpdatesQueryUpdateLimit()) {
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -584,6 +571,59 @@ export default class UpdatePublish extends EasCommand {
...
+  >['updates'] = [],
+  offset: number = 0
+): Promise<any> {
+  const fetchMoreValue = '_fetchMore';
+
+  const { updates } = await ensureBranchExistsAsync({
+    appId: projectId,
+    name: branchName,
+    offset,
+  });
```
ide, line 590: This is where I think we should pass in a page size. The code in charge of pagination and UX should determine the page size instead of ensureBranchExistsAsync.

This will also make it easier to fetch pageSize + 1 updates, where the extra update is a way to see if there are more to fetch. Right now there's a minor bug where if there are exactly a multiple of pageSize updates, we'll show "Next page..." even though there are no more pages.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -584,6 +571,59 @@ export default class UpdatePublish extends EasCommand {
...
+      `There are no updates on branch "${branchName}" published on the platform(s) ${platformFlag}. Did you mean to publish a new update instead?`
+    );
+  }
+
+  if (updates.length && updates.length === getViewBranchUpdatesQueryUpdateLimit()) {
+    updateGroups.push({ title: 'Next page...', value: fetchMoreValue });
+  }
+
+  const selectedUpdateGroup = await selectAsync<string>(
+    'which update would you like to republish?',
```
ide, line 612:
```suggestion
    'Which update would you like to republish?',
```

ide (changes requested):

packages/eas-cli/src/commands/update/index.ts
kgc00: i think this makes sense. i was trying to avoid the case where both items were 0... but that should never happen regardless.

ide: Gotcha... the pageSize+1 suggestion I left in earlier comments should avoid this case too.

packages/eas-cli/src/commands/update/index.ts
kgc00: Great point about fetching an extra update to avoid that final page bug!

I'm not sure I fully understand the suggestion here (re page size) so I'll state what I think you're saying and let me know if it is inaccurate:
- page size (currently denoted as limit) should be explicitly defined here, rather using the global default of `getViewBranchUpdatesQueryUpdateLimit()`

Don't feel strongly either way, happy to make this change. I think we can avoid the bug you mentioned by just displaying `getViewBranchUpdatesQueryUpdateLimit() - 1` updates, rather than fetching a different number of updates altogether.

ide: I am saying the page size should always be passed in here rather than relying on the default. A page size of 50 or 100 might be more usable.

Separately, the number of updates we want to display per page should determine the number of updates to fetch rather than the other way around. It doesn't affect the server much to fetch 300 vs 301 updates. But displaying 299 updates per page is awkward.

kgc00: got it! i'll make the change shortly. thanks for clarifying.

ide (approved):

ide:
The Node 18 tests look like the became flaky recently. May need to look into the memory usage of each test file to see if the newer tests are using a lot of memory and aren't freeing it.

PR #1152, opened 2022-06-09
[eas-cli] improve update terminal output

kgc00:
# Checklist

# Why

We'd like to ensure our users have a better experience using `eas-cli`. Contains various UX improvements listed in  https://linear.app/expo/issue/ENG-4589/improve-publish-command

# How

- Updated `uploadAssetsAsync` to now return the # of unique assets uploaded and use that value to inform the user of the CLI run status
- Added a bunch of info to the log output
- Added a 'fallback' behavior for unsupported terminals for the `ensureProjectExistsAsync` console output

# Test Plan

I've updated the automated tests for `uploadAssetsAsync` to include the new return value.

You should be able to run the CLI as normal but now:
- in an unsupported terminal by [terminal link](https://www.npmjs.com/package/terminal-link), URLs should not contain `)`
- Console output should capitalize all the outputs (branch => Branch)
- Console output should contain Android and iOS specific update IDs
- Console output should contain a link to the website
- Console output should show how many assets we uploaded
- ~Console output should show a warning when assets uploaded near the server limit~

### No assets to upload
image

### Some new assets
image

### Videos

https://user-images.githubusercontent.com/15132641/173119112-aac4c950-1f53-44b7-9129-4abbb553933d.mov

https://user-images.githubusercontent.com/15132641/173119131-7c56b6c1-534e-4d99-8570-bbaf964bf1f4.mov

### DIFF
--- a/packages/eas-cli/src/build/utils/url.ts
+++ b/packages/eas-cli/src/build/utils/url.ts
@@ -35,3 +35,14 @@ export function getInternalDistributionInstallUrl(build: BuildFragment): string 
 
   return build.artifacts.buildUrl;
 }
+
+export function getUpdateGroupUrl(
+  accountName: string,
+  projectName: string,
+  updateGroupId: string
+): string {
+  return new URL(
+    `/accounts/${accountName}/projects/${projectName}/updates/${updateGroupId}`,
+    getExpoWebsiteBaseUrl()
+  ).toString();
+}
--- a/packages/eas-cli/src/build/utils/url.ts
+++ b/packages/eas-cli/src/build/utils/url.ts
@@ -42,7 +42,9 @@ export function getUpdateGroupUrl(
   updateGroupId: string
 ): string {
   return new URL(
-    `/accounts/${accountName}/projects/${projectName}/updates/${updateGroupId}`,
+    `/accounts/${encodeURIComponent(accountName)}/projects/${encodeURIComponent(
+      projectName
+    )}/updates/${encodeURIComponent(updateGroupId)}`,
     getExpoWebsiteBaseUrl()
   ).toString();
 }
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -249,9 +249,16 @@ export async function filterOutAssetsThatAlreadyExistAsync(
   return missingAssets;
 }
 
+type AssetUploadResult = {
+  assetCount: number;
+  uniqueAssetCount: number;
+  uniqueUploadedAssetCount: number;
+};
+
 export async function uploadAssetsAsync(
-  assetsForUpdateInfoGroup: CollectedAssets
-): Promise<number> {
+  assetsForUpdateInfoGroup: CollectedAssets,
+  updateSpinnerText?: (updatedText: string) => void
+): Promise<AssetUploadResult> {
   let assets: RawAsset[] = [];
   let platform: keyof CollectedAssets;
   for (platform in assetsForUpdateInfoGroup) {
@@ -261,6 +268,7 @@ export async function uploadAssetsAsync(
       ...assetsForUpdateInfoGroup[platform]!.assets,
     ];
   }
+  updateSpinnerText?.(`${assets.length} present`);
 
   const assetsWithStorageKey = await Promise.all(
     assets.map(async asset => {
@@ -275,6 +283,7 @@ export async function uploadAssetsAsync(
       storageKey: string;
     }
   >(assetsWithStorageKey, asset => asset.storageKey);
+  updateSpinnerText?.(`${uniqueAssets.length} unique assets found`);
 
   let missingAssets = await filterOutAssetsThatAlreadyExistAsync(uniqueAssets);
   const uniqueUploadedAssetCount = missingAssets.length;
@@ -289,6 +298,7 @@ export async function uploadAssetsAsync(
     })
   );
 
+  updateSpinnerText?.(`${missingAssets.length} missing assets being uploaded`);
   // Wait up to TIMEOUT_LIMIT for assets to be uploaded and processed
   const start = Date.now();
   let timeout = 1;
@@ -302,5 +312,9 @@ export async function uploadAssetsAsync(
       throw new Error('Asset upload timed out. Please try again.');
     }
   }
-  return uniqueUploadedAssetCount;
+  return {
+    assetCount: assets.length,
+    uniqueAssetCount: uniqueAssets.length,
+    uniqueUploadedAssetCount,
+  };
 }
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -16,10 +16,6 @@ import { expoCommandAsync } from '../utils/expoCli';
 import uniqBy from '../utils/expodash/uniqBy';
 
 export const TIMEOUT_LIMIT = 60_000; // 1 minute
-export const MAX_ASSETS_PER_UPLOAD = 600; // https://github.com/expo/universe/blob/main/server/www/src/graphql/mutations/AssetMutation.ts#L14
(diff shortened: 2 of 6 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -16,6 +16,10 @@ import { expoCommandAsync } from '../utils/expoCli';
 import uniqBy from '../utils/expodash/uniqBy';
 
 export const TIMEOUT_LIMIT = 60_000; // 1 minute
+export const MAX_ASSETS_PER_UPLOAD = 600; // https://github.com/expo/universe/blob/main/server/www/src/graphql/mutations/AssetMutation.ts#L14
```
kgc00, line 19: just copy/pasted the # found in `www`. we'll have to remember to keep these two numbers in sync... not sure there's a better way to handle this?

packages/eas-cli/src/project/publish.ts
```diff
@@ -245,7 +249,9 @@ export async function filterOutAssetsThatAlreadyExistAsync(
   return missingAssets;
 }
 
-export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAssets): Promise<void> {
+export async function uploadAssetsAsync(
```
ide, line 252: Document that this returns the number of uploaded assets, including assets that have already been uploaded.

packages/eas-cli/src/project/publish.ts
```diff
@@ -271,6 +277,7 @@ export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAsset
   >(assetsWithStorageKey, asset => asset.storageKey);
 
   let missingAssets = await filterOutAssetsThatAlreadyExistAsync(uniqueAssets);
+  const uniqueUploadedAssetCount = missingAssets.length;
   const { specifications } = await PublishMutation.getUploadURLsAsync(
     missingAssets.map(ma => ma.contentType)
   );
```
ide, line 283: In the future we'll probably want to upload the assets in smaller batches, a few at a time, and show interactive progress.

packages/eas-cli/src/project/publish.ts
```diff
@@ -295,4 +302,5 @@ export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAsset
       throw new Error('Asset upload timed out. Please try again.');
     }
   }
+  return uniqueUploadedAssetCount;
```
ide, line 305: I think it would be useful to return three values: the total number of assets (assets.length), the number of unique assets (uniqueAssets.length), and the number of newly uploaded assets (uniqueUploadedAssetCount). This way we can show a more detailed breakdown.

kbrandwijk:
It seems we know the number of assets _before_ we upload. Could we make the 'Uploading assets' spinner (https://github.com/expo/eas-cli/pull/1152/files#diff-eabf745ea6a7d52d4573e94ea26a1634f849c8232205268bf97a5801ae32792bR457) include that number, so people know why they're waiting a long time (versus only showing when it's done that there were 400 assets).

packages/eas-cli/src/project/publish.ts
kgc00: how would you feel about adding some field to the gql return return value for `PublishMutation.getUploadURLsAsync` or `PublishQuery.getAssetMetadataAsync`? I don't feel like either of those queries really match up with returning `MAX_ASSETS_PER_UPLOAD`, but it feels better than adding an additional endpoint & request just for this data.

packages/eas-cli/src/project/publish.ts
kgc00: added a type with named properties that should resolve this without resorted to docstrings

kgc00:
> It seems we know the number of assets _before_ we upload. Could we make the 'Uploading assets' spinner (https://github.com/expo/eas-cli/pull/1152/files#diff-eabf745ea6a7d52d4573e94ea26a1634f849c8232205268bf97a5801ae32792bR457) include that number, so people know why they're waiting a long time (versus only showing when it's done that there were 400 assets).

added some videos that show this behavior to the PR description.

packages/eas-cli/src/project/publish.ts
kgc00: I do have the interactive progress hooked up. happy to add in the smaller batches functionality as well but that should probably be its own PR

packages/eas-cli/src/project/publish.ts
kgc00: I just updated the command to update the spinner text in real-time as Kim suggested. So it's no longer necessary to return the additional two values from this function, but I left them in the object because it doesn't hurt 🤷

packages/eas-cli/src/project/publish.ts
ide: Agreed

packages/eas-cli/src/build/utils/url.ts
```diff
@@ -35,3 +35,14 @@ export function getInternalDistributionInstallUrl(build: BuildFragment): string
...
   return build.artifacts.buildUrl;
 }
+
+export function getUpdateGroupUrl(
+  accountName: string,
+  projectName: string,
+  updateGroupId: string
+): string {
+  return new URL(
+    `/accounts/${accountName}/projects/${projectName}/updates/${updateGroupId}`,
```
ide, line 45:
```suggestion
    `/accounts/${encodeURIComponent(accountName)}/projects/${encodeURIComponent(projectName)}/updates/${encodeURIComponent(updateGroupId)}`,
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -282,6 +298,7 @@ export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAsset
     })
   );
 
+  updateSpinnerText?.(`${missingAssets.length} missing assets being uploaded`);
```
ide, line 301:
```suggestion
  updateSpinnerText?.(`${missingAssets.length} new assets being uploaded`);
```

packages/eas-cli/src/project/publish.ts
ide: I think this is something to discuss with @wschurman. Yesterday he started optimizing the update-publishing flow, which is bottlenecked on the number of assets and has several checks for the asset count.

I could see two approaches that are mutually compatible: expose the asset limits through a GraphQL API, or allow existing GraphQL APIs to return an error message or warning message that is EAS CLI is then responsible for displaying.

In either case, you may want to back out these changes from this PR and move them to another PR since everything else here looks ready to land.

packages/eas-cli/src/project/publish.ts
```diff
@@ -282,6 +294,7 @@ export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAsset
     })
   );
 
+  updateSpinnerText?.(`${missingAssets.length} new assets being uploaded`);
```
ide, line 297: Thinking about this some more, the phrasing of "being uploaded" is a bit unwieldy compared to saying "uploading X new assets" but the issue with that is we prefix the line with "Uploading assets..." and "Uploading assets... uploading X new assets" is awkward too.

There also needs to be `${missingAssets.length === 1 ? 'asset' : 'assets'}` in there too.

ide (approved):

packages/eas-cli/src/project/publish.ts
kgc00: is there some npm `plural()` package 😄

kgc00: all of the awkwardness you mentioned is real. i prefer: `Uploading assets... 23 new assets uploading`. It keeps with this convention of (prefix) (number) (description)

ide: 'Uploading assets... 23 new assets uploading" sounds fine to me

ide (approved):

PR #1153, opened 2022-06-10
[eas-cli] Limit concurrent asset uploads

wschurman:
# Checklist

# Why

Google gets overloaded when too many assets are uploaded at once. 

Closes ENG-5232.

# How

This limits it to 50 at a time. Previously there was no limit but we had another artificial limit of 400 on the server side, so the max concurrent uploads was 400. We could do 400 here, but I figure 50 is probably a bit more reliable.

# Test Plan

With eas-cli changes running locally and www changes running locally from https://github.com/expo/universe/pull/9852, with an app with 1600 assets (temporarily set MAX_ASSETS_PER_UPDATE to 2000 on local server)
```
export EXPO_LOCAL=1
EXPO_DEBUG=1 ~/expo/eas-cli/packages/eas-cli/bin/run update --branch preview --message "Updating the app"
```
see it finishes successfully.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Match bundle identifier capabilities more accurately. ([#1112](https://github.com/expo/eas-cli/pull/1112) by [@EvanBacon](https://github.com/EvanBacon))
-- Limit concurrent asset uploads to google. ([#1153](https://github.com/expo/eas-cli/pull/1153) by [@wschurman](https://github.com/wschurman))
+- Limit concurrent asset uploads. ([#1153](https://github.com/expo/eas-cli/pull/1153) by [@wschurman](https://github.com/wschurman))
 
 ### 🧹 Chores
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Match bundle identifier capabilities more accurately. ([#1112](https://github.com/expo/eas-cli/pull/1112) by [@EvanBacon](https://github.com/EvanBacon))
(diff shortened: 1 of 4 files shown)


wschurman:
/changelog-entry bug-fix Limit concurrent asset uploads to google

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Match bundle identifier capabilities more accurately. ([#1112](https://github.com/expo/eas-cli/pull/1112) by [@EvanBacon](https://github.com/EvanBacon))
+- Limit concurrent asset uploads to google. ([#1153](https://github.com/expo/eas-cli/pull/1153) by [@wschurman](https://github.com/wschurman))
```
ide, line 14: GCS is an implementation detail
```suggestion
- Limit concurrent asset uploads. ([#1153](https://github.com/expo/eas-cli/pull/1153) by [@wschurman](https://github.com/wschurman))
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -275,10 +276,14 @@ export async function uploadAssetsAsync(assetsForUpdateInfoGroup: CollectedAsset
     missingAssets.map(ma => ma.contentType)
   );
 
+  const assetUploadPromiseLimit = promiseLimit(50);
```
ide, line 279: I suspect the bottleneck will usually be the developer's upload bandwidth, not GCS's ability to receive data and store it to disk. In other works, I don't think a high-level of parallelization helps that much and a lower limit like 5? 10? 15? might be faster since there's less congestion.

wschurman: The only piece of concrete data I have is that it seemed like I was rate limited when I tried to upload 1500 assets at once (even on a fast connection with lots of upload bandwidth). 

I'll make this 15 though since it does make sense that a lower-level of parallelization makes sense.

wschurman, line 279:
```suggestion
  const assetUploadPromiseLimit = promiseLimit(15);
```

ide (approved):

PR #1156, opened 2022-06-13
[eas-cli] Auto-configure eas.json with metadata when missing

byCedric:
# Checklist

# Why

When `eas.json` is missing from the project, it fails with `Could not find eas.json`. With this change, we generate the default `eas.json` file.

# How

Reused the `eas build` `ensureProjectConfiguredAsync` after validating the user's local Expo/RN project.

# Test Plan

### DIFF
--- a/packages/eas-cli/src/commands/metadata/pull.ts
+++ b/packages/eas-cli/src/commands/metadata/pull.ts
@@ -1,6 +1,7 @@
 import { getConfig } from '@expo/config';
 import { Flags } from '@oclif/core';
 
+import { ensureProjectConfiguredAsync } from '../../build/configure';
 import EasCommand from '../../commandUtils/EasCommand';
 import { CredentialsContext } from '../../credentials/context';
 import Log, { learnMore } from '../../log';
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/metadata/configure.ts
```diff
@@ -43,6 +44,7 @@ export default class MetadataConfigure extends EasCommand {
     const projectDir = await findProjectRootAsync();
     const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
```
wkozyra95, line 45: use getExpoConfig wrapper instead

wkozyra95 (commented):
we do not run that configure for other commands (only for build and build:configure)

We can have a discussion about adding that, but if we do it should be consistent for all commands e.g. submit is also failing with that error.

kbrandwijk:
What part of the eas.json configuration is needed for `eas metadata` to function?

byCedric:
As discussed, see PR #1136

kbrandwijk:
> As discussed, see PR #1136

?

PR #1177, opened 2022-06-23
[ci] Add run in band for coverage testing

byCedric:
# Checklist

# Why

This should limit the maximum memory used in different workers. Doing so should make CI a bit more stable

# How

- `--runInBand` on Node 18 coverage tests

# Test Plan

See CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       - run: yarn typecheck
       - run: yarn test
         if: ${{ !matrix.coverage }}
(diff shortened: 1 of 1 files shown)


byCedric:
Interesting seems like that reproduces the OOM instead of fixing it. Will probably do some debugging to see why, and also why `node:1742) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit` is popping up

PR #1193, opened 2022-07-01
[eas-cli] Drop deprecated idfa and price tier from store config

byCedric:
# Checklist

# Why

This drops both `apple.idfa` and `apple.priceTier`.
- For IDFA, that's because it has been fully deprecated in ASC in favor of [AppTrackTransparency](https://developer.apple.com/documentation/apptrackingtransparency)
- For price tier, that's because there is a v2 that is completely different and needs to be re-implemented.

# How

Dropped both configuration options from store config schema

# Test Plan

Only a change in the store config schema


EvanBacon (approved):

byCedric:
/changelog-entry chore Remove deprecated idfa and price tier from store config

PR #1194, opened 2022-07-01
[eas-cli] Add version selector to store config for metadata

byCedric:
# Checklist

# Why

Fixes ENG-5286

This allows users to create or select an editable app version through the store config.

# How

- Added `apple.version` to store config schema
- Added find and update existing editable app version
- Added create new editable app version

# Test Plan

- See added tests

### DIFF
--- a/packages/eas-cli/schema/metadata-0.json
+++ b/packages/eas-cli/schema/metadata-0.json
@@ -572,11 +572,10 @@
       "description": "Configuration that is specific to the Apple App Store.",
       "additionalProperties": false,
       "properties": {
-        "advisory": {
-          "$ref": "#/definitions/apple/AppleAdvisory"
-        },
-        "categories": {
-          "$ref": "#/definitions/apple/AppleCategory"
+        "version": {
+          "type": "string",
+          "description": "The version string to edit, or create, if it doesn't exist.",
+          "minLength": 1
         },
         "copyright": {
           "type": "string",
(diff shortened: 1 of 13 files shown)


packages/eas-cli/schema/metadata-0.json
```diff
@@ -606,11 +606,10 @@
...
       "additionalProperties": false,
       "properties": {
-        "advisory": {
-          "$ref": "#/definitions/apple/AppleAdvisory"
-        },
-        "categories": {
-          "$ref": "#/definitions/apple/AppleCategory"
+        "version": {
+          "type": "string",
+          "description": "The version string to edit, or create, if it doesn't exist.",
```
EvanBacon, line 611: Maybe an example of the version string could be helpful so users don't confuse it with the build number.

packages/eas-cli/src/metadata/apple/tasks/app-version.ts
```diff
@@ -156,3 +194,58 @@ async function resolveVersionAsync(
...
+            AppStoreState.INVALID_BINARY,
+          ].join(','),
+        },
+      },
+    });
+
+    const version = versions.find(model => model.attributes.versionString === options.version);
+    if (version) {
+      return version;
+    }
```
EvanBacon, line 227: Maybe a `debug` log showing available versions could be useful here.

EvanBacon (approved):

packages/eas-cli/schema/metadata-0.json
byCedric: Done, default behavior is using the latest editable version

packages/eas-cli/src/metadata/apple/tasks/app-version.ts
byCedric: I need to add debug in some other places, but that sounds like a good idea

PR #1197, opened 2022-07-06
[eas-cli] Add pagination to supported GQL queries

kgc00:
# Checklist

# Why

Closes [ENG-5322](https://linear.app/expo/issue/ENG-5322/add-pagination-to-eas-cli-for-eas-update)

# How

Unfortunately, this is a pretty beefy PR 🐮 

IMO, we had two options to add pagination to our CLI. The first option was to copy something like the [aws-cli pagination options](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html) and add flags for each command that control that logic (e.g. `--offset` and `--pageSize`). The second option was to create an interactive selector for each command, where users could select whether they wanted to fetch more pages.

The pros of the first option are its simplicity, and that it lends itself well to automation. The pros of the second option are that it's a nicer experience for humans, especially newer users, and that it doesn't force me to change the flags on a bunch of commands.

I ended up going with the second option.

I created a generic-ish function `performPaginatedQuery` that will operate on and return whatever `T` the caller provides (think GQL types like `BuildFragment` or `Update`). The function will recursively call itself while incrementing its offset until no more items are returned. I provided a callback where users can define how the specific query should be executed and what results should be saved from it.

`performPaginatedQuery` is accepts an arg `promptOptions` that is of type `confirm` or `select`. This controls how the user interacts with the CLI. `select` prompts require a user to select a list item and `performPaginatedQuery` return that value when the function finishes execution. `confirm` prompts require a user to approve querying for an additional page of items if there are more pages left, and return always an empty array of `QueryReturnType`. 

Shoutouts to @FiberJW who paired with me on some of this 👍 

# Test Plan

I manually tested each flow for each command that was affected. There's quite a bit here, so I may have missed some flow somewhere. Commands affected:
- branch/list.ts
- branch/view.ts
- build/cancel.ts
- build/list.ts
- channel/list.ts
- channel/rollout.ts
- channel/view.ts
- update/index.ts
- update/list.ts

+ `--non-interactive` + `--json` if those flags are applicable to those commands. (Assuming that prompting the user for input when `--json` is present will not break anything.)

### Confirm (list and view commands)

https://user-images.githubusercontent.com/15132641/177445637-91c35951-85a8-4597-b116-76d4767ef711.mov

### Select 

https://user-images.githubusercontent.com/15132641/177616248-3c8fdb7b-2a49-4f1c-a8e7-d56031511445.mov

### DIFF
--- a/packages/eas-cli/src/commands/channel/list.ts
+++ b/packages/eas-cli/src/commands/channel/list.ts
@@ -1,4 +1,5 @@
 import { Flags } from '@oclif/core';
+import chalk from 'chalk';
 import gql from 'graphql-tag';
 
 import EasCommand from '../../commandUtils/EasCommand';
@@ -12,31 +13,41 @@ import { getExpoConfig } from '../../project/expoConfig';
 import { findProjectRootAsync, getProjectIdAsync } from '../../project/projectUtils';
 import formatFields from '../../utils/formatFields';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+import { PaginatedQueryResponse, performPaginatedQueryAsync } from '../../utils/queries';
+import { BRANCHES_LIMIT } from '../branch/list';
 import { logChannelDetails } from './view';
 
-const CHANNEL_LIMIT = 10_000;
+const CHANNEL_LIMIT = 50;
 
 async function getAllUpdateChannelForAppAsync({
   appId,
-}: {
-  appId: string;
-}): Promise<GetAllChannelsForAppQuery> {
+  channelLimit = CHANNEL_LIMIT,
+  channelOffset = 0,
+  branchLimit = BRANCHES_LIMIT,
+  branchOffset = 0,
+}: GetAllChannelsForAppQueryVariables): Promise<GetAllChannelsForAppQuery> {
   return await withErrorHandlingAsync(
     graphqlClient
       .query<GetAllChannelsForAppQuery, GetAllChannelsForAppQueryVariables>(
         gql`
-          query GetAllChannelsForApp($appId: String!, $offset: Int!, $limit: Int!) {
+          query GetAllChannelsForApp(
+            $appId: String!
+            $channelOffset: Int!
+            $channelLimit: Int!
+            $branchLimit: Int!
+            $branchOffset: Int!
+          ) {
             app {
               byId(appId: $appId) {
                 id
-                updateChannels(offset: $offset, limit: $limit) {
+                updateChannels(offset: $channelOffset, limit: $channelLimit) {
                   id
                   name
                   branchMapping
-                  updateBranches(offset: 0, limit: $limit) {
+                  updateBranches(offset: $branchOffset, limit: $branchLimit) {
                     id
                     name
-                    updates(offset: 0, limit: 10) {
+                    updates(offset: 0, limit: 1) {
                       id
                       group
                       message
@@ -59,7 +70,7 @@ async function getAllUpdateChannelForAppAsync({
             }
           }
         `,
-        { appId, offset: 0, limit: CHANNEL_LIMIT },
+        { appId, channelOffset, channelLimit, branchLimit, branchOffset },
         { additionalTypenames: ['UpdateChannel', 'UpdateBranch', 'Update'] }
       )
       .toPromise()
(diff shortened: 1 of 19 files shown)


kgc00:
I noticed we don't use `/Users/kirby/Expo/eas-cli/packages/eas-cli/src/graphql/queries/SubmissionQuery.ts` anywhere. Should we remove it?

packages/eas-cli/src/update/utils.ts
```diff
@@ -74,3 +76,115 @@ export function ensureValidVersions(exp: ExpoConfig, platform: RequestedPlatform
...
+};
+
+/**
+ * Return an empty array when the promptOptions type is PaginatedQueryConfirmPrompt
+ * Returns an array of the selected items when the promptOptions type is PaginatedQuerySelectPrompt
+ */
+export async function performPaginatedQueryAsync<QueryReturnType extends Record<string, any>>(
+  { pageSize, offset, queryToPerform, promptOptions }: PaginatedQueryArguments<QueryReturnType>,
+  accumulator?: QueryReturnType[]
+): Promise<QueryReturnType[]> {
```
kgc00, line 131: Spent quite a bit of time trying to get a [conditional return type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) working here where if `promptOptions['type']` was `confirm` we return `void`, otherwise return `QueryReturnType`, but I couldn't quite get it functioning. Would love to see that behavior, if anyone has any thoughts on how to do it.

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -88,18 +99,77 @@ export default class ChannelList extends EasCommand {
...
+          channel.updateBranches = branchesForChannels[channel.id];
+        });
+      }
+
+      paginatedQuery.queryResponseRawLength = channels.length;
+      branchOffset += branchLimit + 1;
+    } while (!queriedAllBranches);
+
+    return paginatedQuery;
+  };
```
kgc00, line 163: we need to query all branches for each channel, unfortunately, to display the proper `branchMapping` table to the user.

kind of nasty, but there is no way around it client-side. if we changed how the back-end returned or sorted branches, we could just query for the first 1 or 2. That seems like a good next task.

packages/eas-cli/src/utils/queries.ts
```diff
@@ -0,0 +1,107 @@
...
+  | PaginatedQuerySelectPrompt<QueryReturnType>;
+
+/**
+ * Return an empty array when the promptOptions type is PaginatedQueryConfirmPrompt
+ * Returns an array of the selected items when the promptOptions type is PaginatedQuerySelectPrompt
+ */
+export async function performPaginatedQueryAsync<QueryReturnType extends Record<string, any>>(
+  { pageSize, offset, queryToPerform, promptOptions }: PaginatedQueryArguments<QueryReturnType>,
+  accumulator?: QueryReturnType[]
+): Promise<QueryReturnType[]> {
```
kgc00, line 48: Spent quite a bit of time trying to get a [conditional return type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) working here where if promptOptions['type'] was confirm we return void, otherwise return QueryReturnType, but I couldn't quite get it functioning. Would love to see that behavior if anyone wants to take a crack.

wkozyra95 (changes requested):
The idea behind `--json` option is that people can process command output end do sth with it after eas command finishes, it should be available only in non interactive mode, or in case of :list commands it was always available because there was nothing interactive there.

In your PR you are adding interactivity to commands that did not have them, but you are not handling case when users wants to sue those commands in non interactive way:

I would suggest to:
- add `--non-interactive` flag to all commands that you add interactivity 
- json flag should only be allowed if non-interactive is set(or setting JSON flag could automatically default non-interactive to true, but the non-interactive flag should still exist)
- add cli params for limit and offset to all commands that support paging and can be run in a non-interactive way(or if it's not necessary for specific command we could just print last n entires)

to verify whether specific behavior makes sense I recommend checking how cli behaves if you pipe results to e.g. `eas build:list --json | jq '.' > somefile.json`

kgc00:
i've broken this into other, smaller PRs. closing this one since it's redundant.

PR #1211, opened 2022-07-14
[cli] Change wording on build warning for bundle id/package name

kbrandwijk:
# Checklist

# Why

The warning about package name / bundle identifier also shows up after running `expo prebuild` or `expo run`, in which case it's expected that app.json contains these fields. The warning that it will be ignored once the native code exists is okay, but the deprecation notice should be removed.

# How

Old vs new:
image

# Test Plan

Tested by running `expo prebuild` followed by `eas build`.

### DIFF
--- a/packages/eas-cli/src/project/android/applicationId.ts
+++ b/packages/eas-cli/src/project/android/applicationId.ts
@@ -161,7 +161,7 @@ export function warnIfAndroidPackageDefinedInAppConfigForBareWorkflowProject(
     Log.warn(
       `Specified value for "android.package" in ${getProjectConfigDescription(
         projectDir
-      )} is ignored if native project folders are detected.\n` +
+      )} is ignored because native project folders were detected.\n` +
         'EAS Build will use the value found in the native code.'
     );
     warnPrinted = true;
--- a/packages/eas-cli/src/project/android/applicationId.ts
+++ b/packages/eas-cli/src/project/android/applicationId.ts
@@ -159,10 +159,10 @@ export function warnIfAndroidPackageDefinedInAppConfigForBareWorkflowProject(
 ): void {
   if (AndroidConfig.Package.getPackage(exp) && !warnPrinted) {
     Log.warn(
-      `Specifying "android.package" in ${getProjectConfigDescription(
+      `Specified value for "android.package" in ${getProjectConfigDescription(
         projectDir
-      )} is deprecated for bare workflow projects.\n` +
-        'EAS Build depends only on the value in the native code. Please remove the deprecated configuration.'
+      )} is ignored if native project folders are detected.\n` +
+        'EAS Build will use the value found in the native code.'
     );
     warnPrinted = true;
   }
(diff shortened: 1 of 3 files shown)


kbrandwijk:
/changelog-entry chore Change wording on build warning for bundle id/package name

dsokal (approved):

packages/eas-cli/src/project/android/applicationId.ts
```diff
@@ -159,10 +159,10 @@ export function warnIfAndroidPackageDefinedInAppConfigForBareWorkflowProject(
 ): void {
   if (AndroidConfig.Package.getPackage(exp) && !warnPrinted) {
     Log.warn(
-      `Specifying "android.package" in ${getProjectConfigDescription(
+      `Specified value for "android.package" in ${getProjectConfigDescription(
         projectDir
-      )} is deprecated for bare workflow projects.\n` +
-        'EAS Build depends only on the value in the native code. Please remove the deprecated configuration.'
+      )} is ignored if native project folders are detected.\n` +
```
wkozyra95, line 164: maybe 

```suggestion
      )} is ignored because native project folders were detected.\n` +
```

Also, this warning will be printed if the native directory exists and it is not gitignored, I think calling that bare workflow is more precise here, but I do not have a strong opinion on that

```suggestion
      )} is ignored because this is a bare workflow project.\n` +
```

This message has added benefit of pointing out to users that they have a bare project and a lot of them lately are ejecting unknowingly

wkozyra95 (approved):

packages/eas-cli/src/project/android/applicationId.ts
brentvatne, line 164:
```suggestion
      )} is ignored because an ${chalk.bold('android')} directory was detected in the project, indicating that this is a bare workflow project.\n` +
```

a bit wordy but
- lets people know what triggered this detection (although we could possibly be more precise - detected dir and wasn't gitignored)
- factors in @wkozyra95's comment about bare workflow

kbrandwijk: I specifically moved away from the old 'bare workflow' wording and opted for 'native project folders' because running prebuild is not the same as bare workflow. Bare workflow is where you have made manual changes to native files.

kbrandwijk: See my comment above. I specifically moved away from saying 'bare workflow'. I do like the more specific folder naming.

wkozyra95: I agree that running prebuild is not the same as bare workflow, but not in the way you described. Bare workflow is when you run prebuild and android/ios are not gitignored. Also, this issue is present only in bare workflow, not every time you run prebuild.

kbrandwijk: If I init a project, define my identifiers, run `expo prebuild`, then `eas build`, I get this warning. One of the most common causes for confusion is that we advertise `expo prebuild` and `expo run` as something that can repeatedly be run _in the managed worfklow_, so we shouldn't mention 'bare workflow' here at all imho.
Bottom line here is: there's a native folder, so we're using it (unless you manually gitignore or easignore it, which is not by default in our templates). Mentioning 'bare workflow' without saying 'there's a native folder' is confusing for users, our daily support questions make that very clear.

kbrandwijk:
image
Going with the middle road of the more specific wording.

PR #1270, opened 2022-08-11
[eas-cli] Add js metadata config file support

byCedric:
# Checklist

# Why

This adds support for `store.config.js` files support, for `eas metadata:push`. It supports returned `object`, `function` or `async function`.

# How

Added `loadConfigAsync` and `saveConfigAsync` that handles language specific things, like reading/writing JSON or JavaScript content.

# Test Plan

See added tests

### DIFF
--- a/packages/eas-cli/src/metadata/__tests__/config.test.ts
+++ b/packages/eas-cli/src/metadata/__tests__/config.test.ts
@@ -41,9 +41,9 @@ describe(getStaticConfigFile, () => {
 describe(loadConfigAsync, () => {
   const projectDir = path.resolve(__dirname, 'fixtures');
 
-  it(`throws when file doesn't exists`, async () => {
+  it(`throws when file doesn't exist`, async () => {
     await expect(
-      loadConfigAsync({ projectDir, metadataPath: 'doesnt-exists.json' })
+      loadConfigAsync({ projectDir, metadataPath: 'doesnt-exist.json' })
     ).rejects.toThrow('file not found');
   });
 
--- a/packages/eas-cli/src/metadata/__tests__/config.test.ts
+++ b/packages/eas-cli/src/metadata/__tests__/config.test.ts
@@ -108,4 +108,30 @@ describe(loadConfigAsync, () => {
       loadConfigAsync({ projectDir, metadataPath: 'invalid.async.js' })
     ).resolves.toMatchObject({ configVersion: -1 });
   });
+
+  it(`throws invalid config type from "invalid-type.config.json"`, async () => {
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid-type.config.json' })
+    ).rejects.toThrow(
+      `store config returned an unkown value, should be an object: "${[{ configVersion: -1 }]}"`
+    );
+  });
+
+  it(`throws invalid config type from "invalid-type.config.js"`, async () => {
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid-type.config.js' })
+    ).rejects.toThrow(`store config returned an unkown value, should be an object: "${1337}"`);
+  });
+
+  it(`throws invalid config type from "invalid-type.function.js"`, async () => {
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid-type.function.js' })
+    ).rejects.toThrow(`store config returned an unkown value, should be an object: "${true}"`);
+  });
+
+  it(`throws invalid config type from "invalid-type.async.js"`, async () => {
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid-type.async.js' })
+    ).rejects.toThrow(`store config returned an unkown value, should be an object: "${null}"`);
+  });
 });
--- a/packages/eas-cli/src/metadata/__tests__/config.test.ts
+++ b/packages/eas-cli/src/metadata/__tests__/config.test.ts
@@ -54,30 +54,58 @@ describe(loadConfigAsync, () => {
     ).rejects.toThrow('errors found');
   });
 
-  it(`returns json config`, async () => {
+  it(`returns config from "store.config.json"`, async () => {
     await expect(
       loadConfigAsync({ projectDir, metadataPath: 'store.config.json' })
-    ).resolves.toHaveProperty('apple.copyright', '2022 ACME');
+    ).resolves.toHaveProperty('apple.copyright', 'ACME');
   });
 
-  it(`returns js config`, async () => {
+  it(`returns config from "store.config.js"`, async () => {
     const year = new Date().getFullYear();
     await expect(
       loadConfigAsync({ projectDir, metadataPath: 'store.config.js' })
     ).resolves.toHaveProperty('apple.copyright', `${year} ACME`);
   });
 
-  it(`returns json config when skipping validation`, async () => {
+  it(`returns config from "store.function.js"`, async () => {
+    const year = new Date().getFullYear();
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'store.function.js' })
+    ).resolves.toHaveProperty('apple.copyright', `${year} ACME`);
+  });
+
+  it(`returns config from "store.async.js"`, async () => {
+    const year = new Date().getFullYear();
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'store.async.js' })
+    ).resolves.toHaveProperty('apple.copyright', `${year} ACME`);
+  });
+
+  it(`returns invalid config from "invalid.config.json`, async () => {
     jest.mocked(confirmAsync).mockResolvedValue(true);
     await expect(
       loadConfigAsync({ projectDir, metadataPath: 'invalid.config.json' })
     ).resolves.toMatchObject({ configVersion: -1 });
   });
 
-  it(`returns js config when skipping validation`, async () => {
+  it(`returns invalid config from "invalid.config.js`, async () => {
     jest.mocked(confirmAsync).mockResolvedValue(true);
     await expect(
       loadConfigAsync({ projectDir, metadataPath: 'invalid.config.js' })
+    ).resolves.toMatchObject({ configVersion: -1 });
+  });
+
+  it(`returns invalid config from "invalid.function.js`, async () => {
+    jest.mocked(confirmAsync).mockResolvedValue(true);
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid.function.js' })
+    ).resolves.toMatchObject({ configVersion: -1 });
+  });
+
+  it(`returns invalid config from "invalid.async.js`, async () => {
+    jest.mocked(confirmAsync).mockResolvedValue(true);
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'invalid.async.js' })
     ).resolves.toMatchObject({ configVersion: -1 });
   });
 });
--- a/packages/eas-cli/src/metadata/__tests__/config.test.ts
+++ b/packages/eas-cli/src/metadata/__tests__/config.test.ts
@@ -1,11 +1,42 @@
-import fs from 'fs';
 import path from 'path';
-import tempy from 'tempy';
 
 import { confirmAsync } from '../../prompts';
-import { loadConfigAsync, saveConfigAsync } from '../config';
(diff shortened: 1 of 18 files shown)


byCedric:
/changelog-entry new-feature Add metadata support for dynamic store.config.js files

byCedric:
What's up with the installer tests?

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +19,74 @@ export interface MetadataConfig {
...
+      validationErrors
+    );
+
+    logMetadataValidationError(error);
+    Log.newLine();
+    Log.warn(
+      'Without further updates, the current store configuration may fail to be synchronized with the App Store or pass App Store review.'
+    );
+
+    if (await confirmAsync({ message: 'Do you still want to push the store configuration?' })) {
```
dsokal, line 52: Maybe it's better to warn the user that if they don't accept, EAS CLI will fail.

dsokal (commented):
LGTM

packages/eas-cli/src/metadata/config.ts
byCedric: I think it's clear enough that the CLI will exit with an error. It may be a bit repetitive as is though (it just exits with the original error that was printed before). But happy to improve it!

### When prompting

Screenshot 2022-08-12 at 16 33 34 (3)


### When answering no

Screenshot 2022-08-12 at 16 34 35

byCedric:
I just rebased on `main` because we released a new version.

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,81 @@ export interface MetadataConfig {
   apple?: AppleMetadata;
 }
 
+/**
+ * Resolve the dynamic config from the user.
+ * It supports methods, async methods, or objects (json).
+ */
+async function resolveDynamicConfigAsync(configFile: string): Promise<unknown> {
+  const userConfigContent = await import(configFile);
```
byCedric, line 26: I think this was the ESM-styled dynamic imports. I think we can use them bc it's transpiled.

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,81 @@ export interface MetadataConfig {
...
+  const userConfig = userConfigContent.default || userConfigContent;
+
+  if (typeof userConfig === 'function') {
+    return await userConfig();
+  } else if (typeof userConfig === 'object') {
+    return userConfig;
+  }
```
byCedric, line 33: @EvanBacon We can also add more "side wheels" to help users when they return anything other than an `object` from the (async) function? But not sure if necessary./

byCedric: Screw it, did it anyways.

EvanBacon (approved):

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
   apple?: AppleMetadata;
 }
 
+/**
+ * Resolve the dynamic config from the user.
+ * It supports methods, async methods, or objects (json).
+ */
+async function resolveDynamicConfigAsync(configFile: string): Promise<unknown> {
+  const userConfigContent = await import(configFile);
+  let userConfig = userConfigContent.default || userConfigContent;
```
dsokal, line 27:
```suggestion
  let userConfig = userConfigContent.default ?? userConfigContent;
```

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+
+  if (typeof userConfig === 'function') {
+    userConfig = await userConfig();
```
dsokal, line 30: I don't like reusing variables for values that don't have anything in common.

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+  }
```
dsokal, line 31: Maybe sth like:
```suggestion
  let maybeUserConfigFunction = userConfigContent.default ?? userConfigContent;

  let userConfig;
  if (typeof maybeUserConfigFunction === 'function') {
    userConfig = await maybeUserConfigFunction();
  }
```

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+
+  if (typeof userConfig === 'object' && userConfig !== null && !Array.isArray(userConfig)) {
+    return userConfig;
+  }
+
+  throw new MetadataValidationError(
+    `Metadata store config returned an unkown value, should be an object: "${userConfig}"`
```
dsokal, line 38:
```suggestion
    `Metadata store config returned an unknown value, should be an object: "${userConfig}"`
```

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
   apple?: AppleMetadata;
 }
 
+/**
+ * Resolve the dynamic config from the user.
+ * It supports methods, async methods, or objects (json).
+ */
+async function resolveDynamicConfigAsync(configFile: string): Promise<unknown> {
```
dsokal, line 25: Why does it return `unknown`?

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+  throw new MetadataValidationError(
+    `Metadata store config returned an unkown value, should be an object: "${userConfig}"`
+  );
+}
+
+/**
+ * Get the static configuration file path, based on the metadata context.
+ * This uses any custom name provided, but swaps out the extension for `.json`.
+ */
+export function getStaticConfigFile({
```
dsokal, line 46:
```suggestion
export function getStaticConfigFilePath({
```

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+  if (!valid) {
+    const error = new MetadataValidationError(
+      `Metadata store config errors found`,
+      validationErrors
+    );
+
+    logMetadataValidationError(error);
+    Log.newLine();
+    Log.warn(
+      'Without further updates, the current store configuration may fail to be synchronized with the App Store or pass App Store review.'
```
dsokal, line 83: @brentvatne taught me this :D 

```suggestion
      'Without further updates, the current store configuration can fail to be synchronized with the App Store or pass App Store review.'
```

<a href="https://imgflip.com/i/42bw1b"></a>

packages/eas-cli/src/metadata/__tests__/config.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const expectedFile = path.join(projectDir, 'custom-name.json');
+
+    expect(getStaticConfigFile({ projectDir, metadataPath })).toBe(expectedFile);
+  });
+});
+
+describe(loadConfigAsync, () => {
+  const projectDir = path.resolve(__dirname, 'fixtures');
+
+  it(`throws when file doesn't exists`, async () => {
```
dsokal, line 44:
```suggestion
  it(`throws when file doesn't exist`, async () => {
```

packages/eas-cli/src/metadata/__tests__/config.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    await expect(
+      loadConfigAsync({ projectDir, metadataPath: 'doesnt-exists.json' })
```
dsokal, line 46:
```suggestion
      loadConfigAsync({ projectDir, metadataPath: 'doesnt-exist.json' })
```

dsokal (approved):

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+ */
+async function resolveDynamicConfigAsync(configFile: string): Promise<unknown> {
+  const userConfigContent = await import(configFile);
+  let userConfig = userConfigContent.default || userConfigContent;
+
+  if (typeof userConfig === 'function') {
+    userConfig = await userConfig();
+  }
+
+  if (typeof userConfig === 'object' && userConfig !== null && !Array.isArray(userConfig)) {
```
wkozyra95, line 33: The next call after this function will be validation, and by returning unknown you are losing the information you are checking here.

Either run full validation as part of this function or do not validate at all.

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+  return path.join(projectDir, `${path.basename(configFile, configExtension)}.json`);
+}
+
+/**
+ * Load the store configuration from a metadata context.
+ * This can load `.json` and `.js` config files, using `require`.
+ * It throws MetadataValidationErrors when the file doesn't exist, or contains errors.
+ * The user is prompted to try anyway when errors are found.
+ */
+export async function loadConfigAsync({
```
wkozyra95, line 62: This function is util for reading local configuration, but during the error handling, you are assuming that it will be only called when pushing an update to store. This function should only return the config or throw an error, those error logs and prompt to continue should be separated.

packages/eas-cli/src/metadata/config.ts
```diff
@@ -12,6 +18,83 @@ export interface MetadataConfig {
...
+}
+
+/**
+ * Get the static configuration file path, based on the metadata context.
+ * This uses any custom name provided, but swaps out the extension for `.json`.
+ */
+export function getStaticConfigFile({
+  projectDir,
+  metadataPath,
+}: Pick<MetadataContext, 'projectDir' | 'metadataPath'>): string {
```
wkozyra95, line 49: nit
If you are passing the entire context either way, then I would prefer:
```suggestion
}: MetadataContext): string {
```
or if you want to narrow down the type (and pass only those 2 props on call-site)
```suggestion
}: { projectDir: string, metadataPath: string }): string {
```

wkozyra95 (approved):

packages/eas-cli/src/metadata/config.ts
byCedric: When I wrote this, the idea was that the user configuration should be strongly typed only after we've validated that it matches the schema.

So this method just returns the data from the user config without it being validated. That's what we do in `loadConfigAsync`. We can just change this, but I think the `unknown` reflects the type better here.

packages/eas-cli/src/metadata/config.ts
byCedric: Fair point, will split it up.

packages/eas-cli/src/metadata/config.ts
brentvatne: 🤣

packages/eas-cli/src/metadata/config.ts
byCedric: Seems good to me. I think in other places I've used something like `xOrY`, so I'll change it to:

```
const userConfigOrFunction = ...
let userConfig
```

Does that seem reasonable?

packages/eas-cli/src/metadata/config.ts
byCedric: Let's continue in your proposal discussion

packages/eas-cli/src/metadata/config.ts
byCedric: I think I used `MetadataContext` when writing this first. But, when I worked on the tests, it was kind of annoying to mock out the whole `MetadataContext` every time. So I think this is how I got to that point: to make testing easier.

Let's go for the `{ projectDir, metadataPath }` option.

packages/eas-cli/src/metadata/config.ts
dsokal: Yes, makes sense.

packages/eas-cli/src/metadata/config.ts
byCedric: Done

packages/eas-cli/src/metadata/config.ts
byCedric: Done, validation is now moved to `loadConfigAsync`, and `skipValidation` option is added ([just like the CLI options](https://github.com/expo/expo/search?l=TypeScript&q=skip)).

PR #1272, opened 2022-08-12
[workspace] improve DX while working on Windows

Simek:
# Why

This PR aims to improve DX while working on the `eas-cli` on Windows.

# How

The following changes have been made:
* `rm` calls replaced with `rimraf`
* most of the shell script has been rewritten as Node scripts and commands

I have skipped rewriting `pretarball-ci.sh` script for now, since we are not running `release` workflow on Windows.

> I have also spotted that there are some changes to generated GraphQL code, which are not related to this PR. Let me know if I should revert those changes.

# Test Plan

Yarn scripts like `generate-graphql-code` and `prepack` from `eas-cli` works as expected on Windows. Going to check this on macOS soon and I would like to see the CI results.


packages/eas-cli/scripts/prepack.js
```diff
@@ -0,0 +1,14 @@
+const { exec } = require('child_process');
+
+if (process.env.CLI_SIZE_CHECK) {
+  exec('yarn oclif manifest', (error, stdout, stderr) => {
+    if (error || stderr) {
+      // eslint-disable-next-line no-console
+      console.error(error?.message || stderr);
+      return;
```
dsokal, line 8:
```suggestion
      process.exit(1);
```

packages/eas-json/package.json
```diff
@@ -18,6 +18,7 @@
   "devDependencies": {
     "@types/fs-extra": "9.0.13",
     "memfs": "3.4.7",
+    "rimraf": "^3.0.2",
```
wkozyra95, line 21: use strict versions

.prettierrc
```diff
@@ -4,5 +4,6 @@
   "singleQuote": true,
   "bracketSameLine": true,
   "trailingComma": "es5",
-  "arrowParens": "avoid"
+  "arrowParens": "avoid",
+  "endOfLine": "auto"
```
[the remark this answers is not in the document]
Simek: Auto works a bit weird, so probably only files created on Windows will have the incorrect line ending:
* https://prettier.io/docs/en/options.html#end-of-line

Unfortunately forcing the `lf` line endings does not work on Windows, and warnings are produced even with correct EoL set:
Screenshot 2022-08-12 151212

On macOS with the same setting added check ends successfully. If there are some EoL sensitive files in the repo we probably can use `.gitattributes` to ensure that they always have a correct EoL, or just et the correct matchers in there:
* https://git-scm.com/docs/gitattributes#_end_of_line_conversion

packages/eas-cli/scripts/prepack.js
```diff
@@ -0,0 +1,14 @@
+const { exec } = require('child_process');
```
wkozyra95, line 1: use @expo/spawn-async, it's already in deps

packages/eas-cli/scripts/annotate-graphql-codegen.js
```diff
@@ -1,13 +1,15 @@
-#!/usr/bin/env sh
+const { promises: fs } = require('fs');
```
wkozyra95, line 1:
```suggestion
const fs = require('fs/promises');
```

.prettierrc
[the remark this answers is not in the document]
dsokal: So can you please configure `.gitattributes` and make sure it works correctly?

kbrandwijk:
What's the motivation to add this support? Is anyone actively working on this repo that insists on doing so on Windows? It adds to the burden of making sure things keep working to fully support another platform.

packages/eas-cli/scripts/prepack.js
Simek: I have initially went with that package, but it produces an ESLint error which need to be suppressed, so I just used Node API instead:
Untitled

dsokal (approved):

PR #1287, opened 2022-08-18
[eas-cli] Limit the number of SignedAssetUploadSpecifications fetched at a time

wschurman:
# Checklist

# Why

Generating a ton of these at once on the server creates 1) requests that are huge and that have huge responses, and 2) a lot of load to google to generate the specs. Both of these issues were fixed on the server in https://github.com/expo/universe/pull/10394, but we still need to do batching on the client in order to not get timeouts. 

# How

I think doing 100 at a time is fine. For 3k assets, this took ~10 seconds or so to get all the specifications when I tested locally.

# Test Plan

For an app with 3k assets:
```
export EXPO_STAGING=1
~/expo/eas-cli/packages/eas-cli/bin/run update --branch preview --message "Updating the app"
```
See that it works!!

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -13,6 +13,7 @@ import { PresignedPost } from '../graphql/mutations/UploadSessionMutation';
 import { PublishQuery } from '../graphql/queries/PublishQuery';
 import { uploadWithPresignedPostWithRetryAsync } from '../uploads';
 import { expoCommandAsync } from '../utils/expoCli';
+import chunk from '../utils/expodash/chunk';
 import uniqBy from '../utils/expodash/uniqBy';
 
 export type PublishPlatform = Extract<'android' | 'ios', Platform>;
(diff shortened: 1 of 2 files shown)


wschurman:
/changelog-entry bug-fix Limit the number of SignedAssetUploadSpecifications fetched at a time

packages/eas-cli/src/project/publish.ts
```diff
@@ -285,12 +286,18 @@ export async function uploadAssetsAsync(
...
 
   updateSpinnerText?.(totalAssets, totalAssets);
-
   let missingAssets = await filterOutAssetsThatAlreadyExistAsync(uniqueAssets);
   const uniqueUploadedAssetCount = missingAssets.length;
-  const { specifications } = await PublishMutation.getUploadURLsAsync(
-    missingAssets.map(ma => ma.contentType)
-  );
+
+  const missingAssetChunks = chunk(missingAssets, 100);
```
ide, line 292: How well do we handle 1000?

ide (approved):

dsokal (approved):

packages/eas-cli/src/project/publish.ts
wschurman: We handle 1400 in prod relatively ok (haven't seen failures), 100 is more just setting this to a value that doesn't make this too slow for normal apps (1-2 requests for most apps I assume) while keeping us far away from hitting the request timeout.

ide: OK. My thinking is that we want to find a sweet spot that uses batching efficiently while also not overloading a replica and causing it to run for too long (~1 second would be a high target IMO).

PR #1282, opened 2022-08-16
[eas-cli] Configure EAS project ID in app config in all cases

wschurman:
# Checklist

- [ ] I've added an entry to [CHANGELOG.md](https://github.com/expo/eas-cli/blob/main/CHANGELOG.md#main) if necessary. You can comment this pull request with `/changelog-entry [breaking-change|new-feature|bug-fix|chore] [message]` and CHANGELOG.md will be updated automatically.

# Why

Some background:
- We used to choose which type of manifest to serve in development based on the `extra.eas.projectId` field.
- At one point, code was added to configure the `extra.eas.projectId` field upon running commands that needed the project ID. https://github.com/expo/eas-cli/pull/402
- This caused those commands to change the manifest type used, and the new manifest type wasn't yet ready for prime time.
- https://github.com/expo/eas-cli/pull/412 was added to gate the auto-configuration-write temporarily.
- At some point shortly later, we changed the mechanism used to determine the type of manifest to use to either a flag or looking at the `updates.url` property of the config.
- We forgot to revert https://github.com/expo/eas-cli/pull/412

This PR reverts https://github.com/expo/eas-cli/pull/412.

Closes ENG-6076.

# How

Revert it.

# Test Plan

Run `~/expo/eas-cli/packages/eas-cli/bin/run update:configure` in a new project, ensure `extra.eas.projectId` is added to the app config.

PR #1301, opened 2022-08-22
[eas-cli] Prevent throwing dynamic app config write error when configuring project id

wschurman:
# Checklist

# Why

Blame: https://github.com/expo/eas-cli/pull/1282

- This PR made it so that whenever the project ID is created/fetched from the server, it is put in the app config (app.json). Automatically adding this to a dynamic app config (app.config.js) doesn't work, so an error was thrown.
- While some commands require an explicit project ID (the reason for the blame PR), most do not; instead, most commands just call this method to get the project ID and don't need it to be configured in the manifest for later in their process.
- This caused CI breakage for projects that use dynamic app config and don't synchronize their project ID locally.

In the future, we may require explicitly setting up the project ID before any command can be run, or may make all commands run the setup and throw if they can't set the project ID in the dynamic config case. The benefits of synchronizing the project ID locally are:
    - Less network
    - More explicit behavior. Less prone to error when server is switched, app is transferred to a new account, or account is renamed. This one is important. `owner`/`slug` as a unique identifier is deprecated as it is less reliable.

https://github.com/expo/expo/pull/18744#issuecomment-1222879919

Fixes https://github.com/expo/eas-cli/issues/1296. Fixes 

# How

Separate out the methods, and surround the setting of the field with a try/catch to just warn in the case that it can't be set.

# Test Plan

Run tests.

In local project, `~/expo/eas-cli/packages/eas-cli/bin/run build -p android` (see that it no longer throws)

### DIFF
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -83,19 +83,15 @@ export async function findProjectRootAsync({
   }
 }
 
-export async function setProjectIdAsync(
+/**
+ * Save an EAS project ID to the appropriate field in the app config.
+ */
+export async function saveProjectIdToAppConfigAsync(
   projectDir: string,
+  projectId: string,
   options: { env?: Env } = {}
-): Promise<ExpoConfig | undefined> {
+): Promise<void> {
   const exp = getExpoConfig(projectDir, options);
-
-  const privacy = toAppPrivacy(exp.privacy);
-  const projectId = await ensureProjectExistsAsync({
-    accountName: getProjectAccountName(exp, await ensureLoggedInAsync()),
-    projectName: exp.slug,
-    privacy,
-  });
-
   const result = await modifyConfigAsync(projectDir, {
     extra: { ...exp.extra, eas: { ...exp.extra?.eas, projectId } },
   });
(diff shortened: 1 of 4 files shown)


wschurman:
/changelog-entry bug-fix Prevent throwing dynamic app config write error when configuring project ID

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -140,19 +152,20 @@ export async function getProjectIdAsync(
...
-  const newLocalProjectId = newExp?.extra?.eas?.projectId;
-  if (!newLocalProjectId) {
-    // throw if we still can't locate the projectId
-    throw new Error('Could not retrieve project ID from app.json');
+  const projectId = await fetchProjectIdFromServerAsync(exp);
+
+  try {
+    await saveProjectIdToAppConfigAsync(projectDir, projectId, options);
+  } catch (e: any) {
+    Log.warn(`Failed to save EAS project ID to app config: ${e.message}`);
```
tsapeta, line 165: We should probably say something more about that so people clearly know they have to take an action on that. Also things like what is it used for, is the project ID static/secret, etc. Maybe make a FYI guide here https://github.com/expo/fyi and link to that in the warning?

rlam3: It would also be helpful if this requirement is mentioned in https://docs.expo.dev/versions/latest/config/app/#dynamic-configuration-with-appconfigjs). Maybe an extra column for required or optional properties.

ide: Brent once posted some good guidance on error messages. I forget what he wrote exactly but it was something like: say what's wrong, why it happened, and what to do about it next.

wschurman: Added a comment to the code, but a large set of instructions will have just been printed out by the time this hits so I just changed this to just print whatever is in the warning.

PR #1312, opened 2022-08-24
[eas-cli] warn about EAS outages

szdziedzic:
# Checklist

# Why

Needed for [ENG-4932](https://linear.app/expo/issue/ENG-4932/dealing-with-platform-specific-issues-on-eas).
Warns users about EAS outages in CLI, when using `eas build`, `eas build --auto-submit` and `eas submit` commands.

I will probably also add the same feature for `eas update` after consulting with the team.

# How

Added query to fetch `status.expo.dev` data through our backend and used this data to display warnings.
I check for info about outages (and display warnings when service status != `Operational`) as a first thing when running the commands.

# Test Plan

### without outages
Screenshot 2022-08-25 at 08 10 14

### with EAS Build outage 
Screenshot 2022-08-25 at 08 09 42

### with EAS Build and EAS Submit outage when running with `--auto-submit` flag
Screenshot 2022-08-25 at 08 34 14

### without outages
Screenshot 2022-08-25 at 08 20 18

### with EAS Submit outage
Screenshot 2022-08-25 at 08 21 33

## Edit: updated version of warning communicate looks like this
Screenshot 2022-08-25 at 13 37 50

### DIFF
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -98,23 +98,30 @@ export default class Build extends EasCommand {
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Build);
 
+    if (rawFlags.json) {
+      enableJsonOutput();
+    }
+    const flags = Build.sanitizeFlags(rawFlags);
+
+    const projectDir = await findProjectRootAsync();
+
     await maybeWarnAboutEasOutagesAsync(
-      rawFlags['auto-submit'] || rawFlags['auto-submit-with-profile']
+      flags.autoSubmit
         ? [StatuspageServiceName.EasBuild, StatuspageServiceName.EasSubmit]
         : [StatuspageServiceName.EasBuild]
     );
 
-    if (rawFlags.json) {
-      enableJsonOutput();
+    if (!flags.requestedPlatform) {
+      flags.requestedPlatform = await selectRequestedPlatformAsync();
+      Build.sanitizePlatform(rawFlags, flags.requestedPlatform);
     }
-    const flags = await this.sanitizeFlagsAsync(rawFlags);
 
-    const projectDir = await findProjectRootAsync();
-
-    await runBuildAndSubmitAsync(projectDir, flags);
+    await runBuildAndSubmitAsync(projectDir, flags as BuildFlags);
   }
 
-  private async sanitizeFlagsAsync(flags: RawBuildFlags): Promise<BuildFlags> {
+  private static sanitizeFlags(
+    flags: RawBuildFlags
+  ): Omit<BuildFlags, 'requestedPlatform'> & { requestedPlatform?: RequestedPlatform } {
     const nonInteractive = flags['non-interactive'];
     if (!flags.local && flags.output) {
       Errors.error('--output is allowed only for local builds', { exit: 1 });
@@ -126,20 +133,13 @@ export default class Build extends EasCommand {
       Errors.error('--json is allowed only when building in non-interactive mode', { exit: 1 });
     }
 
-    const requestedPlatform = await selectRequestedPlatformAsync(flags.platform);
-    if (flags.local) {
-      if (flags['auto-submit'] || flags['auto-submit-with-profile'] !== undefined) {
-        // TODO: implement this
-        Errors.error('Auto-submits are not yet supported when building locally', { exit: 1 });
-      }
-
-      if (requestedPlatform === RequestedPlatform.All) {
-        Errors.error('Builds for multiple platforms are not supported with flag --local', {
-          exit: 1,
-        });
-      } else if (process.platform !== 'darwin' && requestedPlatform === RequestedPlatform.Ios) {
-        Errors.error('Unsupported platform, macOS is required to build apps for iOS', { exit: 1 });
-      }
+    const requestedPlatform =
+      flags.platform &&
+      Object.values(RequestedPlatform).includes(flags.platform.toLowerCase() as RequestedPlatform)
+        ? (flags.platform.toLowerCase() as RequestedPlatform)
+        : undefined;
+    if (requestedPlatform) {
+      Build.sanitizePlatform(flags, requestedPlatform);
     }
 
     if (flags['skip-credentials-check']) {
@@ -184,5 +184,22 @@ export default class Build extends EasCommand {
       userInputResourceClass: flags['resource-class'] ?? UserInputResourceClass.DEFAULT,
       message,
     };
+  }
+
+  private static sanitizePlatform(flags: RawBuildFlags, platform: RequestedPlatform): void {
+    if (flags.local) {
+      if (flags['auto-submit'] || flags['auto-submit-with-profile'] !== undefined) {
+        // TODO: implement this
+        Errors.error('Auto-submits are not yet supported when building locally', { exit: 1 });
+      }
+
+      if (platform === RequestedPlatform.All) {
+        Errors.error('Builds for multiple platforms are not supported with flag --local', {
+          exit: 1,
+        });
+      } else if (process.platform !== 'darwin' && platform === RequestedPlatform.Ios) {
+        Errors.error('Unsupported platform, macOS is required to build apps for iOS', { exit: 1 });
+      }
+    }
   }
 }
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -4,10 +4,12 @@ import path from 'path';
 import { BuildFlags, runBuildAndSubmitAsync } from '../../build/runBuildAndSubmit';
 import { UserInputResourceClass } from '../../build/types';
 import EasCommand from '../../commandUtils/EasCommand';
+import { StatuspageServiceName } from '../../graphql/generated';
 import Log from '../../log';
 import { RequestedPlatform, selectRequestedPlatformAsync } from '../../platform';
 import { findProjectRootAsync } from '../../project/projectUtils';
 import { enableJsonOutput } from '../../utils/json';
+import { warnAboutEasOutagesAsync } from '../../utils/statuspageService';
 
 interface RawBuildFlags {
   platform?: string;
@@ -95,6 +97,13 @@ export default class Build extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Build);
+
+    await warnAboutEasOutagesAsync(
+      rawFlags['auto-submit'] || rawFlags['auto-submit-with-profile']
+        ? [StatuspageServiceName.EasBuild, StatuspageServiceName.EasSubmit]
+        : [StatuspageServiceName.EasBuild]
+    );
+
     if (rawFlags.json) {
       enableJsonOutput();
     }
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -113,10 +113,9 @@ export default class Build extends EasCommand {
       );
     }
 
-    flags.requestedPlatform = await selectRequestedPlatformAsync(flags.requestedPlatform);
-    this.validateFlagsWithPlatform(rawFlags, flags.requestedPlatform);
+    const flagsWithPlatform = await this.ensurePlatformSelectedAsync(flags);
 
-    await runBuildAndSubmitAsync(projectDir, flags as BuildFlags);
+    await runBuildAndSubmitAsync(projectDir, flagsWithPlatform);
   }
 
   private sanitizeFlags(
@@ -138,8 +137,6 @@ export default class Build extends EasCommand {
       Object.values(RequestedPlatform).includes(flags.platform.toLowerCase() as RequestedPlatform)
         ? (flags.platform.toLowerCase() as RequestedPlatform)
         : undefined;
-
-    this.validateFlagsWithPlatform(flags, requestedPlatform);
 
     if (flags['skip-credentials-check']) {
       Log.warnDeprecatedFlag(
@@ -185,9 +182,12 @@ export default class Build extends EasCommand {
     };
   }
 
-  private validateFlagsWithPlatform(flags: RawBuildFlags, platform?: RequestedPlatform): void {
-    if (flags.local) {
-      if (flags['auto-submit'] || flags['auto-submit-with-profile'] !== undefined) {
+  private validateFlagsWithPlatform(
+    flags: Omit<BuildFlags, 'requestedPlatform'> & { requestedPlatform?: RequestedPlatform },
+    platform?: RequestedPlatform
+  ): void {
+    if (flags.localBuildOptions.enable) {
+      if (flags.autoSubmit) {
         // TODO: implement this
         Errors.error('Auto-submits are not yet supported when building locally', { exit: 1 });
       }
@@ -200,5 +200,16 @@ export default class Build extends EasCommand {
         Errors.error('Unsupported platform, macOS is required to build apps for iOS', { exit: 1 });
       }
     }
+  }
+
+  private async ensurePlatformSelectedAsync(
+    flags: Omit<BuildFlags, 'requestedPlatform'> & { requestedPlatform?: RequestedPlatform }
+  ): Promise<BuildFlags> {
+    const requestedPlatform = await selectRequestedPlatformAsync(flags.requestedPlatform);
+    this.validateFlagsWithPlatform(flags, requestedPlatform);
+    return {
+      ...flags,
+      requestedPlatform,
+    };
   }
 }
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -101,25 +101,25 @@ export default class Build extends EasCommand {
     if (rawFlags.json) {
       enableJsonOutput();
     }
-    const flags = Build.sanitizeFlags(rawFlags);
+    const flags = this.sanitizeFlags(rawFlags);
 
     const projectDir = await findProjectRootAsync();
 
-    await maybeWarnAboutEasOutagesAsync(
-      flags.autoSubmit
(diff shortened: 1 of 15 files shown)


packages/eas-cli/src/commands/build/index.ts
```diff
@@ -95,6 +97,13 @@ export default class Build extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Build);
+
+    await warnAboutEasOutagesAsync(
```
dsokal, line 101:
```suggestion
    await maybeWarnAboutEasOutageAsync(
```

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,49 @@
...
+} from '../graphql/generated';
+import { StatuspageServiceQuery } from '../graphql/queries/StatuspageServiceQuery';
+import Log from '../log';
+
+export async function warnAboutEasOutagesAsync(
+  serviceNames: StatuspageServiceName[]
+): Promise<void> {
+  const services = await getStatuspageServiceAsync(serviceNames);
+
+  services.forEach(warnAboutServiceOutage);
```
dsokal, line 14:
```suggestion
  for (const service of services) {
    warnAboutServiceOutage(service);
  }
```

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,49 @@
...
+        humanReadableServiceName[service.name]
+      } is currently in a degraded state. The service may temporarily not function as expected.`
+    );
+
+    if (service.incidents.length > 0) {
+      const [currentIncident] = service.incidents;
+      Log.warn(`Reason: ${currentIncident.name}`);
+    }
+
+    Log.warn('Please check https://status.expo.dev/ for more information.');
```
dsokal, line 36: See https://github.com/expo/eas-cli/pull/1250
```suggestion
    Log.warn('Check https://status.expo.dev/ for more information.');
```

dsokal, line 36: @brentvatne review needed :)

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,51 @@
...
+  const humanReadableServiceName: Record<StatuspageServiceName, string> = {
+    [StatuspageServiceName.EasBuild]: 'EAS Build',
+    [StatuspageServiceName.EasSubmit]: 'EAS Submit',
+    [StatuspageServiceName.EasUpdate]: 'EAS Update',
+  };
+
+  if (service.status !== StatuspageServiceStatus.Operational) {
+    Log.warn(
+      `⚠️  ${
+        humanReadableServiceName[service.name]
```
wkozyra95, line 29:
```suggestion
        chalk.bold(humanReadableServiceName[service.name])
```

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,51 @@
...
+      } is currently in a degraded state. The service may temporarily not function as expected.`
+    );
+
+    if (service.incidents.length > 0) {
+      const [currentIncident] = service.incidents;
+      Log.warn(`Reason: ${currentIncident.name}`);
+    }
+
+    Log.warn('Check https://status.expo.dev/ for more information.');
```
wkozyra95, line 38:
```suggestion
    Log.warn(`Check ${link('https://status.expo.dev/')} for more information`);
```

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,51 @@
...
+function warnAboutServiceOutage(service: StatuspageServiceFragment): void {
+  const humanReadableServiceName: Record<StatuspageServiceName, string> = {
+    [StatuspageServiceName.EasBuild]: 'EAS Build',
+    [StatuspageServiceName.EasSubmit]: 'EAS Submit',
+    [StatuspageServiceName.EasUpdate]: 'EAS Update',
+  };
+
+  if (service.status !== StatuspageServiceStatus.Operational) {
+    Log.warn(
+      `⚠️  ${
```
wkozyra95, line 28: it's fine to inline icons this way, but try using the `figures` package, it might provide better fallbacks on systems where icons are not available.

wkozyra95 (approved):

packages/eas-cli/src/utils/statuspageService.ts
szdziedzic: Screenshot 2022-08-25 at 13 37 50
This is how it looks like after the update

dsokal (approved):

packages/eas-cli/src/utils/statuspageService.ts
wkozyra95: the previous icon was definitely looking better on macos, on my system previous version and the current one look the same. I'm not sure what would happen if icon fonts are missing.

I do not have any strong opinion on which approach to take.

szdziedzic: I think that the one using the `figures` package is definitely safer because it has a fallback which is `!!`, so I would say that we can go with it, although maybe @brentvatne has a different opinion 🤔

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -95,6 +97,13 @@ export default class Build extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Build);
+
+    await maybeWarnAboutEasOutagesAsync(
```
wschurman, line 101: move this after L112. This has a the benefit of not hitting our API if the flags/project root dir are incorrect and the command will fail anyways.

packages/eas-cli/src/commands/submit.ts
```diff
@@ -91,6 +92,9 @@ export default class Submit extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Submit);
+
+    await maybeWarnAboutEasOutagesAsync([StatuspageServiceName.EasSubmit]);
```
wschurman, line 96: Same comment, move this after L105.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -228,6 +230,9 @@ export default class UpdatePublish extends EasCommand {
         'non-interactive': nonInteractive,
       },
     } = await this.parse(UpdatePublish);
+
+    await maybeWarnAboutEasOutagesAsync([StatuspageServiceName.EasUpdate]);
```
wschurman, line 234: Same, move after L249

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,53 @@
...
+): Promise<void> {
+  const services = await getStatuspageServiceAsync(serviceNames);
+
+  for (const service of services) {
+    warnAboutServiceOutage(service);
+  }
+}
+
+function warnAboutServiceOutage(service: StatuspageServiceFragment): void {
+  const humanReadableServiceName: Record<StatuspageServiceName, string> = {
```
wschurman, line 22: can move this into module scope to avoid needing to create it each time the function is called

wschurman (approved):
Just a few minor comments, looks good to me otherwise though.

packages/eas-cli/src/graphql/queries/StatuspageServiceQuery.ts
```diff
@@ -0,0 +1,41 @@
...
+          `,
+          { serviceNames },
+          {
+            additionalTypenames: ['StatuspageService', 'StatuspageIncident'],
+          }
+        )
+        .toPromise()
+    );
+
+    return data.statuspageService.byServiceNames;
```
kgc00, line 39: does this not need `data?.`

kgc00 (approved):

packages/eas-cli/src/commands/build/index.ts
szdziedzic: I agree that this improvement makes sense, although my only doubt is that in such a case this warning won't be the first thing that the user sees on the screen (because when `platform` flag is empty we ask the user about it). Example:

Screenshot 2022-08-26 at 06 16 10

I will consult it with @brentvatne

brentvatne: ideally we can both run this **after** flags/project root dir validation and **before** we show any output or prompts. if that looks like it might take a bit too much work for this PR, then let's move it as @wschurman suggests and create a task on linear to follow up on that

packages/eas-cli/src/utils/statuspageService.ts
brentvatne:
```suggestion
    Log.warn('All information on service status and incidents available at https://status.expo.dev/');
```

packages/eas-cli/src/utils/statuspageService.ts
brentvatne: i think we can leave out the icon and just show the text :) the color is a strong enough attention draw

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,49 @@
...
+    [StatuspageServiceName.EasBuild]: 'EAS Build',
+    [StatuspageServiceName.EasSubmit]: 'EAS Submit',
+    [StatuspageServiceName.EasUpdate]: 'EAS Update',
+  };
+
+  if (service.status !== StatuspageServiceStatus.Operational) {
+    Log.warn(
+      `⚠️  ${
```
brentvatne:
```suggestion
      `Service status warning: ${
```

packages/eas-cli/src/graphql/queries/StatuspageServiceQuery.ts
szdziedzic:
```
export async function withErrorHandlingAsync<T>(promise: Promise<OperationResult<T>>): Promise<T> {
  const { data, error } = await promise;

  if (error) {
    if (error.graphQLErrors.some(e => e?.extensions?.isTransient)) {
      Log.error(`We've encountered a transient error. Try again shortly.`);
    }
    throw error;
  }

  // Check for malformed response. This only checks the root query existence,
  // It doesn't affect returning responses with empty resultset.
  if (!data) {
    throw new Error('Returned query result data is null!');
  }

  return data;
}
```
I think that it doesn't because of how `withErrorHandlingAsync` works

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,49 @@
...
+        humanReadableServiceName[service.name]
+      } is currently in a degraded state. The service may temporarily not function as expected.`
```
brentvatne:
```suggestion
      } may temporarily not function as expected.`
```

ideally we'd be able to say what kind of outage is reported - partial or major. so, something like

> "is experiencing a partial outage."

or

> "is experiencing a major outage."

brentvatne (commented):
😎 this is going to be very helpful

packages/eas-cli/src/utils/statuspageService.ts
szdziedzic: That sounds like a good idea, thank you!

szdziedzic: Screenshot 2022-08-26 at 10 38 47
This is how the communicate looks after all of the changes

packages/eas-cli/src/commands/build/index.ts
szdziedzic: I think I was able to successfully refactor the code so we can do flags/project root dir validation first, then we show outage warning and finally we ask for inputs 🙂 Please, let me know what you think about this change.

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const humanReadableServiceName: Record<StatuspageServiceName, string> = {
+  [StatuspageServiceName.EasBuild]: 'EAS Build',
+  [StatuspageServiceName.EasSubmit]: 'EAS Submit',
+  [StatuspageServiceName.EasUpdate]: 'EAS Update',
+};
+
+function warnAboutServiceOutage(service: StatuspageServiceFragment): void {
+  if (service.status !== StatuspageServiceStatus.Operational) {
```
dsokal, line 28: Invert to get rid of one level of indentation.
```suggestion
  if (service.status === StatuspageServiceStatus.Operational) {
    return;
  }
```

packages/eas-cli/src/commands/submit.ts
```diff
@@ -91,12 +92,20 @@ export default class Submit extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags: rawFlags } = await this.parse(Submit);
-    const flags = await this.sanitizeFlagsAsync(rawFlags);
+
+    const flags = await Submit.sanitizeFlags(rawFlags);
```
dsokal, line 96: I think this is not an async function:
```suggestion
    const flags = Submit.sanitizeFlags(rawFlags);
```

packages/eas-cli/src/commands/submit.ts
```diff
@@ -174,4 +184,23 @@ export default class Submit extends EasCommand {
       nonInteractive,
     };
   }
+
+  private static sanitizePlatform(
```
dsokal, line 188: If you don't care about the result of a function like this, it should be named `validate...`:
```suggestion
  private static validatePlatform(
```

dsokal:
```
// validate only
validateXyz(xyz);

// sanitize
const xyz = sanitizeXyz(rawXyz);
```

dsokal: Also, a nit - if I was writing this I'd drop `static`. I think there isn't any benefit from the keyword here. The method is only used in non-static methods.

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -176,4 +185,21 @@ export default class Build extends EasCommand {
       message,
     };
   }
+
+  private static sanitizePlatform(flags: RawBuildFlags, platform: RequestedPlatform): void {
```
dsokal, line 189:
```suggestion
  private validatePlatform(flags: RawBuildFlags, platform: RequestedPlatform): void {
```

dsokal (approved):

packages/eas-cli/src/commands/build/index.ts
wkozyra95: I think both names are incorrect this function does not validate if selected platform is correct, maybe `validateFlagsWithPlatform`?

Personally, I would create the function ensurePlatformSelected that receives BuildFlags object with optional requestedPlatform and returns BuildFlags.

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -95,17 +97,31 @@ export default class Build extends EasCommand {
...
+    await maybeWarnAboutEasOutagesAsync(
+      flags.autoSubmit
+        ? [StatuspageServiceName.EasBuild, StatuspageServiceName.EasSubmit]
+        : [StatuspageServiceName.EasBuild]
+    );
+
+    if (!flags.requestedPlatform) {
+      flags.requestedPlatform = await selectRequestedPlatformAsync();
+      Build.sanitizePlatform(rawFlags, flags.requestedPlatform);
+    }
```
wkozyra95, line 117:
```suggestion
    flags.requestedPlatform = await selectRequestedPlatformAsync(flags.requestedPlatform);
    Build.sanitizePlatform(rawFlags, flags.requestedPlatform);
```

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -117,20 +133,13 @@ export default class Build extends EasCommand {
...
-      } else if (process.platform !== 'darwin' && requestedPlatform === RequestedPlatform.Ios) {
-        Errors.error('Unsupported platform, macOS is required to build apps for iOS', { exit: 1 });
-      }
+    const requestedPlatform =
+      flags.platform &&
+      Object.values(RequestedPlatform).includes(flags.platform.toLowerCase() as RequestedPlatform)
+        ? (flags.platform.toLowerCase() as RequestedPlatform)
+        : undefined;
+    if (requestedPlatform) {
+      Build.sanitizePlatform(flags, requestedPlatform);
```
wkozyra95, line 142: This is not necessary here, that validation can happen inside `sanitizePlatform` function even if the platform is passed as a param.

Also sanitizePlatform currently only validates flags for local builds, and in that case we probably do not need to show outages.

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function warnAboutServiceOutage(service: StatuspageServiceFragment): void {
+  if (service.status === StatuspageServiceStatus.Operational) {
+    return;
+  }
+
+  const outageType = service.status === StatuspageServiceStatus.MajorOutage ? 'major' : 'partial';
+
+  Log.warn(
+    `Service status warning: ${chalk.bold(
```
brentvatne, line 35:
```suggestion
    `${chalk.bold(
```

changed my mind on this after seeing the screenshot 😅

packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -0,0 +1,61 @@
...
+      humanReadableServiceName[service.name]
+    )} is experiencing a ${outageType} outage. The service may temporarily not function as expected.`
```
brentvatne, line 37: i'd put the "is experiencing a ${outageType} outage" part inside of the bold.

also, i think we can just remove the other text now that we let people know what kind of outage is occurring:

```suggestion
    )} is experiencing a ${outageType} outage.`
```

brentvatne (approved):

szdziedzic:
/changelog-entry new-feature Warn about EAS outages when running `build`, `submit`, and `update` commands

dsokal (approved):

PR #1323, opened 2022-08-31
[eas-cli] use paginated query API for update commands

kgc00:
# Checklist

# Why

Introduces paginated queries for fetching data from our backend for the `update` commands. Also refactors some behavior in `branch` commands that touch on paginated queries. 

# How

 - replaced usages of raw queries with paginated queries where it makes sense
 - replaced prompts for users to type input into the CLI with single select prompts on paginated query results to improve UX
- added some edge-case handling for `getPlatformsForGroup`

# Test Plan

Tested locally:
- ran each affected command with / without `offset`, `limit`, `non-interactive`, `json`, and required args (like`branch`)
- added some automated tests to ensure 
  - paginatedQueryOptions are handled by `selectUpdateGroupOnBranchAsync` (mostly the pagination logic should already have tests at `packages/eas-cli/src/utils/__tests__/queries-test.ts`)
  - `getPlatformsForGroup` can handled undefined values (which can sneak in due to improper GQL typing) #1326
  
Please lmk if ya'll have suggestions for additional tests. Would be happy to add some more.


packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -6,17 +6,40 @@ import {
...
+export type UpdateByGroupObject = ViewUpdatesByGroupQuery['updatesByGroup'];
+
 export type AppUpdateObject = ViewAllUpdatesQuery['app']['byId']['updates'][number];
 
 export type UpdateObject = BranchUpdateObject | AppUpdateObject;
 
+export type UpdateGroupObject =
+  | UpdateByGroupObject
+  | AppUpdateGroupObject
+  | BranchUpdateGroupObject;
```
kgc00, line 41: planning to update these to use fragments in a follow up PR. seems a better DX.

packages/eas-cli/src/update/utils.ts
```diff
@@ -19,40 +21,46 @@ export type UpdateGroupDescription = FormatUpdateParameter & {
...
 
-export function getUpdateGroupsWithPlatforms<
-  UpdateFragment extends Pick<Update, 'platform' | 'group'>
->(updates: UpdateFragment[]): (UpdateFragment & { platforms: string })[] {
-  return Object.values(groupBy(updates, updates => updates.group)).map(updateGroup => ({
-    ...updateGroup[0],
-    platforms: formatPlatformForUpdateGroup(updateGroup),
-  }));
-}
+export const UPDATE_COLUMNS_WITH_BRANCH = ['Branch', ...UPDATE_COLUMNS];
```
kgc00, line 42: not sure if folks prefer a different pattern here. I think standardizing our column/table layout is worthwhile and went with what seemed like the simplest option of multiple types.

packages/eas-cli/src/commandUtils/pagination.ts
```diff
@@ -27,19 +27,27 @@ const parseFlagInputStringAsInteger = (
...
+}: {
+  defaultTo: number;
+  limit: number;
+}): any =>
+  Flags.integer({
+    description: `The number of items to fetch each query. Defaults to ${defaultTo} and is capped at ${limit}. Using a lower value may help increase command speed.`,
+    // eslint-disable-next-line async-protect/async-suffix
+    parse: async input => parseFlagInputStringAsInteger(input, 'limit', 1, limit),
+  });
+
```
kgc00, line 42: turns out with the introduction of update groups we now have different limits per query object. branches/updates are capped at 100, update groups can be 2x the size, so they're capped at 50. 

I added a configurable limit flag here to match that. Alternatively we could:
- raise/lower the entity limit on www for all types
- match all limits here on CLI to that

packages/eas-cli/src/branch/queries.ts
```diff
@@ -14,24 +14,29 @@ import {
 
 export const BRANCHES_LIMIT = 50;
 
-export async function selectBranchFromPaginatedQueryAsync(
-  projectId: string,
-  promptTitle: string,
-  options: PaginatedQueryOptions
-): Promise<UpdateBranchFragment> {
-  if (options.nonInteractive) {
+export async function selectBranchOnAppAsync({
```
kgc00, line 17: i tried to standardize the names and format of these queries. since I'm coming back to work on this after some time away and I'm adding in more use-cases it is a little easier to understand good conventions. 

- the names are shortened to include the action (select/list) followed by the data source (on app / on branch)
- the arguments are now an object. 
  - this lends itself to many required args
  - it's easier to remove/add new arguments without unintentional breaking other uses
  - it's more explicit

packages/eas-cli/src/commandUtils/pagination.ts
```diff
@@ -27,19 +27,27 @@ const parseFlagInputStringAsInteger = (
...
 
+export const getLimitFlagWithCustomValues = ({
+  defaultTo,
+  limit,
+}: {
+  defaultTo: number;
+  limit: number;
+}): any =>
+  Flags.integer({
+    description: `The number of items to fetch each query. Defaults to ${defaultTo} and is capped at ${limit}. Using a lower value may help increase command speed.`,
```
wkozyra95, line 38: Consider dropping the last sentence, it's a bit long for a help description, plus the default limit should always be small enough that time improvements would be negligible.

packages/eas-cli/src/commands/branch/delete.ts
```diff
@@ -87,17 +88,18 @@ export default class BranchDelete extends EasCommand {
...
 
   async runAsync(): Promise<void> {
     let {
-      args: { name },
-      flags: { json: jsonFlag },
+      args: { name: branchName },
+      flags,
     } = await this.parse(BranchDelete);
+    const paginatedQueryOptions = getPaginatedQueryOptions(flags);
+    const { json: jsonFlag, 'non-interactive': nonInteractiveFlag } = flags;
```
wkozyra95, line 102: if both flags exist in other places `--json` can only be passed if `--non-interactive` is already set.

It would be good to extract it to some common util, e.g. you could check that in getPaginatedQueryOptions (you should rename it to sth more appropriate in that case)

packages/eas-cli/src/commands/branch/list.ts
```diff
@@ -14,15 +14,15 @@ export default class BranchList extends EasCommand {
 
   async runAsync(): Promise<void> {
     const { flags } = await this.parse(BranchList);
-    const options = getPaginatedQueryOptions(flags);
+    const paginatedQueryOptions = getPaginatedQueryOptions(flags);
```
wkozyra95, line 17: the same as above, both `--json` and `--non-interactive` flags

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -208,28 +169,27 @@ export default class UpdatePublish extends EasCommand {
       description: `File containing the PEM-encoded private key corresponding to the certificate in expo-updates' configuration. Defaults to a file named "private-key.pem" in the certificate's directory.`,
       required: false,
     }),
-    'non-interactive': Flags.boolean({
-      default: false,
-      description: 'Run command in non-interactive mode',
-    }),
+    // do not expose limit/offset since they don't make sense on this command
```
wkozyra95, line 172: that comment is redundant, a lot fo commands do not have pagination and are non-interactive or can provide json output. Those options are not related to each other. Maybe consider rethinking how EasPaginatedQueryFlags looks and separating those.

packages/eas-cli/src/commands/branch/delete.ts
wkozyra95: The same applies to other commands from this PR and some form the previous one.

wkozyra95: Also you are using inconsistent naming, sometimes it's json/nonInteractive and sometimes jsonFlag/nonInteractiveFlag

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
+      },
+    });
+
+    if (!updateGroup || updateGroup.length === 0) {
+      throw new Error(`Could not find any branches for project "${projectId}"`);
+    }
+
+    return updateGroup;
+  }
+}
```
dsokal, line 120: You can consider this as a nitpick but it seems those 3 could be refactored/simplified. They look mostly the same. You could create another, generic function and use it everywhere.

packages/eas-cli/src/update/utils.ts
```diff
@@ -19,40 +21,46 @@ export type UpdateGroupDescription = FormatUpdateParameter & {
...
 
 export function getPlatformsForGroup({
-  group,
-  updates,
+  group = '',
+  updates = [],
```
dsokal, line 46: Can you explain the defaults? They look odd and unnecessary.

kgc00: sure. What I'm thinking here is that functionally we want to have `getPlatformsForGroup` return `N/A` when handling null/undefined/invalid input. It also seems to make sense to have `formatPlatformForUpdateGroup` always be what returns the string.

So with that in mind, we'll either need to default values here or default them in the function body. `groupBy` cannot operate on `undefined`, it needs some kind of array. And we cannot index a record on `undefined`, we need some kind of string.

This seemed to me the simplest option but happy to change it.

packages/eas-cli/src/commands/branch/delete.ts
kgc00: hey! so, lmk if I'm misunderstanding your comment:

> if both flags exist in other places --json can only be passed if --non-interactive is already set.

Are you saying that we want to explicitly not support json output for interactive commands? In that case, I think we should be making use of the `dependsOn` property that ocliff flags provides.

packages/eas-cli/src/update/queries.ts
kgc00: happy to get into refactoring these methods, but if it's alright with you I'd prefer to do it in a follow-up once we've got pagination in all our commands.

packages/eas-cli/src/update/utils.ts
dsokal: Sorry for not being clear. I mostly meant `group = ''`. This particular default looks strange.
Why? Let's say you have passed `group = undefined`. As a result, line 52 will try to access `groupedUpdates['']`. It's obviously `undefined`. `formatPlatformForUpdateGroup` support `undefined` but in my opinion, this is quite a weird flow. I mean you're treating `''` as a falsy value here. It's better to use more meaningful falsy values like `null`  or `undefined`, and them properly.

packages/eas-cli/src/update/utils.ts
```diff
@@ -19,40 +21,46 @@ export type UpdateGroupDescription = FormatUpdateParameter & {
...
 }: {
-  group: string;
-  updates: { group: string; platform: string }[];
+  group: string | undefined;
+  updates: { group: string; platform: string }[] | undefined;
 }): string {
   const groupedUpdates = groupBy(updates, update => update.group);
   return formatPlatformForUpdateGroup(groupedUpdates[group]);
```
dsokal, line 52:
```suggestion
  return group && formatPlatformForUpdateGroup(groupedUpdates[group]);
```
+ drop `= ''` default from `group`.

packages/eas-cli/src/update/utils.ts
dsokal: Also see the comment below.

dsokal (commented):
LGTM but leaving the decision on approving the PR to @wschurman

packages/eas-cli/src/update/queries.ts
dsokal: Sounds good.

packages/eas-cli/src/commands/branch/delete.ts
wkozyra95: > Are you saying that we want to disallow json output for commands run in interactive mode?

Yes, that is how all commands with both nonInteractive and json flags worked before.

If there is any prompt in json only output it would not work correctly. Another alternative would be to implicitly set nonInteractive if json flag is present, I'm fine with this approach too, but you would need to also change all the existing commands that have both of those flags.

> Sure. So my intention was that every line of code I've touched now uses the xFlag convention

In more complicated commands we usually build some context/options object based on flag and other input. This convention might be problematic because you can name sth nonInteractiveFlag when operating inside top-level command code, but for example inside some util function or as part of the context object this name would not be correct and you would have to use `nonInteractive`. As a result, you will have a bunch of `nonInteractive: nonInteractiveFlag` in your code. If this separation is intentional, I'm fine with it.

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -0,0 +1,19 @@
...
+      dependsOn: ['non-interactive'],
+    }),
+    'non-interactive': Flags.boolean({
+      description: 'Run the command in non-interactive mode.',
+    }),
+  },
+  jsonOnly: {
+    json: Flags.boolean({
+      description: 'Enable JSON output, non-JSON messages will be printed to stderr.',
+      dependsOn: ['non-interactive'],
```
wkozyra95, line 16: If jsonOnly is used there is no non-interactive flag to set.

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -0,0 +1,19 @@
+import { Flags } from '@oclif/core';
+
+export const EasNonInteractiveFlags = {
+  nonInteractive: {
+    json: Flags.boolean({
+      description: 'Enable JSON output, non-JSON messages will be printed to stderr.',
+      dependsOn: ['non-interactive'],
```
wkozyra95, line 7: dependsOn requires that other flag is set, but it does not require to be set to a specific value, Are you sure that it's not possible to set `--non-interactive=false` or sth like that?

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -0,0 +1,19 @@
+import { Flags } from '@oclif/core';
+
+export const EasNonInteractiveFlags = {
+  nonInteractive: {
```
wkozyra95, line 4: The naming here is not really clear, it suggests that it's about the noninteractive mode, but only one option has that param.

Maybe st like this?
```
export const EasJsonOutputFlags = {
  withNonInteractive: {
  ...
  jsonOnly: {
```

or maybe drop that wrapper object and just have separate values
```
export const EasNonInteractiveFlag ...
export const EasJsonOnlyFlag...
export const EasNonInteractiveAndJsonFlags...
```

wkozyra95 (approved):

packages/eas-cli/src/commandUtils/flags.ts
kgc00: sgtm

packages/eas-cli/src/commandUtils/flags.ts
kgc00: haha, must've slipped past me. nice catch!

packages/eas-cli/src/update/utils.ts
kgc00: we want this function to return `N/A` if input is invalid, so `group && ...` wouldn't work, but i inlined a ternary. this should resolve the issue.

packages/eas-cli/src/branch/queries.ts
```diff
@@ -89,21 +97,16 @@ function renderPageOfBranches(
...
+      ...currentPage.map(branch => [
+        branch.name,
+        formatUpdateMessage(branch.updates[0]),
+        branch.updates[0]?.runtimeVersion ?? 'N/A',
+        branch.updates[0]?.group ?? 'N/A',
+        getPlatformsForGroup({
+          group: branch.updates[0]?.group,
+          updates: branch.updates,
+        }),
+      ])
```
wschurman, line 109: Just confirming, this is a paginated table of branches that shows the branch name and info about most recent published update on the branch? I guess using `branch:view` is the way to see rest of the updates on the branch.

Also I'm assuming we have a limit how many updates we query per branch so technically the platforms column could be incorrect, but I don't think this will happen in most cases.

Either way, I think this is a lot cleaner if we hoist all the undefined handling out of the functions:

```suggestion
      ...currentPage.map(branch => {
        if (branch.updates.length === 0) {
          return [branch.name, 'N/A', 'N/A', 'N/A', 'N/A'];
        }

        const latestUpdateOnBranch = branch.updates[0];
        const latestUpdateGroup = latestUpdateOnBranch.group;
        const allUpdatesInGroup = branch.updates.filter(u => u.group === latestUpdateGroup);
        return [
          branch.name,
          formatUpdateMessage(latestUpdateOnBranch),
          latestUpdateOnBranch.runtimeVersion,
          latestUpdateOnBranch.group,
          getPlatformsForGroup(allUpdatesInGroup),
        ];
      })
```

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
-): Promise<void> {
-  if (options.nonInteractive) {
-    const updates = await queryUpdateGroupsForBranchAsync(
-      options.limit ?? UPDATES_LIMIT,
-      options.offset,
-      projectId,
-      branchName
-    );
-    renderUpdateGroups(updates, options, branchName);
+export async function listAndRenderUpdatesGroupsOnAppAsync({
```
wschurman, line 29:
```suggestion
export async function listAndRenderUpdateGroupsOnAppAsync({
```

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
+      offset: paginatedQueryOptions.offset,
       queryToPerform: (limit, offset) =>
-        queryUpdateGroupsForBranchAsync(limit, offset, projectId, branchName),
+        queryUpdateGroupsOnAppAsync({ limit, offset, appId: projectId }),
       promptOptions: {
-        title: 'Load more branches?',
-        renderListItems: updates => renderUpdateGroups(updates, options, branchName),
+        title: 'Load more update groups?',
+        renderListItems: updateGroup =>
+          renderUpdateGroupsAsTableWithBranchColumn(updateGroup, paginatedQueryOptions),
```
wschurman, line 52:
```suggestion
        renderListItems: updateGroups =>
          renderUpdateGroupsAsTableWithBranchColumn(updateGroups, paginatedQueryOptions),
```

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
+    renderUpdateGroupsAsTable(updateGroups, paginatedQueryOptions);
+  } else {
+    await paginatedQueryWithConfirmPromptAsync({
+      limit: paginatedQueryOptions.limit ?? UPDATE_GROUPS_LIMIT,
+      offset: paginatedQueryOptions.offset,
+      queryToPerform: (limit, offset) =>
+        queryUpdateGroupsOnBranchAsync({ limit, offset, appId: projectId, branchName }),
+      promptOptions: {
+        title: 'Load more update groups?',
+        renderListItems: updates => renderUpdateGroupsAsTable(updates, paginatedQueryOptions),
```
wschurman, line 83:
```suggestion
        renderListItems: updateGroups => renderUpdateGroupsAsTable(updateGroups, paginatedQueryOptions),
```

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -112,4 +135,142 @@ export const UpdateQuery = {
...
+          {
+            appId,
+            limit,
+            offset,
+            branchName,
+            filter,
+          },
+          { additionalTypenames: ['Update'] }
+        )
+        .toPromise()
```
wschurman, line 226: It looks like in other queries where the target data is nested it's common to dig in here rather than at the callsite. For example, see `EnvironmentSecretsQuery.byAppIdAsync`. If you still want this to throw when the app ID returns nothing that can also be done in this method.

This would make the result type more closely match the what the name of the function indicates is returned.

(ditto with function below)

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
+  args: ViewUpdateGroupsOnBranchForAppQueryVariables
+): Promise<BranchUpdateGroupObject[]> {
+  const { app } = await UpdateQuery.viewUpdateGroupsOnBranchAsync(args);
+
+  const { updateBranchByName } = app.byId;
+  if (!updateBranchByName) {
+    throw new Error(`Could not find branch "${args.branchName}"`);
+  }
+
+  return updateBranchByName.updateGroups;
```
wschurman, line 132: see other comment about moving this logic into `viewUpdateGroupsOnBranchAsync`.

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
 
-  const updateBranch = app?.byId.updateBranchByName;
-  if (!updateBranch) {
-    throw new Error(`Could not find branch "${branchName}"`);
+  if (!app) {
+    throw new Error(`Could not find app with id "${appId}"`);
   }
 
-  return updateBranch.updates;
+  return app.byId.updateGroups;
```
wschurman, line 154: ditto about moving this into `viewUpdateGroupsAsync`.

packages/eas-cli/src/update/queries.ts
```diff
@@ -93,3 +189,33 @@ function renderUpdateGroups(
...
+
+  if (json) {
+    printJsonOnlyOutput({ ...branch, currentPage: updateGroupDescriptions });
+  } else {
+    const updateGroupsTable = new CliTable({
+      head: UPDATE_COLUMNS_WITH_BRANCH,
+      wordWrap: true,
+    });
+
+    updateGroupDescriptions.forEach(({ branch, message, runtimeVersion, group, platforms }) => {
```
wschurman, line 212: this should get the branch from each group?

packages/eas-cli/src/update/queries.ts
```diff
@@ -93,3 +189,33 @@ function renderUpdateGroups(
...
 }
+
+function renderUpdateGroupsAsTableWithBranchColumn(
+  updateGroups: (BranchUpdateGroupObject | AppUpdateGroupObject)[],
+  { json }: PaginatedQueryOptions
+): void {
+  const branch = {
+    name: updateGroups[0]?.[0]?.branch.name ?? 'N/A',
+    id: updateGroups[0]?.[0]?.branch.id ?? 'N/A',
+  };
```
wschurman, line 200: remove?

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
 
-  return updateBranch.updates;
+  return app.byId.updateGroups;
 }
 
-function renderUpdateGroups(
-  updates: BranchUpdateObject[],
-  { json }: PaginatedQueryOptions,
-  branchName: string
+function renderUpdateGroupsAsTable(
```
wschurman, line 157: add a docblock (or runtime assertion might be even better) that checks that all updateGroups passed into this are for the same branch

packages/eas-cli/src/update/queries.ts
```diff
@@ -2,84 +2,180 @@ import chalk from 'chalk';
...
+  branchName,
+  paginatedQueryOptions: painatedQueryOptions,
+}: {
+  projectId: string;
+  branchName: string;
+  paginatedQueryOptions: PaginatedQueryOptions;
+}): Promise<BranchUpdateGroupObject> {
+  if (painatedQueryOptions.nonInteractive) {
+    throw new Error('Unable to select an update in non-interactive mode.');
+  } else {
```
wschurman, line 100: no need for the else (can just move all the code in the else out of the clause since the if throws)

packages/eas-cli/src/commands/branch/delete.ts
kgc00: re `xFlag`: my motivation was aid readability by making the variable name more explicit. In the scenario you're talking about where you pass some collection of flags to a utility method (`commands/build/index.ensurePlatformSelected`) i think it's already pretty explicit that your dealing with flags because there is a `flag.X` syntax going on.

I don't have a strong preference for one way or the other, so if you prefer `x` over `xFlag` for top-level command code I'm happy to stick with that. When I started working on the code-base there was a mix of both, so I just chose one approach and stuck with it.

packages/eas-cli/src/update/utils.ts
```diff
@@ -19,40 +21,46 @@ export type UpdateGroupDescription = FormatUpdateParameter & {
...
 
 export function getPlatformsForGroup({
   group,
-  updates,
+  updates = [],
 }: {
-  group: string;
-  updates: { group: string; platform: string }[];
+  group: string | undefined;
+  updates: { group: string; platform: string }[] | undefined;
```
wschurman, line 49: once other comment is addressed, I think it makes the most sense to change the signature of this function to be `getPlatformsForGroup(updateGroup: { group: string; platform: string }[]): string` and then remove the groupBy and select (hoist those to callsites).

(we may not even need this function anymore and can just use `formatPlatformForUpdateGroup` which should probably be called `formatPlatformsStringForUpdateGroup`)

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -112,4 +135,142 @@ export const UpdateQuery = {
...
+                byId(appId: $appId) {
+                  id
+                  updateGroups(limit: $limit, offset: $offset, filter: $filter) {
+                    id
+                    group
+                    message
+                    createdAt
+                    runtimeVersion
+                    platform
+                    manifestFragment
```
wschurman, line 255: It doesn't look like this is used on this type, but if you want to keep querying it so that the shape of this data is the same as the other queries that's ok. Might be worth looking into a named fragment though for simplicity.

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
```diff
@@ -112,4 +135,142 @@ export const UpdateQuery = {
...
+            query ViewUpdatesByGroup($groupId: ID!) {
+              updatesByGroup(group: $groupId) {
+                id
+                group
+                runtimeVersion
+                platform
+                message
+                createdAt
+                manifestFragment
+                actor {
```
wschurman, line 152: it looks like `getActorDisplayName` requires `__typename`, so we may want to explicitly query that here and update the type of `getActorDisplayName` to match. I think overall there's a lot of work that can be done around types though, so if we want to punt on that to a follow-up PR we can.

wschurman (changes requested):

packages/eas-cli/src/branch/queries.ts
kgc00: > Just confirming, this is a paginated table of branches that shows the branch name and info about most recent published update on the branch? I guess using branch:view is the way to see rest of the updates on the branch.

You got it, man.

> Also I'm assuming we have a limit how many updates we query per branch so technically the platforms column could be incorrect, but I don't think this will happen in most cases.

Yeah, this is a good point to consider. In `graphql/types/UpdateBranch.ts` we define `UpdateBranchFragment` getting the branch name, id, and the first 10 updates on each branch. So the platforms column will be correct in all cases. Additionally, `UpdateBranchFragment`  is only used in this query (`listBranchesAsync`/`listAndRenderBranchesOnAppAsync`).

Taking a closer look here, I think we should update the fragment to grab the first update group, instead of raw updates. There shouldn't really be breaking changes anyway, since it's only used in the `branches:list`.

packages/eas-cli/src/update/queries.ts
kgc00: cool

kgc00: branch without updates:
image

branch with updates:
image

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
kgc00: 👍 yeah, i mentioned above that i'm planning to follow this PR with one to replace update/updategroup schemas with fragments to keep things standardized.

packages/eas-cli/src/update/queries.ts
kgc00: i think this is resolved?

packages/eas-cli/src/graphql/queries/UpdateQuery.ts
kgc00: yeah! I mentioned above I'm planning to follow this up with a PR to introduce fragments etc
https://linear.app/expo/issue/ENG-6202/refactor-improve-gql-typing-around-updates

packages/eas-cli/src/update/utils.ts
kgc00: This sounds great, but I'm worried about exploding this PR ins cope. This method is also used by `channel:view` and `branch:list` which work with the `UpdateBranchFragment` that fetches updates not update groups.

https://linear.app/expo/issue/ENG-6203/refactor-improve-updatebranchfragment-and-surrounding-code

wschurman (approved):

packages/eas-cli/src/update/queries.ts
wschurman: Ah yep, it's just a bit jarring that the inner branch destructured variable shadows the branch variable in the outer scope.

kgc00:
/changelog-entry new-feature Add pagination and interactivity to update commands

packages/eas-cli/src/commandUtils/flags.ts
wkozyra95: if it's not possible to set it to false, then I think dependsOn is fine

PR #1326, opened 2022-08-31
Fix viewing branch list when a branch has no associated platforms

hbiede:
# Checklist

# Why

Branch list would error when the following command was performed: `eas branch:create "test" && eas branch:list`

# How

Fix error where platform-less branches would throw a type error

# Test Plan

Run `eas branch:create "test" && eas branch:list` with `v1.2.0`, see an error, run again with this branch and see it succeed

### DIFF
--- a/packages/eas-cli/src/branch/queries.ts
+++ b/packages/eas-cli/src/branch/queries.ts
@@ -89,16 +89,21 @@ function renderPageOfBranches(
     });
 
     table.push(
-      ...currentPage.map(branch => [
-        branch.name,
-        formatUpdate(branch.updates[0]),
-        branch.updates[0]?.runtimeVersion ?? 'N/A',
-        branch.updates[0]?.group ?? 'N/A',
-        getPlatformsForGroup({
-          updates: branch.updates,
-          group: branch.updates[0]?.group ?? [],
-        }),
-      ])
+      ...currentPage.map(branch => {
+        const update: UpdateBranchFragment['updates'][0] | undefined = branch.updates[0];
+        return [
+          branch.name,
+          formatUpdate(update),
+          update?.runtimeVersion ?? 'N/A',
+          update?.group ?? 'N/A',
+          update?.group
+            ? getPlatformsForGroup({
+                updates: branch.updates,
+                group: update?.group ?? '',
+              })
+            : 'N/A',
+        ];
+      })
     );
 
     Log.addNewLineIfNone();
--- a/packages/eas-cli/src/branch/queries.ts
+++ b/packages/eas-cli/src/branch/queries.ts
@@ -96,7 +96,7 @@ function renderPageOfBranches(
         branch.updates[0]?.group ?? 'N/A',
         getPlatformsForGroup({
           updates: branch.updates,
-          group: branch.updates[0]?.group,
+          group: branch.updates[0]?.group ?? [],
         }),
       ])
     );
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/branch/queries.ts
```diff
@@ -96,7 +96,7 @@ function renderPageOfBranches(
         branch.updates[0]?.group ?? 'N/A',
         getPlatformsForGroup({
           updates: branch.updates,
-          group: branch.updates[0]?.group,
+          group: branch.updates[0]?.group ?? [],
```
wschurman, line 99: I think this will have an incorrect type? The type of the group property is string, so this would pass an array.

I played around with this locally and think the best solution is to explicitly declare the result of `branch.updates[0]` to be typed as potentially undefined.

```typescript
  table.push(
    ...currentPage.map(branch => {
      const update = branch.updates[0] as UpdateBranchFragment['updates'][0] | undefined;
      return [
        branch.name,
        formatUpdate(update),
        update?.runtimeVersion ?? 'N/A',
        update?.group ?? 'N/A',
        update?.group
          ? getPlatformsForGroup({
              updates: branch.updates,
              group: update.group,
            })
          : 'N/A',
      ];
    })
  );
```

packages/eas-cli/src/update/utils.ts
```diff
@@ -47,14 +47,14 @@ export function getPlatformsForGroup({
 }
 
 export function formatPlatformForUpdateGroup(
-  updateGroup: {
+  updateGroup?: {
```
wschurman, line 50: then, with the changes above, you should be able to revert the changes to this file

wschurman (changes requested):
Nice find! @kgc00 should also take a look.

packages/eas-cli/src/branch/queries.ts
hbiede: Yep, I noticed that, fixed it in a different way, committed it, and forgot to push 🤦🏼

packages/eas-cli/src/update/utils.ts
```diff
@@ -52,7 +52,7 @@ export function formatPlatformForUpdateGroup(
     platform: string;
   }[]
 ): string {
-  return updateGroup.length === 0
+  return updateGroup!.length === 0
```
wschurman, line 55:
```suggestion
  return updateGroup.length === 0
```

packages/eas-cli/src/branch/queries.ts
```diff
@@ -89,16 +89,21 @@ function renderPageOfBranches(
...
+        const update: UpdateBranchFragment['updates'][0] | undefined = branch.updates[0];
+        return [
+          branch.name,
+          formatUpdate(update),
+          update?.runtimeVersion ?? 'N/A',
+          update?.group ?? 'N/A',
+          update?.group
+            ? getPlatformsForGroup({
+                updates: branch.updates,
+                group: update?.group ?? '',
```
wschurman, line 102:
```suggestion
                group: update.group,
```

wschurman (approved):
Just a few more changes before commit, but looks good to me otherwise.

kgc00 (approved):
Looks good to me! Thanks for submitting. Fwiw, I think `getPlatformsForGroup` should be updated to handle the `undefined` case (and can roll that into a PR I'm working on).

wkozyra95 (approved):

PR #1327, opened 2022-08-31
[eas-cli] Add noImplicitOverride to tsconfigs

wschurman:
# Checklist

# Why

See the article describing the benefits: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag

This is used in www and entity repos and has been helpful during refactorings.

# How

Add the flag, fix errors.

# Test Plan

`yarn start`


wkozyra95 (approved):
I do not have anything against that change because it's almost zero cost to maintain that, but I'm not sure how much help it will be here. We use inheritance very sparingly, so mainly it's used in command classes from oclif.

dsokal (approved):

wschurman:
> I do not have anything against that change because it's almost zero cost to maintain that, but I'm not sure how much help it will be here. We use inheritance very sparingly, so mainly it's used in command classes from oclif.

Yeah, I agree that how much help it gives us is limited, but even for something like oclif I could see it being helpful. For example, hypothetically let's imagine that oclif renamed one of its base class properties that we override. While they'd put it as a breaking change in the changelog, it's also nice to have a "test" (tsc) fail that tells us that their change broke ours as well.

PR #1346, opened 2022-09-07
Small updates to the language of the Device Create command

justinmcbride:
Small language updates that I noticed while running the `eas device:create` command to add some iOS devices to our provisioning profile.


packages/eas-cli/src/devices/manager.ts
```diff
@@ -72,7 +72,7 @@ export class AccountResolver {
     }
 
     const useProjectAccount = await confirmAsync({
-      message: `You're inside the project directory. Would you like to use ${chalk.underline(
+      message: `You're inside the project directory. Would you like to use the ${chalk.underline(
```
justinmcbride, line 75: The addition of `the` is to make it consistent with line 67: https://github.com/expo/eas-cli/pull/1346/files#diff-546c508e29c605e958c66fc78b76c3c46ea8f5b921a406a511bdf3925039227dR67

dsokal (approved):
Thanks!

PR #1348, opened 2022-09-07
[eas-cli] Standardize GQL Update type schema using Fragment 

kgc00:
# Checklist

# Why

There are a lot of different queries that fetch `Updates`. When we then use that data elsewhere in the code-base we need to match sure that the typing in those functions matches the query type. This gets complicated and unnecessarily difficult to handle. We should simplify the typing so that every schema has the same format and we just use that one format when referring to updates in our typing.

# How

Most of the standardizing was already done in previous PRs. This PR introduces a standardized fragment and replaces in-line schemas with a spread of the fragment.

# Test Plan

Ran commands locally to confirm the GQL changes didn't break our fetches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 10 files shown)


dsokal (approved):

wkozyra95 (approved):

wschurman (approved):

kgc00:
/changelog-entry chore Standardize the GQL 'Update' schema using UpdateFragmentNode

PR #1352, opened 2022-09-09
[eas-cli] Add pagination to channel commands

kgc00:
# Checklist

# Why

closes https://linear.app/expo/issue/ENG-6260/add-pagination-to-channel-commands

We'd like to use a standardized pagination format across our CLI.

# How

- reworked some GQL queries 
- introduced paginated queries for channel commands
- refactored some names and typing around updates and branches to be more consistent
- made `getBranchByName` throw when the branch does not exist to keep it in line with how other fetches work

# Test Plan

Ran affected commands locally with different options and arguments. (offset/non-interactive/json/limit)


packages/eas-cli/src/graphql/queries/BranchQuery.ts
```diff
@@ -47,9 +50,13 @@ export const BranchQuery = {
         )
         .toPromise()
     );
-    return branch;
+    const { updateBranchByName } = response.app.byId;
+    if (!updateBranchByName) {
+      throw new BranchNotFoundError(`Could not find a branch named "${name}".`);
+    }
+    return updateBranchByName;
```
kgc00, line 57: there were roughly equal amounts of callsites that threw after `getBranchByNameAsync` did not return a branch as callsites that did not throw (3 vs 2, if i remember correctly). So I just chose to throw here, because that is generally what we do for other gql queries that don't return the expected object.

I don't feel too strongly about this, lmk if folks prefer throwing at the callsites instead.

packages/eas-cli/src/graphql/queries/ChannelQuery.ts
```diff
@@ -29,10 +36,10 @@ export const ChannelQuery = {
                     name
                     createdAt
                     branchMapping
-                    updateBranches(offset: 0, limit: 1000) {
+                    updateBranches(offset: 0, limit: 5) {
                       id
                       name
-                      updates(offset: 0, limit: 10) {
+                      updateGroups(offset: 0, limit: 1) {
```
kgc00, line 42: we only render the most recent update group, no need to fetch more

kgc00:
/changelog-entry new-feature Add pagination to channel commands

packages/eas-cli/src/channel/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+          : []),
+        message,
+        runtimeVersion,
+        group,
+        platforms,
+      ]);
+    });
+  }
+  Log.log(updateChannelsTable.toString());
+}
```
kgc00, line 112: diff for this section is here https://github.com/expo/eas-cli/pull/1352/files#diff-d9539cd71f4ffb77a01e54dbaaf128a10588b0de93cd44f9e6f392e915557678L37

packages/eas-cli/src/commands/project/init.ts
```diff
@@ -19,14 +19,14 @@ export default class ProjectInit extends EasCommand {
 
     if (exp.extra?.eas?.projectId) {
       Log.error(
-        `app.json is already linked to project with ID: ${chalk.bold(exp.extra?.eas?.projectId)}`
+        `app.json is already linked to app with ID: ${chalk.bold(exp.extra?.eas?.projectId)}`
```
kgc00, line 22: renamed `project` to `app`. `app` is the more modern, preferred term, right?

packages/eas-cli/src/commands/channel/list.ts
```diff
@@ -78,27 +31,9 @@ export default class ChannelList extends EasCommand {
...
+      projectId,
+      paginatedQueryOptions,
     });
-    const channels = getAllUpdateChannelForAppResult.app?.byId.updateChannels;
-    if (!channels) {
-      throw new Error(`Could not find channels on project with id ${projectId}`);
-    }
-
-    if (jsonFlag) {
-      printJsonOnlyOutput(channels);
```
wkozyra95, line 90: This change will make it very hard to make runAsync return value that should be printed in json only mode, as proposed here https://github.com/expo/eas-cli/pull/1360/files

You should probably discuss that with @wschurman to avoid large refactors back and forth.

wkozyra95 (approved):
LGTM
I'm not familiar with updates logic too much, so mostly just skimmed through the code.

packages/eas-cli/src/commands/channel/list.ts
wschurman: I wouldn't worry too much. I think my proposed change would require significant refactoring either way so I'd say it's fine to move forward with this.

packages/eas-cli/src/commands/project/init.ts
wschurman: We use them somewhat interchangeably all over the place. I think project is still preferred.

packages/eas-cli/src/graphql/queries/BranchQuery.ts
wschurman: I think it's fine to throw here

wschurman (approved):

PR #1366, opened 2022-09-16
[eas-cli] Use confirmation when canceling builds

kgc00:
# Checklist

# Why

When running `build:cancel` for a single build, the results of the confirm prompt are ignored.

image

# How

I changed the behavior slightly, to always have the user select from a list of builds (even if there is just one) and to always ask for confirmation that 'yes, we should delete the build'. Happy to change this back to how it was with just the bugfix, but this seemed better behavior to me because:
- it's more consistent across use-cases 
- it always asks for confirmation

# Test Plan

Ran the command locally

### DIFF
--- a/packages/eas-cli/src/commands/build/cancel.ts
+++ b/packages/eas-cli/src/commands/build/cancel.ts
@@ -59,11 +59,12 @@ function formatUnfinishedBuild(
   return `${platform} Started at: ${startTime}, Status: ${status}, Id: ${build.id}`;
 }
 
-async function selectBuildToCancelAsync(
+export async function selectBuildToCancelAsync(
   projectId: string,
   projectFullName: string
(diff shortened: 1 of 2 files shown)


kgc00:
/changelog-entry bug-fix No longer cancel a build when user answers no to confirm prompt

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -112,7 +106,12 @@ async function selectBuildToCancelAsync(
         value: build.id,
       }))
     );
-    return buildId;
+
+    return (await confirmAsync({
```
wkozyra95, line 110: I would drop confirm prompt, but don't have strong opinions either way.

wkozyra95 (approved):

PR #1368, opened 2022-09-16
[1/n][eas-cli] Remove some of getProjectAccountNameAsync

wschurman:
# Checklist

- [ ] I've added an entry to [CHANGELOG.md](https://github.com/expo/eas-cli/blob/main/CHANGELOG.md#main) if necessary. You can comment this pull request with `/changelog-entry [breaking-change|new-feature|bug-fix|chore] [message]` and CHANGELOG.md will be updated automatically.

# PR stack Why

These methods rely upon the owner field to determine the account that owns the project, but the source of truth is the project ID (for example, if a project is transferred). We're trying to remove all reliance upon the owner field for this reason.

This series of PRs will remove `getProjectAccountNameAsync`, `getProjectAccountName`, `getProjectUsername` and other methods of that sort.

# Why

This converts secret management to use the source of truth for the owning account.

# How

Access the owning account via GraphQL.

# Test Plan

```
$ ~/expo/eas-cli/packages/eas-cli/bin/run secret:list
Secrets for this account and project:
┌──────┬──────┬───────┬────┬────────────┐
│ Name │ Type │ Scope │ ID │ Updated at │
└──────┴──────┴───────┴────┴────────────┘

$ ~/expo/eas-cli/packages/eas-cli/bin/run secret:create
✔ Secret name: … test
✔ Select secret type › string
✔ Secret value: … wat
✔ ️Created a new secret test with value wat on project @wschurman141/turtlewat.

$ ~/expo/eas-cli/packages/eas-cli/bin/run secret:list
Secrets for this account and project:
┌──────┬────────┬─────────┬──────────────────────────────────────┬─────────────────┐
│ Name │ Type   │ Scope   │ ID                                   │ Updated at      │
├──────┼────────┼─────────┼──────────────────────────────────────┼─────────────────┤
│ test │ string │ project │ 361d34b2-d71d-4947-aa37-bda9f2b796ac │ Sep 16 14:33:52 │
└──────┴────────┴─────────┴──────────────────────────────────────┴─────────────────┘

$ ~/expo/eas-cli/packages/eas-cli/bin/run secret:delete --id=361d34b2-d71d-4947-aa37-bda9f2b796ac

You are about to permanently delete secret with id: "361d34b2-d71d-4947-aa37-bda9f2b796ac".
This action is irreversible.

✔ Are you sure you wish to proceed? … no / yes
✔ ️Deleted secret with id "361d34b2-d71d-4947-aa37-bda9f2b796ac".

[...]

PR #1370, opened 2022-09-16
[3/n][eas-cli] getUsername -> getUsernameForActor

wschurman:
# Checklist

# Why

https://github.com/expo/eas-cli/pull/1368/files#r973460073

Part 3 of https://github.com/expo/eas-cli/pull/1368.

# How

Rename method, remove exp arg since this function is not project-dependent.

# Test Plan

`yarn start` (typecheck), audit callsites, run tests.

### DIFF
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -18,17 +18,11 @@ import { getVcsClient } from '../vcs';
 import { getExpoConfig } from './expoConfig';
 import { fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync } from './fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync';
 
-export function getUsername(exp: ExpoConfig, user: Actor): string | undefined {
+export function getUsernameForActor(user: Actor): string | undefined {
   switch (user.__typename) {
     case 'User':
       return user.username;
     case 'Robot':
-      // owner field is necessary to run `expo prebuild`
-      if (!exp.owner) {
-        throw new Error(
-          'The "owner" manifest property is required when using robot users. See: https://docs.expo.dev/versions/latest/config/app/#owner'
-        );
-      }
       // robot users don't have usernames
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -36,21 +36,11 @@ export function getProjectAccountName(exp: ExpoConfig, user: Actor): string {
...
- *               fetching it via the App.ownerAccount GraphQL field.
- */
-export function getUsername(exp: ExpoConfig, user: Actor): string | undefined {
+export function getUsernameForActor(user: Actor): string | undefined {
   switch (user.__typename) {
     case 'User':
       return user.username;
     case 'Robot':
-      // owner field is necessary to run `expo prebuild`
-      if (!exp.owner) {
```
wkozyra95, line 49: A bit of context on why it was this way:

When you are running prebuild, expo-cli is using the owner field for some stuff, and if it is not present it fallbacks to the username. In the case of EAS Build, we are running prebuild with unauthenticated expo-cli on our own servers, so we are passing this username as a fallback, in case the owner is not present.
Even If everything is migrated already and nothing in prebuilds process or in SDK depends on the owner field, we still need to support older projects.

With the current state of this PR if someone is building using a bot account, does not have an owner field in app.json, and has a project old enough to relay on the owner field they will get the same result as if they ran prebuild without logging in.

The approach I would recommend is to:
- for older projects that still use the owner field(no idea if there is a way to detect that), if you don't want that check here, then add validation somewhere in the build process e.g https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/build/runBuildAndSubmit.ts#L260
- for new projects, I'm assuming that we don't need that validation, but if it's hard to detect I don't see anything wrong with throwing that error for robot users and requiring an explicit owner field.

wschurman:
(rebased this to the top of the stack and landed the rest since this one requires the most discussion)

wschurman:
Abandoning

PR #1371, opened 2022-09-16
[4/n][eas-cli] Remove getProjectAccountName

wschurman:
# Checklist

# Why

Part 4 of https://github.com/expo/eas-cli/pull/1368.

# How

Remove the function, replace with functions that fetch the account via GraphQL.

# Test Plan

Run tests.

### DIFF
--- a/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
+++ b/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
@@ -4,7 +4,7 @@ import assert from 'assert';
 
 import Log, { learnMore } from '../../log';
 import { GradleBuildContext, resolveGradleBuildContextAsync } from '../../project/android/gradle';
-import { getProjectAccountName, getProjectIdAsync } from '../../project/projectUtils';
+import { getOwnerAccountForProjectIdAsync, getProjectIdAsync } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
 import { findAccountByName } from '../../user/Account';
 import { ensureActorHasUsername, ensureLoggedInAsync } from '../../user/actions';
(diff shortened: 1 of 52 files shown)


packages/eas-cli/src/credentials/manager/ManageAndroid.ts
```diff
@@ -61,8 +61,14 @@ export class ManageAndroid {
...
     });
+
+    const getAccountNameForProjectAsync = async (): Promise<string> => {
+      const projectId = await getProjectIdAsync(ctx.exp, { nonInteractive: false });
+      return (await getOwnerAccountForProjectIdAsync(projectId)).name;
+    };
+
     const accountName = ctx.hasProjectContext
-      ? getProjectAccountName(ctx.exp, ctx.user)
+      ? await getAccountNameForProjectAsync()
```
wkozyra95, line 71: If your goal is to remove any reliance on the owner field then getting read of `findAccountByName` calls should be a part of it. That method does not do anything that inherently depends on the owner field in app.json, but as far as I'm aware this is the only case where we know the account name and it's not just a primary account.

In this case, you could do sth like this

```
let account
if (ctx.hasProjectContext) {
      const projectId = await getProjectIdAsync(ctx.exp, { nonInteractive: false });
      account = (await getOwnerAccountForProjectIdAsync(projectId));
} else {
     account = ctx.user.primaryAccount 
}
```

wkozyra95 (approved):

packages/eas-cli/src/credentials/manager/ManageAndroid.ts
wschurman: Added to task: https://linear.app/expo/issue/ENG-6278#comment-e1966bc6

PR #1373, opened 2022-09-17
[6/n][eas-cli] Clean up some remaining project utils methods

wschurman:
# Checklist

# Why

Part 6 of https://github.com/expo/eas-cli/pull/1368.

# How

Inline a method we never want to be used anywhere but during project linking. Grep for "owner" to verify that there are no more uses of this field.

# Test Plan

tsc


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -143,9 +126,23 @@ export async function getProjectIdAsync(
 
   Log.warn('EAS project not configured.');
 
+  const getAccountNameForEASProjectSync = (exp: ExpoConfig, user: Actor): string => {
+    if (exp.owner) {
```
wkozyra95, line 130: Not sure if it should be part of this PR, but there needs to be a better way to select an account. To do it now, you need to define it in app.json, run a command that adds projected, and then remove it.

wkozyra95 (approved):

PR #1376, opened 2022-09-19
[7/n][eas-cli] Improve `eas init` command

wschurman:
# Checklist

# Why

This PR removes the last dependence on the owner field. It can still be present and is used to bypass one prompt in this command, but it is not required.

# How

- Add ability to specify `id` flag to command. This will probably be the instructions on the website rather than setting the owner field.
- Add wizard flow to command when `id` flag is not specified.

# Test Plan

```
$ ~/expo/eas-cli/packages/eas-cli/bin/run init --id=1234
✔ Linked project to ID: 1234

$ ~/expo/eas-cli/packages/eas-cli/bin/run init --id=12345
✔ Project is already linked to  a different ID: 1234. Do you wish to overwrite it? … no
Aborting

$ ~/expo/eas-cli/packages/eas-cli/bin/run init --id=12345
✔ Project is already linked to  a different ID: 1234. Do you wish to overwrite it? … yes
✔ Linked project to ID: 12345

$ ~/expo/eas-cli/packages/eas-cli/bin/run init --id=123456 --force
✔ Linked project to ID: 123456

$ ~/expo/eas-cli/packages/eas-cli/bin/run init --id=1234567 --non-interactive
    Error: Project is already linked to a different ID: 123456. Use --force flag to overwrite.

$ git checkout -- .

$ ~/expo/eas-cli/packages/eas-cli/bin/run init
✔ Which account would you like to link this project to? › wschurman141
✔ Existing EAS project found for @wschurman141/turtlewat (id = 1a7e84d3-978d-476d-97cb-08a992acd17e). Link this project? … yes
✔ Linked project to ID: 1a7e84d3-978d-476d-97cb-08a992acd17e

$ git checkout -- .

$ ~/expo/eas-cli/packages/eas-cli/bin/run init
✔ Which account would you like to link this project to? › wschurman141
✔ Existing EAS project found for @wschurman141/turtlewat (id = 1a7e84d3-978d-476d-97cb-08a992acd17e). Link this project? … no
    Error: EAS project ID configuration canceled. Re-run the command to select a different account/project.

$ ~/expo/eas-cli/packages/eas-cli/bin/run init
✔ Which account would you like to link this project to? › real-org-will
    Error: You don't have permission to create a new project on the real-org-will account and no matching project already exists on the account.

$ ~/expo/eas-cli/packages/eas-cli/bin/run init
✔ Which account would you like to link this project to? › wschurman141
✔ Existing EAS project found for @wschurman141/turtlewat (id = 1a7e84d3-978d-476d-97cb-08a992acd17e). Link this project? … no
    Error: EAS project ID configuration canceled. Re-run the command to select a different account/project.

$ ~/expo/eas-cli/packages/eas-cli/bin/run init
✔ Which account would you like to link this project to? › wschurman141
✔ Existing EAS project found for @wschurman141/turtlewat (id = 1a7e84d3-978d-476d-97cb-08a992acd17e). Link this project? … yes
✔ Linked project to ID: 1a7e84d3-978d-476d-97cb-08a992acd17e
```


packages/eas-cli/src/commands/project/init.ts
```diff
@@ -1,27 +1,226 @@
...
   static override aliases = ['init'];
 
-  async runAsync(): Promise<void> {
+  static override flags = {
+    id: Flags.string({
+      description: 'ID of the EAS project to link',
+    }),
+    force: Flags.boolean({
+      description: 'Whether to overwrite any existing project ID',
+      dependsOn: ['id'],
```
wkozyra95, line 38: why require an id? I think it would be useful to resolve id based on the owner field. Dominik suggested in one of the previous PRs to add envs that skip projected requirements, but using `eas init --force && eas build ...` would solve that issue.

e.g.  https://github.com/expo/turtle-v2-example/blob/master/managed/app.config.ts#L9 We need sth like this to be able to build the same project for different accounts and environments, but with --force flag we wouldn't need to define those ids locally.
Also, our current solution has the problem, where the project is not defined during the build, but that is not a problem for us, because we never run that project, just build for testing.

wkozyra95 (approved):

packages/eas-cli/src/commands/project/init.ts
wschurman: Explicit IDs is the direction we're moving in and where we want to end up (deprecating owner and slug fields entirely). Put another way, if we were building EAS from scratch, knowing the requirements we know now we wouldn't add the owner field or the slug field at all. Instead, everything would just be ID based. While it's not as beautiful, it's far more robust, less prone to error, and requires far less engineering attention.

Inferring the account from the owner field in this PR should be considered a fallback and legacy. In the near future, we'll most likely just tell people to run `eas init --id <project_id>` instead of adding an owner field.

For the differing project ID per environment, I think a map in a dynamic config is the most correct and robust approach. There's no question what project this is referring to, and no chance for error if and env is assumed but incorrectly applied. (for example, if some the env is forgotten on a sub-command it fails on the server rather than accidentally apply to a different project implicitly).

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -161,7 +161,7 @@ export default class DeviceDelete extends EasCommand {
 
   async chooseDevicesToDeleteAsync(
     appleDevices: AppleDeviceQueryResult[],
-    udids: string[]
+    udids: string[] | undefined
```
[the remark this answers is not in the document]
wschurman: Ah, whoops, yeah, will remove. This was leftover from me trying to use `@oclif/test` and needing to upgrade `@oclif/core` which revealed this type error. We'll need to make this change when we upgrade.

packages/eas-cli/src/commands/project/init.ts
wkozyra95: Sure, that is fair. Our internal use case is a bit of hack on it's own and I see why we would not want to support it.

brentvatne (commented):
here are my suggestions in terms of copy.

### Before

```
easd init
✔ Which account would you like to link this project to? › brents
✔ Existing EAS project found for @brents/williapp (id = 3e5310ac-6900-4b9e-b0a1-e9ed2a346e5c). Link this project? … yes
✔ Linked project to ID: 3e5310ac-6900-4b9e-b0a1-e9ed2a346e5c
```

### After

```
easd init
✔ Which account should own this project? › brents
✔ Existing project found: @brents/williapp (ID: 3e5310ac-6900-4b9e-b0a1-e9ed2a346e5c). Link this project? … yes
✔ Linked to @brents/williapp (ID: 3e5310ac-6900-4b9e-b0a1-e9ed2a346e5c) (modified app.json)
```

- in general let's provide the account name / slug next to the id each time we output it. I know the point of uuids is to move to an actually stable identifier that isn't dependent on account name, but for users I think it's valuable to see the project name next to it.
- i changed prompts to be a bit less wordy
- when we indicate that we linked the project, it's worth saying how we did that. i added `(modified app.json)` here - vercel does `(created .vercel)`

PR #1381, opened 2022-09-21
[eas-cli] Add pagination to device:* commands

kgc00:
# Checklist

# Why

This PR finishes adding support for paginated queries to the EAS CLI.

# How

Following in the steps of previous PRs I 
- introduced helper methods to query for items
- updated the gql schema for relevant queries
- did some light rewriting/refactoring

As a part of this PR, I removed support for deleting multiple devices from an Apple team in a single `delete` command. That requires multi-select pagination and we are choosing to hold off an adding that functionality due to UX concerns.

# Test Plan

Ran the affected commands (using Brent's account 🙏 ). Confirmed each combination of flags and command was working properly.

### device:list

https://user-images.githubusercontent.com/15132641/192421718-948410c7-bd39-4c17-ae46-e89ca1763a88.mov

### device:delete

https://user-images.githubusercontent.com/15132641/192421879-4a12f915-62fa-417d-ba2e-681f0a456ecb.mov

(non-interactive not shown)

### DIFF
--- a/packages/eas-cli/src/commands/device/delete.ts
+++ b/packages/eas-cli/src/commands/device/delete.ts
@@ -3,28 +3,38 @@ import { Flags } from '@oclif/core';
 import assert from 'assert';
 
 import EasCommand, { EASCommandProjectConfigContext } from '../../commandUtils/EasCommand';
-import { chooseDevicesToDeleteAsync } from '../../credentials/ios/actions/DeviceUtils';
+import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
+import {
+  EasPaginatedQueryFlags,
+  PaginatedQueryOptions,
+  getPaginatedQueryOptions,
+} from '../../commandUtils/pagination';
 import { AppleDeviceMutation } from '../../credentials/ios/api/graphql/mutations/AppleDeviceMutation';
 import {
   AppleDeviceQuery,
   AppleDeviceQueryResult,
-  AppleDevicesByTeamIdentifierQueryResult,
 } from '../../credentials/ios/api/graphql/queries/AppleDeviceQuery';
-import { AppleTeamQuery } from '../../credentials/ios/api/graphql/queries/AppleTeamQuery';
 import { authenticateAsync, getRequestContext } from '../../credentials/ios/appstore/authenticate';
+import {
+  selectAppleDeviceOnAppleTeamAsync,
+  selectAppleTeamOnAccountAsync,
+} from '../../devices/queries';
 import formatDevice from '../../devices/utils/formatDevice';
 import { AppleDevice, Maybe } from '../../graphql/generated';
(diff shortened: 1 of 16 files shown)


packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -3,251 +3,172 @@ import { Flags } from '@oclif/core';
...
-    if (!deleteOnApple) {
-      return;
-    }
-
     const ctx = await authenticateAsync({ teamId: appleTeamIdentifier });
     const context = getRequestContext(ctx);
 
     Log.addNewLineIfNone();
-    const removeAppleSpinner = ora('Disabling devices on Apple').start();
+    const removeAppleSpinner = ora('Disabling device on Apple').start();
```
kgc00, line 114: how does ora interact with json? do i need to use pretty print json output here or not print anything at all?

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -3,251 +3,172 @@ import { Flags } from '@oclif/core';
...
     assert(appleTeamIdentifier, 'No team identifier is specified');
 
-    const appleDevicesResult = await this.getDevicesForTeamAsync(account.name, appleTeamIdentifier);
-
-    if (!appleDevicesResult) {
-      return;
-    }
-
-    const { appleTeamName, appleDevices } = appleDevicesResult;
+    const chosenDevice = udid
```
kgc00, line 72: Just making a note here that we are dropping support for deleting multiple devices in a single command run. 

Based on my discussion in and beyond #1385, we're choosing to drop support for multi-select prompts + pagination. The idea is that we'd prefer a better UX if we were to support it and we don't have the bandwidth to do that just now. So we'll remove the ability and re-enable it at some later date.

kgc00:
/changelog-entry breaking-change Adds paginated support to device commands. Removes the ability to delete multiple devices at once.

packages/eas-cli/src/commands/device/delete.ts
```diff
@@ -3,251 +3,167 @@ import { Flags } from '@oclif/core';
...
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
 export default class DeviceDelete extends EasCommand {
   static override description = 'remove a registered device from your account';
 
   static override flags = {
-    'apple-team-id': Flags.string(),
-    udid: Flags.string({ multiple: true }),
+    'apple-team-id': Flags.string({ description: 'The Apple team ID on which to find the device' }),
+    udid: Flags.string({ description: 'The Apple device ID to disable' }),
```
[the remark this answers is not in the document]
kgc00: it's called disable throughout our prompts to the user, so at least we're consistent. I'll leave deciding on the proper term to others.

FiberJW (approved):
anime-boy

PR #1383, opened 2022-09-21
[10/n][eas-cli] Add new on-demand context concept for commands

wschurman:
# Checklist

# Why

The main idea is that we need a better way to declaratively specify dependencies/requirements for commands in a type-safe way. This PR creates a new concept similar in typesystem pattern to flags/parse, but for things like projectId, authenticated actor, project directory, etc.

It is the evolution of https://github.com/expo/eas-cli/pull/1360 but allows commands to be more fine-grained in what they specify/require.

# How

Create new Context concept following the patten used by flags/parse (pass the class constructor to the `getContextAsync` method to infer types).

This allows classes to specify what they need. The idea is that we'll remove `getProjectIdAsync` (in the next PR) and make it so that all commands that need the `projectId` must get it through this new context concept. This will work similarly with things like requiring project directory, etc.

This allows us to do things like validate the project setup (see linked task) on all commands that require it.

# Test Plan

Run tests.

```
$ ~/expo/eas-cli/packages/eas-cli/bin/run branch:list

Branches:
...
```

Run the other commands modified (especially the ones that can be run outside a project folder).

### DIFF
--- a/packages/eas-cli/src/commandUtils/EasCommand.ts
+++ b/packages/eas-cli/src/commandUtils/EasCommand.ts
@@ -13,20 +13,97 @@ import {
   logEvent,
 } from '../analytics/rudderstackClient';
 import { learnMore } from '../log';
-import { findProjectRootAsync } from '../project/projectUtils';
-import { getUserAsync } from '../user/User';
+import { getExpoConfig } from '../project/expoConfig';
+import { findProjectRootAsync, getProjectIdAsync } from '../project/projectUtils';
+import { Actor, getUserAsync } from '../user/User';
 import { ensureLoggedInAsync } from '../user/actions';
 import { easCliVersion } from '../utils/easCli';
 import { setVcsClient } from '../vcs';
 import GitClient from '../vcs/clients/git';
 
-export default abstract class EasCommand extends Command {
+export interface CommandConfiguration {
   /**
    * When user data is unavailable locally, determines if the command will
-   * force the user to log in
+   * force the user to log in. By default, all commands require authentication.
    */
-  protected requiresAuthentication = true;
-  protected mustBeRunInsideProject = true;
+  allowUnauthenticated?: boolean;
+
+  /**
+   * Whether a command can be run outside of an Expo project directory.
+   * By default, all commands must be run inside an Expo project directory.
+   */
+  canRunOutsideProject?: boolean;
+}
+
+export interface ContextOptions {
+  nonInteractive: boolean;
+}
+
+export abstract class ContextField<T> {
+  abstract getValueAsync(options: ContextOptions): Promise<T>;
+}
+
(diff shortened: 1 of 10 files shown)


packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -13,45 +13,124 @@ import {
...
+  }
 
   protected abstract runAsync(): Promise<any>;
 
   // eslint-disable-next-line async-protect/async-suffix
   async run(): Promise<any> {
     await initAnalyticsAsync();
 
-    if (this.mustBeRunInsideProject) {
+    if (!this.commandConfiguration.canRunOutsideProject) {
```
wkozyra95, line 114: if the context is supposed to be the only source of truth, then we should also handle a case where the command does not have to be run inside the project, but if it is, then it can use that information.

e.g. `eas credentials` should work like that, outside of the project it is managing account credentials, and if it's in the project it can be more specific. Running outside the projects seems to be already broken on main, so it's not related to this PR, but in a case like that should be supported

packages/eas-cli/src/commands/analytics.ts
```diff
@@ -7,7 +7,9 @@ export default class AnalyticsView extends EasCommand {
 
   static override args = [{ name: 'STATUS', options: ['on', 'off'] }];
 
-  protected override requiresAuthentication = false;
+  protected override commandConfiguration: CommandConfiguration = {
+    allowUnauthenticated: true,
```
wkozyra95, line 11: this should work outside of the project context

wkozyra95 (approved):

packages/eas-cli/src/commands/analytics.ts
wschurman: Yep, with this new context concept it'll just work since it doesn't specify the project context.

packages/eas-cli/src/commandUtils/EasCommand.ts
wschurman: Yep, later on in this stack I introduce a OptionalProjectIdContext, which addresses this exact case.

packages/eas-cli/src/commands/analytics.ts
wkozyra95: I don't see why `account:view` needs to have `canRunOutsideProject: true` but here it's not necessary?

wschurman: Because the command doesn't declare a context that requires a project (projectId, projectDir), those checks won't be performed and it therefore won't have access to project data.

wkozyra95: That was my point, `account:view` also does not declare the context that needs projectId

PR #1384, opened 2022-09-21
[11/n][eas-cli] Use new context for logged-in actor

wschurman:
# Checklist

# Why

This uses the actor context introduced in https://github.com/expo/eas-cli/pull/1383.

This will allow commands to more declaratively specify that they require a logged-in user rather than needing the `allowUnauthenticated` property on the command class. The benefit of this over the `allowUnauthenticated` property is that it is type-safe. For a command to use actor, they need to specify the context dependency. Any command that doesn't need actor doesn't need to specify the context dependency, and is therefore by definition allowUnauthenticated.

# How

Remove most callsites to ensureLoggedInAsync (the remaining ones are in EasCommand and getProjectIdAsync which are going to be refactored out in the subsequent PRs).

# Test Plan

`yarn test`. Sanity check by running all changed commands.

### DIFF
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -7,7 +7,7 @@ import path from 'path';
 
 import { BuildFlags, runBuildAndSubmitAsync } from '../../build/runBuildAndSubmit';
 import { UserInputResourceClass } from '../../build/types';
-import EasCommand from '../../commandUtils/EasCommand';
+import EasCommand, { EASCommandLoggedInContext } from '../../commandUtils/EasCommand';
 import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
 import { StatuspageServiceName } from '../../graphql/generated';
 import Log, { link } from '../../log';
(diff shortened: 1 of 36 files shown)


packages/eas-cli/src/commands/build/index.ts
```diff
@@ -104,6 +108,10 @@ export default class Build extends EasCommand {
     }
     const flags = this.sanitizeFlags(rawFlags);
 
+    const { actor } = await this.getContextAsync(Build, {
```
wkozyra95, line 111: I don't think we ever had a use-case where this value couldn't be global, I understand the reluctance to use the global state for anything, but if we would need to pass it everywhere it might be a bit messy

wkozyra95 (approved):

packages/eas-cli/src/commands/build/index.ts
wschurman: In general global state for logged-in user makes a few things nearly impossible:
- switching users (switching logged-in user in a running program)
- mocking users (though jest is ok at this, injection is still a better IMO)

One other argument against global state is that we'd have to type it as `Actor | null` globally, while this context allows us to type it affirmatively.

PR #1385, opened 2022-09-21
[eas-cli] add paginated multi-select

kgc00:
# Checklist

# Why

#1381 requires pagination on a multi-select prompt. Split it out to its own PR to isolate changes.

closes https://linear.app/expo/issue/ENG-6422/add-multiselect-support-on-paginated-queries

# How

- added in a new utility method `paginatedQueryWithMultiSelectPromptAsync`
- added explicit `ExpoChoice` typing for `paginatedQueryWithSelect*` methods

# Test Plan

Added an automated test. Manually tested by changing a random select prompt to a multiselect prompt. 

### Demo
Initial approach -- require an 'enter' press after selecting a new page:

https://user-images.githubusercontent.com/15132641/191625689-b0240d26-c921-4377-b875-e5ef8e48096d.mov

Alternative approach -- auto-submit when a user selects a new page:

https://user-images.githubusercontent.com/15132641/192012158-777c5d96-68f2-4b7c-9f49-a73b20366593.mov

### DIFF
--- a/packages/eas-cli/src/prompts.ts
+++ b/packages/eas-cli/src/prompts.ts
@@ -55,7 +55,22 @@ export async function selectAsync<T>(
   );
   return value ?? null;
 }
-
+export async function multiselectAsync<T>(
+  message: string,
+  choices: ExpoChoice<T>[],
+  options?: Options
+): Promise<T[]> {
+  const { value } = await promptAsync(
+    {
+      message,
+      choices,
+      name: 'value',
+      type: 'multiselect',
+    },
+    options
+  );
+  return value ?? null;
+}
 /**
  * Create a more dynamic yes/no confirmation that can be cancelled.
  *
--- a/packages/eas-cli/src/prompts.ts
+++ b/packages/eas-cli/src/prompts.ts
@@ -1,5 +1,9 @@
+// @ts-ignore
+import { MultiSelect } from 'enquirer';
 import { constants } from 'os';
 import prompts, { Answers, Choice, Options, PromptType, PromptObject as Question } from 'prompts';
+
+import { PAGINATION_FETCH_MORE_VALUE } from './utils/queryConstants';
 
 export { PromptType, Question, Choice };
 
@@ -55,22 +59,7 @@ export async function selectAsync<T>(
   );
   return value ?? null;
 }
-export async function multiselectAsync<T>(
-  message: string,
-  choices: ExpoChoice<T>[],
-  options?: Options
-): Promise<T[]> {
-  const { value } = await promptAsync(
-    {
-      message,
-      choices,
-      name: 'value',
-      type: 'multiselect',
-    },
-    options
-  );
-  return value ?? null;
-}
+
 /**
  * Create a more dynamic yes/no confirmation that can be cancelled.
  *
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/utils/queries.ts
```diff
@@ -131,3 +127,70 @@ async function paginatedQueryWithSelectPromptInternalAsync<
...
+    selected: selectedIdsAccumulator.includes(promptOptions.getIdentifierForQueryItem(queryItem)),
+  }));
+  if (areMorePagesAvailable) {
+    selectionPromptListItems.push({ title: 'Next page...', value: fetchMoreValue });
+  }
+  if (selectionPromptListItems.length === 0) {
+    return;
+  }
+
+  const valueOfUserSelectedListItems = await multiselectAsync<string>(
```
wkozyra95, line 172: It's hard to see from the recording, but I understand that the way it works to switch to the next page you need to press space to mark it and enter to submit, which displays new prompt with a next page.

I don't think this is a good experience, I have 2 alternatives:
- When the next page is marked with space immediately go to the next page. All items that were already checked should also show up plus entire new page of unchecked items(This assumes that list displayed to users might grow).
- When the next page is marked with space immediately go to the next page, but on every page after the first one, also display the previous page entry.

I think the first approach would be cleaner from users perspective.

As for how to implement that, I think `onState` callback should provide that info. And if it's not possible with `prompts`, maybe consider different lib for multiselect.

kgc00: I also thought: "this doesn't seem ideal"! But it seemed the best we could do. Thanks for the alternative approaches. I'll investigate to see if I can cook up something that feels a bit better to use.

wkozyra95: I tried to test that, but it seems you are not using that code in any of the commands, can you add it to the intended command or update the recording?

wkozyra95: In general, if the alternative implementation is a problem I would be leaning into keeping old implementation and setting a default limit to sth most people would never hit.

packages/eas-cli/src/prompts.ts
```diff
@@ -1,3 +1,7 @@
+// @ts-ignore
+// the docs mention we can create custom prompts by sublclassing built-in prompts:
+// https://github.com/enquirer/enquirer#-custom-prompts
+import MultiSelect from 'enquirer/lib/prompts/multiselect';
```
wkozyra95, line 4:
```suggestion
import { MultiSelect } from 'enquirer';
```
Also, I don't see what this comment and link are there, initially I assumed that this is an explanation for using that direct import, but it does not seem necessary.

kgc00: good catch! the typing in the module is non-existant, so I got a ts error when I tried to import multiselect from the package in the way you suggest. However, when I actually runt he code, it does import properly. I'll remove the comment and adopt this import style.

packages/eas-cli/src/utils/queries.ts
kgc00: Sorry, I forgot to record a video. I think the UX of this approach is what we were hoping for. However, the library is untyped and subclassing to get the logic we want seems unnecessarily verbose and maybe a little fragile. Added a video to the description.

However, I think this is better than setting a super high limit. We've seen hundreds on apps/accounts and this will soon lead to timeouts for those users when they try to run these commands.

wkozyra95: > I think setting a super high limit for device queries is the wrong approach.

I did not suggest increasing that to huge values, just to sth that the average user would not hit. I did not realize that the default limit is already at 50, if that is the case I'm fine with either approach

> However, the library is untyped and subclassing to get the logic we want seems unnecessarily verbose and maybe a little fragile.

I don't think this is a problem, I was more concerned about size with adding one more library, but it seems it did not affect much.

wkozyra95: In terms of UX it's not exactly what I had in mind. You are currently creating a new prompt for every page, that is left in the terminal. Ideally, that logic should be inside multi-select, so when you fetch the next page state of the existing prompt is updated instead, but I'm not sure how possible it is with this lib.

wkozyra95: cc @brentvatne any opinion on that?

wkozyra95 (approved):

kgc00:
After speaking with the team about this I'll be closing the PR. The UX on a paginated multi-select is inherently confusing. The way to make it work would be to have some kind of infinite scroll functionality, where you don't ever "select" a new page, it's just auto fetched. We'd have to implement that ourselves, which we don't have the bandwidth for at the moment. So I'll instead use a normal non-multi-select for the base PR.

packages/eas-cli/src/utils/queries.ts
brentvatne: (i followed up with kirby and suggested skipping this functionality)

PR #1386, opened 2022-09-21
[eas-cli] Pass through platform to export command

brentvatne:
# Checklist

# Why

When publishing an update for a specific platform, eg: `eas update -p ios`, we should only build the bundle for that platform. Currently, we build for both platforms regardless. So, `eas update -p ios` will build both iOS and Android bundles. This was a problem for Fernando because his Android codebase wasn't compiling at the time, but iOS worked fine.

# How

Pass through the platform flag. `expo-cli export` and `npx expo export` both support `--platform ios|android|all` so this was straightforward.

# Test Plan

Run the command with a single platform and it should work:

image

Ah, nevermind. Server changes needed :) Opened this PR in WIP state until that is ready.

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -147,9 +147,11 @@ export async function buildUnsortedUpdateInfoGroupAsync(
 export async function buildBundlesAsync({
   projectDir,
   inputDir,
+  platform = 'all',
 }: {
   projectDir: string;
   inputDir: string;
+  platform?: PublishPlatform | 'all';
 }): Promise<void> {
   const packageJSON = JsonFile.read(path.resolve(projectDir, 'package.json'));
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -147,9 +147,11 @@ export async function buildUnsortedUpdateInfoGroupAsync(
 export async function buildBundlesAsync({
   projectDir,
   inputDir,
+  platform = 'all',
 }: {
   projectDir: string;
   inputDir: string;
+  platform?: PublishPlatform | 'all';
```
EvanBacon, line 154:
```suggestion
  platform,
}: {
  projectDir: string;
  inputDir: string;
  platform: PublishPlatform;
```
Doesn't _need_ to be optional.

PR #1387, opened 2022-09-21
[12/n][eas-cli] Use new context for project ID

wschurman:
# Checklist

# Why

Same as https://github.com/expo/eas-cli/pull/1384 but for `projectId`.

# How

Replace most of `getProjectIdAsync` with the new context param.

# Test Plan

`yarn test`

Run all changed commands as a sanity check.

### DIFF
--- a/packages/eas-cli/src/build/createContext.ts
+++ b/packages/eas-cli/src/build/createContext.ts
@@ -10,7 +10,7 @@ import { Analytics, BuildEvent } from '../analytics/events';
 import { CredentialsContext } from '../credentials/context';
 import { BuildResourceClass } from '../graphql/generated';
 import { getExpoConfig } from '../project/expoConfig';
-import { getOwnerAccountForProjectIdAsync, getProjectIdAsync } from '../project/projectUtils';
+import { getOwnerAccountForProjectIdAsync } from '../project/projectUtils';
 import { resolveWorkflowAsync } from '../project/workflow';
 import { Actor } from '../user/User';
 import { createAndroidContextAsync } from './android/build';
@@ -31,6 +31,7 @@ export async function createBuildContextAsync<T extends Platform>({
   resourceClass,
   message,
   actor,
+  projectId,
 }: {
   buildProfileName: string;
   buildProfile: BuildProfile<T>;
@@ -44,14 +45,11 @@ export async function createBuildContextAsync<T extends Platform>({
   resourceClass: BuildResourceClass;
   message?: string;
   actor: Actor;
+  projectId: string;
 }): Promise<BuildContext<T>> {
   const exp = getExpoConfig(projectDir, { env: buildProfile.env });
 
   const projectName = exp.slug;
-  const projectId = await getProjectIdAsync(exp, {
-    env: buildProfile.env,
-    nonInteractive,
-  });
   const account = await getOwnerAccountForProjectIdAsync(projectId);
   const workflow = await resolveWorkflowAsync(projectDir, platform);
   const accountId = account.id;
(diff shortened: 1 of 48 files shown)


packages/eas-cli/src/build/createContext.ts
```diff
@@ -44,14 +45,11 @@ export async function createBuildContextAsync<T extends Platform>({
   resourceClass: BuildResourceClass;
   message?: string;
   actor: Actor;
+  projectId: string;
```
wkozyra95, line 48: When you build with the `--platform=all` it's entirely possible that android and ios builds will have different projectIds based on envs, so projected here needs to read from `exp` from the build context that was resolved based on envs from this build profile.

wkozyra95 (changes requested):

packages/eas-cli/src/build/createContext.ts
wschurman: Good point. I need to rethink these particular commands as far as context goes.

wschurman:
@wkozyra95 - Thanks for the review of this. That was the critical piece that was needed to refine this to make it work for all commands. 

I updated this to encompass all config (getExpoConfig, getProjectId, and the dynamic versions of those two via a injected function). Let me know what you think.

packages/eas-cli/src/commands/metadata/pull.ts
```diff
@@ -27,18 +26,20 @@ export default class MetadataPull extends EasCommand {
...
 
     const { flags } = await this.parse(MetadataPull);
-    const { actor } = await this.getContextAsync(MetadataPull, { nonInteractive: false });
+    const { actor, getDynamicProjectConfigAsync } = await this.getContextAsync(MetadataPull, {
+      nonInteractive: false,
+    });
 
     const projectDir = await findProjectRootAsync();
-    const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
+    const { exp } = await getDynamicProjectConfigAsync({ skipSDKVersionRequirement: true });
```
wkozyra95, line 42: this is a bug, `getConfig` should not be used directly and when you use getExpoConfig  skipSDKVersionRequirement is passed there by default. You can use exp from EASCommandProjectConfigContext

packages/eas-cli/src/commands/metadata/push.ts
```diff
@@ -25,18 +24,20 @@ export default class MetadataPush extends EasCommand {
...
 
     const { flags } = await this.parse(MetadataPush);
-    const { actor } = await this.getContextAsync(MetadataPush, { nonInteractive: false });
+    const { actor, getDynamicProjectConfigAsync } = await this.getContextAsync(MetadataPush, {
+      nonInteractive: false,
+    });
 
     const projectDir = await findProjectRootAsync();
-    const { exp } = getConfig(projectDir, { skipSDKVersionRequirement: true });
+    const { exp } = await getDynamicProjectConfigAsync({ skipSDKVersionRequirement: true });
```
wkozyra95, line 40: The same as above

packages/eas-cli/src/submit/context.ts
```diff
@@ -41,15 +41,15 @@ export async function createSubmissionContextAsync<T extends Platform>(params: {
...
   profile: SubmitProfile<T>;
   projectDir: string;
-  projectId: string;
   applicationIdentifier?: string;
   actor: Actor;
+  getDynamicProjectConfigAsync: DynamicConfigContextFn;
 }): Promise<SubmissionContext<T>> {
   const { applicationIdentifier, projectDir, nonInteractive, actor } = params;
-  const exp = getExpoConfig(projectDir, { env: params.env });
+  const { exp, projectId } = await params.getDynamicProjectConfigAsync({ env: params.env });
```
wkozyra95, line 49: It might be a bit cleaner to pass exp and projected
- When this code is called from submit command you would run the same code as here
- When this code is run as a part of the build command with auto-submit option you already have those value in buildCtx

wschurman: Good suggestion! Will do in a follow-up: https://linear.app/expo/issue/ENG-6471/clean-up-getdynamicprojectconfigasync-in-submit-context

PR #1389, opened 2022-09-22
[14/n][eas-cli] Simplify credentials manager and device create for within-project-only running

wschurman:
# Checklist

# Why

These commands don't specify `canRunOutsideProject` (previously `mustBeRunInsideProject = false`), so they always are run inside a project. Therefore, a lot of the complex runtime switching can be removed.

This was noticed due to the typechecker telling us this with the new projectId context in the commands (and passing projectId in from the root commands rather than getting it at the callsite).

I have a feeling not specifying `mustBeRunInsideProject = false` previously was an oversight and there is a bug in prod, but I'm not familiar with this stuff. If it was a bug, this PR should be able to be safely ignored.

# How

Remove non-project code.

# Test Plan

Run commands (`eas credentials`, `eas device:create`) in non-project. See that output is the same as production `eas-cli`.

### DIFF
--- a/packages/eas-cli/src/build/android/build.ts
+++ b/packages/eas-cli/src/build/android/build.ts
@@ -91,7 +91,7 @@ export async function prepareAndroidBuildAsync(
           ctx.easJsonCliConfig?.appVersionSource === AppVersionSource.REMOTE
             ? false
             : ctx.buildProfile.autoIncrement,
-        nonInteractive: ctx.nonInteractive,
+        projectId: ctx.projectId,
       });
     },
     prepareJobAsync: async (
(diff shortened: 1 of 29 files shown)


wkozyra95 (commented):
credentials manager was working in the past outside of the project directory, not sure if this was an intentional change
probably it was introduced here https://github.com/expo/eas-cli/commit/76d75e9c17411e901da5815283b248321c44b79f

It's probably up to @quinlanj whether we want to keep old behavior or not.

wschurman:
Alright, will move forward assuming we'll want the credentials and device manager among other commands to be able to run not in a project directory. Note that if written correctly, commands will automatically declare whether they need this context so there will be no potential for bugs like the one that has been in prod for 6 months.

PR #1395, opened 2022-09-22
[18/n][eas-cli] Remove unused arguments in some context util functions

wschurman:
# Checklist

# Why

These arguments are unused in the production callsites, so remove them.

# How

Remove unused args.

# Test Plan

Run tests and tsc.


packages/eas-cli/src/commandUtils/context/__tests__/ProjectDirContextField-test.ts
```diff
@@ -56,7 +40,7 @@ describe(ProjectDirContextField['findProjectRootAsync'], () => {
       },
       '/app'
     );
-    const projectRoot = await ProjectDirContextField['findProjectRootAsync']({ cwd: '/app/src' });
+    const projectRoot = await ProjectDirContextField['findProjectRootAsync']();
```
wkozyra95, line 43: The point of this test was to check if `process.cwd` is in the subdirectory will the function return directory with package.json. If you mock the entire pkg-dir then this test does not check anything at this point.

packages/eas-cli/src/commandUtils/context/__tests__/ProjectDirContextField-test.ts
```diff
@@ -23,31 +25,13 @@ describe(ProjectDirContextField['findProjectRootAsync'], () => {
       },
       '/app'
     );
-    await expect(ProjectDirContextField['findProjectRootAsync']({ cwd: '/app' })).rejects.toThrow(
+    await expect(ProjectDirContextField['findProjectRootAsync']()).rejects.toThrow(
```
wkozyra95, line 28: If I understand correctly process.cwd is preserved inside the test, so detecting project root here fails not because there is no packagae.json in the directory, but because cwd is totally outside the `/app` dir.

wkozyra95 (commented):
I didn't see any good way to mock process.cwd, so either keep the optional param to override cwd in findProjectRoot or delete all the tests from ProjectDirContextField-test.ts

wschurman:
Abandoning this one.

PR #1401, opened 2022-09-26
[ENG-4932][eas-cli] handle errors form new platform-specific kill switches

szdziedzic:
# Checklist

# Why

https://linear.app/expo/issue/ENG-4932/dealing-with-platform-specific-issues-on-eas
Companion to https://github.com/expo/universe/pull/10576.

# How

Handle errors from new platform-specific kill switches to stop accepting builds for a given platform, so users see an accurate message for which platform we are not accepting builds vs just seeing generic information that the whole free tier is disabled.

# Test Plan

Tested locally

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -147,27 +147,27 @@ function handleBuildRequestError(error: any, platform: Platform): never {
     throw new Error('Build request failed.');
   } else if (error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED') {
     Log.error(
-      `EAS Build free tier is temporarily disabled. Try again later. Check ${link(
-        'https://status.expo.dev/'
-      )} for updates. ${learnMore('https://expo.fyi/eas-build-queues')}`
+      `EAS Build free tier is temporarily disabled and we are not accepting new builds. Try again later. ${learnMore(
+        'https://expo.fyi/eas-build-queues'
+      )}`
     );
     throw new Error('Build request failed.');
   } else if (
     error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED_IOS'
   ) {
     Log.error(
-      `EAS Build free tier is temporarily disabled for iOS. Try again later. Check ${link(
-        'https://status.expo.dev/'
-      )} for updates. ${learnMore('https://expo.fyi/eas-build-queues')}`
+      `EAS Build free tier is temporarily disabled for iOS and we are not accepting new builds. Try again later. ${learnMore(
+        'https://expo.fyi/eas-build-queues'
+      )}`
     );
     throw new Error('Build request failed.');
   } else if (
     error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED_ANDROID'
   ) {
     Log.error(
-      `EAS Build free tier is temporarily disabled for Android. Try again later. Check ${link(
-        'https://status.expo.dev/'
-      )} for updates. ${learnMore('https://expo.fyi/eas-build-queues')}`
+      `EAS Build free tier is temporarily disabled for Android and we are not accepting new builds. Try again later. ${learnMore(
+        'https://expo.fyi/eas-build-queues'
+      )}`
     );
     throw new Error('Build request failed.');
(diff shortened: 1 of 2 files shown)


wkozyra95 (approved):

packages/eas-cli/src/build/build.ts
```diff
@@ -140,12 +140,34 @@ function handleBuildRequestError(error: any, platform: Platform): never {
...
+      )} for updates. ${learnMore('https://expo.fyi/eas-build-queues')}`
+    );
+    throw new Error('Build request failed.');
+  } else if (
+    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED_ANDROID'
+  ) {
+    Log.error(
+      `EAS Build free tier is temporarily disabled for Android. Try again later. Check ${link(
+        'https://status.expo.dev/'
+      )} for updates. ${learnMore('https://expo.fyi/eas-build-queues')}`
```
szdziedzic, line 170: @brentvatne Are these error messages correct? As far as I understand, these are the user-facing ones 🙂

kbrandwijk: I'm not 100% sold on the wording, as the situation is that we stop accepting new builds, not that we've disabled EAS Build, as all queued builds will be processed, right?

Second, would we always have a Statuspage incident when using these kill switches? The error message makes that a requirement. And how would this interact with the other feature we added for showing the user any statuspage incident if there is one?

kbrandwijk: I would drop the status page part.

packages/eas-cli/src/build/build.ts
```diff
@@ -140,12 +140,34 @@ function handleBuildRequestError(error: any, platform: Platform): never {
...
+        'https://expo.fyi/eas-build-queues'
+      )}`
+    );
+    throw new Error('Build request failed.');
+  } else if (
+    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED_ANDROID'
+  ) {
+    Log.error(
+      `EAS Build free tier is temporarily disabled for Android and we are not accepting any new builds. Try again later. ${learnMore(
+        'https://expo.fyi/eas-build-queues'
```
szdziedzic, line 169: @kbrandwijk What do you think about it?

kbrandwijk: I like

kbrandwijk (approved):

packages/eas-cli/src/build/build.ts
```diff
@@ -140,12 +140,34 @@ function handleBuildRequestError(error: any, platform: Platform): never {
...
+      `EAS Build free tier is temporarily disabled for iOS and we are not accepting any new builds. Try again later. ${learnMore(
+        'https://expo.fyi/eas-build-queues'
+      )}`
+    );
+    throw new Error('Build request failed.');
+  } else if (
+    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_FREE_TIER_DISABLED_ANDROID'
+  ) {
+    Log.error(
+      `EAS Build free tier is temporarily disabled for Android and we are not accepting any new builds. Try again later. ${learnMore(
```
brentvatne, line 168:
```suggestion
      `Free Android builds are temporarily disabled. ${learnMore(
```

what do you guys think about this? no fluff, just gets right to the point

brentvatne: (also slightly more accurate because you can be on an "on demand" plan and see this message)

brentvatne (approved):
i slightly prefer the wording that i suggested but i am ok with either

packages/eas-cli/src/build/build.ts
szdziedzic: I usually prefer error messages which get straight to the point, so I like this one, although I don't have a strong opinion, and the version proposed by @kbrandwijk is also good for me.

brentvatne: @kbrandwijk - your call

kbrandwijk: Given how sensitive users (suddenly) are to nuance when it comes to something that's not working/not available, it'd go with the longer version we discussed above in my comment.

szdziedzic:
/changelog-entry chore Handle errors from platform-specific kill switches to disable free tier builds

PR #1416, opened 2022-09-30
[eas-cli] Add metadata validation command with complex rules

byCedric:
# Checklist

# Why

Fixes ENG-5287
Fixes ENG-5798

# How

- Added complex custom rule validation
- Updated AJV to latest, with "human errors" plugin to make json schema errors more readable
- Allow creating metadata context without authenticating (for lint command)
- Added custom `eas metadata:lint` command

# Test Plan

- `$ eas metadata:lint`
- `$ eas metadata:lint --json`


packages/eas-cli/src/commands/metadata/lint.ts
```diff
@@ -0,0 +1,68 @@
+import { getConfig } from '@expo/config';
+import { Flags } from '@oclif/core';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import Log from '../../log';
+import { loadConfigAsync } from '../../metadata/config/resolve';
+import { createMetadataContextAsync } from '../../metadata/context';
+import { MetadataValidationError, logMetadataValidationError } from '../../metadata/errors';
+import { findProjectRootAsync } from '../../project/projectUtils';
```
EvanBacon, line 9: Maybe I'm missing something but I don't see this function in the import file

packages/eas-cli/package.json
```diff
@@ -29,9 +29,11 @@
     "@expo/spawn-async": "1.7.0",
     "@expo/timeago.js": "1.0.0",
     "@oclif/core": "1.13.10",
+    "@segment/ajv-human-errors": "^2.1.2",
     "@urql/core": "2.6.1",
     "@urql/exchange-retry": "0.3.3",
-    "ajv": "^6.12.6",
+    "ajv": "^8.11.0 ",
```
byCedric, line 35:
```suggestion
    "ajv": "^8.11.0",
```

packages/eas-cli/src/commands/metadata/lint.ts
byCedric: Yeah, this was the previous API, before Will's changes.

byCedric:
/changelog-entry new-feature Add custom metadata validation command with more complex rules

packages/eas-cli/src/metadata/utils/array.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Filter the array using a truthy predicate, returning a type-safe truthy array.
+ */
+export function coalesce<T>(list: (T | undefined | null)[]): T[] {
+  return list.filter(Boolean) as T[];
+}
```
dsokal, line 6: - The function name is misleading. Usually, "coalesce" is used for applying fallback values to potentially falsy values, sth like `const coalesce = (x?: number) => x ?? 2137;`. I'd be explicit here and change the name to sth like "filterOutFalsyValues" (this can be too explicit, leaving it up to you).
- Don't use `Boolean()`. `Boolean(0) === false` :) 
  Screenshot 2022-10-11 at 11 05 46
- You don't need `as T[]` part, see below.
- Maybe move the function to the `expodash` folder? I can see it can be used in other places too.

```suggestion
/**
 * Filter the array using a truthy predicate, returning a type-safe truthy array.
 */
export function CHANGETHENAME<T>(list: (T | undefined | null)[]): T[] {
  return list.filter<T>((i): i is T =>i !== undefined && i !== null);
}
```

packages/eas-cli/src/metadata/utils/ajv.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @see https://github.com/ajv-validator/ajv-formats
+ */
+export function createValidator(): Ajv {
+  const validator = new Ajv({
+    strict: false, // The metadata schema is shared with vscode, including vscode-only properties
+    verbose: true, // Required for `ajv-human-errors`
+    allErrors: true, // Required for `ajv-human-errors`
+  });
+
+  return addFormats(validator).addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'));
```
dsokal, line 18: Move the require to the top-level scope. This way if something changes in `ajv`, we can know about it earlier than when calling the function.

packages/eas-cli/src/metadata/utils/ajv.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function getReadableErrors(errors: ErrorObject[] = []): HumanError[] {
+  if (!errors.length) {
```
dsokal, line 22: Don't use implicit coercion, you can read more about it here - https://www.oreilly.com/library/view/you-dont-know/9781491905159/ch04.html
```suggestion
  if (errors.length !== 0) {
```

packages/eas-cli/src/metadata/errors.ts
```diff
@@ -55,7 +55,14 @@ export function logMetadataValidationError(error: MetadataValidationError): void
...
-    Log.log(error.errors.map(err => `  - ${chalk.bold(err.dataPath)} ${err.message}`).join('\n'));
+    Log.log(
+      error.errors
+        .map(err => {
+          const dataPath = err.path.join('.');
+          return `  - ${chalk.bold(`$.${dataPath}`)} ${err.message}`;
+        })
+        .join('\n')
+    );
   }
```
dsokal, line 66: Refactor to avoid using map inside Log.log:
```
for (const err of error.errors) {
  Log.log(...);
}
```

packages/eas-cli/src/metadata/context.ts
```diff
@@ -52,8 +53,8 @@ export async function createMetadataContextAsync(params: {
     params.profileName
   );
 
-  const exp = params.exp;
-  const user = params.credentialsCtx.user;
+  const exp = params.exp ?? getExpoConfig(params.projectDir);
+  const user = params.credentialsCtx ? params.credentialsCtx.user : undefined;
```
dsokal, line 57: I think this is the same:
```suggestion
  const user = params.credentialsCtx?.user;
```

packages/eas-cli/src/metadata/apple/rules/infoRestrictedWords.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Apple restricts certain words from being used in name, description, or keywords.
+ * Using these words likely result in a rejection.
+ */
+export const infoRestrictedWords: IssueRule = {
+  id: 'apple.info.restrictedWords',
+  severity: 1,
+  validate(config) {
+    if (!config.apple?.info || !Object.keys(config.apple.info).length) {
```
dsokal, line 18:
```suggestion
    if (!config.apple?.info || Object.keys(config.apple.info).length === 0) {
```

packages/eas-cli/src/metadata/config/resolve.ts
```diff
@@ -67,38 +60,21 @@ export async function loadConfigAsync({
...
-  return { valid, errors: validator.errors || [] };
-}
-
 /** Create a versioned deserializer to fetch App Store data from the store configuration. */
 export function createAppleReader(config: MetadataConfig): AppleConfigReader {
   assert(config.configVersion === 0, 'Unsupported store configuration version');
-  assert(config.apple, 'No apple configuration found');
-  return new AppleConfigReader(config.apple);
+  assert(config.apple !== undefined, 'No apple configuration found');
+  return new AppleConfigReader(config.apple!);
```
dsokal, line 77: Don't use `!`, this is an antipattern.

packages/eas-cli/src/metadata/config/resolve.ts
```diff
@@ -67,38 +60,21 @@ export async function loadConfigAsync({
   const configData = await resolveDynamicConfigAsync(configFile);
 
   if (!skipValidation) {
-    const { valid, errors: validationErrors } = validateConfig(configData);
+    const issues = validateConfig(configData);
 
-    if (!valid) {
-      throw new MetadataValidationError(`Metadata store config errors found`, validationErrors);
+    if (issues.length) {
```
dsokal, line 65:
```suggestion
    if (issues.length > 0) {
```

dsokal (changes requested):

packages/eas-cli/src/metadata/config/validate.ts
```diff
@@ -0,0 +1,50 @@
+import { appleRules } from '../apple/rules';
+import { createValidator, getReadableErrors } from '../utils/ajv';
+import { Issue, IssueRule, IssueSeverity } from './issue';
+import { MetadataConfig } from './schema';
+
+/** The pre-compiled AJV validator instance, with the metadata spec */
+const validator = createValidator().compile(require('../../../schema/metadata-0.json'));
```
wkozyra95, line 7: why does this need to be global? I'm not sure how heavy this part is but by doing it here you are adding overhead to every command in cli. If I understand correctly oclif does not lazy load commands, so this code will run every time.(I might be wrong about it)

byCedric:
/changelog-entry new-feature Add custom metadata validation command with more complex rules

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This is the log of notable changes to EAS CLI and related packages.
 - Improve `eas init` command. ([#1376](https://github.com/expo/eas-cli/pull/1376) by [@wschurman](https://github.com/wschurman))
 - Warn about outdated channel configuration before build when using eas update. ([#1397](https://github.com/expo/eas-cli/pull/1397) by [@kbrandwijk](https://github.com/kbrandwijk))
 - Validate that owner and projectId and slug all are in alignment. ([#1405](https://github.com/expo/eas-cli/pull/1405) by [@wschurman](https://github.com/wschurman))
+- Add custom metadata validation command with more complex rules. ([#1416](https://github.com/expo/eas-cli/pull/1416) by [@byCedric](https://github.com/byCedric))
```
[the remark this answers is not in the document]
byCedric: Done, see https://github.com/expo/eas-cli/pull/1416#issuecomment-1277527891

packages/eas-cli/src/metadata/utils/array.ts
byCedric: How about naming it just "filter"? It's the underlying action, while doing some typescript cleanups.

I'll move it to expodash

packages/eas-cli/src/metadata/errors.ts
byCedric: The main reason I used 1 log is performance, but I might be wrong here. I'll change it to a simple loop.

packages/eas-cli/src/metadata/utils/array.ts
dsokal: `filter` is fine

packages/eas-cli/src/metadata/errors.ts
dsokal: Performance shouldn't be a problem here :)

packages/eas-cli/src/commands/metadata/lint.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { flags } = await this.parse(MetadataLint);
+    const { projectDir } = await this.getContextAsync(MetadataLint, {
+      nonInteractive: false,
+    });
+
+    if (flags.json) {
+      enableJsonOutput();
+    }
+
+    const { exp } = getConfig(projectDir);
```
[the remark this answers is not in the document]
dsokal: So maybe we need the `ProjectConfig` context that doesn't require authentication? Can you talk with @wschurman about that? 🙏

packages/eas-cli/src/metadata/utils/ajv.ts
byCedric: @wkozyra95 made a good point about the EAS CLI not being lazy-loaded ([see comment](https://github.com/expo/eas-cli/pull/1416#discussion_r993154086)). I think it doesn't make sense to load these JSON files every time, but I don't have a strong opinion on that though.

packages/eas-cli/src/commands/metadata/lint.ts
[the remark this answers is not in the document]
byCedric: Yep, will do!

packages/eas-cli/src/metadata/config/validate.ts
byCedric: You are right, it doesn't need to be. AJV does recommend it, but I think that's mostly for backends. Will move it to the function instead

packages/eas-cli/src/metadata/utils/ajv.ts
wkozyra95: I was thinking there about calling compile in the global scope(I'm not sure how heavy it is). Importing in global scope should be fine.

packages/eas-cli/src/commands/metadata/lint.ts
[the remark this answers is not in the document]
byCedric: > Why do you need to have a project config to validate the metadata file? 

It's required to resolve the right `bundleIdentifier`, either from the profile or resolved Expo config. But we don't use that when linting.

We could split that context up different resolve functions, for example:
- The base information for metadata (including store config file, profile, platform)
- App store authentication for metadata (the base + bundleIdentifier - the app ID in the app store, credentials ctx)

packages/eas-cli/src/metadata/context.ts
```diff
@@ -41,8 +42,8 @@ export type MetadataAppStoreAuthentication = {
  */
 export async function createMetadataContextAsync(params: {
   projectDir: string;
-  credentialsCtx: CredentialsContext;
-  exp: ExpoConfig;
+  credentialsCtx?: CredentialsContext;
+  exp?: ExpoConfig;
```
wschurman, line 46: I think this can be left as non-nullable and the fallback below be removed?

packages/eas-cli/src/utils/expodash/filter.ts
```diff
@@ -0,0 +1,4 @@
+/** Filter the array from `null` and `undefined` values, returning a type-safe array. */
+export default function filter<T>(list: (T | undefined | null)[]): T[] {
+  return list.filter<T>((i): i is T => i !== undefined && i !== null);
+}
```
wschurman, line 4: optional: alternatively you could use predicates in combination with the builtin `filter` method. From www:
```typescript
export function truthy<TValue>(value: TValue | null | undefined): value is TValue {
  return !!value;
}

const b = a.filter(truthy);
```

packages/eas-cli/src/commands/metadata/lint.ts
```diff
@@ -0,0 +1,69 @@
...
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectDir,
+  };
+
+  async runAsync(): Promise<void> {
+    Log.warn('EAS Metadata is in beta and subject to breaking changes.');
+
+    const { flags } = await this.parse(MetadataLint);
+    const { projectDir } = await this.getContextAsync(MetadataLint, {
+      nonInteractive: false,
```
wschurman, line 34: optional: this command looks fairly non-interactive so this could maybe be set to true?

byCedric: Yes, we are blocked on setting non-interactive to true for both `eas metadata:pull` and `eas metadata:push`, caused by the `credentials/ios/appstore/AppStoreApi.ts` (used in the `CredentialsContext`) using methods that could prompt users.

But because the `eas metadata:lint` command doesn't use the credentials part, we can set this to non interactive.

packages/eas-cli/src/utils/expodash/filter.ts
byCedric: I like that pattern, but I'm unsure how it would fit in the `expodash` functions. We can just drop them in there and go from there? 😄 (cc @dsokal)

packages/eas-cli/src/commands/metadata/lint.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { flags } = await this.parse(MetadataLint);
+    const { projectDir } = await this.getContextAsync(MetadataLint, {
+      nonInteractive: true,
+    });
+
+    if (flags.json) {
+      enableJsonOutput();
+    }
+
+    const { exp } = getConfig(projectDir);
```
wschurman, line 41: This should be fine, just add a note in a comment above it about why this isn't gotten from a command context.

packages/eas-cli/src/metadata/context.ts
wschurman: Ah, cool, yeah splitting it up would work as well.

wschurman (commented):
I'll leave review to the people more familiar with metadata. Thanks for the info about why project config context wasn't correct for this particular command!

packages/eas-cli/src/utils/expodash/filter.ts
dsokal: I think it fits in expodash well, so go ahead :)

dsokal (approved):
LGTM 
Last outstanding comment - https://github.com/expo/eas-cli/pull/1416/files#r992056426

packages/eas-cli/src/metadata/utils/ajv.ts
byCedric: Ah gotcha. Well, in that case, we can import the JSON files top level. I will make that change after I split up the metadata context.

packages/eas-cli/src/metadata/context.ts
byCedric: Done!

packages/eas-cli/src/commands/metadata/lint.ts
byCedric: Done, and `getConfig` is removed.

packages/eas-cli/src/utils/expodash/filter.ts
byCedric: Done!

byCedric: Done!

packages/eas-cli/src/commands/metadata/lint.ts
```diff
@@ -0,0 +1,79 @@
...
+    profile: Flags.string({
+      description:
+        'Name of the submit profile from eas.json. Defaults to "production" if defined in eas.json.',
+    }),
+  };
+
+  static override contextDefinition = {
+    // The metadata lint command is created to integrate in other dev tooling, like vscode-expo.
+    // These integrations might spam this command, so we avoid communicating with our services here.
+    // Note that this is an exception and you should normally use `ProjectConfig` instead.
```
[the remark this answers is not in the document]
byCedric: I think it reflected on using `ProjectDir` instead of `ProjectConfig`, right? Can remove it, but I believe this was put to discourage using `ProjectDir` directly.

Issue #1349, opened 2022-09-08
[tvOS] eas-cli build doesn't create a proper Provisioning Profile for tvOS


danychi:
I was trying to generate a tvOS build using `eas-cli build --platform="ios"` and the process was returning this kind of error:
```
 error: No profile for team 'TEAM' matching '*[expo] bundle.id AppStore 2022-08-25T15:51:50.365Z' found: Xcode couldn't find any provisioning profiles matching 'TEAM/*[expo] bundle.id AppStore 2022-08-25T15:51:50.365Z'. Install the profile (by dragging and dropping it onto Xcode's dock item) or select a different one in the Signing & Capabilities tab of the target editor. (in target 'TARGET-tvOS' from project 'TARGET')
``` 
The reason is that by default eas creates an `App Store` provisioning profile type, while the project needs a `tvOS App Store` provisioning profile type. 

<img width="1063" alt="Screenshot 2022-09-08 at 14 13 56" src="https://user-images.githubusercontent.com/45874461/189119640-fed2d7c3-dfd9-4353-8af7-7986693495a0.png">

So after creating the provisioning profile on https://developer.apple.com/account/resources/profiles/list
and uploading it to https://expo.dev/accounts/TEAM/projects/PROJECT_NAME/credentials then everything works perfectly 👌

quinlanj:
@danychi just to sanity check since tvOS support was added not too long ago:
- what version of `eas-cli` do you have? 
- did you ever run `eas build` for your tvOS project on a cli version earlier than 2.1.0?

quinlanj:
Closing this issue for now. @danychi if you run into the same problem again, please open up another issue and:
- reference this issue
- let us know what version of `eas-cli` you are running
- let us know what changes you've made in your project to build for tvOS (configuration changes, XCode build settings, etc)

PR #1417, opened 2022-09-30
[cli] more robust platform check

quinlanj:
# Checklist

# Why

We don't have any docs on how our users should build their project for tvOS yet (assuming it will be released after we complete https://linear.app/expo/issue/ENG-5930/research-requirements-to-make-apple-tv-targets-buildable). I'm not really sure what users are doing to build for tvOS but having a more robust check seems like a good improvement and will probably help https://github.com/expo/eas-cli/issues/1349  .

After talking to @douglowder , we should be looking at the following Xcode build settings to determine which Apple Platform we should be building for:
- SDK_ROOT
- TARGETED_DEVICE_FAMILY

# Test Plan

new tests pass

### DIFF
--- a/packages/eas-cli/src/project/ios/__tests__/target-test.ts
+++ b/packages/eas-cli/src/project/ios/__tests__/target-test.ts
@@ -1,5 +1,6 @@
 import { testTarget } from '../../../credentials/__tests__/fixtures-ios';
 import { ApplePlatform } from '../../../credentials/ios/appstore/constants';
+import { Target } from '../../../credentials/ios/types';
 import { getApplePlatformFromDeviceFamily, getApplePlatformFromSdkRoot } from '../target';
 
 function getApplePlatformWithSdkRoot(sdkRoot: string): ApplePlatform | null {
--- a/packages/eas-cli/src/project/ios/__tests__/target-test.ts
+++ b/packages/eas-cli/src/project/ios/__tests__/target-test.ts
@@ -1,8 +1,8 @@
 import { testTarget } from '../../../credentials/__tests__/fixtures-ios';
 import { ApplePlatform } from '../../../credentials/ios/appstore/constants';
-import { getApplePlatformFromSdkRoot } from '../target';
+import { getApplePlatformFromDeviceFamily, getApplePlatformFromSdkRoot } from '../target';
 
-function getApplePlatform(sdkRoot: string): ApplePlatform {
+function getApplePlatformWithSdkRoot(sdkRoot: string): ApplePlatform | null {
   const target = {
     ...testTarget,
     buildSettings: {
@@ -10,6 +10,18 @@ function getApplePlatform(sdkRoot: string): ApplePlatform {
     },
   };
   return getApplePlatformFromSdkRoot(target);
+}
+
+function getApplePlatformWithDeviceFamily(
+  deviceFamily: string | number | undefined
+): ApplePlatform | null {
+  const target = {
+    ...testTarget,
+    buildSettings: {
+      TARGETED_DEVICE_FAMILY: deviceFamily,
+    },
+  };
+  return getApplePlatformFromDeviceFamily(target);
 }
 
 describe(getApplePlatformFromSdkRoot, () => {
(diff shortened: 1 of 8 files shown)


quinlanj:
/changelog-entry bug-fix Make Apple Platform detection more robust.

packages/eas-cli/src/project/ios/__tests__/target-test.ts
```diff
@@ -21,13 +33,24 @@ describe(getApplePlatformFromSdkRoot, () => {
...
+    expect(getApplePlatformWithSdkRoot('macosx11.1')).toBe(ApplePlatform.MAC_OS);
+    expect(getApplePlatformWithSdkRoot('appletvos14.3')).toBe(ApplePlatform.TV_OS);
+    expect(getApplePlatformWithSdkRoot('appletvsimulator14.3')).toBe(null);
+    expect(getApplePlatformWithSdkRoot('watchos7.2')).toBe(null);
+    expect(getApplePlatformWithSdkRoot('watchsimulator7.2')).toBe(null);
+  });
+});
+
+describe(getApplePlatformFromDeviceFamily, () => {
+  test('apple platform can be obtained from device family types', () => {
```
douglowder, line 48: Most common case will be "1,2" (universal iPhone and iPad). Can you add a test for that?

douglowder (changes requested):
See inline

packages/eas-cli/src/project/ios/target.ts
```diff
@@ -244,6 +256,32 @@ export function getApplePlatformFromSdkRoot(target: Target): ApplePlatform {
...
+  }
+}
+
+export function getApplePlatformFromDeviceFamily(target: Target): ApplePlatform | null {
+  const deviceFamily = target.buildSettings?.TARGETED_DEVICE_FAMILY;
+  if (!deviceFamily) {
+    return null;
+  }
+  if (typeof deviceFamily === 'string') {
+    const deviceFamilyNumber = parseInt(deviceFamily, 10);
```
dsokal, line 269: After applying Doug's comment (https://github.com/expo/eas-cli/pull/1417/files#r984991660) I think it's better to change it so that we don't interpret "1,2" as "1".
Screenshot 2022-10-03 at 11 27 21

More general comment: it's good practice to avoid using `parseInt` as the algorithm is not as obvious as people think. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt. Use `Number("some string")` if you need to convert a string to a number.

dsokal (commented):
LGTM but leaving the decision on details to @douglowder

quinlanj:
ok so based on the apple docs and an arcane stack overflow comment, the value of `TARGETED_DEVICE_FAMILY` can be:
- a csv string (in the case where there are multiple devices)
- a number for a single device
- undefined

packages/eas-cli/src/project/ios/target.ts
```diff
@@ -244,6 +256,43 @@ export function getApplePlatformFromSdkRoot(target: Target): ApplePlatform {
...
+ * https://stackoverflow.com/questions/39677524/xcode-8-how-to-change-targeted-device-family#comment100316573_39677659
+ *
+ * @returns - Apple Platform when known, defaults to null when unknown
+ */
+export function getApplePlatformFromDeviceFamily(target: Target): ApplePlatform | null {
+  const deviceFamily = target.buildSettings?.TARGETED_DEVICE_FAMILY;
+  if (!deviceFamily) {
+    return null;
+  }
+  if (typeof deviceFamily === 'string') {
```
douglowder, line 277: I would maybe suggest a simple lookup from a map... maybe make this change?

```patch
diff --git a/packages/eas-cli/src/project/ios/target.ts b/packages/eas-cli/src/project/ios/target.ts
index 0ccda087..0ee86e88 100644
--- a/packages/eas-cli/src/project/ios/target.ts
+++ b/packages/eas-cli/src/project/ios/target.ts
@@ -275,24 +275,18 @@ export function getApplePlatformFromDeviceFamily(target: Target): ApplePlatform
     return null;
   }
   if (typeof deviceFamily === 'string') {
-    const devices = deviceFamily.split(',');
-    const arbitraryDevice = devices[0];
-    const deviceFamilyNumber = Number(arbitraryDevice);
-    return deviceFamilyToPlatform(deviceFamilyNumber);
+    return deviceFamilyToPlatformMap[deviceFamily] || null;
   } else if (typeof deviceFamily === 'number') {
-    return deviceFamilyToPlatform(deviceFamily);
+    return deviceFamilyToPlatformMap[`${deviceFamily}`] || null;
   }
   throw new Error(
     `Unexpected device family type in XCode build settings: ${JSON.stringify(deviceFamily)}`
   );
 }
 
-function deviceFamilyToPlatform(deviceFamily: number): ApplePlatform | null {
-  if (deviceFamily === 1 || deviceFamily === 2) {
-    return ApplePlatform.IOS;
-  } else if (deviceFamily === 3) {
-    return ApplePlatform.TV_OS;
-  } else {
-    return null;
-  }
-}
+const deviceFamilyToPlatformMap: { [index: string]: ApplePlatform } = {
+  '1,2': ApplePlatform.IOS,
+  '1': ApplePlatform.IOS,
+  '2': ApplePlatform.IOS,
+  '3': ApplePlatform.TV_OS,
+};
```

packages/eas-cli/src/project/ios/__tests__/target-test.ts
```diff
@@ -12,6 +12,18 @@ function getApplePlatform(sdkRoot: string): ApplePlatform {
   return getApplePlatformFromSdkRoot(target);
 }
 
+function getApplePlatformWithDeviceFamily(
+  deviceFamily: string | number | undefined
+): ApplePlatform | null {
+  const target = {
```
dsokal, line 18:
```suggestion
  const target: Target = {
```

dsokal (approved):

packages/eas-cli/src/project/ios/target.ts
wkozyra95: I don't think this lookup is a safe approach, this is a comma-separated list of values, so `2,1` is also valid

https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html

wkozyra95 (approved):

packages/eas-cli/src/project/ios/target.ts
douglowder: Ah true... I've never seen `2,1` appear but it is indeed valid. Good catch :) I withdraw my suggestion

douglowder (approved):

PR #1427, opened 2022-10-07
[eas-cli] Update: improve DX on first run, correct error message

Simek:
# Checklist

# Why

Fixes ENG-6013
Fixes ENG-6014

# How

This changeset adds the prompt for the user when there is no runtime version specified, which asks if setup the updates automatically for them (via `eas update:configure`).

If user do not want this the CLI exits as before.

If user selects "Yes", we are programmatically running `eas update:configure`, and then re-reading the modified config, so the process can continue without the interruption or command exit.

As always, the copy/messages content probably can be written better. 😉 

# Test Plan

The changes have been tested locally in few project by running `dev` bin of EAS CLI.

# Preview

Screenshot 2022-10-07 at 21 43 45

Screenshot 2022-10-07 at 21 40 31

### DIFF
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -6,6 +6,7 @@ import chalk from 'chalk';
 
 import { getEASUpdateURL } from '../../api';
 import EasCommand from '../../commandUtils/EasCommand';
+import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
 import { AppPlatform } from '../../graphql/generated';
 import Log, { learnMore } from '../../log';
 import { RequestedPlatform, appPlatformDisplayNames } from '../../platform';
@@ -70,6 +71,7 @@ export default class UpdateConfigure extends EasCommand {
       exp: updatedExp,
       platform,
       workflows,
+      graphqlClient,
     });
 
     Log.addNewLineIfNone();
@@ -296,19 +298,20 @@ async function configureNativeFilesForEASUpdateAsync({
   platform,
   workflows,
   projectId,
-}: ConfigureForEASUpdateArgs): Promise<void> {
+  graphqlClient,
+}: ConfigureForEASUpdateArgs & { graphqlClient: ExpoGraphqlClient }): Promise<void> {
   if (
     [RequestedPlatform.Android, RequestedPlatform.All].includes(platform) &&
     workflows.android === Workflow.GENERIC
   ) {
-    await syncAndroidUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
+    await syncAndroidUpdatesConfigurationAsync(graphqlClient, projectDir, exp, projectId);
     Log.withTick(`Configured ${chalk.bold('AndroidManifest.xml')} for EAS Update`);
   }
   if (
     [RequestedPlatform.Ios, RequestedPlatform.All].includes(platform) &&
     workflows.ios === Workflow.GENERIC
   ) {
-    await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
+    await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, exp, projectId);
     Log.withTick(`Configured ${chalk.bold('Expo.plist')} for EAS Update`);
   }
 }
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -1,5 +1,5 @@
 import { ExpoConfig, modifyConfigAsync } from '@expo/config';
-import { Platform, Workflow } from '@expo/eas-build-job';
+import { Workflow } from '@expo/eas-build-job';
 import { Flags } from '@oclif/core';
 import assert from 'assert';
 import chalk from 'chalk';
@@ -13,7 +13,8 @@ import {
   installExpoUpdatesAsync,
   isExpoUpdatesInstalledOrAvailable,
 } from '../../project/projectUtils';
-import { resolveWorkflowAsync } from '../../project/workflow';
+import { getCombinedWorkflowDataAsync } from '../../project/workflow';
+import type { CombinedWorkflowData } from '../../project/workflow';
 import { syncUpdatesConfigurationAsync as syncAndroidUpdatesConfigurationAsync } from '../../update/android/UpdatesModule';
 import { syncUpdatesConfigurationAsync as syncIosUpdatesConfigurationAsync } from '../../update/ios/UpdatesModule';
 
@@ -54,34 +55,28 @@ export default class UpdateConfigure extends EasCommand {
       await installExpoUpdatesAsync(projectDir);
     }
 
-    const [androidWorkflow, iosWorkflow] = await Promise.all([
-      resolveWorkflowAsync(projectDir, Platform.ANDROID),
-      resolveWorkflowAsync(projectDir, Platform.IOS),
-    ]);
-
+    const workflows = await getCombinedWorkflowDataAsync(projectDir);
     const updatedExp = await configureAppJSONForEASUpdateAsync({
       projectDir,
+      projectId,
       exp,
       platform,
-      workflows: {
-        android: androidWorkflow,
-        ios: iosWorkflow,
-      },
-      projectId,
+      workflows,
     });
+
     Log.withTick(`Configured ${chalk.bold('app.json')} for EAS Update`);
 
     // configure native files for EAS Update
     if (
       [RequestedPlatform.Android, RequestedPlatform.All].includes(platform) &&
-      androidWorkflow === Workflow.GENERIC
+      workflows.android === Workflow.GENERIC
     ) {
       await syncAndroidUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
       Log.withTick(`Configured ${chalk.bold('AndroidManifest.xml')} for EAS Update`);
     }
     if (
       [RequestedPlatform.Ios, RequestedPlatform.All].includes(platform) &&
-      iosWorkflow === Workflow.GENERIC
+      workflows.ios === Workflow.GENERIC
     ) {
       await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
       Log.withTick(`Configured ${chalk.bold('Expo.plist')} for EAS Update`);
@@ -92,31 +87,31 @@ export default class UpdateConfigure extends EasCommand {
   }
 }
 
-async function configureAppJSONForEASUpdateAsync({
+type ConfigureAppJSONForEASUpdateArgs = {
+  projectDir: string;
+  exp: ExpoConfig;
+  platform: RequestedPlatform;
+  workflows: CombinedWorkflowData;
+  projectId: string;
+};
+
+export async function configureAppJSONForEASUpdateAsync({
   projectDir,
   exp,
   platform,
   workflows,
   projectId,
-}: {
-  projectDir: string;
-  exp: ExpoConfig;
-  platform: RequestedPlatform;
-  workflows: {
-    [key in RequestedPlatform.Android | RequestedPlatform.Ios]: Workflow;
-  };
-  projectId: string;
-}): Promise<ExpoConfig> {
+}: ConfigureAppJSONForEASUpdateArgs): Promise<ExpoConfig> {
   // this command is non-interactive in the way it was designed
   const easUpdateURL = getEASUpdateURL(projectId);
   const updates = { ...exp.updates, url: easUpdateURL };
 
   const androidDefaultRuntimeVersion =
-    workflows['android'] === Workflow.GENERIC
+    workflows.android === Workflow.GENERIC
       ? DEFAULT_BARE_RUNTIME_VERSION
       : DEFAULT_MANAGED_RUNTIME_VERSION;
   const iosDefaultRuntimeVersion =
-    workflows['ios'] === Workflow.GENERIC
+    workflows.ios === Workflow.GENERIC
       ? DEFAULT_BARE_RUNTIME_VERSION
       : DEFAULT_MANAGED_RUNTIME_VERSION;
 
@@ -277,9 +272,9 @@ async function configureAppJSONForEASUpdateAsync({
         )}`
       );
       Log.warn(
-        `In order to finish configuring your project for EAS Update, you are going to need manually add the following to your app.config.js:\n${learnMore(
-          'https://expo.fyi/eas-update-config.md'
-        )}\n`
+        `In order to finish configuring your project for EAS Update, you are going to need manually add the following to your ${chalk.bold(
+          'app.config.js'
+        )}:\n${learnMore('https://expo.fyi/eas-update-config.md')}\n`
       );
       Log.log(chalk.bold(JSON.stringify(newConfigOnlyAddedValues, null, 2)));
       Log.addNewLineIfNone();
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -292,7 +292,7 @@ export async function configureAppJSONForEASUpdateAsync({
   return result.config.expo;
 }
 
-async function configureNativeFilesForEASUpdateAsync({
+export async function configureNativeFilesForEASUpdateAsync({
   projectDir,
   exp,
   platform,
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -56,6 +56,7 @@ export default class UpdateConfigure extends EasCommand {
     }
 
     const workflows = await getCombinedWorkflowDataAsync(projectDir);
+
     const updatedExp = await configureAppJSONForEASUpdateAsync({
       projectDir,
       projectId,
@@ -64,30 +65,20 @@ export default class UpdateConfigure extends EasCommand {
       workflows,
     });
 
-    Log.withTick(`Configured ${chalk.bold('app.json')} for EAS Update`);
-
-    // configure native files for EAS Update
-    if (
-      [RequestedPlatform.Android, RequestedPlatform.All].includes(platform) &&
-      workflows.android === Workflow.GENERIC
-    ) {
-      await syncAndroidUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
-      Log.withTick(`Configured ${chalk.bold('AndroidManifest.xml')} for EAS Update`);
-    }
-    if (
-      [RequestedPlatform.Ios, RequestedPlatform.All].includes(platform) &&
-      workflows.ios === Workflow.GENERIC
-    ) {
-      await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
-      Log.withTick(`Configured ${chalk.bold('Expo.plist')} for EAS Update`);
-    }
+    await configureNativeFilesForEASUpdateAsync({
+      projectDir,
+      projectId,
+      exp: updatedExp,
+      platform,
+      workflows,
+    });
 
     Log.addNewLineIfNone();
     Log.log(`🎉 Your app is configured to run EAS Update!`);
   }
 }
 
-type ConfigureAppJSONForEASUpdateArgs = {
+type ConfigureForEASUpdateArgs = {
   projectDir: string;
   exp: ExpoConfig;
   platform: RequestedPlatform;
@@ -101,7 +92,7 @@ export async function configureAppJSONForEASUpdateAsync({
   platform,
   workflows,
   projectId,
-}: ConfigureAppJSONForEASUpdateArgs): Promise<ExpoConfig> {
+}: ConfigureForEASUpdateArgs): Promise<ExpoConfig> {
   // this command is non-interactive in the way it was designed
   const easUpdateURL = getEASUpdateURL(projectId);
   const updates = { ...exp.updates, url: easUpdateURL };
@@ -295,7 +286,32 @@ export async function configureAppJSONForEASUpdateAsync({
   }
   assert(result.config, 'A successful result should have a config');
 
+  Log.withTick(`Configured ${chalk.bold('app.json')} for EAS Update`);
+
   return result.config.expo;
+}
+
+async function configureNativeFilesForEASUpdateAsync({
+  projectDir,
+  exp,
+  platform,
+  workflows,
+  projectId,
+}: ConfigureForEASUpdateArgs): Promise<void> {
+  if (
+    [RequestedPlatform.Android, RequestedPlatform.All].includes(platform) &&
+    workflows.android === Workflow.GENERIC
+  ) {
+    await syncAndroidUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
+    Log.withTick(`Configured ${chalk.bold('AndroidManifest.xml')} for EAS Update`);
+  }
+  if (
+    [RequestedPlatform.Ios, RequestedPlatform.All].includes(platform) &&
+    workflows.ios === Workflow.GENERIC
+  ) {
+    await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, updatedExp, projectId);
+    Log.withTick(`Configured ${chalk.bold('Expo.plist')} for EAS Update`);
+  }
 }
 
 function isRuntimeEqual(
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -1,5 +1,5 @@
 import { ExpoConfig, modifyConfigAsync } from '@expo/config';
-import { Workflow } from '@expo/eas-build-job';
+import { Platform, Workflow } from '@expo/eas-build-job';
 import { Flags } from '@oclif/core';
 import assert from 'assert';
 import chalk from 'chalk';
@@ -13,8 +13,7 @@ import {
   installExpoUpdatesAsync,
   isExpoUpdatesInstalledOrAvailable,
 } from '../../project/projectUtils';
-import { getCombinedWorkflowDataAsync } from '../../project/workflow';
-import type { CombinedWorkflowData } from '../../project/workflow';
+import { resolveWorkflowPerPlatformAsync } from '../../project/workflow';
 import { syncUpdatesConfigurationAsync as syncAndroidUpdatesConfigurationAsync } from '../../update/android/UpdatesModule';
 import { syncUpdatesConfigurationAsync as syncIosUpdatesConfigurationAsync } from '../../update/ios/UpdatesModule';
 
@@ -55,7 +54,7 @@ export default class UpdateConfigure extends EasCommand {
       await installExpoUpdatesAsync(projectDir);
     }
 
-    const workflows = await getCombinedWorkflowDataAsync(projectDir);
+    const workflows = await resolveWorkflowPerPlatformAsync(projectDir);
 
     const updatedExp = await configureAppJSONForEASUpdateAsync({
       projectDir,
@@ -82,7 +81,7 @@ type ConfigureForEASUpdateArgs = {
   projectDir: string;
   exp: ExpoConfig;
   platform: RequestedPlatform;
-  workflows: CombinedWorkflowData;
+  workflows: Record<Platform, Workflow>;
   projectId: string;
 };
 
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -595,7 +595,9 @@ function transformRuntimeVersions(exp: ExpoConfig, platforms: Platform[]): Recor
       platform,
       nullthrows(
         Updates.getRuntimeVersion(exp, platform),
-        `Unable to determine runtime version for ${requestedPlatformDisplayNames[platform]}. Learn more: https://docs.expo.dev/eas-update/runtime-versions/`
+        `Unable to determine runtime version for ${
+          requestedPlatformDisplayNames[platform]
+        }. ${chalk.dim('Learn more: https://docs.expo.dev/eas-update/runtime-versions/')}`
       ),
     ])
   );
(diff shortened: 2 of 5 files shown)


Simek:
/changelog-entry new-feature Update: Ask user to run `eas update:configure` automatically, if `runtimeVersion` is missing.
/changelog-entry chore Update: Improve the error message if `runtimeVersion` is missing.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +608,32 @@ async function getRuntimeVersionObjectAsync(
...
-  );
+  try {
+    return transformRuntimeVersions(exp, platforms);
+  } catch (error: any) {
+    ora().start().fail(error.message);
+
+    const runConfig = await selectAsync(`Do you want us to run 'eas update:configure' for you?`, [
+      { title: 'Yes', value: true },
+      {
+        title: 'No, I will edit files or run the command manually (EAS CLI exits)',
```
kbrandwijk, line 619:
```suggestion
        title: 'No, I will set the runtime version manually (EAS CLI exits)',
```

Simek: How strong is you opinion on that phrasing? 

The `eas update:configure` besides setting the runtime version performs edit of other files, to prepare the project for EAS Update, that's why I have decided to went with that sentence.

kbrandwijk: The 'I will edit files' part is a bit inconclusive, and you are only checking for runtime version, so it's not guaranteed that they need to run that command, they could have made any change that causes this. So if you're checking runtime version, I think the instructions should be as narrow as possible to set runtime version. Otherwise, instead of saying 'cannot detect runtime version' you should say 'it seems your project is not configured for EAS Update yet' (but I prefer to just stick to the runtime version.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +608,32 @@ async function getRuntimeVersionObjectAsync(
...
+        value: false,
+      },
+    ]);
+
+    if (!runConfig) {
+      Errors.exit(1);
+    }
+
+    const { default: UpdateConfigure } = await import('./configure');
```
wschurman, line 628: this is the first time we've done this in the CLI so it may be setting a bit of a precedent. I see that https://oclif.io/docs/running_programmatically says it's doable, but it also mentions that factoring it out is preferred and I tend to agree. The main reason I think we should factor out the underlying code is that that there's no typechecking on this passing of flags, so if we add a flag to the configure command, we'll most likely forget to inspect this callsite to ensure the flag is passed. By factoring out the code, we can type the method strictly.

wschurman (changes requested):
Couple issues I see:
1. We'll need to handle the non-interactive flag here (fail same as before if the flag is set).
2. The configs fetched on L203 and L207 are now stale, so they'll need to be re-fetched.
3. The installation of expo-updates in L215-L239 now I believe can be removed since it's done in the call of configure?

packages/eas-cli/src/commands/update/index.ts
Simek: According to dynamic imports, Expo CLI use this pattern for the performance, I also saw one usage in this codebase already for loading config file:
* https://github.com/expo/eas-cli/blob/024c699637f46914d0ba9766f321faaf5fc4efb4/packages/eas-cli/src/metadata/config.ts#L23

wschurman: Ah, sorry, I didn't mean the dynamic import, I meant the calling of another command programmatically instead of just factoring out the code. Dynamic imports are totally cool.

packages/eas-cli/src/commands/update/index.ts
Simek: What do you think about: 
```
"No, I will set the runtime version or run the command manually (EAS CLI exits)"
```
since the question above refers only to that action?

And then, we can rephrase the question above: 
```
"Running 'eas update:configure' will attempt to fix this issue. Do you want us to run the command for you?"
```
Or:
```
"Running 'eas update:configure' will try to automatically add the missing configuration. Do you want us to run the command for you?"
```
Or even something else, along those lines.

Simek:
1. Nice catch, addressed.
2. Public config is now passed from the transform method, and updated if needed. There is no need to refetch the private config, since it's only used once, before the transform method call.
3. I was still able to trigger this warning when using `--non-interactive` flag in the first version of this PR, but now, after refactor, we only call the `eas.json` edit part, so this is definitely still needed.

packages/eas-cli/src/commands/update/index.ts
Simek: After the latest changes the content of this message has been altered, since I have changed the logic. Let's resolve that for now, and we can get back to this in the re-review, if I get confirmation, that only calling the `app.json` edit was a correct approach here.

packages/eas-cli/src/project/workflow.ts
```diff
@@ -34,3 +34,16 @@ export async function resolveWorkflowAsync(
   }
   return Workflow.MANAGED;
 }
+
+export type CombinedWorkflowData = {
+  [value in Platform]: Workflow;
+};
```
wkozyra95, line 40: maybe inline that type in the function return value or name it sth more descriptive e.g. `WorkflowPerPlatform`, plus you probably should use `Record<Platform, Workflow>` instead

packages/eas-cli/src/project/workflow.ts
```diff
@@ -34,3 +34,16 @@ export async function resolveWorkflowAsync(
...
+
+export async function getCombinedWorkflowDataAsync(
+  projectDir: string
+): Promise<CombinedWorkflowData> {
```
dsokal, line 44:
```suggestion
): Promise<Record<Platform, Wokrflow>> {
```

packages/eas-cli/src/project/workflow.ts
dsokal, line 40:
```suggestion
```

packages/eas-cli/src/project/workflow.ts
```diff
@@ -34,3 +34,16 @@ export async function resolveWorkflowAsync(
...
+  return {
+    android: await resolveWorkflowAsync(projectDir, Platform.ANDROID),
+    ios: await resolveWorkflowAsync(projectDir, Platform.IOS),
+  };
```
dsokal, line 48:
```suggestion
  const [android, ios] = await Promise.all([
    resolveWorkflowAsync(projectDir, Platform.ANDROID),
    resolveWorkflowAsync(projectDir, Platform.IOS),
  ]);
  return { android, ios };
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +613,43 @@ async function getRuntimeVersionObjectAsync(
...
-    ])
-  );
+  try {
+    return [transformRuntimeVersions(exp, platforms), null];
+  } catch (error: any) {
+    if (nonInteractive) {
+      Errors.error(error, { exit: 1 });
+    }
+
+    ora().start().fail(error.message);
```
dsokal, line 623: What's that? Do you want to print the `x` sign? You can do it explicitly. See how we implemented `Log.succeed` - https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/log.ts#L48

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -198,7 +199,7 @@ export default class UpdatePublish extends EasCommand {
     // If a group was specified, that means we are republishing it.
     republish = republish || !!group;
 
-    const { exp, projectId, projectDir } = await getDynamicProjectConfigAsync({
+    let { exp, projectId, projectDir } = await getDynamicProjectConfigAsync({
```
dsokal, line 202: Don't use `let`

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +613,43 @@ async function getRuntimeVersionObjectAsync(
...
+          value: false,
+        },
+      ]
+    );
+
+    if (!runConfig) {
+      Errors.exit(1);
+    }
+
+    const { configureAppJSONForEASUpdateAsync } = await import('./configure');
```
dsokal, line 642: Why do you need the dynamic import?

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +613,43 @@ async function getRuntimeVersionObjectAsync(
...
-        Updates.getRuntimeVersion(exp, platform),
-        `Unable to determine runtime version for ${platform}`
-      ),
-    ])
-  );
+  try {
+    return [transformRuntimeVersions(exp, platforms), null];
+  } catch (error: any) {
+    if (nonInteractive) {
+      Errors.error(error, { exit: 1 });
```
wkozyra95, line 620: You are still in a file where a command is defined, so it might be ok, but in general, you should avoid exiting anywhere other than top-level code and use exceptions instead.

packages/eas-cli/src/commands/update/index.ts
Simek: Performance, it's the common pattern in Expo CLI.

packages/eas-cli/src/commands/update/index.ts
Simek: I also prefer `const`, but Will requested that change:
> The configs fetched on L203 and L207 are now stale, so they'll need to be re-fetched.

Using `let` here is justified I think.

packages/eas-cli/src/commands/update/index.ts
Simek: There was not a helper for that, so I have used Ora directly.

I always attempt to reduce the amount of lines changed, if not needed, but probably we can add that helper, since this PR evolved fince it's first form already.

packages/eas-cli/src/project/workflow.ts
Simek: Why?

packages/eas-cli/src/commands/update/index.ts
dsokal: It's not Expo CLI, please don't use dynamic import in this repo.

packages/eas-cli/src/project/workflow.ts
dsokal: Performance :D

Simek: Can you alaborate?

packages/eas-cli/src/commands/update/index.ts
Simek: Also, this is "need" IHMO, since this path will be triggered only on non-setup or invalid setup projects which make the percentile of `update` calls, and with static import we will always load whole configure step, which affects the command stat-up and TTI time.

dsokal: Can you prepare some benchmarks on what is the performance gain? I'd say it is negligible.

packages/eas-cli/src/commands/update/index.ts
dsokal: Let's change that

packages/eas-cli/src/project/workflow.ts
dsokal: Ok, I'll ask another question. You changed it from `Promise.all` to the non-`Promise.all` version. Why?

dsokal: With `Promise.all` those two can run concurrently.

wkozyra95: Your implementation is running those methods one after another, Dominik's approach is running them at the same time.

packages/eas-cli/src/commands/update/index.ts
dsokal: I'm not saying about benchmarks in general. I'm saying that I want to see benchmarks for this particular case. As Wojtek pointed out, there is no real gain here as probably all other modules imported in the `configure` module are imported anyway. 
The problem with the code that you proposed is that using dynamic imports in **EAS CLI** would be a precedence. In this case, I'm afraid, it's not at all justified.

wkozyra95: Also unless ocliff is doing lazy loading per command there might be no impact on performance at all regardless of the size of the imported part.

Simek: https://github.com/expo/eas-cli/pull/1427#discussion_r991500606

https://github.com/expo/eas-cli/blob/024c699637f46914d0ba9766f321faaf5fc4efb4/packages/eas-cli/src/metadata/config.ts#L23

dsokal: This line is reading the config from the user's project. It's not the same. 
In you're code you're importing internal modules.

Simek: > there is no real gain here as probably all other modules imported in the configure module are imported anyway.

There are no other `configure` imports or usage in the `update` command, and imports are loaded per command on run, so I think that the sentence above is incorrect.

dsokal: Just compare the imports in those files.
Screenshot 2022-10-12 at 12 53 31
Screenshot 2022-10-12 at 12 53 17

I won't believe there's any performance gain until I see data :)

wschurman: Now that this is just a normal function being imported I think it it makes sense to just use a normal import (we do this for pretty much all business logic functions). 

The key though is that we should move the `configureAppJSONForEASUpdateAsync` out of the command module and into its own module. If it's in the command module, a lot of extra stuff pertaining to commands and command static stuff is loaded and I'd be willing to bet that the dynamic import is more optimal since it wouldn't be loaded in most cases. If it's in its own module, then none of the command-related stuff will be loaded and the overhead of loading it normally is negligible. If it's in its own module, it is pretty much the same as any other function being imported in any command module.

packages/eas-cli/src/log.ts
```diff
@@ -45,6 +45,10 @@ export default class Log {
     Log.warn(`› ${chalk.bold('--' + flag)} flag is deprecated. ${message}`);
   }
 
+  public static failed(message: string): void {
```
wschurman, line 48:
```suggestion
  public static fail(message: string): void {
```
(to match tense of `succeed`. I'd also be open to renaming `succeed` to `success` and make this one `failure`)

packages/eas-cli/src/commands/update/index.ts
wschurman: Yep, the common pattern for this is to just throw a normal exception. (grep for `if (nonInteractive)`).

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -586,15 +610,43 @@ async function getRuntimeVersionObjectAsync(
...
+        { title: 'Yes', value: true },
+        {
+          title: 'No, I will set the runtime version manually (EAS CLI exits)',
+          value: false,
+        },
+      ]
+    );
+
+    if (!runConfig) {
+      Errors.exit(1);
```
wschurman, line 636: ditto about throwing an exception rather than using `Errors.exit` (can grep for both to see the relative usage of each, and tend to prefer the more common pattern)

wschurman (changes requested):
Looking a lot better and more well-typed (thanks for making that change)! 

The main change requested in this round of reviews is to move the factored-out-functions into a module outside of the command (which should alleviate concerns about using normal imports over dynamic imports).

I don't really have a good sense of dynamic import speed, but generally I opt for the more common pattern in a codebase when adding new code. I suggest sticking to the more common convention for feature PRs in order to reduce unrelated discussion on the PR and to get your feature landed more quickly. It was my bad for my comment conveying that it should be used; instead, I was more was conveying that that wasn't the issue with the line, but instead the issue was the import of the whole command (which is now fixed).

That being said, if there's a convention that you believe is better (and in this case have some perf numbers), I definitely encourage you to put up some PRs to change the pattern throughout the codebase. I tend to do this often. I often start with a small proposal PR showing the idea, gather comments, and then a few larger PRs to change the convention once there's buy in.

packages/eas-cli/src/commands/update/index.ts
Simek: This is a conscious decision to exit without an error here, since the user have chose this option. Throwing an another error here will not add any value or information, can only confuse user.

https://github.com/expo/eas-cli/blob/123615212058bf841dec00edcc16da284c27e148/packages/eas-cli/src/commands/build/index.ts#L300-L307

wkozyra95: > This is a conscious decision to exit without an error here,

`Errors.exit(1)` is not exiting without an error.

>  can only confuse user.

I don't see how, if you throw sth along the lines of "Aborting ..."

> Also this pattern has been taken from the codebase...

Similar patterns are not really a good argument, sth that is fine there might not be fine here. This "pattern" is used in very different contexts:
- In build, it's used in code that is run at the very start of the process, as part of the input/config validation. That function is handling deprecated config for something very specific.
- In update code that exit is run in the util function that someone might want to extract to common code in the future.

With that said I think leaving exit might be fine, but in this case, I think it's a bit borderline.

wschurman: Under the hood all `Errors.exit` does is throw an error: https://github.com/oclif/core/blob/main/src/errors/index.ts#L18

I don't feel too strongly here, but now that I grep for similar cases I see a mix of `throw`, `Errors.exit` and `process.exit`. `process.exit` is definitely incorrect since it will cause the tear down code of oclif commands not to run (see EasCommand.finally). `Errors.exit` and `throw` are equivalent, so it's just a matter of style.

I'd say this is fine here, but really what we should do in a follow-up is a brief audit of all cases followed by a potential lint rule to ensure that this remains consistent going forward across the codebase. That way there's no attention cost to this anymore.

For this PR, it's fine to keep it as you have it.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -247,8 +249,15 @@ export default class UpdatePublish extends EasCommand {
       }
     }
 
-    const runtimeVersions = await getRuntimeVersionObjectAsync(exp, platformFlag, projectDir);
-    await checkEASUpdateURLIsSetAsync(exp, projectId);
+    const [runtimeVersions, expUpdated] = await getRuntimeVersionObjectAsync(
```
wschurman, line 252: Prefer to keep this named `exp` and renaming the one before it `expBeforeRuntimeVersionUpdate` (or using `let` is fine as well). The reason for this is that L429 is missing a change to `expUpdated`.

wschurman (approved):

jonsamp (commented):
Screen Shot 2022-10-17 at 1 32 13 PM

The decisions + code in this PR are right to me. I tested this locally and see that I run into an error (pictured above): `Error: An SDK version must be defined when using the 'sdkVersion' runtime policy.`. --However, when I run the same `eas update` command again, it works fine.

Once we get this to publish with the written runtime version, we should be good to ship this.

Simek:
@jonsamp Last commit added after your comment should fix those issues, as we discussed, it would be great if you can confirm that the problem with runtime policy no longer occurs.

wschurman (changes requested):
https://github.com/expo/eas-cli/pull/1467 brings up a really good point that I hadn't thought of: installing or configuring expo-updates (or changing existing configuration) requires a new build. Because of this, doing these things at the beginning of `eas update` (including installing expo-updates) might not make as much sense as I once thought, since even if the update is published the user would still need to do a build which would embed a newer manifest in it and the published update would never be used (the clients select the most recent update, and in this case the build's embedded update would be newer than this published update). To me, this behavior seems more confusing than just throwing an error and telling people to run `eas update:configure`, which with https://github.com/expo/eas-cli/pull/1467 will tell the user they need to run a build before published updates will work.

jonsamp:
We talked about this feature in a meeting for awhile. @brentvatne put together a summary:

> should eas update run eas update:configure automatically if not yet configured? imo yes -> but let people know afterwards that the update may not apply to any existing builds.

> should eas update and eas build run eas init automatically if not yet configured? imo yes -> similar to trying to access an authenticated portion of a sign before signing in. i think it's a bad first time experience for a user to run a command and have it fail and tell them to run another command manually

> should we configure eas update and build in eas init? yes, we should prompt people for this

> should we configure eas update when you run eas build? yes, we should use the same sort of prompt we use around push notifications to let people say yes/no/don't ask me again

I think that we should prompt people on EAS Update and as the first one says "let people know afterwards that the update may not apply to any existing builds". Developers can still view published updates in Expo Go and development builds even if they did not have EAS Update configured until they published.

@wschurman, the underlying premise of this PR is right to me. If you agree, and you think the code is good, we should approve this one.

jonsamp (approved):
I just tested this and it is working great. Nice work Simek, this is a big improvement!

brentvatne (approved):

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+    return [transformRuntimeVersions(exp, platforms), exp];
+  } catch (error: any) {
+    if (nonInteractive) {
+      throw error;
+    }
+
+    Log.fail(error.message);
+
+    const runConfig = await selectAsync(
+      `Do you want us to run automatic ${chalk.bold(
```
brentvatne, line 643:
```suggestion
      `Configure runtime version in ${chalk.bold(
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+        'app.json'
+      )} configuration for EAS Update for you?`,
```
brentvatne, line 645:
```suggestion
      )} automatically for EAS Update?`,
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+      exp: newConfig,
+      projectDir,
+      projectId,
+      platform: platformFlag as RequestedPlatform,
+      workflows,
+      graphqlClient,
+    });
+
+    const continueWithChanges = await selectAsync(
+      `Would you like to continue update process with uncommitted changes in repository?`,
```
brentvatne, line 680:
```suggestion
      `Continue update process with uncommitted changes in repository?`,
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+      [
+        { title: 'Yes', value: true },
+        {
+          title: 'No, I will commit the modified files first (EAS CLI exits)',
```
brentvatne, line 684:
```suggestion
          title: 'No, I will commit the modified files first',
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+    Log.fail(error.message);
+
+    const runConfig = await selectAsync(
+      `Do you want us to run automatic ${chalk.bold(
+        'app.json'
+      )} configuration for EAS Update for you?`,
+      [
+        { title: 'Yes', value: true },
+        {
+          title: 'No, I will set the runtime version manually (EAS CLI exits)',
```
brentvatne, line 649:
```suggestion
          title: 'No, I will set the runtime version manually',
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,69 @@ async function getRuntimeVersionObjectAsync(
...
+    const newConfig: ExpoConfig = { ...exp, ...configUpdate };
+
+    await configureNativeFilesForEASUpdateAsync({
+      exp: newConfig,
+      projectDir,
+      projectId,
+      platform: platformFlag as RequestedPlatform,
+      workflows,
+      graphqlClient,
+    });
```
wschurman, line 677: Based on the code in `configure`, I think the result of `configureAppJSONForEASUpdateAsync` is the source of truth for `newConfig`.

```suggestion
    const newConfig = await configureAppJSONForEASUpdateAsync({
      exp,
      projectDir,
      projectId,
      platform: platformFlag as RequestedPlatform,
      workflows,
    });

    await configureNativeFilesForEASUpdateAsync({
      exp: newConfig,
      projectDir,
      projectId,
      platform: platformFlag as RequestedPlatform,
      workflows,
      graphqlClient,
    });
```

packages/eas-cli/src/commands/update/configure.ts
```diff
@@ -54,69 +55,55 @@ export default class UpdateConfigure extends EasCommand {
       await installExpoUpdatesAsync(projectDir);
     }
 
-    const [androidWorkflow, iosWorkflow] = await Promise.all([
-      resolveWorkflowAsync(projectDir, Platform.ANDROID),
-      resolveWorkflowAsync(projectDir, Platform.IOS),
-    ]);
+    const workflows = await resolveWorkflowPerPlatformAsync(projectDir);
```
wschurman, line 58: optional: can put all three of these methods into a single utility method to ensure they are kept consistent between the two commands
```typescript
function configureAsync(...) {
  const workflows = await resolveWorkflowPerPlatformAsync(projectDir);
  const updatedExp = await configureAppJSONForEASUpdateAsync(...);
  await configureNativeFilesForEASUpdateAsync(...)
}
```
and then call this function from both `configure` and `update`.

wschurman (approved):
We'll want to ensure @FiberJW's changes are included in the common factored out code so that this happens on both `configure` and `update`
> let people know afterwards that the update may not apply to any existing builds

packages/eas-cli/src/commands/update/index.ts
Simek: @wschurman I have stated the reasoning for how this code works in conversation with Jon (should post a comment here probably), let me quote it:

> the problem was that `configureAppJSONForEASUpdateAsync` returns updated config, but it was only a partial with the changes, so instead of using result directly we need to merge it with full config

It was a root cause why `"Error: Unable to determine runtime version for android"` started appearing after one of the PR changes. The problem has been mentioned in this comment:
* https://github.com/expo/eas-cli/pull/1427#pullrequestreview-1144562679

wschurman: Looking at `modifyConfigAsync` source, it looks like it returns exactly what's on disk? Maybe we need to run `transformRuntimeVersions` again or something?

Simek: I don't think that `modifyConfigAsync` is a source of problem there, it already receives config partial:

* https://github.com/expo/eas-cli/blob/b9c505bbdc30af90b17633028c171465e9ec8df2/packages/eas-cli/src/commands/update/configure.ts#L128-L129

packages/eas-cli/src/commands/update/index.ts
Simek: This pattern of adding `(EAS CLI exits)` at the end of exiting option has been taken from the existing code.

I we want to remove it, I feel like this should be done in the whole codebase, in the separate PR, but personally I like this addition, and I think that make it clear for the user what will happed.

packages/eas-cli/src/commands/update/index.ts
wschurman: Ah, I see. In that case we should re-fetch the full config using `getDynamicProjectConfigAsync`. Doing just an object merge will overwrite keys (doesn't do a deep merge).

```suggestion
    const configUpdate = await configureAppJSONForEASUpdateAsync({
      exp,
      projectDir,
      projectId,
      platform: platformFlag as RequestedPlatform,
      workflows,
    });

    const newConfig = (await getDynamicProjectConfigAsync({ isPublicConfig: true })).exp;

    await configureNativeFilesForEASUpdateAsync({
      exp: newConfig,
      projectDir,
      projectId,
      platform: platformFlag as RequestedPlatform,
      workflows,
      graphqlClient,
    });
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -600,15 +630,67 @@ async function getRuntimeVersionObjectAsync(
...
-      platform,
-      nullthrows(
-        Updates.getRuntimeVersion(exp, platform),
-        `Unable to determine runtime version for ${platform}`
-      ),
-    ])
-  );
+  try {
+    return [transformRuntimeVersions(exp, platforms), exp];
+  } catch (error: any) {
```
wschurman, line 638: catching all errors assumes that they are all runtime version errors vs other classes of errors (code errors, other errors). There doesn't seem to be a very good way of detecting just the class of error though other than error message introspection (which I don't recommend), so I think this is okay.

The only thing we risk is that it is already configured and `transformRuntimeVersions` throws another error and we clobber their configuration accidentally.

wschurman (approved):
Approving with the a few inline comments (just saw you already addressed one of them and the other isn't actionable).

packages/eas-cli/src/commands/update/configure.ts
Simek: I think that @byCedric plan to refactor the `configure` code anyway, and if this is out of scope of his planned changes I can tackle that in a separate PR.

packages/eas-cli/src/commands/update/index.ts
Simek: Only error that `transformRuntimeVersions` can throw is a one defined by us via `nullthrows` (it's a very basic method), so I think that is safe to use the generic catch here, and assume that every error is a missing runtime version.

packages/eas-cli/src/commands/update/index.ts
brentvatne: It looks like there is only one other place where we do this and it's in a fairly obscure place: [https://github.com/expo/eas-cli/blob/2aade4c0c82665f12cd8053d339122fb35f56168/packages/eas-cli/src/commands/build/index.ts#L305](https://github.com/expo/eas-cli/blob/2aade4c0c82665f12cd8053d339122fb35f56168/packages/eas-cli/src/commands/build/index.ts#L305)



I'm not entirely opposed to letting users know that the process will exit but I found this wording to be confusinghttps://github.com/expo/eas-cli/blob/2aade4c0c82665f12cd8053d339122fb35f56168/packages/eas-cli/src/commands/build/index.ts#L305. It'd be better to say something like (this process will exit)

packages/eas-cli/src/commands/update/index.ts
wkozyra95: Sorry, it appears I was wrong (I relied on someone else comment without verifying that myself), It's fine to use `Errors.error` or `Errors.exit` anywhere in the code, it throws an exception not exiting immediately as I thought before.

jonsamp:
I would like to get this merged whenever you have a chance @Simek. This PR is in a good spot and I'd love to see it in production sooner than later. I really appreciate your work on this and your responses to all the comments.

PR #1444, opened 2022-10-14
[eas-cli] Initialize analytics consistently and move analytics into context

wschurman:
# Checklist

# Why

The analytics system has been very fragile in the past. Things like always logging with a random uuid have plagued us numerous times.

This PR makes the analytics system non-global and instead injects it, which allows for consistent initialization and identification of the user to rudderstack/amplitude.

# How

This does a few things:
1. Make AnalyticsManager an interface, and only expose that interface to the commands. The internals of the manager system will switch between a no-op logger and the real one based on the user's analytics opt-out preferences.
2. Instantiate that interface when the command runs and inject it into contexts and commands. This way we can guarantee that it is initialized and the user identification has already occurred.
3. Only expose that interface to commands via a context for them to ensure that it is initialized as well.

# Test Plan

1. Run tests (they cover some of this).

# Manual test

Manually test to ensure that the user attribution is correct in all scenarios:

1. `export EXPO_STAGING=1`
2. Turn on live events in https://analytics.amplitude.com/expo/activity (with `Unified - dev` project selected)
3. While logged-in, run some commands, see that my user is getting `action` events.
4. Log out.
5. Run a command that requires login, see that once I log in the `action` event is correctly attributed to my existing amplitude user.
6. Log out.
7. Delete my `~/.expo-staging/state.json` and `~/Library/Preferences/eas-cli-nodejs/user-settings.json` (preference files) `analyticsDeviceId` key (remove persistent device IDs)
8. Run a command that requires login, see that once I log in the `action` event is correctly attributed to my existing amplitude user even though I have a new persistent device ID.
9. Log out
10. Repeat step 7 (delete persistent IDs)
11. Run a command that doesn't require login (`diagnostics`), see in amplitude that a new user is created.
12. Run a command that does require login, see that new user that was created is merged into my old user.


packages/eas-cli/src/analytics/AnalyticsManager.ts
```diff
@@ -0,0 +1,230 @@
+import RudderAnalytics from '@expo/rudder-sdk-node';
+import os from 'os';
+import { URL } from 'url';
+import { v4 as uuidv4 } from 'uuid';
+
+import { Actor, getActorDisplayName } from '../user/User';
+import UserSettings from '../user/UserSettings';
+import { easCliVersion } from '../utils/easCli';
+
+const PLATFORM_TO_ANALYTICS_PLATFORM: { [platform: string]: string } = {
```
dsokal, line 10:
```suggestion
const PLATFORM_TO_ANALYTICS_PLATFORM: Partial<Record<NodeJS.Platform, string>> = {
```

packages/eas-cli/src/analytics/AnalyticsManager.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  private getRudderStackContext(): Record<string, any> {
+    const platform = PLATFORM_TO_ANALYTICS_PLATFORM[os.platform()] || os.platform();
+    return {
+      os: { name: platform, version: os.release() },
+      device: { type: platform, model: platform },
+      app: { name: 'eas cli', version: easCliVersion ?? undefined },
+    };
+  }
+}
```
dsokal, line 230: Github won't let me comment on the file below (`NoOpAnalyticsManager.ts`) - the file is empty and not referenced anywhere. Is this intentional?

packages/eas-cli/src/build/createContext.ts
```diff
@@ -48,6 +49,7 @@ export async function createBuildContextAsync<T extends Platform>({
   message?: string;
   actor: Actor;
   graphqlClient: ExpoGraphqlClient;
+  analyticsManager: IAnalyticsManager;
```
dsokal, line 52: Not a strong opinion but maybe we could rename this from `analyticsManager` to `analytics`? To me, the word "manager" in code is almost always meaningless. The interface name can stay as it is.

packages/eas-cli/src/analytics/AnalyticsManager.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+export interface IAnalyticsManager {
+  logEvent(name: AnalyticsEvent, properties: Record<string, any>): void;
+}
+
+/**
+ * The interface for commands to use to orchestrate the analytics system. Should only be necessary to use
+ * this within EASCommand.
+ */
+export interface IAnalayticsManagerWithOrchestration extends IAnalyticsManager {
```
dsokal, line 85: Typo:
```suggestion
export interface IAnalyticsManagerWithOrchestration extends IAnalyticsManager {
```

packages/eas-cli/src/analytics/AnalyticsManager.ts
```diff
@@ -0,0 +1,230 @@
...
+  setActor(actor: Actor): void;
+  flushAsync(): Promise<void>;
+}
```
dsokal, line 88: I'd prefer not to use this convention of starting the interface name with `I`. 

1. At the moment, it's not forbidden by ESLint but it should be.
2. We don't use this convention elsewhere in the CLI.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -106,37 +115,46 @@ export default abstract class EasCommand extends Command {
...
+   * If needed in a subclass, use the SessionManager ContextOption.
+   */
+  private sessionManager!: SessionManager;
+
+  /**
+   * The analytics manager. Used for logging analytics.
+   * It is set up here to ensure a consistent set up.
    */
-  private readonly sessionManager = new SessionManager();
+  private analyticsManager!: IAnalayticsManagerWithOrchestration;
```
dsokal, line 144: In other places we use the following pattern to get rid of `!`:
Screenshot 2022-10-17 at 09 25 49

dsokal (approved):

packages/eas-cli/src/build/build.ts
```diff
@@ -87,12 +88,16 @@ export async function prepareBuildRequestForPlatformAsync<
...
   );
 
-  await withAnalyticsAsync(async () => await builder.syncProjectConfigurationAsync(ctx), {
-    attemptEvent: BuildEvent.CONFIGURE_PROJECT_ATTEMPT,
-    successEvent: BuildEvent.CONFIGURE_PROJECT_SUCCESS,
-    failureEvent: BuildEvent.CONFIGURE_PROJECT_FAIL,
-    trackingCtx: ctx.trackingCtx,
-  });
+  await withAnalyticsAsync(
+    ctx.analyticsManager,
```
wkozyra95, line 92: It feels like we are passing 2 contexts for analytics (analyticsManager, trackingCtx), maybe it should be merged?
Also maybe this helper `withAnalyticsAsync` should be part of the manager interface.

wkozyra95 (approved):

packages/eas-cli/src/commandUtils/EasCommand.ts
wschurman: Had this at first, but `_`-prefixed private members are disallowed in this repo so I went with this. I'll change back and choose a different name.

packages/eas-cli/src/build/build.ts
wschurman: Will make this change in a follow-up. The tracking context is really just an arbitrary key-value so it's probably a misnomer rather than a context.

wschurman:
/changelog-entry bug-fix Initialize analytics consistently and move analytics into context

PR #1461, opened 2022-10-19
feat: support uploading different bundle combos

EvanBacon:
# Checklist

# Why

Expo Router defaults to shipping with universal Metro support (ios, android, web) which means `eas update` breaks by default if it asserts that `web` is not supported.

# How

- This PR also adds the ability to upload single platforms.
- The `--platform` flag is a bit more flexible/sensible now. If your project only has `android` and you run `eas update --platform ios --skip-bundler` then the error will be about how no bundles were uploaded rather than some mysterious metadata validation issue.
- Simplified spinners a bit.
- Made a few errors more contextual.

# Test Plan

Web:

- `npx create-react-native-app -t with-router`
- `eas update` → works

Single platform:

- `yarn expo export -p ios`
- `eas update -p ios --skip-bundler` → works

Better errors:

- `rm -rf dist`
- `eas update --skip-bundler` → tells you to either change the `--input-dir` or disable `--skip-bundler`.


EvanBacon:
/changelog-entry new-feature support uploading different platform combinations

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -601,13 +621,18 @@ async function getRuntimeVersionObjectAsync(
...
-    platforms.map(platform => [
-      platform,
-      nullthrows(
-        Updates.getRuntimeVersion(exp, platform),
-        `Unable to determine runtime version for ${platform}`
-      ),
-    ])
+    platforms.map(platform => {
+      if (platform === 'web') {
+        return ['web', 'n/a'];
```
wschurman, line 626:
```suggestion
        return ['web', null];
```
?

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -391,45 +400,53 @@ export default class UpdatePublish extends EasCommand {
 
       // build bundle and upload assets for a new publish
       if (!skipBundler) {
-        const bundleSpinner = ora().start('Building bundle...');
+        const bundleSpinner = ora().start('Exporting...');
         try {
           await buildBundlesAsync({ projectDir, inputDir });
-          bundleSpinner.succeed('Built bundle!');
+          bundleSpinner.succeed('Exported platforms');
```
ide, line 406: This should say something about exporting updates, not platforms. Also, if you decide to include the word "platforms" in the message, make sure to render it as "platform" singular when there's just one target output platform.

packages/eas-cli/src/commands/update/index.ts
ide: Agree that using null is a good idea -- if there's a special sentinel value we need to check for, it helps to make the type checker aware of it, and `null` is the most built-in way to do that

packages/eas-cli/src/project/publish.ts
```diff
@@ -193,43 +199,62 @@ export function loadMetadata(distRoot: string): Metadata {
   if (metadata.bundler !== 'metro') {
     throw new Error('Only bundles created with Metro are currently supported');
   }
+  const platforms = Object.keys(metadata.fileMetadata);
+  if (platforms.length === 0) {
+    Log.warn('No platforms were exported');
```
ide, line 204:
```suggestion
    Log.warn('No updates were exported for any platform');
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -47,19 +49,20 @@ type ManifestFragment = {
...
   bundle: Joi.string().required(),
-}).required();
+}).optional();
 export const MetadataJoi = Joi.object({
   version: Joi.number().required(),
   bundler: Joi.string().required(),
   fileMetadata: Joi.object({
+    web: fileMetadataJoi,
     android: fileMetadataJoi,
     ios: fileMetadataJoi,
```
ide, line 67: style nit: order platforms alphabetically

packages/eas-cli/src/project/publish.ts
```diff
@@ -193,43 +199,62 @@ export function loadMetadata(distRoot: string): Metadata {
...
-  inputDir: string;
-  platforms: PublishPlatform[];
-}): Promise<CollectedAssets> {
-  const distRoot = await resolveInputDirectoryAsync(inputDir);
-  const metadata = loadMetadata(distRoot);
-
-  const assetsFinal: CollectedAssets = {};
-  for (const platform of platforms) {
-    assetsFinal[platform] = {
+export function filterPlatforms<T extends Partial<Record<Platform, any>>>(
```
ide, line 210: Since the error thrown by this function references metadata.json specifically, allude to that coupling by including "exported" in the method name. I think it is also helpful to make it more visible that the filtering parameter is a flag.

```suggestion
export function filterExportedPlatformsByFlag<T extends Partial<Record<Platform, any>>>(
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -193,43 +199,62 @@ export function loadMetadata(distRoot: string): Metadata {
...
+): T {
+  if (platformFlag === 'all') {
+    return record;
+  }
+
+  const platform = platformFlag as Platform;
+
+  if (!record[platform]) {
+    throw new Error(
+      `--platform="${platform}" not found in metadata.json. Available platforms: ${Object.keys(
```
ide, line 222:
```suggestion
      `--platform="${platform}" not found in metadata.json. Available platform(s): ${Object.keys(
```

packages/eas-cli/src/project/publish.ts
```diff
@@ -193,43 +199,62 @@ export function loadMetadata(distRoot: string): Metadata {
...
+        record
+      ).join(', ')}`
+    );
+  }
+
+  return { [platform]: record[platform] } as T;
+}
+
+/** Given a directory, load the metadata.json and collect the assets for each platform. */
+export async function collectAssetsAsync(inputDir: string): Promise<CollectedAssets> {
```
ide, line 232: Rename the parameter since "inputDir" already means something specific in these files, and the argument shouldn't be that inputDir anymore (in practice, it's distRoot but this function looks like it's intended to take any directory passed to it):
```suggestion
export async function collectAssetsAsync(dir: string): Promise<CollectedAssets> {
```

ide (changes requested):

ide (approved):

EvanBacon:
https://github.com/expo/eas-cli/issues/1484

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Support uploading different platform combinations. ([#1461](https://github.com/expo/eas-cli/pull/1461) by [@EvanBacon](https://github.com/EvanBacon))
```
dsokal, line 11: Maybe mention this is for `eas update`

packages/eas-cli/src/project/publish.ts
```diff
@@ -200,43 +206,62 @@ export function loadMetadata(distRoot: string): Metadata {
...
+): T {
+  if (platformFlag === 'all') {
+    return record;
+  }
+
+  const platform = platformFlag as Platform;
+
+  if (!record[platform]) {
+    throw new Error(
+      `--platform="${platform}" not found in metadata.json. Available platform(s): ${Object.keys(
```
dsokal, line 229: IMO these error messages sound as if `--platform="${platform}"` (and `--input-dir="${inputDir}"`) was sth you could find in metadata.json. I mean the exact string.

EvanBacon: Thought about this a bit more and the entire error message is generally not helpful. User's probably don't even know what a `metadata.json` is since we don't document it and the format is non-standard. 

Screen Shot 2022-11-08 at 11 39 02 AM

All they do know is that the value they provided `--platform="..."` isn't available.

PR #1467, opened 2022-10-20
[eas-cli] Warn about outdated build deployment when configuring EAS Update

FiberJW:
# Checklist

# Why

Some users can get confused about why EAS Update doesn't work after adding it to an existing app without rebuilding it.

# How

I added a warning to rebuild and resubmit your app after running `eas update:configure`. 

# Test Plan

Screen Shot 2022-10-20 at 13 27 17

### DIFF
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -7,7 +7,6 @@ import chalk from 'chalk';
 import { getEASUpdateURL } from '../../api';
 import EasCommand from '../../commandUtils/EasCommand';
 import { AppPlatform } from '../../graphql/generated';
-import { BuildQuery } from '../../graphql/queries/BuildQuery';
 import Log, { learnMore } from '../../log';
 import { RequestedPlatform, appPlatformDisplayNames } from '../../platform';
 import {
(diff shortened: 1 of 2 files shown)


wschurman (changes requested):
Left a comment on a related PR, might be worth looking: https://github.com/expo/eas-cli/pull/1427#pullrequestreview-1149636318

Overall I think this is a really really good thing to print, and makes me even more confident that we should require people to run `eas update:configure` before running `eas update`, and that `eas update` shouldn't try to configure the app since it's likely a `eas build` (or local build) will need to take place in-between them.

Requesting changes to just always print this instead of conditionally printing it.

packages/eas-cli/src/commands/update/configure.ts
```diff
@@ -87,6 +87,11 @@ export default class UpdateConfigure extends EasCommand {
       Log.withTick(`Configured ${chalk.bold('Expo.plist')} for EAS Update`);
     }
 
+    Log.addNewLineIfNone();
+    Log.warn(
+      `If you have previously deployed this app to users, you will need to rebuild and submit a new version before you can update your app with EAS Update.`
```
wschurman, line 92:
```suggestion
      `All builds of your app going forward will be eligible to receive updates published with EAS Update.`
```

My suggestion might be a bit technical, I'm not sure. Would be good to get feedback from other reviewers. I think the thing I'd wan't this warning to convey is that only new builds that have this configuration will get updates. Maybe we just say that? idk.

FiberJW: I think your suggestion states a fact that the user likely understands at this point in time, whereas the original warning's intention is to make the user aware that they might have to take an extra action to get the outcome they desire

wkozyra95: `All builds of your app` might worry users that somehow all builds will be getting all updates, to avoid that maybe we could change `All` -> `Only`. Of course, it might be desired behavior to introduce a bit of uncertainty to make users check docs.

FiberJW:
I updated the message with Will's suggestion: 
Screen Shot 2022-10-26 at 16 21 32

wschurman (approved):

PR #1473, opened 2022-10-24
[eas-cli] Add flag sanitization in updates command

byCedric:
# Checklist

# Why

This starts ENG-5926 and is the first of a couple of stacked PRs to clean this command up.

# How

- Added `RawUpdateFlags` and `UpdateFlags`
- Added basic validation for incorrect combinations of flags
- Renamed a few variables in `runAsync` to emphasize entity relation
  - `group` → `groupId`
  - `message` → `updateMessage`

# Test Plan

`eas update` (with all flag-variations should work)

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -92,7 +92,7 @@ type UpdateFlags = {
   skipBundler: boolean;
   privateKeyPath?: string;
   json: boolean;
-  nonInteractive?: boolean;
+  nonInteractive: boolean;
 };
 
 async function ensureChannelExistsAsync(
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -92,7 +92,7 @@ type UpdateFlags = {
   skipBundler: boolean;
   privateKeyPath?: string;
   json: boolean;
-  nonInteractive: boolean;
+  nonInteractive?: boolean;
 };
 
 async function ensureChannelExistsAsync(
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -67,6 +67,34 @@ import {
 export const defaultPublishPlatforms: PublishPlatform[] = ['android', 'ios'];
 export type PublishPlatformFlag = PublishPlatform | 'all';
 
+type RawUpdateFlags = {
+  auto: boolean;
+  branch?: string;
+  message?: string;
+  group?: string;
+  republish?: boolean;
+  platform: string;
+  'input-dir': string;
+  'skip-bundler': boolean;
+  'private-key-path'?: string;
+  'non-interactive': boolean;
+  json: boolean;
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/commands/update/index.ts
```diff
@@ -577,6 +602,38 @@ export default class UpdatePublish extends EasCommand {
       }
     }
   }
+
+  private sanitizeFlags(flags: RawUpdateFlags): UpdateFlags {
+    const nonInteractive = flags['non-interactive'] ?? false;
```
wschurman, line 607:
```suggestion
    const nonInteractive = flags['non-interactive'];
```
(since this is type boolean)

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -577,6 +602,38 @@ export default class UpdatePublish extends EasCommand {
...
+
+    const { auto, branch: branchName, message: updateMessage } = flags;
+    if (nonInteractive && !auto && !(branchName && updateMessage)) {
+      Errors.error(
```
wschurman, line 611: these exit/error methods (that return `never`) are pretty oddly-typed. For example, if you pass `{exit: false}` in as the second argument, it doesn't exit and the typescript types are incorrect.

I made a similar comment on another recent PR (https://github.com/expo/eas-cli/pull/1427#discussion_r994887505) but my general feeling is that these add a bit of overhead for the reviewer since it's not immediately clear just from reading that they halt the procedure (like a throw would). That being said, I see that these are used in this param sanitization pretty consistently, so I guess it's okay here. It still took me a bit longer to review though, so just a data point.

wschurman (approved):

packages/eas-cli/src/commands/update/index.ts
byCedric: Sure, no strong opinions on this one. I just added it for clarity; this is the default (just like we do with some other flags). But it is indeed very explicit.

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -61,6 +61,34 @@ import { createUpdateChannelOnAppAsync } from '../channel/create';
...
+  platform: PublishPlatformFlag;
+  branchName?: string;
+  updateMessage?: string;
+  republish: boolean;
+  groupId?: string;
+  inputDir: string;
+  skipBundler: boolean;
+  privateKeyPath?: string;
+  json: boolean;
+  nonInteractive: boolean;
```
byCedric, line 89:
```suggestion
  nonInteractive?: boolean;
```

byCedric: Follow up from the feedback from @wschurman

packages/eas-cli/src/commands/update/index.ts
byCedric: Ah, the `this.getContextAsync` requires a boolean for `nonInteractive`. Since we have to provide a default anyways, I'll revert this change and put it in the `santizeFlags` method.

PR #1478, opened 2022-10-25
[eas-cli] Extract entity utilities from update command class

byCedric:
# Checklist

# Why

Continuation of PR #1473 and ENG-5926.

# How

- Moved ensure methods from `UpdatePublish` to `src/branch|channel/queries.ts`.
- Copied other methods from commands `BranchCreate` and `ChannelCreate`, to keep queries contained in `src/branch|channel/queries.ts` files.
- Validated the logging is similar (still slightly different)
  - _branch logging_
    - removed: `Created branch: <branchName>`
    - why: The branch is both listed in the formatted fields (first field), and the "Channel is pointing to branch" logging, seems not worth it to determine if the branch was created or already existed. (_but if people disagree, I'd be happy to change it_)
  - _channel logging_  
    - from: `Created a channel: <channelName> pointed at branch: <branchName>` (_only logged when created_)
    - to: `Channel: <channelName> pointed at branch: <branchName>` (_always logged_) 
    - why: The channel name is not part of the formatted fields at the end. We either need to display this, or maybe even add it to the formatted fields as item? (_instead of the `Log.withTick`_)

# Test Plan

- `eas update --branch new-branch`
- `eas update --branch existing-branch`

### DIFF
--- a/packages/eas-cli/src/commands/branch/create.ts
+++ b/packages/eas-cli/src/commands/branch/create.ts
@@ -16,6 +16,7 @@ import { getDisplayNameForProjectIdAsync } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
+// NOTE(cedric): copied to src/branch/queries.ts to reuse in multiple commands
 export async function createUpdateBranchOnAppAsync(
   graphqlClient: ExpoGraphqlClient,
   { appId, name }: CreateUpdateBranchForAppMutationVariables
--- a/packages/eas-cli/src/commands/branch/create.ts
+++ b/packages/eas-cli/src/commands/branch/create.ts
@@ -1,52 +1,13 @@
 import chalk from 'chalk';
-import gql from 'graphql-tag';
 
+import { createUpdateBranchOnAppAsync } from '../../branch/queries';
 import { getDefaultBranchNameAsync } from '../../branch/utils';
 import EasCommand from '../../commandUtils/EasCommand';
-import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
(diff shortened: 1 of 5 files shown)


byCedric:
Open question to everyone here:
  1. Do we want to keep exporting these `create<Entity>` or `ensure<Entity>...` methods from other command classes? I don't really like it, since it implicitly creates depending command classes. But, not sure if there is a reason to do it.
  2. Why do we have `src/graphql/queries|mutations` and separate `src/branch|channel/queries.ts` files? Seems like they would be more in place under the graphql client, like most of the non-updates-related entities?
  3. Seeing some inconsistencies with the return types and logic of `ensure<Entity>` methods. Some won't return anything, others just the entity ID. The `ensureChannelExists` also just yolos it and uses potential errors to determine if the channel exists, makes sense in terms of limiting the graphql calls, but it does differ quite a lot compared to `ensureBranchExists`.

packages/eas-cli/src/commands/branch/create.ts
```diff
@@ -16,6 +16,7 @@ import { getDisplayNameForProjectIdAsync } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
+// NOTE(cedric): copied to src/branch/queries.ts to reuse in multiple commands
```
wschurman, line 19: can these methods be removed now that they've been factored out?

wschurman (approved):

packages/eas-cli/src/commands/branch/create.ts
byCedric: Yeah, pending my open questions. I left a comment so I could do a quick search&destroy operation 😄

byCedric: Done!

PR #1479, opened 2022-10-25
[eas-cli] Improve configuration setup for EAS update

byCedric:
# Checklist

# TODO

- Add tests for `src/updates/configure.ts` -> comming in follow up PR

# Why

- Continuation of PR https://github.com/expo/eas-cli/pull/1473 and [ENG-5926](https://linear.app/expo/issue/ENG-5926)
- Fixes ENG-6748, ENG-5806
- Part of ENG-6002

# How

This moves most of the (auto) configuration from both `eas update:configure` and `eas update` into `src/updates/configure.ts` and reuses it on both commands.

> This also includes the change from PR #1467, just for completeness and being able to test the final result with `npx @bycedric/some-cli-test update` - a published version of these stacked PRs.

# Result

## `eas update:configure`

<details><summary>Project without any EAS service/config setup</summary>

image

</details>

<details><summary>Project with EAS Update fully configured</summary>

image

</details>

<details><summary>Project with EAS Update fully configured, except <code>runtimeVersion</code></summary>

image

</details>

## `eas update:configure --non-interactive`

Adding this flag is required to ensure people can set their project up without `extra.eas.projects`. When always forcing non interactive, people will be prompted with the first example here:

<details><summary>Project without any EAS service/config setup</summary>

image

</details>

<details><summary>Project with just <code>extra.eas.projectId</code> setup</summary>

image

</details>

## `eas update --auto`

<details><summary>Project without any EAS service/config setup</summary>

image

</details>

<details><summary>Project with just <code>extra.eas.projectId</code> setup</summary>

image

</details>

<details><summary>Project with EAS Update fully configured</summary>

image

</details>

<details><summary>Project with EAS Update fully configured, except <code>runtimeVersion</code></summary>

image

</details>

# Test Plan

You can use my "fork package" on npm to play with the current state. It's `@bycedric/some-cli-test@3.0.4`, e.g. `npx @bycedric/some-cli-test@3.0.4 update` in a brand new project.

- `eas update:configure`
- `eas update`

### DIFF
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -1,25 +1,9 @@
-import { ExpoConfig, modifyConfigAsync } from '@expo/config';
-import { Platform, Workflow } from '@expo/eas-build-job';
 import { Flags } from '@oclif/core';
-import assert from 'assert';
-import chalk from 'chalk';
 
-import { getEASUpdateURL } from '../../api';
 import EasCommand from '../../commandUtils/EasCommand';
-import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
-import { AppPlatform } from '../../graphql/generated';
-import Log, { learnMore } from '../../log';
-import { RequestedPlatform, appPlatformDisplayNames } from '../../platform';
-import {
-  installExpoUpdatesAsync,
-  isExpoUpdatesInstalledOrAvailable,
-} from '../../project/projectUtils';
-import { resolveWorkflowPerPlatformAsync } from '../../project/workflow';
-import { syncUpdatesConfigurationAsync as syncAndroidUpdatesConfigurationAsync } from '../../update/android/UpdatesModule';
-import { syncUpdatesConfigurationAsync as syncIosUpdatesConfigurationAsync } from '../../update/ios/UpdatesModule';
-
-const DEFAULT_MANAGED_RUNTIME_VERSION = { policy: 'sdkVersion' } as const;
-const DEFAULT_BARE_RUNTIME_VERSION = '1.0.0' as const;
+import Log from '../../log';
+import { RequestedPlatform } from '../../platform';
+import { ensureEASUpdatesIsConfiguredAsync } from '../../update/configure';
 
 export default class UpdateConfigure extends EasCommand {
   static override description = 'configure the project to support EAS Update';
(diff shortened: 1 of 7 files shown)


kbrandwijk:
At first glance, this seems to be somewhat incompatible with https://github.com/expo/eas-cli/pull/1427. Did you consider that PR?

byCedric:
> to be somewhat incompatible with #1427. Did you consider that PR?

Yes, that PR has a different intent in what we want `eas update` to be. Evan and I created the version from this PR as a prototype to help remove friction encountered when (starting to) use EAS Updates. The friction here being mostly:
1. Having to run `eas init`, `eas update:configure`, and `eas update`.
2. Running into multiple errors, some of which are not helpful

Basically, whenever someone runs `eas update`, we assume intent that this user wants to use EAS Updates. With that in mind, we automate most of the setup while providing the user enough context to know that just running that command won't be sufficient. They also need to rebuild their app to start using EAS Updates or open it in Expo Go.

I do like some of the refactors that @Simek brought in. I already told @jonsamp that it's great to merge that PR before this one. I can rebase upon that one. But these stacked PRs take it a bit further, like dropping some exported functions from classes and reusing the same code for both `eas update:configure` and `eas update`.

byCedric:
In the prototype version, I also started refactoring the `eas update` runAsync method, but I kept that out of these PRs to wait until both PR #1427 and #1467 are merged.

byCedric:
/changelog-entry chore Remove friction from the initial EAS update setup

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,294 @@
...
+    runtimeVersion: modifyConfig.runtimeVersion ?? exp.runtimeVersion,
+    updates: { ...exp.updates, url: modifyConfig.updates?.url ?? exp.updates?.url },
+    android: {
+      ...exp.android,
+      runtimeVersion: modifyConfig.android?.runtimeVersion ?? exp.android?.runtimeVersion,
+    },
+    ios: {
+      ...exp.ios,
+      runtimeVersion: modifyConfig.ios?.runtimeVersion ?? exp.ios?.runtimeVersion,
+    },
```
EvanBacon, line 107:
```suggestion
    runtimeVersion: modifyConfig.runtimeVersion ?? exp.runtimeVersion,
    updates: { ...exp.updates, ...modifyConfig.updates },
    android: {
      ...exp.android,
      ...modifyConfig.android
    },
    ios: {
      ...exp.ios,
      ...modifyConfig.ios
    },
```
This should work better

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,294 @@
...
+  };
+
+  // TODO(cedric): check where these properties are coming from, they pop up in `eas update`
+  delete mergedExp.sdkVersion; // Note(cedric): this is always returned from `getConfig`, but we don't want to set it
+  delete mergedExp['_internal'];
+  delete mergedExp.originalFullName;
+  delete mergedExp.currentFullName;
+  delete mergedExp.platforms;
```
EvanBacon, line 115:
```suggestion
```

If we drop the `...exp` then we won't need the naive clean-up phase.

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,294 @@
...
+}): void {
+  if (modifyConfig.updates?.url) {
+    Log.withTick(
+      exp.updates?.url
+        ? `Overwrote updates.url "${exp.updates.url}" with "${modifyConfig.updates.url}"`
+        : `Configured updates.url to "${modifyConfig.updates.url}"`
+    );
+  }
+
+  if (modifyConfig.android?.runtimeVersion ?? modifyConfig.runtimeVersion) {
```
EvanBacon, line 156: not necessary but this block and the next (ios, android) should be a loop.

byCedric: Yeah, to do this, we need to simplify the types. Right now `appPlatformDisplayNames` uses `AppPlatform` (which is all caps `ANDROID`/`IOS`). But, `ExpoConfig` uses lowercase `Platform`.

So to minimize the type of juggling, I kept it the same.

EvanBacon (approved):

packages/eas-cli/src/commands/update/configure.ts
```diff
@@ -39,297 +25,27 @@ export default class UpdateConfigure extends EasCommand {
...
-          },
-          ios: { ...exp.ios, runtimeVersion: undefined },
-          updates,
-        };
-        newConfigOnlyAddedValues = {
-          runtimeVersion: newAndroidRuntimeVersion,
-          ...(exp.android && 'runtimeVersion' in exp.android
-            ? {
-                android: {
-                  runtimeVersion: '<remove this key>',
```
wschurman, line 139: looks like the new version of this doesn't have this in the manual instruction?

packages/eas-cli/src/commands/update/configure.ts
```diff
@@ -39,297 +25,27 @@ export default class UpdateConfigure extends EasCommand {
...
-    workflows.ios === Workflow.GENERIC
-      ? DEFAULT_BARE_RUNTIME_VERSION
-      : DEFAULT_MANAGED_RUNTIME_VERSION;
-
-  const newAndroidRuntimeVersion =
-    exp.android?.runtimeVersion ?? exp.runtimeVersion ?? androidDefaultRuntimeVersion;
-  const newIosRuntimeVersion =
-    exp.ios?.runtimeVersion ?? exp.runtimeVersion ?? iosDefaultRuntimeVersion;
-
-  let newConfig: Partial<ExpoConfig>;
```
wschurman, line 120: would be good to add a unit test for just this piece of code that works before and after your refactor

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+    `In order to finish configuring your project for EAS Update, you are going to need manually add the following to your app.config.js:\n${learnMore(
+      'https://expo.fyi/eas-update-config.md'
+    )}\n`
+  );
+  Log.log(chalk.bold(JSON.stringify(modifyConfig, null, 2)));
+  Log.addNewLineIfNone();
+
+  if (workflows.android === Workflow.GENERIC || workflows.ios === Workflow.GENERIC) {
+    Log.warn(
+      `You will also have to manually edit the projects ${chalk.bold(
```
wschurman, line 191:
```suggestion
      `You will also have to manually edit the project's ${chalk.bold(
```

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+  const { projectChanged, exp: expWithUpdates } =
+    await ensureEASUpdatesIsConfiguredInExpoConfigAsync({
+      exp: expWithoutUpdates,
+      projectDir,
+      projectId,
+      platform,
+      workflows,
+    });
+
+  if (projectChanged || !hasExpoUpdates) {
```
wschurman, line 272: is there a reason we wouldn't want to always do this (like the configure command used to do)?

wschurman (changes requested):
The main thing we should make sure of is that the dynamic config stuff continues to work as we want it to (not saying that the way it worked previously was best, but if we want to change its behavior we should do so intentionally).

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+  platform: RequestedPlatform;
+  workflows: Record<Platform, Workflow>;
+}): Promise<{ projectChanged: boolean; exp: ExpoConfig }> {
+  const modifyConfig: Partial<ExpoConfig> = {};
+
+  if (exp.updates?.url !== getEASUpdateURL(projectId)) {
+    modifyConfig.updates = { url: getEASUpdateURL(projectId) };
+  }
+
+  let androidRuntimeVersion = exp.android?.runtimeVersion ?? exp.runtimeVersion;
```
EvanBacon, line 67: Not super important but we had a lot of this kinda logic in `expo export` and `expo publish` where everything was duplicated twice. Prefer loops even though there are only two values.

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+    updates: { ...exp.updates, ...modifyConfig.updates },
+    android: {
+      ...exp.android,
+      ...modifyConfig.android,
+    },
+    ios: {
+      ...exp.ios,
+      ...modifyConfig.ios,
+    },
+  };
```
EvanBacon, line 107: Would be cool if this was split into a different function that returned `mergedExp` so you could easily test the results and ensure the clobbering always works as expected. I could see a situation where someone changes the `modifyConfig` to define a value like `runtimeVersion: undefined` which then removes the `runtimeVersion` altogether.

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+  workflows: Record<Platform, Workflow>;
+}): void {
+  Log.addNewLineIfNone();
+  Log.warn(
+    `It looks like you are using a dynamic configuration! ${learnMore(
+      'https://docs.expo.dev/workflow/configuration/#dynamic-configuration-with-appconfigjs)'
+    )}`
+  );
+  Log.warn(
+    `In order to finish configuring your project for EAS Update, you are going to need manually add the following to your app.config.js:\n${learnMore(
```
EvanBacon, line 182:
```suggestion
    `Finish configuring your project for EAS Update by adding the following to the project app.config.js:\n${learnMore(
```

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+      'https://expo.fyi/eas-update-config.md'
```
EvanBacon, line 183:
```suggestion
      'https://expo.fyi/eas-update-config'
```

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+    )}\n`
+  );
+  Log.log(chalk.bold(JSON.stringify(modifyConfig, null, 2)));
+  Log.addNewLineIfNone();
+
+  if (workflows.android === Workflow.GENERIC || workflows.ios === Workflow.GENERIC) {
+    Log.warn(
+      `You will also have to manually edit the projects ${chalk.bold(
+        'Expo.plist/AndroidManifest.xml'
+      )}. ${learnMore('https://expo.fyi/eas-update-config.md#native-configuration')}`
```
EvanBacon, line 193:
```suggestion
      chalk`The native config files {bold Expo.plist & AndroidManifest.xml} must be updated to support EAS Update. ${learnMore('https://expo.fyi/eas-update-config.md#native-configuration')}`
```

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+    projectDir: string;
+    platform: RequestedPlatform;
+  }
+): Promise<{ projectChanged: boolean }> {
+  const hasExpoUpdates = isExpoUpdatesInstalledOrAvailable(
+    projectDir,
+    expWithoutUpdates.sdkVersion
+  );
+  if (!hasExpoUpdates) {
+    await installExpoUpdatesAsync(projectDir, { silent: true });
```
EvanBacon, line 258: [sanity] Toggle off silent when debug logging is enabled, makes our job easier in the future.

packages/eas-cli/src/update/configure.ts
byCedric: The main reason for that is to keep `eas update` performance similar to how it was before. We can run most of these `ensureX` commands, but it will slow it down.

That being said, haven't benchmarked it.

packages/eas-cli/src/commands/update/configure.ts
byCedric: Yeah, I like that too. Unfortunately, this method is one of those problematic methods that are really _really_ hard to test without setting up a project from scratch.

It does a lot of things:
  - Detecting if updates is set up in `app.json`
  - Detecting the current type of project (generic vs managed)
  - Make changes relative to the "requested platform" and current `runtimeVersion`
  - Make sure native files are synced relative to "requested platform" and the current type of project
  - Try to log the steps made, relative to the type of project

If we had to write tests for this, there would be so many cases and variables to cover everything. It's also the main reason why I split this out over multiple functions in this PR.

That being said, we can definitely add these tests for the new split functions without having to set up whole projects during testing.

packages/eas-cli/src/commands/update/configure.ts
byCedric: I think you are right. I can compensate for that in the `warnEASUpdatesManualConfig`. We can iterate over the `modifyConfig` and replace `undefined` with this string.

packages/eas-cli/src/update/configure.ts
byCedric: One thing we can do here is changing the order of changes, much like how AST sanitization would work.

Instead of checking ahead of time if `runtimeVersions` are similar, we can do that after.
- Loop per platform first
  - Make required changes
- If requested platform === all && runtimeVersion is identical for all platforms: move it one level up

Does that sound better?

packages/eas-cli/src/update/configure.ts
byCedric: Setting something to `undefined` does indeed remove that property, at least when using `JSON.parse` (there is no `undefined` type in JSON).

I'll split it regardless for future testing.

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/**
+ * Make sure the `app.json` is configured to use EAS Updates.
+ * This does a couple of things:
+ *   - Ensure update URL is set to the project EAS endpoint
+ *   - Ensure runtimeVersion is defined for both or individual platforms
+ *   - Output the changes made, or the changes required to make manually
+ */
+export async function ensureEASUpdatesIsConfiguredInExpoConfigAsync({
```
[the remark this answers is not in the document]
byCedric: Probably for testing later, but I'll remove it now.

packages/eas-cli/src/update/configure.ts
```diff
@@ -0,0 +1,291 @@
...
+    );
+  }
+
+  Log.addNewLineIfNone();
+}
+
+/**
+ * Make sure that the current `app.json` configuration for EAS Updates is set natively.
+ */
+export async function ensureEASUpdatesIsConfiguredNativelyAsync(
```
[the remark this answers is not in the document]
byCedric: Probably for testing later, but I'll remove it now.

packages/eas-cli/src/commands/update/configure.ts
byCedric: Done

byCedric:
@wschurman made the changes, going to merge now. Feel free to reach out if anything comes up

PR #1485, opened 2022-10-27
[ENG-6238][eas-cli] add `build:run` support for Android

szdziedzic:
# Checklist

# Why

https://linear.app/expo/issue/ENG-6238/add-option-to-automatically-download-and-run-ios-simulator-build

# How

Add Android support for the `eas build:run` command. Reuse code from Expo local CLI.

https://user-images.githubusercontent.com/55145344/198262829-cd7641ac-cf34-4159-b5aa-02de4da03fb4.mov

Still needs some improvements in error handling and debugging.

Also I need to test it on Windows device.

# Test Plan

Test manually.

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -31,6 +31,7 @@
     "@oclif/core": "1.19.2",
     "@oclif/plugin-autocomplete": "1.3.4",
     "@segment/ajv-human-errors": "^2.1.2",
+    "@types/unixify": "1.0.0",
     "@urql/core": "3.0.5",
     "@urql/exchange-retry": "1.0.0",
     "ajv": "^8.11.0",
@@ -78,6 +79,7 @@
     "terminal-link": "2.1.1",
     "tslib": "2.4.0",
     "turndown": "7.1.1",
+    "unixify": "1.0.0",
     "untildify": "4.0.0",
     "uuid": "9.0.0",
     "wrap-ansi": "7.0.0"
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -31,7 +31,6 @@
     "@oclif/core": "1.19.2",
     "@oclif/plugin-autocomplete": "1.3.4",
     "@segment/ajv-human-errors": "^2.1.2",
-    "@types/unixify": "1.0.0",
     "@urql/core": "3.0.5",
     "@urql/exchange-retry": "1.0.0",
     "ajv": "^8.11.0",
@@ -79,7 +78,6 @@
     "terminal-link": "2.1.1",
     "tslib": "2.4.0",
     "turndown": "7.1.1",
-    "unixify": "1.0.0",
     "untildify": "4.0.0",
     "uuid": "9.0.0",
     "wrap-ansi": "7.0.0"
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -59,7 +59,7 @@ function formatBuildChoiceValue(value: string | undefined | null): string {
   return value ? chalk.bold(value) : chalk.dim('Unknown');
 }
 
-function formatBuildChoiceAndDescription(build: BuildFragment): {
+function formatBuildChoiceTitleAndDescription(build: BuildFragment): {
   title: string;
   description: string;
 } {
@@ -117,7 +117,7 @@ export async function listAndSelectBuildsOnAppAsync(
     message: `Select simulator build to run for ${projectDisplayName} app`,
     name: 'selectedSimulatorBuild',
     choices: builds.map(build => {
-      const buildChoice = formatBuildChoiceAndDescription(build);
+      const buildChoice = formatBuildChoiceTitleAndDescription(build);
       return {
         title: buildChoice.title,
         description: buildChoice.description,
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -6,6 +6,7 @@ import { AppPlatform, BuildFilter, BuildFragment } from '../graphql/generated';
 import { BuildQuery } from '../graphql/queries/BuildQuery';
 import Log from '../log';
 import { promptAsync } from '../prompts';
+import { fromNow } from '../utils/date';
 import { printJsonOnlyOutput } from '../utils/json';
 import { paginatedQueryWithConfirmPromptAsync } from '../utils/queries';
 import { formatGraphQLBuild } from './utils/formatBuild';
@@ -58,7 +59,10 @@ function formatBuildChoiceValue(value: string | undefined | null): string {
   return value ? chalk.bold(value) : chalk.dim('Unknown');
 }
 
-function formatBuildChoiceTitle(build: BuildFragment): string {
+function formatBuildChoiceAndDescription(build: BuildFragment): {
+  title: string;
+  description: string;
+} {
   const splittedCommitMessage = build.gitCommitMessage?.split('\n');
   const formattedCommitData =
     build.gitCommitHash && splittedCommitMessage && splittedCommitMessage.length > 0
@@ -67,14 +71,16 @@ function formatBuildChoiceTitle(build: BuildFragment): string {
         )}"`
       : 'Unknown';
 
-  return [
-    `ID: ${chalk.dim(build.id)}`,
-    `\tVersion: ${formatBuildChoiceValue(build.appVersion)}`,
-    `\t${
-      build.platform === AppPlatform.Ios ? 'Build number' : 'Version code'
-    }: ${formatBuildChoiceValue(build.appBuildVersion)}`,
-    `\tCommit: ${formattedCommitData}`,
-  ].join('\n');
+  return {
+    title: `ID: ${chalk.dim(build.id)} (${chalk.dim(`${fromNow(new Date(build.updatedAt))} ago`)})`,
+    description: [
+      `\tVersion: ${formatBuildChoiceValue(build.appVersion)}`,
+      `\t${
+        build.platform === AppPlatform.Ios ? 'Build number' : 'Version code'
+      }: ${formatBuildChoiceValue(build.appBuildVersion)}`,
+      `\tCommit: ${formattedCommitData}`,
+    ].join('\n'),
+  };
 }
 
 export async function listAndSelectBuildsOnAppAsync(
@@ -110,11 +116,15 @@ export async function listAndSelectBuildsOnAppAsync(
     type: 'select',
     message: `Select simulator build to run for ${projectDisplayName} app`,
     name: 'selectedSimulatorBuild',
-    choices: builds.map(build => ({
-      title: formatBuildChoiceTitle(build),
-      value: build,
-      disabled: selectPromptDisabledFunction?.(build),
-    })),
+    choices: builds.map(build => {
+      const buildChoice = formatBuildChoiceAndDescription(build);
+      return {
+        title: buildChoice.title,
+        description: buildChoice.description,
+        value: build,
+        disabled: selectPromptDisabledFunction?.(build),
+      };
+    }),
     warn: warningMessage,
   });
   return selectedSimulatorBuild;
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -63,11 +63,11 @@ function formatBuildChoiceTitleAndDescription(build: BuildFragment): {
   title: string;
   description: string;
 } {
-  const splittedCommitMessage = build.gitCommitMessage?.split('\n');
+  const splitCommitMessage = build.gitCommitMessage?.split('\n');
   const formattedCommitData =
-    build.gitCommitHash && splittedCommitMessage && splittedCommitMessage.length > 0
+    build.gitCommitHash && splitCommitMessage && splitCommitMessage.length > 0
       ? `${chalk.dim(build.gitCommitHash.slice(0, 7))} "${chalk.bold(
-          splittedCommitMessage[0] + (splittedCommitMessage.length > 1 ? '…' : '')
+          splitCommitMessage[0] + (splitCommitMessage.length > 1 ? '…' : '')
         )}"`
       : 'Unknown';
 
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 
 import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
 import { PaginatedQueryOptions } from '../commandUtils/pagination';
-import { BuildFilter, BuildFragment } from '../graphql/generated';
+import { AppPlatform, BuildFilter, BuildFragment } from '../graphql/generated';
 import { BuildQuery } from '../graphql/queries/BuildQuery';
 import Log from '../log';
 import { appPlatformDisplayNames } from '../platform';
@@ -87,7 +87,9 @@ export async function listAndSelectBuildsOnAppAsync(
     choices: builds.map(build => ({
       title: `id: ${build.id}, platform: ${appPlatformDisplayNames[build.platform]}, version: ${
         build.appVersion
-      }, build number: ${build.appBuildVersion}`,
+      }, ${build.platform === AppPlatform.Ios ? 'buildNumber' : 'versionCode'}: ${
+        build.appBuildVersion
+      }`,
       value: build,
     })),
   });
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -85,6 +85,7 @@ function formatBuildChoiceTitleAndDescription(build: BuildFragment): {
 
 export async function listAndSelectBuildsOnAppAsync(
   graphqlClient: ExpoGraphqlClient,
+  selectedPlatform: AppPlatform,
   {
     projectId,
     projectDisplayName,
@@ -114,7 +115,9 @@ export async function listAndSelectBuildsOnAppAsync(
 
   const { selectedSimulatorBuild } = await promptAsync({
     type: 'select',
-    message: `Select simulator build to run for ${projectDisplayName} app`,
+    message: `Select ${selectedPlatform === AppPlatform.Ios ? 'iOS' : 'Android'} ${
+      selectedPlatform === AppPlatform.Ios ? 'simulator' : 'emulator'
+    } build to run for ${projectDisplayName} app`,
     name: 'selectedSimulatorBuild',
     choices: builds.map(build => {
       const buildChoice = formatBuildChoiceTitleAndDescription(build);
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -59,9 +59,12 @@ function formatBuildChoiceValue(value: string | undefined | null): string {
 }
 
 function formatBuildChoiceTitle(build: BuildFragment): string {
+  const splittedCommitMessage = build.gitCommitMessage?.split('\n');
   const formattedCommitData =
-    build.gitCommitHash && build.gitCommitMessage
-      ? `${chalk.dim(build.gitCommitHash.slice(0, 7))} "${chalk.bold(build.gitCommitMessage)}"`
+    build.gitCommitHash && splittedCommitMessage && splittedCommitMessage.length > 0
+      ? `${chalk.dim(build.gitCommitHash.slice(0, 7))} "${chalk.bold(
+          splittedCommitMessage[0] + (splittedCommitMessage.length > 1 ? '…' : '')
+        )}"`
       : 'Unknown';
 
   return [
--- a/packages/eas-cli/src/build/queries.ts
+++ b/packages/eas-cli/src/build/queries.ts
@@ -5,7 +5,6 @@ import { PaginatedQueryOptions } from '../commandUtils/pagination';
 import { AppPlatform, BuildFilter, BuildFragment } from '../graphql/generated';
 import { BuildQuery } from '../graphql/queries/BuildQuery';
 import Log from '../log';
-import { appPlatformDisplayNames } from '../platform';
 import { promptAsync } from '../prompts';
 import { printJsonOnlyOutput } from '../utils/json';
 import { paginatedQueryWithConfirmPromptAsync } from '../utils/queries';
@@ -55,6 +54,26 @@ export async function listAndRenderBuildsOnAppAsync(
   }
 }
 
+function formatBuildChoiceValue(value: string | undefined | null): string {
+  return value ? chalk.bold(value) : chalk.dim('Unknown');
+}
+
+function formatBuildChoiceTitle(build: BuildFragment): string {
+  const formattedCommitData =
+    build.gitCommitHash && build.gitCommitMessage
+      ? `${chalk.dim(build.gitCommitHash.slice(0, 7))} "${chalk.bold(build.gitCommitMessage)}"`
+      : 'Unknown';
+
+  return [
+    `ID: ${chalk.dim(build.id)}`,
+    `\tVersion: ${formatBuildChoiceValue(build.appVersion)}`,
+    `\t${
+      build.platform === AppPlatform.Ios ? 'Build number' : 'Version code'
+    }: ${formatBuildChoiceValue(build.appBuildVersion)}`,
+    `\tCommit: ${formattedCommitData}`,
+  ].join('\n');
+}
+
 export async function listAndSelectBuildsOnAppAsync(
   graphqlClient: ExpoGraphqlClient,
   {
@@ -62,11 +81,15 @@ export async function listAndSelectBuildsOnAppAsync(
     projectDisplayName,
     filter,
     queryOptions,
+    selectPromptDisabledFunction,
+    warningMessage,
   }: {
     projectId: string;
     projectDisplayName: string;
     filter?: BuildFilter;
     queryOptions: PaginatedQueryOptions;
+    selectPromptDisabledFunction?: (build: BuildFragment) => boolean;
+    warningMessage?: string;
   }
 ): Promise<BuildFragment> {
   const builds = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
@@ -85,13 +108,11 @@ export async function listAndSelectBuildsOnAppAsync(
     message: `Select simulator build to run for ${projectDisplayName} app`,
     name: 'selectedSimulatorBuild',
     choices: builds.map(build => ({
-      title: `id: ${build.id}, platform: ${appPlatformDisplayNames[build.platform]}, version: ${
-        build.appVersion
-      }, ${build.platform === AppPlatform.Ios ? 'buildNumber' : 'versionCode'}: ${
-        build.appBuildVersion
-      }`,
+      title: formatBuildChoiceTitle(build),
       value: build,
+      disabled: selectPromptDisabledFunction?.(build),
     })),
+    warn: warningMessage,
   });
   return selectedSimulatorBuild;
 }
--- a/packages/eas-cli/src/commands/build/run.ts
+++ b/packages/eas-cli/src/commands/build/run.ts
@@ -143,7 +143,7 @@ async function resolvePlatformAsync(platform?: string): Promise<AppPlatform> {
     choices: [
       { title: 'Android', value: AppPlatform.Android },
       { title: 'iOS', value: AppPlatform.Ios },
-    ].filter(choice => choice.value !== AppPlatform.Ios || process.platform === 'darwin'),
+    ],
   });
   return selectedPlatform;
 }
@@ -165,7 +165,7 @@ async function maybeGetBuildAsync(
     !flags.runArchiveFlags.url &&
     !flags.runArchiveFlags.latest
   ) {
-    return await listAndSelectBuildsOnAppAsync(graphqlClient, {
+    return await listAndSelectBuildsOnAppAsync(graphqlClient, flags.selectedPlatform, {
       projectId,
       projectDisplayName: await getDisplayNameForProjectIdAsync(graphqlClient, projectId),
       filter: {
--- a/packages/eas-cli/src/commands/build/run.ts
+++ b/packages/eas-cli/src/commands/build/run.ts
@@ -99,7 +99,7 @@ export default class Run extends EasCommand {
     const selectedPlatform = await resolvePlatformAsync(platform);
 
     if (platform === 'ios' && process.platform !== 'darwin') {
-      Errors.error('You can only run iOS build on Apple device', {
+      Errors.error('You can only run iOS build on macOS devices', {
         exit: 1,
       });
     }
@@ -173,7 +173,7 @@ async function maybeGetBuildAsync(
       selectPromptDisabledFunction: build =>
         build.platform === AppPlatform.Ios
           ? false
-          : !build.artifacts?.applicationArchiveUrl?.endsWith('.apk'),
+          : !build.artifacts?.applicationArchiveUrl?.endsWith('.apk') ?? false,
       warningMessage: 'This build is not a simulator/emulator build',
     });
   } else if (flags.runArchiveFlags.latest) {
--- a/packages/eas-cli/src/commands/build/run.ts
+++ b/packages/eas-cli/src/commands/build/run.ts
@@ -98,6 +98,12 @@ export default class Run extends EasCommand {
 
     const selectedPlatform = await resolvePlatformAsync(platform);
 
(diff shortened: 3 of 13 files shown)


packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,139 @@
...
+  const attachedDevices: {
+    pid: string;
+    type: string;
+  }[] = splitItems
+    .slice(1, splitItems.length)
+    .map(line => {
+      // unauthorized: ['FA8251A00719', 'unauthorized', 'usb:338690048X', 'transport_id:5']
+      // authorized: ['FA8251A00719', 'device', 'usb:336592896X', 'product:walleye', 'model:Pixel_2', 'device:walleye', 'transport_id:4']
+      // emulator: ['emulator-5554', 'offline', 'transport_id:1']
+      const [pid] = line.split(' ').filter(Boolean);
```
dsokal, line 78: Search for the other `truthy` usages in the code.
```suggestion
      const [pid] = line.split(' ').filter(truthy);
```

packages/eas-cli/src/run/android/emulator.ts
```diff
@@ -0,0 +1,161 @@
...
+  Log.newLine();
+  Log.log(`Opening emulator ${chalk.bold(emulator.name)}`);
+
+  // Start a process to open an emulator
+  const emulatorProcess = spawnAsync(
+    emulatorExecutable,
+    [
+      `@${emulator.name}`,
+      // disable animation for faster boot -- this might make it harder to detect if it mounted properly tho
+      //'-no-boot-anim'
```
dsokal, line 75:
```suggestion
```

packages/eas-cli/src/run/android/emulator.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    return runningEmulator;
+  }
+
+  const emulators = await getAvaliableAndroidEmulatorsAsync();
+
+  Log.newLine();
+  const { selectedEmulator } = await promptAsync({
+    type: 'select',
+    message: `Select a emulator to run your app on`,
```
dsokal, line 103:
```suggestion
    message: `Select an emulator to run your app on:`,
```

packages/eas-cli/src/run/android/run.ts
```diff
@@ -1,5 +1,17 @@
...
+
+  const selectedEmulator = await emulator.selectEmulatorAsync();
+
+  const bootedEmulator = await emulator.ensureEmulatorBootedAsync(selectedEmulator);
+
+  await emulator.installAppAsync(bootedEmulator, appPath);
+
+  const { packageName, activityName } = await getAptParametersAsync(appPath);
+
+  await emulator.startAppAsync(bootedEmulator, packageName, activityName);
```
dsokal, line 16: :

D


```suggestion
  await assertExecutablesExistAsync();
  const selectedEmulator = await emulator.selectEmulatorAsync();
  const bootedEmulator = await emulator.ensureEmulatorBootedAsync(selectedEmulator);
  await emulator.installAppAsync(bootedEmulator, appPath);
  const { packageName, activityName } = await getAptParametersAsync(appPath);
  await emulator.startAppAsync(bootedEmulator, packageName, activityName);
```

packages/eas-cli/src/run/android/systemRequirements.ts
```diff
@@ -0,0 +1,45 @@
...
+  return stdout.trim();
+}
+
+function assertExecutableExists(executable: string): void {
+  if (executable.includes('/')) {
+    if (!existsSync(executable)) {
+      throw new Error(
+        `Couldn't find ${chalk.bold(
+          executable
+        )} executable in the Android SDK. Please make sure it's installed.`
```
dsokal, line 25: Is it me or this sounds a little weird?

"Please make sure it's installed." - I'm not sure what is "it" in this sentence.

packages/eas-cli/src/run/android/sdk.ts
```diff
@@ -0,0 +1,25 @@
+import { existsSync } from 'fs-extra';
+import os from 'os';
+import path from 'path';
+
+export const ANDROID_DEFAULT_LOCATION: Readonly<Partial<Record<NodeJS.Platform, string>>> = {
+  darwin: path.join(os.homedir(), 'Library', 'Android', 'sdk'),
+  linux: path.join(os.homedir(), 'Android', 'sdk'),
```
wkozyra95, line 7:
```suggestion
  linux: path.join(os.homedir(), 'Android', 'Sdk'),
```

packages/eas-cli/src/run/android/sdk.ts
```diff
@@ -0,0 +1,25 @@
...
+  win32: path.join(os.homedir(), 'AppData', 'Local', 'Android', 'Sdk'),
+};
+
+const defaultLocation = ANDROID_DEFAULT_LOCATION[process.platform];
+
+export function getAndroidSdkRoot(): string | null {
+  if (process.env.ANDROID_HOME && existsSync(process.env.ANDROID_HOME)) {
```
wkozyra95, line 14: Do not use sync methods if you do not have a strong reason to do so.

packages/eas-cli/src/run/android/sdk.ts
```diff
@@ -0,0 +1,25 @@
...
+  } else if (process.env.ANDROID_SDK_ROOT && existsSync(process.env.ANDROID_SDK_ROOT)) {
+    return process.env.ANDROID_SDK_ROOT;
+  } else if (defaultLocation && existsSync(defaultLocation)) {
+    return defaultLocation;
+  } else {
+    return null;
+  }
+}
+
+export const sdkRoot = getAndroidSdkRoot();
```
wkozyra95, line 25: this will execute on import, it might slow down all eas-cli commands.

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -143,6 +144,11 @@ async function maybeGetBuildAsync(
   projectId: string,
   paginatedQueryOptions: PaginatedQueryOptions
 ): Promise<BuildFragment | null> {
+  const distributionType =
+    flags.selectedPlatform === AppPlatform.Ios
+      ? DistributionType.Simulator
+      : DistributionType.Internal;
```
wkozyra95, line 150: why? you can install any android build in emulator, not only internal distribution

packages/eas-cli/src/run/android/aapt.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import Log from '../../log';
+import { sdkRoot } from './sdk';
+
+async function aaptAsync(...options: string[]): Promise<SpawnResult> {
+  try {
+    return await spawnAsync(await getAaptExecutableAsync(), options);
+  } catch (error: any) {
+    const errorMessage = (error.stderr || error.stdout || error.message).trim();
+    error.message = errorMessage;
```
wkozyra95, line 12:
```suggestion
    if (error.stderr) {
      Log.error(error.stderr);
      throw error;
    }
```
Maybe that would be better?
Note that this approach will not be correct if you are catching aapt errors in some cases

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,139 @@
...
+  return sanitizeAdbDeviceName(stdout) ?? null;
+}
+
+// TODO: This is very expensive for some operations.
+export async function getRunningEmulatorsAsync(): Promise<AndroidEmulator[]> {
+  const { stdout } = await adbAsync('devices', '-l');
+
+  const splitItems = stdout.trim().replace(/\n$/, '').split(os.EOL);
+  // First line is `"List of devices attached"`, remove it
+  // @ts-ignore: todo
```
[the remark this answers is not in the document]
szdziedzic: It's not. Thanks!

packages/eas-cli/src/run/android/emulator.ts
```diff
@@ -0,0 +1,161 @@
...
+      // disable animation for faster boot -- this might make it harder to detect if it mounted properly tho
+      //'-no-boot-anim'
+    ],
+    {
+      stdio: 'ignore',
+      detached: true,
+    }
+  );
+
+  emulatorProcess.child.unref();
```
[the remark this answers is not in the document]
szdziedzic: From `child_process` JS Doc:
```
By default, the parent will wait for the detached child to exit. To prevent the parent from waiting for a given subprocess to exit, use the subprocess.unref() method. Doing so will cause the parent's event loop to not include the child in its reference count, allowing the parent to exit independently of the child unless there is an established IPC channel between the child and the parent.
```

In our case, we want the `eas build:run` command to finish once the app is installed and launched and we don't want to wait with it until the emulator is closed.

How would it look like without this line:

https://user-images.githubusercontent.com/55145344/198571339-6e3813b8-f387-4238-b105-f6d731be8a85.mov

I will also add a similar explanation in the code

packages/eas-cli/src/run/android/systemRequirements.ts
```diff
@@ -0,0 +1,45 @@
+import spawnAsync from '@expo/spawn-async';
+import chalk from 'chalk';
+import { existsSync } from 'fs-extra';
+
+import { getAaptExecutableAsync } from './aapt';
+import { adbExecutable } from './adb';
+import { emulatorExecutable } from './emulator';
+
+async function whichAsync(executable: string): Promise<string | null> {
```
wkozyra95, line 9: Do not use `which`, you can check if an executable exists by calling it and checking the error.

packages/eas-cli/src/commands/build/run.ts
szdziedzic: @brentvatne How do you think this filtering should work? We had a conversation about it today with @wkozyra95  and it looks like currently there is no trivial way to retrieve only builds containing `.apk` from the backend. 

I see some potential options:

1. **Add server-side filtering**: We can create an index on the `artifacts` column in the database (which is `JSONB` type) and filter builds on this column using regex to see if the `applicationArchiveUrl` contains `.apk`, which doesn't sound like a perfect solution, to be honest. It feels like it would cause some performance issues.
2. **Add client-side filtering**: We can query for all Android builds and filter them on the `build.artifacts.applicationArchiveUrl` field as above. In some cases, we would need to hit the backend multiple times so the number of builds left after filtering is equal to the `limit` parameter. The factor limiting the number of queries we make to the backend, in this case, is the fact that the archives for builds older than 30 days get deleted, so it should be a reasonable number. It's also not a perfect solution.
3. **Use some easier, but more inaccurate filtering**: We can use simple backend filters like `distributionType === INTERNAL`, but it can result in query results containing `.aab` builds and missing on some `.apk` builds as well. It feels like it would work correctly for the majority of the builds, although it's still an inaccurate solution and it might be confusing for the users.

What do you think about it?

CC @dsokal

wkozyra95: Ad 1. I'm reworking that part currently in www that value can URL/bucket key. We would need to have an index on a column with a json object, which is not great.
Ad. 2. Client-side filtering is a bad idea in most cases, but here it could work. Normally the problem is that if you want to show 10 builds you would need to query in a loop until you get those builds or check all builds. In most cases, it would not be acceptable to abuse API like that, but here we only need builds that artifacts did not expire, so upper limit on number of requests is not that high.
Ad. 3 It's maybe not 100% correct solution, but I think skipping filtering by artifact type is the best approach here. I'm not sure whether we should filter only internal distribution builds or just take all android builds. With this approach, we would just list the latest builds(or the latest internal distribution) without any validation, and if the user selects a build that produced aab, we would just throw an error then.

brentvatne: I like the synchronicity of this discussion! I had this on my tasks for the day:

> Create task: when aab is built for internal distribution, don't let people download and try to install

We had a developer email us with some confusion around installing an internal distribution build on their device. It turned out that they had configured the build to produce an aab rather than the default apk for internal distribution, but on our build page we didn't provide any indication that this may be a problem, and we provided the usual UI for them to download and attempt to launch it on their device. It turned out that they had an app installed that was registered to launch files with the aab extension, so when they opened the downloaded aab it opened another application entirely, creating grave concern.

<hr />

Here's another possible idea - what if we showed all builds in the list but made any aab's unselectable and made that visually clear? This would be workable as a first pass and it would also allow us to determine whether this is annoying for users. It also helps to communicate that aab's are not directly installable to devices and apks are. If we don't show aabs in the list, I could imagine some users running this command and being confused as to why they don't see a particular build. This allows us to communicate that directly, at the cost of having to possibly sort through more builds in the list. Ultimately we will add some filters to this in the future, such as filter by development client builds, and so users can take care of filtering out aabs on their own.

dsokal: > Here's another possible idea - what if we showed all builds in the list but made any aab's unselectable and made that visually clear

I like the idea. My only concern is what if all builds are unselectable? 🤔

packages/eas-cli/src/run/android/aapt.ts
```diff
@@ -0,0 +1,52 @@
...
+    Log.debug('Failed to resolve the Android SDK path, falling back to global aapt executable');
+    return 'aapt';
+  }
+  const aaptPaths = await glob(`${sdkRoot}/build-tools/**/aapt`);
+
+  if (aaptPaths.length === 0) {
+    throw new Error('Failed to resolve the Android aapt path');
+  }
+
+  return aaptPaths.sort().reverse()[0];
```
dsokal, line 29: This is more straightforward:
```suggestion
  const sorted = aaptPaths.sort();
  return sorted[sorted.length - 1];
```

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  return `${sdkRoot}/platform-tools/adb`;
+}
+
+export function sanitizeAdbDeviceName(deviceName: string): string | undefined {
+  return deviceName
+    .trim()
+    .split(/[\r\n]+/)
+    .shift();
```
dsokal, line 44:
```suggestion
  return deviceName
    .trim()
    .split(/[\r\n]+/)[0];
```

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,137 @@
...
+    pid: string;
+    type: string;
+  }[] = splitItems
+    .slice(1, splitItems.length)
+    .map(line => {
+      // unauthorized: ['FA8251A00719', 'unauthorized', 'usb:338690048X', 'transport_id:5']
+      // authorized: ['FA8251A00719', 'device', 'usb:336592896X', 'product:walleye', 'model:Pixel_2', 'device:walleye', 'transport_id:4']
+      // emulator: ['emulator-5554', 'offline', 'transport_id:1']
+      const [pid] = line.split(' ').filter(truthy);
+
```
dsokal, line 79:
```suggestion
```

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,137 @@
...
+  const devicePromises = attachedDevices.map<Promise<AndroidEmulator>>(async props => {
+    const { pid } = props;
+
+    const name = (await getAdbNameForDeviceIdAsync(pid)) ?? '';
+
+    return {
+      pid,
+      name,
+    };
+  });
```
dsokal, line 94:
```suggestion
  const devicePromises = attachedDevices.map<Promise<AndroidEmulator>>(async ({ pid }) => {
    const name = (await getAdbNameForDeviceIdAsync(pid)) ?? '';
    return {
      pid,
      name,
    };
  });
```

packages/eas-cli/src/run/android/sdk.ts
```diff
@@ -0,0 +1,23 @@
...
+import os from 'os';
+import path from 'path';
+
+export const ANDROID_DEFAULT_LOCATION: Readonly<Partial<Record<NodeJS.Platform, string>>> = {
+  darwin: path.join(os.homedir(), 'Library', 'Android', 'sdk'),
+  linux: path.join(os.homedir(), 'Android', 'Sdk'),
+  win32: path.join(os.homedir(), 'AppData', 'Local', 'Android', 'Sdk'),
+};
+
+const defaultLocation = ANDROID_DEFAULT_LOCATION[process.platform];
```
dsokal, line 11:
```suggestion
const ANDROID_DEFAULT_LOCATION_FOR_CURRENT_PLATFORM = ANDROID_DEFAULT_LOCATION[process.platform];
```

packages/eas-cli/src/run/android/systemRequirements.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export async function assertExecutablesExistAsync(): Promise<void> {
+  for (const executable of [
+    await getAdbExecutableAsync(),
+    await getEmulatorExecutableAsync(),
+    await getAaptExecutableAsync(),
+  ]) {
+    await assertExecutableExistsAsync(executable);
+  }
```
dsokal, line 47:
```suggestion
  const xyz = [
    await getAdbExecutableAsync(),
    await getEmulatorExecutableAsync(),
    await getAaptExecutableAsync(),
  ];
  for (const executable of xyz) {
    await assertExecutableExistsAsync(executable);
  }
```

packages/eas-cli/src/commands/build/run.ts
wkozyra95: > My only concern is what if all builds are unselectable?

We are already doing that for submits, when artifacts expire we still show those builds but it's crossed out.

szdziedzic: I also like this idea, although I wonder if this is a problem if the command behavior is inconsistent between iOS and Android. Meaning that on Android we show all builds with some of them in an unselectable state, but on iOS, we show only filtered simulator builds. I'm wondering if this is acceptable in your opinion 🤔 (I assume that it is, although I want to double-check).

szdziedzic: >We are already doing that for submits, when artifacts expire we still show those builds but it's crossed out.

@wkozyra95 I think we log an error in such a case.

Here is the code which handles it: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/submit/ArchiveSource.ts#L260

Screenshot 2022-11-03 at 11 17 19

szdziedzic: This is how it looks now:

Screenshot 2022-11-04 at 09 45 39
Screenshot 2022-11-04 at 09 45 51


Please notice, you don't see the underline issues on these screen shots because I'm using VSCode. The problem still exists for other terminals.
https://github.com/expo/eas-cli/pull/1502#issuecomment-1302127360

szdziedzic (commented):
https://user-images.githubusercontent.com/55145344/203332637-7e15da54-c2cd-4e2c-8f74-2abf195e0ed9.mov

This is how the CLI flow looks like after the updates (removed downloads progress bar and added progress tracker - same we use for uploads, updated choice formatting when selecting builds)

packages/eas-cli/src/run/android/aapt.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+export async function getAaptExecutableAsync(): Promise<string> {
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (!sdkRoot) {
+    Log.debug('Failed to resolve the Android SDK path, falling back to global aapt executable');
+    return 'aapt';
+  }
+  const aaptPaths = await glob(`${sdkRoot}/build-tools/**/aapt`);
```
dsokal, line 24: Use `path.join` instead of interpolation.

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export async function getAdbExecutableAsync(): Promise<string> {
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (!sdkRoot) {
+    Log.debug('Failed to resolve the Android SDK path, falling back to global adb executable');
+    return 'adb';
+  }
+
+  return `${sdkRoot}/platform-tools/adb`;
```
dsokal, line 37: `path.join`

packages/eas-cli/src/run/android/emulator.ts
```diff
@@ -0,0 +1,159 @@
...
+  waitForEmulatorToBeBootedAsync,
+} from './adb';
+import { getAndroidSdkRootAsync } from './sdk';
+
+export const EMULATOR_MAX_WAIT_TIMEOUT = 60 * 1000 * 3;
+
+export async function getEmulatorExecutableAsync(): Promise<string> {
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (sdkRoot) {
+    return `${sdkRoot}/emulator/emulator`;
```
dsokal, line 23: `path.join`

packages/eas-cli/src/run/android/emulator.ts
```diff
@@ -0,0 +1,159 @@
...
+import {
+  AndroidEmulator,
+  adbAsync,
+  getFirstRunningEmulatorAsync,
+  isEmulatorBootedAsync,
+  waitForEmulatorToBeBootedAsync,
+} from './adb';
+import { getAndroidSdkRootAsync } from './sdk';
+
+export const EMULATOR_MAX_WAIT_TIMEOUT = 60 * 1000 * 3;
```
dsokal, line 18:
```suggestion
export const EMULATOR_MAX_WAIT_TIMEOUT_MS = 3 * 60 * 1000;
```

packages/eas-cli/src/utils/download.ts
```diff
@@ -94,34 +92,51 @@ async function downloadFileWithProgressBarAsync(
...
 
-  const tmpArchivePathDir = path.join(getTmpDirectory(), uuidv4());
-  await fs.promises.mkdir(tmpArchivePathDir, { recursive: true });
-
-  const tmpArchivePath = path.join(tmpArchivePathDir, `${uuidv4()}.tar.gz`);
-
-  await downloadFileWithProgressBarAsync(url, tmpArchivePath, 'Downloading app archive...');
-  await tarExtractAsync(tmpArchivePath, outputDir);
+  if (platform === AppPlatform.Android) {
+    const apkFilePath = path.join(outputDir, `${uuidv4()}.apk`);
```
dsokal, line 103: Is it always an .apk?

packages/eas-cli/src/build/queries.ts
```diff
@@ -55,18 +55,50 @@ export async function listAndRenderBuildsOnAppAsync(
...
 
+function formatBuildChoiceValue(value: string | undefined | null): string {
+  return value ? chalk.bold(value) : chalk.dim('Unknown');
+}
+
+function formatBuildChoiceTitleAndDescription(build: BuildFragment): {
+  title: string;
+  description: string;
+} {
+  const splittedCommitMessage = build.gitCommitMessage?.split('\n');
```
dsokal, line 66: https://livexp.com/blog/is-it-split-or-splitted-and-some-other-similar-irregular-verbs/

packages/eas-cli/src/utils/download.ts
szdziedzic: I've always seen `.apk` for my apps built from the example project, although you are right I will make it more elastic and support both `apk` and tarball for Android 👍 .

packages/eas-cli/src/build/queries.ts
szdziedzic: Thanks!

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  return sanitizeAdbDeviceName(stdout) ?? null;
+}
+
+// TODO: This is very expensive for some operations.
+export async function getRunningEmulatorsAsync(): Promise<AndroidEmulator[]> {
+  const { stdout } = await adbAsync('devices', '-l');
+
+  const splitItems = stdout.trim().replace(/\n$/, '').split(os.EOL);
```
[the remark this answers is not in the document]
szdziedzic: It was in Evan's code, although I deleted it and it still seems to work just fine, so I suppose it isn't really needed 🤔

packages/eas-cli/src/run/android/systemRequirements.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (executable.includes('/')) {
+    if (!(await pathExists(executable))) {
+      throw new Error(
+        `Couldn't find ${chalk.bold(
+          executable
+        )} executable in the Android SDK. Please make sure ${chalk.bold(executable)} is installed.`
+      );
+    }
+  } else {
+    if (!(await checkIfGlobalExecutableExistsAsync('addd'))) {
```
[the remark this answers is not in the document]
szdziedzic: Right, sorry, good catch

packages/eas-cli/src/run/android/aapt.ts
dsokal: It's better than interpolation because you don't have to think whether `sdkRoot` ends with `/` or not.

packages/eas-cli/src/run/android/aapt.ts
```diff
@@ -0,0 +1,58 @@
...
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (!sdkRoot) {
+    Log.debug('Failed to resolve the Android SDK path, falling back to global aapt executable');
+    return 'aapt';
+  }
+  const aaptPaths = await glob(
+    `${unixify(sdkRoot)}/build-tools/**/${
+      ['win32', 'win64'].includes(process.platform) ? 'aapt.exe' : 'aapt'
+    }`
+  );
```
szdziedzic, line 29: After fixing this part I was able to successfully run the command on my Windows machine 🎉

szdziedzic: I will add few UX improvements for Windows users and I think than it should be good to go 😄

wkozyra95: where did you get win64 from? there is no such platform, also you did not address Dominik's comment about using path.join instead of interpolation

szdziedzic: @wkozyra95 I thought `win64` is a thing, so sorry 🤦.  It should be fixed now.

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -132,7 +139,7 @@ async function resolvePlatformAsync(platform?: string): Promise<AppPlatform> {
     choices: [
       { title: 'Android', value: AppPlatform.Android },
       { title: 'iOS', value: AppPlatform.Ios },
-    ],
+    ].filter(choice => choice.value !== AppPlatform.Ios || process.platform === 'darwin'),
```
szdziedzic, line 142: Do we want to prompt non-Mac users to select a platform having only one option available? Or just always select Android for them if the platform flag is not specified?

I suppose we want to just select Android but what is your opinion?

packages/eas-cli/src/run/android/aapt.ts
```diff
@@ -0,0 +1,59 @@
...
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (!sdkRoot) {
+    Log.debug('Failed to resolve the Android SDK path, falling back to global aapt executable');
+    return 'aapt';
+  }
+  const aaptPaths = await glob(
+    unixify(
+      path.join(sdkRoot, 'build-tools', '**', process.platform === 'win32' ? 'aapt.exe' : 'aapt')
+    )
+  );
```
szdziedzic, line 30: This is an improved glob, working both on Windows and Mac

packages/eas-cli/src/commands/build/run.ts
brentvatne: let's select android by default

brentvatne (approved):

packages/eas-cli/src/run/android/aapt.ts
wkozyra95, line 30:
```suggestion
  const aaptPaths = await glob(
      path.posix.join('build-tools/**', process.platform === 'win32' ? 'aapt.exe' : 'aapt')`,
      { cwd: sdkRoot, absolute: true }
  );
```
unixify is loosing information e.g. both `C:\\sth` and `D:\\sth` would be resolved to `/sth`. I have no idea how fast-glob is handling patterns that are absolute path on windows, but even if it works correctly we should not rely on that.

packages/eas-cli/src/run/android/adb.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export async function getAdbExecutableAsync(): Promise<string> {
+  const sdkRoot = await getAndroidSdkRootAsync();
+  if (!sdkRoot) {
+    Log.debug('Failed to resolve the Android SDK path, falling back to global adb executable');
+    return 'adb';
+  }
+
+  return path.join(sdkRoot, 'platform-tools', 'adb');
```
wkozyra95, line 38:
```suggestion
  return path.join(sdkRoot, 'platform-tools/adb');
```

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -97,13 +98,19 @@ export default class Run extends EasCommand {
 
     const selectedPlatform = await resolvePlatformAsync(platform);
 
+    if (platform === 'ios' && process.platform !== 'darwin') {
+      Errors.error('You can only run iOS build on Apple device', {
```
dsokal, line 102:
```suggestion
      Errors.error('You can only run iOS build on macOS devices', {
```

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -156,21 +166,22 @@ async function maybeGetBuildAsync(
...
-          BuildDistributionType.SIMULATOR
-        ),
+        distribution: distributionType,
         status: BuildStatus.Finished,
       },
       queryOptions: paginatedQueryOptions,
+      selectPromptDisabledFunction: build =>
+        build.platform === AppPlatform.Ios
+          ? false
+          : !build.artifacts?.applicationArchiveUrl?.endsWith('.apk'),
```
dsokal, line 176:
```suggestion
          : !build.artifacts?.applicationArchiveUrl?.endsWith('.apk') ?? REPLACE_WITH_APPROPRIATE_BOOLEAN,
```

dsokal: It's weird that TS didn't catch that.

packages/eas-cli/src/run/android/aapt.ts
dsokal: Agreed with Wojtek, we shouldn't need to use `unixify`.

packages/eas-cli/package.json
```diff
@@ -78,6 +79,7 @@
     "terminal-link": "2.1.1",
     "tslib": "2.4.0",
     "turndown": "7.1.1",
+    "unixify": "1.0.0",
```
dsokal, line 82: See the other comment, I think we don't need this package.

packages/eas-cli/src/commands/build/run.ts
szdziedzic: Good catch!

dsokal (approved):

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -97,13 +98,19 @@ export default class Run extends EasCommand {
 
     const selectedPlatform = await resolvePlatformAsync(platform);
 
+    if (platform === 'ios' && process.platform !== 'darwin') {
+      Errors.error('You can only run iOS build on macOS devices', {
```
wkozyra95, line 102: Copy paste? You are not building here

packages/eas-cli/src/commands/build/run.ts
```diff
@@ -151,26 +165,27 @@ async function maybeGetBuildAsync(
...
-        ),
+        distribution: distributionType,
         status: BuildStatus.Finished,
       },
       queryOptions: paginatedQueryOptions,
+      selectPromptDisabledFunction: build =>
+        build.platform === AppPlatform.Ios
+          ? false
+          : !build.artifacts?.applicationArchiveUrl?.endsWith('.apk') ?? false,
+      warningMessage: 'This build is not a simulator/emulator build',
```
wkozyra95, line 181:
```suggestion
      warningMessage: 'This is not a simulator/emulator build',
```

wkozyra95 (approved):

packages/eas-cli/src/commands/build/run.ts
szdziedzic: Not a copy-paste 😄 I can do:
```suggestion
      Errors.error('You can only use an iOS simulator to run apps on macOS devices', {
```
Is it okay?

wkozyra95: :+1:

szdziedzic:
/changelog-entry new-feature Add `eas build:run` support for Android

PR #1499, opened 2022-11-01
[eas-cli][update] Attach Git Commit to EAS Update

FiberJW:
# Checklist

# Why

We'd like for users to be able to associate an Update with a Git commit.

# How

I added `gitCommitHash` to the input we send to our GraphQL API for creating Update records. I also had to fix a few GraphQL codegen issues because we did a refactor which changed where some GraphQL mutations were defined.

I use the same method that we use to get the git commit hash for builds: `await vcsClient.getCommitHashAsync()`

# Test Plan

Using a local version of the CLI and website, I published a new update on Staging and used the website to render the new commit data (and GitHub link 👀 )

Screen Shot 2022-11-01 at 16 11 30
Screen Shot 2022-11-01 at 16 10 53

Updated log:

Screen Shot 2022-11-01 at 16 33 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Add Git commit to EAS Update record. ([#1499](https://github.com/expo/eas-cli/pull/1499) by [@fiberjw](https://github.com/fiberjw))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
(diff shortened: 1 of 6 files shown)


wschurman (approved):
Would also be good to point to how it's done for builds in the PR summary (to ensure it's the same definition of a commit). lgtm though.

dsokal (approved):

wkozyra95 (approved):
In builds, we also report if the git working tree is dirty
https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/build/metadata.ts#L50

As far as I remember that value is not shown to the users there, but it might be helpful to mark that somehow on the website both for updates and builds e.g. by displaying  `19199af2` for clean state and `19199af2*` dirty.

FiberJW:
@wkozyra95 that's a good idea! I can follow up with a change to the API to let us record that info

PR #1500, opened 2022-11-01
refactor(update): deprecate the --auto flag for eas update

byCedric:
# Checklist

# Why

Users can define a custom `--branch` and `--message` value when publishing an update. Previously, when one of these wasn't provided, it would fail unless you added the `--auto` flag.

With this change, you don't need to provide `--auto`.

# How

I did not use the `deprecated: true` option on the flag, which always logs a warning (even when not using the flag)
image

Instead, I moved this logging to `sanitizeFlags` and only warn people not to use it when they actually use it.

When used | When omitted
--- | ---
image | image

# Test Plan

You can use my test version of EAS CLI with `npx @bycedric/some-cli-test@3.0.5 <command>`, or build it yourself.

- `eas update`
- `eas update --auto`
- `eas update --republish --groupId`

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -38,7 +38,11 @@ import {
 import { resolveWorkflowAsync } from '../../project/workflow';
 import { promptAsync } from '../../prompts';
 import { selectUpdateGroupOnBranchAsync } from '../../update/queries';
-import { checkEASUpdateURLIsSetAsync, formatUpdateMessage } from '../../update/utils';
+import {
+  checkEASUpdateURLIsSetAsync,
+  formatUpdateMessage,
+  truncateString as truncateUpdateMessage,
+} from '../../update/utils';
 import {
   checkManifestBodyAgainstUpdateInfoGroup,
   getCodeSigningInfoAsync,
@@ -182,7 +186,7 @@ export default class UpdatePublish extends EasCommand {
     const [runtimeVersions, exp] = await getRuntimeVersionObjectAsync(
       expBeforeRuntimeVersionUpdate,
       platformFlag,
-      projectDir,
+      projectDir
       // projectId,
       // nonInteractive,
       // graphqlClient,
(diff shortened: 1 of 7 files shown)


byCedric:
/changelog-entry chore Deprecate `--auto` flag on `eas update` and always default `--branch` or `--message` from git

EvanBacon (approved):

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -532,12 +533,12 @@ export default class UpdatePublish extends EasCommand {
...
     const { auto, branch: branchName, message: updateMessage } = flags;
-    if (nonInteractive && !auto && !(branchName && updateMessage)) {
-      Errors.error(
-        '--auto or both --branch and --message are required when updating in non-interactive mode',
-        { exit: 1 }
+
+    if (auto) {
+      // When `branchName` or `updateMessage` is omitted, we always (auto) default to git
+      Log.warn(
+        '--auto flag is deprecated. When --branch and --message are not provided, both flags default to git branch and commit message.'
```
wschurman, line 541:
```suggestion
        '--auto flag is deprecated. When --branch and --message are not provided, the flags default to the git branch name and commit message respectively.'
```

wschurman (approved):

kbrandwijk:
> Previously, when one of these wasn't provided, it would fail unless you added the --auto flag.

It wouldn't fail, it would just ask for the missing inputs. Does this make the interactive route that asks for these fields unreachable? If so, that should probably be cleaned up. Although I'm not sure this is a good DX at all.

If you decide to go through with this, _this is not a chore but a breaking change_, as a user who is normally used to get an interactive prompt suddenly gets a different behavior that they don't expect.

brentvatne:
that's a good point @kbrandwijk - what do you think @byCedric @EvanBacon?

byCedric:
> It wouldn't fail, it would just ask for the missing inputs.

Yes, I specifically meant the `--non-interactive` mode. There it would fail since it can't ask the user anything.

> Does this make the interactive route that asks for these fields unreachable? If so, that should probably be cleaned up.

Indeed, but falls out of scope for this PR. It's a huge task to refactor this command. [See both this comment](https://github.com/expo/eas-cli/pull/1500/files#diff-eabf745ea6a7d52d4573e94ea26a1634f849c8232205268bf97a5801ae32792bR552), and [this task](https://linear.app/expo/issue/ENG-5926/refactor-update-command).

> _this is not a chore but a breaking change_

It's a behavior change, but nothing breaks, and is backward compatible (does not error when using `--auto`). People can still create new updates, delete old updates, or even just (re)publish again with `--branch` or `--message`.

> a user who is normally used to getting an interactive prompt suddenly gets a different behavior that they don't expect.

> we probably want to change the way we log this to the user, maybe log upfront 'Using branch 'xyz' and update message '...'. Do you want to continue?'

The idea is that the defaults are set to "sensible defaults". Branches were introduced in EAS Update, named after git branches. This only enforces that as default.

Asking users up front to set the default isn't an improvement over this PR or the current state IMHO. We already ask if the branch or message isn't defined, and those questions give more control when users want to answer "no" to the proposed question.

PR #1507, opened 2022-11-07
[eas-cli] Only create a channel with branch name when branch is created

byCedric:
# Checklist

# Why

I noticed that whenever you have your own custom branch/channels mapping set up, it still created a new channel with that same branch name. Even when it's not necessary. This change was introduced in PR #1478, while trying to better surface important behavior within the command (instead of obfuscating it in side utility functions)

# How

Preferably, I would implement something like:

```
const existingBranch = BranchQuery.getBranchByName({ name });
if (!existingBranch) {
  // create channel
}
```

But we throw errors in that case, which makes implementing this a bit harder. Instead, I opted for a `isCreated` boolean in the `ensureBranchExists` utility class.

# Test Plan


dsokal (commented):
Nit: `branchIsCreated` -> `createdBranch`. Maybe this is me but `branchIsCreated` should be always true after calling `ensureBranchExistsAsync`.

byCedric:
> Nit: `branchIsCreated` -> `createdBranch`. Maybe this is me but `branchIsCreated` should be always true after calling `ensureBranchExistsAsync`.

Makes sense, done!

byCedric:
/changelog-entry bug-fix Only create a new channel when update branch is new

PR #1531, opened 2022-11-18
[ENG-6941][eas-json] add better error message for when invalid image is set in the `eas.json`

szdziedzic:
# Checklist

# Why

https://linear.app/expo/issue/ENG-6941/add-error-message-for-when-image-validation-fails-in-the-cli

# How

Add a custom error message handler when the `image` field is invalid.

Screenshot 2022-11-21 at 14 59 51

# Test Plan

Test locally.

### DIFF
--- a/packages/eas-json/src/accessor.ts
+++ b/packages/eas-json/src/accessor.ts
@@ -3,11 +3,32 @@ import assert from 'assert';
 import chalk from 'chalk';
 import fs from 'fs-extra';
 import * as fleece from 'golden-fleece';
+import { ValidationError } from 'joi';
 import path from 'path';
 
 import { InvalidEasJsonError, MissingEasJsonError } from './errors';
+import { link } from './log';
 import { EasJsonSchema } from './schema';
 import { EasJson } from './types';
+
+const customErrorMessageHandlers: ((err: ValidationError) => void)[] = [
+  // Ask user to upgrade eas-cli version or check the docs when image is invalid.
+  (err: ValidationError) => {
+    for (const detail of err.details) {
+      if (detail.path[detail.path.length - 1] === 'image') {
+        throw new InvalidEasJsonError(
+          chalk.red(
+            `Specified build image '${
+              detail?.context?.value
+            }' is not recognized. Please update your EAS CLI and see ${link(
+              'https://docs.expo.dev/build-reference/infrastructure/'
+            )} for supported build images.`
+          )
+        );
+      }
+    }
+  },
+];
 
 export class EasJsonAccessor {
   private easJsonPath: string;
@@ -39,6 +60,10 @@ export class EasJsonAccessor {
       noDefaults: true,
     });
     if (error) {
+      for (const handler of customErrorMessageHandlers) {
+        handler(error);
+      }
+
       const errorMessages = error.message.split('. ');
       throw new InvalidEasJsonError(
         `${chalk.bold('eas.json')} is not valid.\n- ${errorMessages.join('\n- ')}`
--- a/packages/eas-json/src/log.ts
+++ b/packages/eas-json/src/log.ts
@@ -1,0 +1,16 @@
+import chalk from 'chalk';
+import terminalLink from 'terminal-link';
+
+// link function from packages/eas-cli/src/log.ts
+export function link(
+  url: string,
+  { text = url, dim = true }: { text?: string; dim?: boolean } = {}
+): string {
+  let output: string;
+  if (terminalLink.isSupported) {
+    output = terminalLink(text, url);
+  } else {
+    output = `${text === url ? '' : text + ': '}${chalk.underline(url)}`;
+  }
+  return dim ? chalk.dim(output) : output;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 - Better wording in eas submit. ([#1527](https://github.com/expo/eas-cli/pull/1527) by [@dsokal](https://github.com/dsokal))
 - Upload project sources for EAS Build and archives for EAS Submit to GCS. ([#1524](https://github.com/expo/eas-cli/pull/1524) by [@wkozyra95](https://github.com/wkozyra95))
+- Add a better error message for when an invalid `image` is set in `eas-json`. ([#1531](https://github.com/expo/eas-cli/pull/1531) by [@szdziedzic](https://github.com/szdziedzic))
 
 ## [2.7.1](https://github.com/expo/eas-cli/releases/tag/v2.7.1) - 2022-11-16
 
--- a/packages/eas-json/package.json
+++ b/packages/eas-json/package.json
@@ -14,6 +14,7 @@
     "joi": "17.6.4",
     "log-symbols": "4.1.0",
     "semver": "7.3.8",
+    "terminal-link": "2.1.1",
     "tslib": "2.4.0"
   },
   "devDependencies": {


packages/eas-json/src/accessor.ts
```diff
@@ -39,6 +40,19 @@ export class EasJsonAccessor {
       noDefaults: true,
     });
     if (error) {
+      if (error.message.includes('.image"')) {
+        // eslint-disable-next-line no-console
+        console.log(
+          chalk.red(
+            `The image you specified in the ${chalk.bold(
+              'eas.json'
+            )} is invalid for this eas-cli version. Please check if your eas-cli version is up to date. Otherwise visit ${link(
```
dsokal, line 49:
```suggestion
            )} is invalid for this EAS CLI version. Please check if your EAS CLI version is up to date. Otherwise visit ${link(
```

packages/eas-json/src/accessor.ts
```diff
@@ -39,6 +40,19 @@ export class EasJsonAccessor {
       noDefaults: true,
     });
     if (error) {
+      if (error.message.includes('.image"')) {
```
dsokal, line 43: This looks really fragile. Can we detect it more reliably?

packages/eas-json/src/log.ts
```diff
@@ -0,0 +1,16 @@
+import chalk from 'chalk';
```
dsokal, line 1: In the past, we tried to avoid printing logs from eas-json. Move error detection to eas-cli.

packages/eas-json/src/accessor.ts
dsokal: Also, this code doesn't belong to eas-json. Move to eas-cli.

dsokal (changes requested):

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
```diff
@@ -0,0 +1,38 @@
...
+  (err: ValidationError) => {
+    for (const detail of err.details) {
+      if (detail.path[detail.path.length - 1] === 'image') {
+        Log.error(
+          `The image you specified in ${chalk.bold(
+            'eas.json'
+          )} is invalid for this EAS CLI version. Please check if your EAS CLI version is up to date. Othervise visit ${link(
+            'https://docs.expo.dev/build-reference/infrastructure/'
+          )} to learn more about supported images.`
+        );
```
szdziedzic, line 19: @brentvatne @kbrandwijk Is this message correct?

packages/eas-json/src/errors.ts
```diff
@@ -7,7 +8,11 @@ class NamedError extends Error {
   }
 }
 
-export class InvalidEasJsonError extends NamedError {}
+export class InvalidEasJsonError extends NamedError {
+  constructor(message?: string, public readonly joiError?: ValidationError) {
```
dsokal, line 12:
```suggestion
  constructor(message?: string, public readonly validationError?: ValidationError) {
```

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
kbrandwijk: 'invalid for this EAS CLI version' sounds a bit cryptic. 



'Specified build image 'xyz' is not recognized. Please update your EAS CLI and see '...' for supported build images'.



Maybe something like this, which I'm also not 100% happy with, so I'll think about it for a bit.

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -130,7 +130,7 @@ export async function runBuildAndSubmitAsync(
     projectDir,
     nonInteractive: flags.nonInteractive,
   });
-  const easJsonAccessor = new EasJsonAccessor(projectDir);
+  const easJsonAccessor = new EasJsonAccessorWithCustomErrorMessages(projectDir);
```
dsokal, line 133: IMO it's worthwhile figuring out how to enforce using `EasJsonAccessorWithCustomErrorMessages` in eas-cli. So we use it consistently across the codebase. I think it should be possible with some eslint rule.

dsokal (approved):

packages/eas-cli/src/build/runBuildAndSubmit.ts
szdziedzic: Sounds like a good idea, I will try to look at it

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
kbrandwijk: Also, can we swallow the server response, because the list of images returned is actually the outdated list from the current eas-cli version eas.json schema definition, which adds to the confusion.

packages/eas-cli/.eslintrc.js
```diff
@@ -32,6 +32,14 @@ module.exports = {
...
       },
     ],
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: "NewExpression[callee.name='EasJsonAccessor']",
+        message:
+          "Don't use EasJsonAccessor class, use EasJsonAccessorWithCustomErrors class instead!",
+      },
+    ],
```
szdziedzic, line 42: @dsokal What do you think about it?

szdziedzic: Screenshot 2022-11-18 at 16 27 17

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
szdziedzic: > Also, can we swallow the server response, because the list of images returned is actually the outdated list from the current eas-cli version eas.json schema definition, which adds to the confusion.

@kbrandwijk Sorry, I'm not sure if I understand you correctly - so you are suggesting to not validate the build image based on the offline schema specification attached to every CLI version, but instead, just show the server error we get once submitting the build with an incorrect image? Is this correct?

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
```diff
@@ -0,0 +1,38 @@
+import { EasJson, EasJsonAccessor } from '@expo/eas-json';
+import { InvalidEasJsonError } from '@expo/eas-json/build/errors';
+import chalk from 'chalk';
+import { ValidationError } from 'joi';
+
+import Log, { link } from '../../log';
+
+export const easJsonCustomErrorMessageHandlers: ((err: ValidationError) => void)[] = [
```
wkozyra95, line 8:
```suggestion
const easJsonCustomErrorMessageHandlers: ((err: ValidationError) => void)[] = [
```

packages/eas-cli/.eslintrc.js
```diff
@@ -32,6 +32,14 @@ module.exports = {
         ],
       },
     ],
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: "NewExpression[callee.name='EasJsonAccessor']",
+        message:
+          "Don't use EasJsonAccessor class, use EasJsonAccessorWithCustomErrors class instead!",
```
dsokal, line 40:
```suggestion
          "Use EasJsonAccessorWithCustomErrors instead",
```

packages/eas-cli/.eslintrc.js
dsokal: Looks good!

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
```diff
@@ -0,0 +1,38 @@
...
+          )} is invalid for this EAS CLI version. Please check if your EAS CLI version is up to date. Othervise visit ${link(
+            'https://docs.expo.dev/build-reference/infrastructure/'
+          )} to learn more about supported images.`
+        );
+      }
+    }
+  },
+];
+
+export class EasJsonAccessorWithCustomErrorMessages extends EasJsonAccessor {
```
wkozyra95, line 25: I'm not strongly against it, but if possible I would prefer to avoid it if possible. The main problem with your first version was logging in eas-json package. Alternatively, you could:
- Replace the log with an exception that has a more specific error message,
- If different logging/formatting is necessary and you need to use a logger, I would suggest using a more generic name. `WithCustomErrorMessages` is the correct name, but if it's only the accessor we use, that information is redundant.
- It's a bit more work, but ideally, to combine the benefits of both solutions, you could move validation to eas-json, throw an exception there, and override how errors are handled in oclif so it's formatted in a way logger would do.

@dsokal do you have an opinion on that?

wkozyra95 (approved):

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
wkozyra95: I think Kim wanted to only print custom errors without the original Joi error.

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
dsokal: No

szdziedzic: I think I will go with the first approach in this PR and create a separate task for completing the third approach which would be a good change overall.

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
kbrandwijk: Ah right, yes that is what I meant.

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
szdziedzic: https://linear.app/expo/issue/ENG-6945/improve-eas-cli-error-logging

packages/eas-cli/src/utils/easJson/EasJsonAccessorWithCustomErrors.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ValidationError } from 'joi';
+
+import { link } from '../../log';
+
+const easJsonCustomErrorMessageHandlers: ((err: ValidationError) => void)[] = [
+  // Ask user to upgrade eas-cli version or check the docs when image is invalid.
+  (err: ValidationError) => {
+    for (const detail of err.details) {
+      if (detail.path[detail.path.length - 1] === 'image') {
+        throw Error(
```
wkozyra95, line 13: When we discussed offling, we talked about moving this back to eas-json package

wkozyra95 (approved):

szdziedzic:
/changelog-entry chore Add a better error message for when an invalid `image` is set in `eas-json`

PR #1533, opened 2022-11-18
[eas-cli] Split republish command from update command

byCedric:
# Checklist

# Why

Part of ENG-5926

It also fixes or improves a couple of things:
 1. Do not create a new git hash for the republished update
 2. Reuse old code signing from the original update (instead of calculating again)
 3. Explicitly block republishing to a different branch, with helpful next-step commands

# How

Copied existing republish code from `eas update`, and moved it to a new command. This command has now basic tests to validate exit cases and a successful publish (with mocks).

As for fix #1 - We actually ask the VSC client for a new hash when republishing with `eas update`, which is incorrect. It should use the existing git reference from the original update.

As for improvement #2 - @wschurman, I believe this is the correct behavior (specifically for republishing).

As for fix #3 - When using environment variables to load the branch name within the app, this will be incorrect when reusing the exact same bundle to publish on another branch. Because of that, I added this block with commands to still achieve what the user might want to do.

# Test Plan

Instead of:

```bash
$ eas update --republish --branch <name> --group <id>
```

You can now use any of:

```bash
# Assumes the branch from existing group
$ eas update:republish --group <id>

# Asks users to pick an update from the branch
$ eas update:republish --branch <name>

# Both are also possible, but branch is validated to be the same (see #3)
$ eas update:republish --branch <name> --group <id>
```


packages/eas-cli/src/commands/update/__tests__/republish.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    expect(PublishMutation.setCodeSigningInfoAsync).toHaveBeenCalledWith(
+      expect.any(Object), // graphql client
+      'update-new',
+      expect.objectContaining(codeSigning)
+    );
+  });
+});
+
+/** Create a new in-memory project, based on src/commands/project/__tests__/init.test.ts */
+function mockTestProject({
```
wschurman, line 178: could be worth factoring this out in a follow-up if it's substantially similar to init.test.ts.

packages/eas-cli/src/commands/update/republish.ts
```diff
@@ -0,0 +1,373 @@
...
+          ', '
+        )}`
+      );
+    }
+
+    // This command only republishes a single update group
+    // The update group properties are the same for all updates
+    const groupId = updatesToPublish[0].groupId;
+    const runtimeVersion = updatesToPublish[0].runtimeVersion;
+    const branchName = flags.branchName ?? updatesToPublish[0].branchName;
```
wschurman, line 130:
```suggestion
    const { groupId, runtimeVersion, branchName } = updatesToPublish[0];
```

packages/eas-cli/src/commands/update/republish.ts
```diff
@@ -0,0 +1,373 @@
...
+
+    // Prevent users from republishing updates to a different branch.
+    // When using environment variables, the branch name is not updated, possibly causing unexpected sideeffects.
+    assertBranchNameIsEqualToExistingUpdateBranch(updatesToPublish, branchName, flags);
```
wschurman, line 134:
```suggestion
```

if we're just going to check that it's equal to the branch name, let's just make the `group` and `branch` flags mutually exclusive (using `exclusive` declaration). Then we can just use the branch name from `updatesToPublish[0].branchName;`.

packages/eas-cli/src/commands/update/republish.ts
```diff
@@ -0,0 +1,373 @@
...
+
+    // If codesigning was created for the original update, we need to add it to the republish
+    const codeSigningByPlatform = await PublishQuery.getCodeSigningInfoFromUpdateGroupAsync(
+      graphqlClient,
+      groupId
+    );
```
wschurman, line 140: instead of doing a separate query here, just query the code signing info in the original update group queries within `getOrAskUpdatesAsync`

packages/eas-cli/src/commands/update/republish.ts
```diff
@@ -0,0 +1,373 @@
...
+      Log.withTick(
+        `The republished update will be signed with the same codesigning as the original update.`
+      );
+    }
+
+    const updateMessage = await getOrAskUpdateMessageAsync(updatesToPublish, flags);
+    const { branchId } = await ensureBranchExistsAsync(graphqlClient, {
+      appId: projectId,
+      branchName,
+    });
```
wschurman, line 152:
```suggestion
```
can just query `Update.branchId` in `getOrAskUpdatesAsync`

wschurman (changes requested):
Looks good as far as the feature goes, just a few comments. Nice test!

> Drop the republish logic from eas update and refer to eas update:republish when using eas update --republish or eas update --group -> Will be a new PR stacked on this one.

yep, let's just make use of the flag throw and inform the user that the command has changed (breaking change).

byCedric:
> Looks good as far as the feature goes, just a few comments. Nice test!
> 
> > Drop the republish logic from eas update and refer to eas update:republish when using eas update --republish or eas update --group -> Will be a new PR stacked on this one.
> 
> yep, let's just make use of the flag throw and inform the user that the command has changed (breaking change).

Done in #1535 :)

packages/eas-cli/src/commands/update/__tests__/republish.test.ts
byCedric: Yeah, might be worth having this in a test utils section. That way, we don't need most boilerplate codes for each test.

packages/eas-cli/src/commands/update/republish.ts
byCedric: Yeah, this was related to https://github.com/expo/eas-cli/pull/1533#discussion_r1030463645, specifically:

> At first, I made it possible to use both `--group` and `--branch` to "promote" and the existing update group to a different branch. But after talking with @jonsamp, we think this is a foot gun. ...

packages/eas-cli/src/commands/update/republish.ts
byCedric: Done!

packages/eas-cli/src/commands/update/republish.ts
byCedric: Done!

packages/eas-cli/src/commands/update/republish.ts
byCedric: Done!

packages/eas-cli/src/commands/update/republish.ts
byCedric: Done!

packages/eas-cli/src/graphql/types/Update.ts
```diff
@@ -9,6 +9,7 @@ export const UpdateFragmentNode = gql`
     runtimeVersion
     platform
     manifestFragment
+    gitCommitHash
```
byCedric, line 12: This also fixes an issue where we "generate" a new git commit hash when republishing older updates.

byCedric:
/changelog-entry breaking-change Move update republish to separate command

byCedric:
/changelog-entry breaking-change Move update republish to separate command

wschurman (approved):

jonsamp (approved):
Nice test! I support this change. Thanks for separating this out into a new command. After this we will need to update a couple spots in the docs

PR #1570, opened 2022-12-07
[eas-cli] Adds channels to build profiles on eas update:configure

jonsamp:
# Checklist

- [x] I've added an entry to [CHANGELOG.md](https://github.com/expo/eas-cli/blob/main/CHANGELOG.md#main) if necessary. You can comment this pull request with `/changelog-entry [breaking-change|new-feature|bug-fix|chore] [message]` and CHANGELOG.md will be updated automatically.

# Why

To make EAS Update easier to set up, we are adding `channel` properties to build profiles when running `eas update:configure`.

The logic is:
- If eas.json is present:
  - if there is not pre-existing channel or release channel, set a channel named after the profile name
  - else do nothing (leave as is)
- else warn that the developer may need to run `eas build:configure`, then re-run `eas update:configure` to set everything up.

In addition, I updated the "Next steps" text, so there's actionable items after running `eas update:configure`.

Screenshot 2022-12-07 at 12 47 10 PM

# Test Plan

Run `easd update:configure` in a project. Check:
- that projects with channels in build profiles are not overwritten
- that projects with release channels in build profiles are not overwritten
- that projects get a channel when none is specified after running `eas update:configure`
- that if **eas.json** is not present, you see a readable error message

PR #1579, opened 2022-12-19
[eas-cli] Fix update --channel and --branch flag validation

byCedric:
# Checklist

# Why

A partner reported an issue when using the following command combination:

```bash
$ eas update --branch "..." --message "..." --non-interactive
> Error: --auto or both --channel or --branch and --message are required
> when updating in non-interactive mode
```

# How

It seems that PR #1570 introduced an incorrect validation combination, where it requires users to provide all flags (`--branch`, `--channel`, _and_ `--message`) when not using `--auto`.

This changes:
- The validation to allow combinations like:
  - `--branch` + `--message`
  - `--channel` + `--message`
  - `--branch` + `--channel` + `--message` (this is caught after flag validation as incorrect combination)
- Change the error message in attempt to improve communicating the allowed combinations
  > ```bash
  > Error: --branch and --message, or --channel and --message are required when updating in non-interactive mode unless --auto is specified
  > ```

# Test Plan

Added basic test for this case, including some `--republish` deprecation tests.

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -511,7 +511,7 @@ export default class UpdatePublish extends EasCommand {
     const { auto, branch: branchName, channel: channelName, message: updateMessage } = flags;
     if (nonInteractive && !auto && !(updateMessage && (branchName || channelName))) {
       Errors.error(
-      '--branch and --message, or --channel and --message are required when updating in non-interactive mode unless --auto is specified',
+        '--branch and --message, or --channel and --message are required when updating in non-interactive mode unless --auto is specified',
         { exit: 1 }
       );
     }
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -511,7 +511,7 @@ export default class UpdatePublish extends EasCommand {
     const { auto, branch: branchName, channel: channelName, message: updateMessage } = flags;
     if (nonInteractive && !auto && !(updateMessage && (branchName || channelName))) {
       Errors.error(
-      '--auto, --branch and --message, or --channel and --message are required when updating in non-interactive mode',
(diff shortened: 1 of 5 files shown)


byCedric:
/changelog-entry bug-fix Fix update `--channel` and `--branch` flag validation

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -509,9 +509,9 @@ export default class UpdatePublish extends EasCommand {
     const nonInteractive = flags['non-interactive'] ?? false;
 
     const { auto, branch: branchName, channel: channelName, message: updateMessage } = flags;
-    if (nonInteractive && !auto && !(branchName && channelName && updateMessage)) {
+    if (nonInteractive && !auto && !(updateMessage && (branchName || channelName))) {
       Errors.error(
-        '--auto or both --channel or --branch and --message are required when updating in non-interactive mode',
+        '--auto or --message and either --branch or --channel are required when updating in non-interactive mode',
```
wkozyra95, line 514: This is an improvement, but I think this error is still not clear. I sounds to me that (auto or message) and (channel  or branch) is required.

wkozyra95 (approved):

packages/eas-cli/src/commands/update/index.ts
byCedric: Open for other suggestions as well :)

wkozyra95:
```suggestion
        '--message and either --branch or --channel are required when updating in non-interactive mode unless --auto flag is specified ',
```
maybe sth like this?

byCedric: Let's combine them:

```suggestion
      '--branch and --message, or --channel and --message are required when updating in non-interactive mode unless --auto is specified',
```

PR #1588, opened 2022-12-24
[eas-cli] use appVersion instead of sdkVersion

jonsamp:
# Checklist

# Why

The `appVersion` runtime version policy is easier to understand and read than `sdkVersion`, and now that android correctly parses values like `1.10`, we can make this the default policy.

In addition, the `appVersion` policy will work better with upcoming EAS website dashboards.

# Test Plan

Run `eas update:configure` and make sure the runtime version policy is `appVersion`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - Add `--channel` option to `update:republish`. ([#1580](https://github.com/expo/eas-cli/pull/1580) by [@byCedric](https://github.com/byCedric))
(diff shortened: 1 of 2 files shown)


jonsamp:
/changelog-entry new-feature use `appVersion` as default runtime version policy when running `eas update:configure`

wschurman (approved):
Does the app version automatically upgrade when `expo upgrade` is run? Or, are we relying upon google and apple to tell the user that it needs to be incremented?

jonsamp:
@wschurman We are relying on apple/google/etc to tell people to upgrade. Adding a note to `expo upgrade` is a good idea. I [made a task](https://linear.app/expo/issue/ENG-7129/add-note-to-to-expo-upgrade-to-upgrade-version) + PR for that. I think a good experience there is to have a note saying "Update the `version` field in your app config, then make new builds" under a "Next steps" section.

PR #1598, opened 2022-12-29
[eas-cli] Makes eas.json configuration to only run on update:configure

jonsamp:
# Checklist

# Why

This PR is a bug fix. I landed a PR where when running `eas update:configure`, we set the `channel` property in build profiles. However, I placed that logic in a place that is also called when running `eas update ...`. (We want to set up expo-updates and configure app.json if needed when running `eas update`)

This is not the desired behavior for this configuration however. We _only_ want to set up channels in build profiles when a developer is running `eas update:configure`. By removing this from the `eas update` command, it allows developers to set a channel in their native code, then use EAS Build, and not have their native channel overwritten.

# How

- Takes the `ensureEASUpdateIsConfiguredInEasJsonAsync()` function out of the shared configuration and call it in the `update:configure` command.

# Test Plan

1. Create an app with `npc create-expo-app myApp`
2. Run `eas build:configure`
3. Run `eas update`. Expect that there will be no channels configured in eas.json, while app.json will be configured.
4. Run `eas update:configure`. Expect that eas.json is configured with channels.

### DIFF
--- a/packages/eas-cli/src/commands/update/configure.ts
+++ b/packages/eas-cli/src/commands/update/configure.ts
@@ -5,7 +5,10 @@ import EasCommand from '../../commandUtils/EasCommand';
 import { EASNonInteractiveFlag } from '../../commandUtils/flags';
 import Log from '../../log';
 import { RequestedPlatform } from '../../platform';
-import { ensureEASUpdateIsConfiguredAsync } from '../../update/configure';
+import {
+  ensureEASUpdateIsConfiguredAsync,
+  ensureEASUpdateIsConfiguredInEasJsonAsync,
+} from '../../update/configure';
 
 export default class UpdateConfigure extends EasCommand {
   static override description = 'configure the project to support EAS Update';
@@ -45,6 +48,8 @@ export default class UpdateConfigure extends EasCommand {
       projectDir,
       platform,
     });
+
+    await ensureEASUpdateIsConfiguredInEasJsonAsync(projectDir);
 
     Log.addNewLineIfNone();
     Log.log(`🎉 Your app is configured with EAS Update!`);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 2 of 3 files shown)


jonsamp:
/changelog-entry bug-fix Makes eas.json configuration to only run on `update:configure`

dsokal (approved):

packages/eas-cli/src/commands/update/configure.ts
```diff
@@ -46,6 +49,8 @@ export default class UpdateConfigure extends EasCommand {
       platform,
     });
 
+    await ensureEASUpdateIsConfiguredInEasJsonAsync(projectDir);
```
wkozyra95, line 52: `update:configure` is supposed to prepare the project for updates, but with this change, you are removing all config steps and only updating eas.json. e.g. You are not checking if expo-updates is installed.

Based on the PR description I would expect that changes in this PR will not affect `update:configure` method and the only difference for the update call would be skipping that eas.json update.

jonsamp: I think there may be a misunderstanding here.

This line (52) is in side the `runAsync()` function on the `update:configure` command. When running `update:configure`, we will run two key functions: 
1. `ensureEASUpdateIsConfiguredAsync()`: This installs expo-updates, configures app.json, etc
2. `ensureEASUpdateIsConfiguredInEasJsonAsync()`: This adds channel names to eas.json

Previously, that second function, `ensureEASUpdateIsConfiguredInEasJsonAsync()`, was inside of `ensureEASUpdateIsConfiguredAsync()`, which made it run whenever anyone ran `update`.

This PR removes the eas.json configuration part to only run on `update:configure`, and makes it so that `update` does not run the eas.json configuration part.

When running `update`, we will still install expo-updates + configure app.json etc. Here's where we call that in the `update` command: https://github.com/expo/eas-cli/blob/37bdbe8af17fc750570d6a1cfa54beef42c8c3a8/packages/eas-cli/src/commands/update/index.ts#L203

wschurman (approved):

jonsamp:
Merging because I got two checkmarks and I think I answered @wkozyra95's question adequately. If you still think that this is an issue, please let me know and we can discuss/revert

packages/eas-cli/src/commands/update/configure.ts
wkozyra95: yes, you are right, sorry. I didn't see that both functions were called in configure

PR #1639, opened 2023-01-19
[eas-cli] Print error message sent from server

wkozyra95:
# Checklist

# Why

Often when reacting to a new situation, we do not have a way to send good error messages to users without an EAS CLI update. This change is making all error messages defined server side.

# How

- Print error returned from www
- Print entire raw error in debug mode

# Test Plan

enable kill switch and try to run build on staging

20230119_12h19m11s_grim


szdziedzic (commented):
I think this approach looks good at the moment, although I'm wondering how will the docs URL be printed when forwarding kill switch errors from turtle v2 https://github.com/expo/turtle-v2/blob/main/src/services/turtle-api/src/api/builds.ts#L556-L559. Ideally, it would be nice to have it clickable (as it is now) 🤔

wkozyra95:
> I think this approach looks good at the moment, although I'm wondering how will the docs URL be printed when forwarding kill switch errors from turtle v2 https://github.com/expo/turtle-v2/blob/main/src/services/turtle-api/src/api/builds.ts#L556-L559. Ideally, it would be nice to have it clickable (as it is now) thinking

It should be clickable either way(at least it works in my terminal), that library we are using is only useful when we want to have sth clickable, but the text displayed is different than url. Otherwise, it does not do anything.

PR #1655, opened 2023-01-26
[ENG-7351][eas-cli] set `m1-medium` resource class for builds which don't specify resource class for SDK 48 and RN 0.71

szdziedzic:
# Checklist

# Why

https://linear.app/expo/issue/ENG-7351/switch-the-default-resourceclass-to-m1-for-sdk-48-projects

# How

If the resource class is unspecified, and the project is using SDK 48 / React Native 0.71, use the `m1-medium` resource class for the build.

# Test Plan

Test locally.

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -66,6 +66,7 @@ import { LocalBuildOptions } from './local';
 import { ensureExpoDevClientInstalledForDevClientBuildsAsync } from './utils/devClient';
 import { printBuildResults, printLogsUrls } from './utils/printBuildInfo';
 import { ensureRepoIsCleanAsync } from './utils/repository';
+import { resolveBuildResourceClassAsync } from './utils/resourceClass';
 
 let metroConfigValidated = false;
 let sdkVersionChecked = false;
@@ -82,81 +83,6 @@ export interface BuildFlags {
   localBuildOptions: LocalBuildOptions;
   resourceClass?: ResourceClass;
   message?: string;
-}
-
-const iosResourceClassToBuildResourceClassMapping: Record<ResourceClass, BuildResourceClass> = {
-  [ResourceClass.DEFAULT]: BuildResourceClass.IosDefault,
-  [ResourceClass.LARGE]: BuildResourceClass.IosLarge,
-  [ResourceClass.M1_EXPERIMENTAL]: BuildResourceClass.IosM1Large,
-  [ResourceClass.M1_MEDIUM]: BuildResourceClass.IosM1Medium,
-  [ResourceClass.M1_LARGE]: BuildResourceClass.IosM1Large,
-  [ResourceClass.INTEL_MEDIUM]: BuildResourceClass.IosIntelMedium,
-  [ResourceClass.MEDIUM]: BuildResourceClass.IosMedium,
-};
-
-const androidResourceClassToBuildResourceClassMapping: Record<
-  Exclude<
-    ResourceClass,
-    | ResourceClass.M1_EXPERIMENTAL
-    | ResourceClass.M1_MEDIUM
-    | ResourceClass.M1_LARGE
-    | ResourceClass.INTEL_MEDIUM
-  >,
-  BuildResourceClass
-> = {
-  [ResourceClass.DEFAULT]: BuildResourceClass.AndroidDefault,
-  [ResourceClass.LARGE]: BuildResourceClass.AndroidLarge,
-  [ResourceClass.MEDIUM]: BuildResourceClass.AndroidMedium,
-};
-
-function resolveBuildResourceClass(
-  profile: ProfileData<'build'>,
-  resourceClassFlag?: ResourceClass
-): BuildResourceClass {
-  if (
-    profile.platform !== Platform.IOS &&
-    resourceClassFlag &&
-    [
-      ResourceClass.M1_EXPERIMENTAL,
-      ResourceClass.M1_MEDIUM,
-      ResourceClass.M1_LARGE,
-      ResourceClass.INTEL_MEDIUM,
-    ].includes(resourceClassFlag)
-  ) {
-    throw new Error(`Resource class ${resourceClassFlag} is only available for iOS builds`);
-  }
-
-  const profileResourceClass = profile.profile.resourceClass;
-  if (profileResourceClass && resourceClassFlag && resourceClassFlag !== profileResourceClass) {
-    Log.warn(
-      `Build profile specifies the "${profileResourceClass}" resource class but you passed "${resourceClassFlag}" to --resource-class.\nUsing the  "${resourceClassFlag}" as the override.`
-    );
-  }
-  const resourceClass = resourceClassFlag ?? profileResourceClass ?? ResourceClass.DEFAULT;
-
-  if (profile.platform === Platform.IOS && resourceClass === ResourceClass.M1_EXPERIMENTAL) {
-    Log.warn(`Resource class ${chalk.bold('m1-experimental')} is deprecated.`);
-  }
-  if (
-    profile.platform === Platform.IOS &&
-    [ResourceClass.LARGE, ResourceClass.M1_LARGE].includes(resourceClass)
-  ) {
-    Log.warn(
-      `Large resource classes are not available for iOS builds yet. Your build will use the medium resource class.`
-    );
-  }
-
-  return profile.platform === Platform.ANDROID
-    ? androidResourceClassToBuildResourceClassMapping[
-        resourceClass as Exclude<
-          ResourceClass,
-          | ResourceClass.M1_EXPERIMENTAL
-          | ResourceClass.M1_MEDIUM
-          | ResourceClass.M1_LARGE
-          | ResourceClass.INTEL_MEDIUM
-        >
-      ]
-    : iosResourceClassToBuildResourceClassMapping[resourceClass];
 }
 
 export async function runBuildAndSubmitAsync(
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,124 @@
...
+    ? androidResourceClassToBuildResourceClassMapping[
+        resourceClass as Exclude<
+          ResourceClass,
+          | ResourceClass.M1_EXPERIMENTAL
+          | ResourceClass.M1_MEDIUM
+          | ResourceClass.M1_LARGE
+          | ResourceClass.INTEL_MEDIUM
+        >
+      ]
+    : iosResourceClassToBuildResourceClassMapping[resourceClass];
```
dsokal, line 123: Extract the logic to two platform-specific functions. It'll be easier to understand them.

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -208,7 +134,12 @@ export async function runBuildAndSubmitAsync(
       flags,
       moreBuilds: platforms.length > 1,
       buildProfile,
-      resourceClass: resolveBuildResourceClass(buildProfile, flags.resourceClass),
+      resourceClass: await resolveBuildResourceClassAsync(
+        buildProfile,
+        projectDir,
+        getDynamicProjectConfigAsync,
```
wkozyra95, line 140: do not call getDynamicProjectConfigAsync more than once if it's not necessary, just resolve resouce class latter when you have exp

packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,135 @@
...
+    ));
+
+  if (resourceClass === ResourceClass.M1_EXPERIMENTAL) {
+    Log.warn(`Resource class ${chalk.bold('m1-experimental')} is deprecated.`);
+  }
+
+  if ([ResourceClass.LARGE, ResourceClass.M1_LARGE].includes(resourceClass)) {
+    Log.warn(
+      `Large resource classes are not available for iOS builds yet. Your build will use the medium resource class.`
+    );
```
wkozyra95, line 107: When those values are specified we still send large to www and display large on website. So even now technically this is not true.
Also this message will also show up after we introduce large classes which will be even more confusing

packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,135 @@
...
+  ) {
+    return ResourceClass.M1_MEDIUM;
+  } else {
+    return ResourceClass.DEFAULT;
+  }
+}
+
+function resolveAndroidResourceClass(
+  profileResourceClass?: ResourceClass,
+  resourceClassFlag?: ResourceClass
```
wkozyra95, line 57: Why 2 options? You already handled a case earlier where both values are specified. You should pass here only one option.
Also didn't we decide to deprecate the flag?(I don't remember, just asking)

packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,135 @@
...
+  if ([ResourceClass.LARGE, ResourceClass.M1_LARGE].includes(resourceClass)) {
+    Log.warn(
+      `Large resource classes are not available for iOS builds yet. Your build will use the medium resource class.`
+    );
+  }
+
+  return iosResourceClassToBuildResourceClassMapping[resourceClass];
+}
+
+export async function resolveBuildResourceClassAsync(
```
dsokal, line 113: Nitpick: Place exported functions above non-exported. When you open a module, you're usually interested in the API, not implementation details.

packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,135 @@
...
+    | ResourceClass.INTEL_MEDIUM
+  >,
+  BuildResourceClass
+> = {
+  [ResourceClass.DEFAULT]: BuildResourceClass.AndroidDefault,
+  [ResourceClass.LARGE]: BuildResourceClass.AndroidLarge,
+  [ResourceClass.MEDIUM]: BuildResourceClass.AndroidMedium,
+};
+
+async function resolveIosDefaultRequestedResourceClassAsync(
```
dsokal, line 37: The same with this function. I'd place it below `resolveIosResourceClassAsync` as it's less generic.

dsokal (approved):

packages/eas-cli/src/build/utils/resourceClass.ts
szdziedzic: The flag is already deprecated, but we still need to handle it

packages/eas-cli/src/build/utils/resourceClass.ts
szdziedzic: >When those values are specified we still send large to www and display large on website. So even now technically this is not true.

This is not true, we don't display `large` on the website in such case, since we changed how the `resourceClass` resolver works https://github.com/expo/universe/blob/main/server/www/src/graphql/types/Build.ts#L317-L318
I also tested it, this is the build submitted with `large` resource class: https://expo.dev/accounts/szdziedzic/projects/testapp/builds/9560591e-b00a-4d92-9f43-ca77545288e5

>Also this message will also show up after we introduce large classes which will be even more confusing

That's right. Do you suggest dropping it? It might be confusing to users who use the deprecated flag with a `large` resource class. They will start the build with the `large` resource class and see `Intel Medium` on the website.

wkozyra95: > Do you suggest dropping it?

yes, there is no large class for iOS in documentation, so it's on users if it's not used correctly, I believe we should think first about the experience of users that use supported stuff and add warnings like that only it does not affect those users.

packages/eas-cli/src/build/utils/resourceClass.ts
wkozyra95: handle yes, but this is not a behavior I would expect from deprecated messages, especially with warnings displayed just before. I'll add a comment there.

Also, the first part of the comment still stands, there is no reason to pass 2 values.

packages/eas-cli/src/build/utils/resourceClass.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export async function resolveBuildResourceClassAsync(
+  profile: ProfileData<'build'>,
+  projectDir: string,
+  getDynamicProjectConfigAsync: DynamicConfigContextFn,
+  resourceClassFlag?: ResourceClass
+): Promise<BuildResourceClass> {
+  const profileResourceClass = profile.profile.resourceClass;
+  if (profileResourceClass && resourceClassFlag && resourceClassFlag !== profileResourceClass) {
+    Log.warn(
```
wkozyra95: In response to https://github.com/expo/eas-cli/pull/1655/files#r1087760626

This message suggest that flag is more important option than eas.json. I don't have an opinion on actual behavior and which field should have priority, but in other places where one config can override another, we are doing it silently.

If you want to keep it in add to this message that flag is deprecated.

packages/eas-cli/src/build/utils/resourceClass.ts
szdziedzic: Sounds reasonable, thanks for this discussion!

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -203,17 +130,27 @@ export async function runBuildAndSubmitAsync(
   const buildCtxByPlatform: { [p in AppPlatform]?: BuildContext<Platform> } = {};
 
   for (const buildProfile of buildProfiles) {
+    const { exp, projectId } = await getDynamicProjectConfigAsync({
+      env: buildProfile.profile.env,
+    });
```
wkozyra95, line 135: I don't have anything specific against, but is there a reason you moved it here?

packages/eas-cli/src/build/utils/resourceClass.ts
szdziedzic: We already print the deprecation warning as the first thing when the user calls `eas build` with the `--resource-class` flag. Do you think it's fine to print the same information twice? Just want to make sure that you are aware of the first warning.

Screenshot 2023-01-26 at 13 31 01

szdziedzic: The whole experience looks like this at the moment:

Screenshot 2023-01-26 at 13 32 59

wkozyra95: then I think this one is not necessary

packages/eas-cli/src/build/runBuildAndSubmit.ts
szdziedzic: https://github.com/expo/eas-cli/blob/b22a8abd96a40f5937f6a912f107dc1ad0bec3fb/packages/eas-cli/src/build/runBuildAndSubmit.ts#L276

The `resourceClass` was required to create build context as well as `getDynamicProjectConfigAsync` (`resolveResourceClass` also uses data from `getDynamicProjectConfigAsync`). We used `getDynamicProjectConfigAsync` inside of the `createBuildContextAsync` function, so another option was to move resolving `resourceClass` to the `prepareAndStartBuild` function https://github.com/expo/eas-cli/blob/b22a8abd96a40f5937f6a912f107dc1ad0bec3fb/packages/eas-cli/src/build/runBuildAndSubmit.ts#L251 or `createBuildContextAsync` function itself. I decided the best place resolve the resource class is the top-level `runBuildAndSubmit` function for better visibility and that's why I extracted this logic above.

Do you have a better place for it in mind?

szdziedzic: What do you think about it now?

wkozyra95 (approved):

szdziedzic:
/changelog-entry new-feature Set `m1-medium` resource class for SDK version `>=48` and RN version `>=0.71` builds with unspecified resource class

Issue #1685, opened 2023-02-09
Command line too long when doing EAS local build on Windows


trajano:
### Build/Submit details page URL

https://forums.expo.dev/t/command-line-too-long-when-doing-eas-local-build-on-windows/69457/1

### Summary

local build is being passed a JWT using the command line which breaks on [Windows if the number of characters exceed 8191-characters](https://stackoverflow.com/a/29359348/242042)

### Managed or bare?

managed

### Environment

> eas-cli/3.5.2 win32-x64 node-v16.18.1

  expo-env-info 1.0.5 environment info:
    System:
      OS: Windows 10 10.0.22621
    Binaries:
      Node: 16.18.1 - ~\.nvm\versions\node\v16.18.1\bin\node.EXE
      Yarn: 1.22.19 - ~\.nvm\versions\node\v16.18.1\bin\yarn.CMD
      npm: 8.19.2 - ~\.nvm\versions\node\v16.18.1\bin\npm.CMD
      Watchman: 20230108.221750.0 - C:\Users\trajano\scoop\shims\watchman.EXE
    IDEs:
      Android Studio: AI-213.7172.25.2113.9123335
    npmPackages:
      babel-preset-expo: ~9.2.1 => 9.2.2
      expo: ^47.0.12 => 47.0.13
      react: 18.1.0 => 18.1.0
      react-dom: 18.1.0 => 18.1.0
      react-native: 0.70.5 => 0.70.5
      react-native-web: ~0.18.7 => 0.18.10
    Expo Workflow: managed

### Error output

I tried to debug this using 
```
DEBUG=* eas build --platform android --profile preview  --local
```

The command that fails in the end is `The command line is too long.`

I hacked the code a bit to show the command it is sending.

```
{
  command: 'npx',
  args: [
    '-y',
    'eas-cli-local-build-plugin@0.0.128',
    'eyJq... this is a very long JWT that contains my keystore...'
  ]
}
```

### Reproducible demo or steps to reproduce from a blank project

This can be done using a private key that is 8192 byte length.

I wrote up the details on the forum as well
https://forums.expo.dev/t/command-line-too-long-when-doing-eas-local-build-on-windows/69457/1

My suggestion to fix this would be to change the local build plugin code so that it accepts the JWT from STDIN (avoid using a file because it contains private keys)

wkozyra95:
> eas-cli/3.5.2 win32-x64 node-v16.18.1

[...]

PR #1698, opened 2023-02-15
[eas-cli] Validate platform for local builds earlier

wkozyra95:
# Checklist

# Why

https://github.com/expo/eas-cli/issues/1685

Sometimes running local builds on windows fails before we even validate the platform, so to users it looks like a bug.
In the above case validation was never reached because spawning a process for local buld plugin failed.

# How

Validate platform in eas-cli before local plugin is executed.

# Test Plan

Not tested, copied the check from the local build plugin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


dsokal (approved):

wkozyra95:
/changelog-entry chore Validate the platform for local builds earlier.

szdziedzic (approved):

PR #1715, opened 2023-02-23
[eas-cli] Add support for asset excludes

wschurman:
# Checklist

# Why

This is a second try of https://github.com/expo/eas-cli/pull/1526 which "removes" assets from the exported bundle (almost as if we were to pass a assetExcludeGlobs to `expo export`).

# How

Exclude excluded assets from `dist` folder when building update.

# Test Plan

Note that this fails (client doesn't work):

1. create app with two images, one of which is included in the EAS asset exclude patterns
2. build app
3. update app to not use either image in App.tsx, publish update
4. reload app, see neither image is shown
5. update app to use both images, publish update, see one is excluded
6. reload app, see only the non-excluded image is shown (this is the issue, it should show both since the excluded image is in the bundle)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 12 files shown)


wschurman:
This doesn't work. Created PR for posterity.

Issue #1714, opened 2023-02-23
Authentication with Apple Developer Portal failed! [Error -36607]


MaxAst:
### Summary

When running `eas build -p ios --auto-submit` and choose `yes` when being asked to log in to my Apple account, I get this error:

```
✖ Logging in...
Authentication with Apple Developer Portal failed!
    Error: Apple Service Error -36607. Unable to sign you in to your Apple ID. Try again later. <a target="_blank" 
    href="https://support.apple.com/kb/HT212522 ">Learn more.<span class="sr-only"> Opens in a new window.</span></a>
```

after getting the error, I do get the Apple ID verification code prompt on my apple devices though, so it might be that the CLI is not able to handle cases where 2fa is required.

eas-cli/3.6.1 darwin-arm64 node-v18.12.1

### Managed or bare?

managed

### Environment

expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 13.2
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
      Yarn: 1.22.19 - ~/.yarn/bin/yarn
      npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
      Watchman: 2023.02.13.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    IDEs:
      Xcode: 14.2/14C18 - /usr/bin/xcodebuild
    npmGlobalPackages:
      eas-cli: 3.6.1
      expo-cli: 6.3.0
    Expo Workflow: managed

### Error output

Authentication with Apple Developer Portal failed!
    Error: Apple Service Error -36607. Unable to sign you in to your Apple ID. Try again later. <a target="_blank" 
    href="https://support.apple.com/kb/HT212522 ">Learn more.<span class="sr-only"> Opens in a new window.</span></a>

### Reproducible demo or steps to reproduce from a blank project

eas build -p ios --auto-submit

MaxAst:
note: I know that [this closed issue](https://github.com/expo/eas-cli/issues/1220) exists, but I think this is a different error

MaxAst:
fwiw, you can bypass this problem by 
[...]

PR #1719, opened 2023-02-24
[eas-cli] bump apple-utils to support hashcash for authentication

EvanBacon:
- fix https://github.com/expo/eas-cli/issues/1714 🤞 
- implements Apple's proprietary implementation of hashcash for signing authentication requests.
- Tested on all my accounts at a rate that would surely trigger rate limit issues otherwise. Didn't get a single failure.


EvanBacon:
/changelog-entry bug-fix Implement Apple's proprietary Hashcash algorithm for signing authentication requests.

PR #1738, opened 2023-03-09
[eas-cli] [ENG-6945] Unify error display method

radoslawkrzemien:
Overriden the `Command`'s `catch` method to catch all the errors thrown by the commands, log the error message there the same way as it was previously handled in `src/build/build.ts` and re-throw generic `Error` instance with predefined base error message. To avoid duplication, old `Log.error()` calls have been removed from `src/build/build.ts`. Also added tests to see if the `catch()` method works as intended.

# Checklist

# Why

There used to be a difference between how the expected/handled errors had been displayed by `eas-cli` when compared to these that were not expected. This change is meant to unify this behaviour
Expected:
Screenshot 2023-03-09 at 16 25 16
Unexpected:
Screenshot 2023-03-09 at 16 25 29

See: https://linear.app/expo/issue/ENG-6945/improve-eas-cli-error-logging

# How

Utilised the `catch()` method of `oclif`'s `Command` class, which is extended by custom base class `EasCommand`, which in turn is extended by all the commands in `eas-cli`. `EasCommand`'s `catch()` now logs the caught error the same way as it was previously done in case of expected errors, and then re-throws a generic error with pre-defined base error message. Code fragments which used to log the errors previously and re-throw generic error now just throw the error and let the `catch()` do the logging for them.
Expected:
Screenshot 2023-03-09 at 16 30 50
Unexpected:
Screenshot 2023-03-09 at 16 31 04

# Test Plan

Added automatic tests to check the `catch()` methods behaviour
Screenshot 2023-03-09 at 16 32 27
Plus tested manually by breaking the project to intentionally cause errors and check how they are displayed

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -173,26 +173,23 @@ function handleBuildRequestError(error: any, platform: Platform): never {
   Log.debug(JSON.stringify(error.graphQLErrors, null, 2));
 
   if (SERVER_SIDE_DEFINED_ERRORS.includes(error?.graphQLErrors?.[0]?.extensions?.errorCode)) {
-    Log.error(error?.graphQLErrors?.[0]?.message);
-    throw new Error('Build request failed.');
+    throw new Error(error?.graphQLErrors?.[0]?.message);
   } else if (
     error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE'
   ) {
-    Log.error(
+    throw new Error(
       `EAS Build is down for maintenance. Try again later. Check ${link(
         'https://status.expo.dev/'
       )} for updates.`
     );
-    throw new Error('Build request failed.');
   } else if (
     error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_TOO_MANY_PENDING_BUILDS'
   ) {
-    Log.error(
+    throw new Error(
       `You have already reached the maximum number of pending ${requestedPlatformDisplayNames[platform]} builds for your account. Try again later.`
     );
-    throw new Error('Build request failed.');
   } else if (error?.graphQLErrors) {
-    Log.error(
+    throw new Error(
       'Build request failed. Make sure you are using the latest eas-cli version. If the problem persists, report the issue.'
     );
   }
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -179,7 +179,7 @@ const SERVER_SIDE_DEFINED_ERRORS: Record<string, typeof EasCommandError> = {
   VALIDATION_ERROR: RequestValidationError,
 };
 
-function handleBuildRequestError(error: any, platform: Platform): never {
+export function handleBuildRequestError(error: any, platform: Platform): never {
   Log.debug(JSON.stringify(error.graphQLErrors, null, 2));
 
   const graphQLErrorCode: string = error?.graphQLErrors?.[0]?.extensions?.errorCode;
@@ -628,7 +628,3 @@ async function updateIosBuildProfilesToUseM1WorkersAsync(projectDir: string): Pr
   await easJsonAccessor.writeAsync();
   Log.withTick('Updated eas.json. Your next builds will run on M1 workers.');
 }
-
-export const testExports = {
-  handleBuildRequestError,
-};
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -36,6 +36,16 @@ import { createProgressTracker } from '../utils/progress';
 import { sleepAsync } from '../utils/promise';
 import { getVcsClient } from '../vcs';
 import { BuildContext } from './context';
+import {
+  EasBuildDownForMaintenanceError,
+  EasBuildFreeTierDisabledAndroidError,
+  EasBuildFreeTierDisabledError,
+  EasBuildFreeTierDisabledIOSError,
+  EasBuildTooManyPendingBuildsError,
+  GenericGraphQLError,
+  RequestValidationError,
+  TurtleDeprecatedJobFormatError,
+} from './errors';
 import { transformMetadata } from './graphql';
 import { LocalBuildMode, runLocalBuildAsync } from './local';
 import { collectMetadataAsync } from './metadata';
@@ -161,35 +171,33 @@ export async function prepareBuildRequestForPlatformAsync<
   };
 }
 
-const SERVER_SIDE_DEFINED_ERRORS = [
-  'TURTLE_DEPRECATED_JOB_FORMAT',
-  'EAS_BUILD_FREE_TIER_DISABLED',
-  'EAS_BUILD_FREE_TIER_DISABLED_IOS',
-  'EAS_BUILD_FREE_TIER_DISABLED_ANDROID',
-  'VALIDATION_ERROR',
-];
+const SERVER_SIDE_DEFINED_ERRORS: { [key: string]: any } = {
+  TURTLE_DEPRECATED_JOB_FORMAT: TurtleDeprecatedJobFormatError,
+  EAS_BUILD_FREE_TIER_DISABLED: EasBuildFreeTierDisabledError,
+  EAS_BUILD_FREE_TIER_DISABLED_IOS: EasBuildFreeTierDisabledIOSError,
+  EAS_BUILD_FREE_TIER_DISABLED_ANDROID: EasBuildFreeTierDisabledAndroidError,
+  VALIDATION_ERROR: RequestValidationError,
+};
 
 function handleBuildRequestError(error: any, platform: Platform): never {
   Log.debug(JSON.stringify(error.graphQLErrors, null, 2));
 
-  if (SERVER_SIDE_DEFINED_ERRORS.includes(error?.graphQLErrors?.[0]?.extensions?.errorCode)) {
-    throw new Error(error?.graphQLErrors?.[0]?.message);
-  } else if (
-    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE'
-  ) {
-    throw new Error(
+  const graphQLErrorCode: string = error?.graphQLErrors?.[0]?.extensions?.errorCode;
+  if (Object.keys(SERVER_SIDE_DEFINED_ERRORS).includes(graphQLErrorCode)) {
+    const ErrorClass: typeof Error = SERVER_SIDE_DEFINED_ERRORS[graphQLErrorCode];
+    throw new ErrorClass(error?.graphQLErrors?.[0]?.message);
+  } else if (graphQLErrorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE') {
+    throw new EasBuildDownForMaintenanceError(
       `EAS Build is down for maintenance. Try again later. Check ${link(
         'https://status.expo.dev/'
       )} for updates.`
     );
-  } else if (
-    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_TOO_MANY_PENDING_BUILDS'
-  ) {
-    throw new Error(
+  } else if (graphQLErrorCode === 'EAS_BUILD_TOO_MANY_PENDING_BUILDS') {
+    throw new EasBuildTooManyPendingBuildsError(
       `You have already reached the maximum number of pending ${requestedPlatformDisplayNames[platform]} builds for your account. Try again later.`
     );
   } else if (error?.graphQLErrors) {
-    throw new Error(
+    throw new GenericGraphQLError(
       'Build request failed. Make sure you are using the latest eas-cli version. If the problem persists, report the issue.'
     );
   }
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -42,6 +42,7 @@ import {
   EasBuildFreeTierDisabledError,
   EasBuildFreeTierDisabledIOSError,
   EasBuildTooManyPendingBuildsError,
+  EasCommandError,
   GenericGraphQLError,
   RequestValidationError,
   TurtleDeprecatedJobFormatError,
@@ -171,7 +172,7 @@ export async function prepareBuildRequestForPlatformAsync<
   };
 }
 
-const SERVER_SIDE_DEFINED_ERRORS: { [key: string]: any } = {
+const SERVER_SIDE_DEFINED_ERRORS: Record<string, typeof EasCommandError> = {
   TURTLE_DEPRECATED_JOB_FORMAT: TurtleDeprecatedJobFormatError,
   EAS_BUILD_FREE_TIER_DISABLED: EasBuildFreeTierDisabledError,
   EAS_BUILD_FREE_TIER_DISABLED_IOS: EasBuildFreeTierDisabledIOSError,
@@ -184,7 +185,7 @@ function handleBuildRequestError(error: any, platform: Platform): never {
 
   const graphQLErrorCode: string = error?.graphQLErrors?.[0]?.extensions?.errorCode;
   if (graphQLErrorCode in SERVER_SIDE_DEFINED_ERRORS) {
-    const ErrorClass: typeof Error = SERVER_SIDE_DEFINED_ERRORS[graphQLErrorCode];
+    const ErrorClass: typeof EasCommandError = SERVER_SIDE_DEFINED_ERRORS[graphQLErrorCode];
     throw new ErrorClass(error?.graphQLErrors?.[0]?.message);
   } else if (graphQLErrorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE') {
     throw new EasBuildDownForMaintenanceError(
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -10,6 +10,7 @@ import { BuildEvent } from '../analytics/AnalyticsManager';
 import { withAnalyticsAsync } from '../analytics/common';
 import { getExpoWebsiteBaseUrl } from '../api';
 import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
+import { EasCommandError } from '../commandUtils/errors';
 import {
   AppPlatform,
   BuildFragment,
@@ -42,7 +43,6 @@ import {
   EasBuildFreeTierDisabledError,
   EasBuildFreeTierDisabledIOSError,
   EasBuildTooManyPendingBuildsError,
-  EasCommandError,
   RequestValidationError,
   TurtleDeprecatedJobFormatError,
 } from './errors';
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -628,3 +628,7 @@ async function updateIosBuildProfilesToUseM1WorkersAsync(projectDir: string): Pr
   await easJsonAccessor.writeAsync();
   Log.withTick('Updated eas.json. Your next builds will run on M1 workers.');
 }
+
+export const testExports = {
+  handleBuildRequestError,
+};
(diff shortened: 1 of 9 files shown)


radoslawkrzemien:
/changelog-entry chore Unify how the command errors are displayed

packages/eas-cli/src/build/build.ts
```diff
@@ -173,26 +173,23 @@ function handleBuildRequestError(error: any, platform: Platform): never {
...
 
   if (SERVER_SIDE_DEFINED_ERRORS.includes(error?.graphQLErrors?.[0]?.extensions?.errorCode)) {
-    Log.error(error?.graphQLErrors?.[0]?.message);
-    throw new Error('Build request failed.');
+    throw new Error(error?.graphQLErrors?.[0]?.message);
   } else if (
     error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE'
   ) {
-    Log.error(
+    throw new Error(
```
wkozyra95, line 180: Consider creating custom error class, so you can latter distinguish error with proper use facing error from all the others

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +177,13 @@ export default abstract class EasCommand extends Command {
     await this.analytics.flushAsync();
     return super.finally(err);
   }
+
+  protected baseErrorMessage = 'Command failed';
+  protected override catch(err: CommandError): Promise<any> {
+    const cleanMessage = err.message.startsWith('[GraphQL] ')
+      ? err.message.replace('[GraphQL] ', '')
+      : err.message;
+    Log.error(cleanMessage);
```
wkozyra95, line 186:
```suggestion
    Log.error(cleanMessage);
    Log.debug(err);
```
To preserve the original stack trace when someone runs with `EXPO_DEBUG=1`

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +177,13 @@ export default abstract class EasCommand extends Command {
     await this.analytics.flushAsync();
     return super.finally(err);
   }
+
+  protected baseErrorMessage = 'Command failed';
+  protected override catch(err: CommandError): Promise<any> {
+    const cleanMessage = err.message.startsWith('[GraphQL] ')
```
wkozyra95, line 183: No strong opinion how to handle it, but maybe id you know that error is from grphql we should somehow mention that API request failed.
If you just want to strip leading graphql text, then this is fine, but if you decide to customize it more when handling graphql error, then I would suggest checking if `err.graphQLErrors` exists instead, instead of checking string prefix.

wkozyra95 (commented):
LGTM

Not necessarily part of this PR, but it would also be good to better handle errors that are a result of the failed spawned process. Currently, the error is just "process exited with exit code x", which is a bit useless, you can check if an error has `stdout` and `stderr` fields and print that instead. It would also be great to print commands that failed, but that might be harder.

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: @wkozyra95 Updated the check for graphQL error, and changed the base error message printed in that case

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: @wkozyra95 Added the log to debug

wkozyra95: `Log.debug(err.message)` will not display stack trace (it would just print message again which does not help here). You need to print actual error object with `Log.debug(err)`

wkozyra95: or alternatively instead of creating new error bellow just do 
```
err.message = isGraphQL ? this.baseGraphQLErrorMessage : this.baseErrorMessage
throw err
``` 
This aproach wouldn't work when you run `EXPO_DEBUG=1`, but it would when you run `DEBUG=*`(oclif is respecting this env)

packages/eas-cli/src/build/build.ts
radoslawkrzemien: @wkozyra95 Created some custom error classes for previously defined errors in case a custom behaviour would be needed in the future

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: Ok, gotcha, fill fix soon

dsokal:
> Not necessarily part of this PR, but it would also be good to better handle errors that are a result of the failed spawned process. Currently, the error is just "process exited with exit code x", which is a bit useless, you can check if an error has stdout and stderr fields and print that instead. It would also be great to print commands that failed, but that might be harder.

FWIW if we use `@expo/spawn-async`, the error messages should include the program name. See my PR  -https://github.com/expo/spawn-async/pull/15

packages/eas-cli/src/build/build.ts
```diff
@@ -161,38 +171,33 @@ export async function prepareBuildRequestForPlatformAsync<
...
 }
 
-const SERVER_SIDE_DEFINED_ERRORS = [
-  'TURTLE_DEPRECATED_JOB_FORMAT',
-  'EAS_BUILD_FREE_TIER_DISABLED',
-  'EAS_BUILD_FREE_TIER_DISABLED_IOS',
-  'EAS_BUILD_FREE_TIER_DISABLED_ANDROID',
-  'VALIDATION_ERROR',
-];
+const SERVER_SIDE_DEFINED_ERRORS: { [key: string]: any } = {
```
dsokal, line 174:
```suggestion
const SERVER_SIDE_DEFINED_ERRORS: { [key: string]: Error } = {
```

packages/eas-cli/src/build/build.ts
```diff
@@ -161,38 +171,33 @@ export async function prepareBuildRequestForPlatformAsync<
...
 
-  if (SERVER_SIDE_DEFINED_ERRORS.includes(error?.graphQLErrors?.[0]?.extensions?.errorCode)) {
-    Log.error(error?.graphQLErrors?.[0]?.message);
-    throw new Error('Build request failed.');
-  } else if (
-    error?.graphQLErrors?.[0]?.extensions?.errorCode === 'EAS_BUILD_DOWN_FOR_MAINTENANCE'
-  ) {
-    Log.error(
+  const graphQLErrorCode: string = error?.graphQLErrors?.[0]?.extensions?.errorCode;
+  if (Object.keys(SERVER_SIDE_DEFINED_ERRORS).includes(graphQLErrorCode)) {
```
dsokal, line 186: If TS won't complain about it then this is better:
```suggestion
  if (graphQLErrorCode in SERVER_SIDE_DEFINED_ERRORS) {
```

packages/eas-cli/src/build/errors.ts
```diff
@@ -0,0 +1,47 @@
+export class TurtleDeprecatedJobFormatError extends Error {
```
dsokal, line 1: General comment for this file: 
1. `Eas-build` -> `EAS Build`
2. Don't we have those error messages already defined somewhere on the server side? I think it's better to copy them over instead of rephrasing.

packages/eas-cli/src/build/queries.ts
```diff
@@ -43,7 +43,7 @@ export async function listAndRenderBuildsOnAppAsync(
       offset: paginatedQueryOptions.offset,
       queryToPerform: (limit, offset) =>
         BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
-          appId: projectId,
+          appId: 111 as unknown as string,
```
dsokal, line 46: Leftover from debugging?

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +177,18 @@ export default abstract class EasCommand extends Command {
     await this.analytics.flushAsync();
     return super.finally(err);
   }
+
+  protected baseErrorMessage = 'Command failed.';
+  protected baseGraphQLErrorMessage = 'GraphQL request failed.';
```
dsokal, line 182: This should be handled/caught by ESLint, but define class properties before methods.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +177,18 @@ export default abstract class EasCommand extends Command {
...
+  protected override catch(err: CommandError): Promise<any> {
+    const isGraphQLError = (err: any): boolean => {
+      return err?.graphQLErrors;
+    };
```
dsokal, line 186: It's usually better to place helpers like this outside methods. Also, I'm not sure if we need this particular helper. It's used only once and it's really simple.

packages/eas-cli/src/build/build.ts
wkozyra95: Sorry should have been more clear. I meant to suggest creating sth like `EasCommandError` for errors that we now are already properly formated and ready to print as it is. So in the code that it prints you could do

```
if (err instanceof EasCommandError) {
  //print
} else {
  // rest of the logic for printing unknown erorrs
}
```

I don't have anything against making it more granular, so I'm fine with that change, but with this implementation there still no way to detect if error needs to be processed or not.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +177,18 @@ export default abstract class EasCommand extends Command {
     await this.analytics.flushAsync();
     return super.finally(err);
   }
+
+  protected baseErrorMessage = 'Command failed.';
```
dsokal, line 181: Maybe there's a way to access the actual command name from some oclif internals?

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: Well, it's not the case with other properties, like on lines 150 or 141, etc.

dsokal: Can you please fix those as well?

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: It is simple, but maybe the logic will change, or we'll need to determine the error type etc. I can remove it too, the same for me

dsokal: I think this is a bit premature for now.

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: Haven't seen that so far, but I first heard about `oclif` yesterday 🤷  so maybe there is

packages/eas-cli/src/build/build.ts
radoslawkrzemien: TS does complain about that - it expects Error instance, not Error type

dsokal: `:typeof Error` then?

packages/eas-cli/src/commandUtils/EasCommand.ts
dsokal: I think this should be `this.id`.
Screenshot 2023-03-10 at 14 16 02

packages/eas-cli/src/build/build.ts
radoslawkrzemien: `TS2322: Type 'typeof TurtleDeprecatedJobFormatError' is not assignable to type 'ErrorConstructor'.   Type 'typeof TurtleDeprecatedJobFormatError' provides no match for the signature '(message?: string | undefined): Error'. `

szdziedzic: I think `Record<string, typeof Error>` should work fine

packages/eas-cli/src/commandUtils/__tests__/EasCommand-test.ts
```diff
@@ -96,5 +101,85 @@ describe(EasCommand.name, () => {
...
+        runAsyncMock.mockImplementation(() => {
+          throw error;
+        });
+        try {
+          await TestEasCommand.run();
+        } catch {}
+
+        expect(logErrorSpy).toBeCalledWith('Unexpected, internal error message');
+        expect(logDebugSpy).toBeCalledWith(error);
+      });
```
szdziedzic, line 121: Thanks for adding tests 👍

packages/eas-cli/src/build/build.ts
dsokal: I checked out your branch locally and `new () => Error` is what should work here. You'll have to change the type for `ErrorClass` a few line below.

szdziedzic: @dsokal I've seen in our codebase that we usually prefer to type such objects as records over explicit declarations. Is this some sort of rule we want to stick to or is it just coincidence and we are fine with both ways of typing such a thing?

dsokal: You're right, `Record<...>` is always better.

packages/eas-cli/src/commandUtils/__tests__/EasCommand-test.ts
radoslawkrzemien: `Record<string, typeof Error>` gives the same error as `{ [key: string]: any }` sadly

dsokal: I think you commented under wrong line but try what I suggested there `Record<string, new () => Error>`.

packages/eas-cli/src/build/build.ts
radoslawkrzemien: `Record<string, typeof Error>` didn't work, but `Record<string, typeof EasCommandError>` did 🤷 
All these error types are sub-types of `EasCommandError` so it's even better/more specific

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: Looks like you're right, updating

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -103,6 +106,21 @@ export default abstract class EasCommand extends Command {
...
+   */
+  private sessionManagerInternal?: SessionManager;
+
+  /**
+   * The analytics manager. Used for logging analytics.
+   * It is set up here to ensure a consistent setup.
+   */
+  private analyticsInternal?: AnalyticsWithOrchestration;
+
+  protected baseErrorMessage: string = '';
```
dsokal, line 121: Why is it an empty string? Maybe make it optional if it makes sense it's optional:
```suggestion
  protected baseErrorMessage?: string;
```

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +183,24 @@ export default abstract class EasCommand extends Command {
     await this.analytics.flushAsync();
     return super.finally(err);
   }
+
+  protected override catch(err: Error): Promise<any> {
+    let baseMessage: string = '';
```
dsokal, line 188: This works too:
```suggestion
    let baseMessage: string;
```

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +183,24 @@ export default abstract class EasCommand extends Command {
...
+    if (err instanceof EasCommandError) {
+      Log.error(err.message);
+      baseMessage = this.baseErrorMessage || `${this.id} command failed.`;
```
dsokal, line 191: Prefer nullish coalescing (`??`) over `||`.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +183,24 @@ export default abstract class EasCommand extends Command {
...
+    } else {
+      const cleanMessage =
+        err instanceof CombinedError && err?.graphQLErrors
+          ? err.message.replace('[GraphQL] ', '')
+          : err.message;
+      Log.error(cleanMessage);
+      baseMessage =
+        err instanceof CombinedError && err?.graphQLErrors
+          ? this.baseGraphQLErrorMessage
+          : this.baseErrorMessage || `${this.id} command failed.`;
```
dsokal, line 201:
```suggestion
          : this.baseErrorMessage ?? `${this.id} command failed.`;
```
but you have to accept my other suggestion first.

szdziedzic (approved):
LGTM! 🚀  Nice 🔥

packages/eas-cli/src/build/errors.ts
```diff
@@ -0,0 +1,53 @@
+export class EasCommandError extends Error {
```
wkozyra95, line 1: This error is not build-specific, move it somewhere close to code that is handling the error, probably in commandUtils directory

packages/eas-cli/src/build/errors.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+export class EasBuildTooManyPendingBuildsError extends EasCommandError {
+  constructor(message?: string) {
+    super(message ?? 'EAS Build has too many pending builds at the moment.');
+  }
+}
+
+export class GenericGraphQLError extends EasCommandError {
```
wkozyra95, line 49: remove(context in the earlier comment)

packages/eas-cli/src/build/errors.ts
```diff
@@ -0,0 +1,53 @@
+export class EasCommandError extends Error {
+  constructor(message?: string) {
+    super(message ?? 'Unknown EAS error occurred.');
+  }
+}
+
+export class TurtleDeprecatedJobFormatError extends EasCommandError {
+  constructor(message?: string) {
```
wkozyra95, line 8: Make argument nonoptional in all those errors, it will avoid the need to provide a fallback string, plus it will be apparent that the server (or calling code) is a source of truth for error messages.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -175,4 +183,24 @@ export default abstract class EasCommand extends Command {
...
+          : err.message;
+      Log.error(cleanMessage);
+      baseMessage =
+        err instanceof CombinedError && err?.graphQLErrors
+          ? this.baseGraphQLErrorMessage
+          : this.baseErrorMessage ?? `${this.id} command failed.`;
+    }
+    Log.debug(err);
+    throw new Error(baseMessage);
+  }
```
wkozyra95, line 205:
```suggestion
    let baseMessage = this.baseErrorMessage ?? `${this.id} command failed.`;
    if (err instanceof EasCommandError) {
      Log.error(err.message);
    } else if (err instanceof CombinedError && err?.graphQLErrors) {
      const cleanMessage = err.message.replace('[GraphQL] ', '')
      Log.error(cleanMessage);
      baseMessage = this.baseGraphQLErrorMessage
    } else {
      Log.error(err.message);
    }
    Log.debug(err);
    throw new Error(baseMessage);
  }
```

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -103,6 +106,21 @@ export default abstract class EasCommand extends Command {
...
+  private sessionManagerInternal?: SessionManager;
+
+  /**
+   * The analytics manager. Used for logging analytics.
+   * It is set up here to ensure a consistent setup.
+   */
+  private analyticsInternal?: AnalyticsWithOrchestration;
+
+  protected baseErrorMessage?: string;
+  protected baseGraphQLErrorMessage: string = 'GraphQL request failed.';
```
wkozyra95, line 122: If you do not need to customize it per command, it should be const in a global scope.

packages/eas-cli/src/build/build.ts
```diff
@@ -623,3 +628,7 @@ async function updateIosBuildProfilesToUseM1WorkersAsync(projectDir: string): Pr
   await easJsonAccessor.writeAsync();
   Log.withTick('Updated eas.json. Your next builds will run on M1 workers.');
 }
+
+export const testExports = {
```
wkozyra95, line 632: revert

wkozyra95 (approved):

szdziedzic (commented):
One tip: we usually don't create changelog entries until the last moment before merging the PR, so we don't have to resolve conflicts manually (in case the new version is released in meantime)

PR #1745, opened 2023-03-15
[eas-cli] Add test case to ensure public config is used as update

byCedric:
# Checklist

# Why

Make sure we are only using the public config when creating updates.

# How

Added test case.

# Test Plan

- See tests
- `$ yarn create expo-app ./test-update && cd ./test-update`
- `$ eas update --auto`
- _Make sure this works as expected_
- Modify the `app.json` to create code-signed updates
- `$ eas update --auto`
- _Make sure code signing works as expected_

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -211,7 +211,7 @@ export default class UpdatePublish extends EasCommand {
       projectId,
     });
 
-    const { exp } = await getDynamicProjectConfigAsync({});
+    const { exp } = await getDynamicProjectConfigAsync({ isPublicConfig: true });
     const codeSigningInfo = await getCodeSigningInfoAsync(expPrivate, privateKeyPath);
 
     let realizedPlatforms: PublishPlatform[] = [];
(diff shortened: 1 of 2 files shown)


douglowder (approved):
Looks good, can be merged if @wschurman is ok with it

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -211,7 +211,7 @@ export default class UpdatePublish extends EasCommand {
       projectId,
     });
 
-    const { exp } = await getDynamicProjectConfigAsync({});
+    const { exp } = await getDynamicProjectConfigAsync({ isPublicConfig: true });
```
wschurman, line 214: I was curious about the blame, looks like the blame rev was 8d0d34c497459c9bd36cb3ea89cb433d5e5dd52b. Looking at the PR, this would have been super hard to catch in code review (neither author nor reviewer error). I'd argue that the default for this is wrong. It should default closed, require override to open; i.e. default to public.

My guess is that there are probably a lot of other places we've made this error due to this API.

wschurman (approved):
In a follow-up, we may want to audit callsites to these getConfig methods and their derivatives to ensure this is correct elsewhere. Furthermore, I recommend we either switch defaults (default to public) or remove the argument entirely from the method and just have two separate methods (getPublicConfig, getPrivateConfig). It would be even more helpful if they were distinct typescript types so that APIs that needed private config would have typescript errors if a public config is passed in and vice-versa.

I don't really have the full context about what a private vs public config is. Would be helpful to describe in the PR summary.

PR #1760, opened 2023-03-29
[eas-cli] Do not require commits in eas build:internal

wkozyra95:
# Checklist

# Why

When building from git with the option "requireCommit" set to true, if pre-install hooks or install itself are leaving the local repo in a dirty state, the `eas build:internal` fails while attempting to commit the changes.

# How

- applyCliConfigAsync is changing the global vcs client, so it shouldn't be run as a part of a DynamicProjectConfig, because it would be triggered at different parts of the build process. All calls to that function should finish before `getContextAsync` returns.
  - I considered 2 alternative approaches: 
     - wrapping call to applyCliConfigAsync in `once`
     - moving call to the findProjectDirAndVerifyProjectSetupAsync in a [DynamicProjectConfig](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commandUtils/context/DynamicProjectConfigContextField.ts#L20) otside of the arrow function

- update vcs client to git no commit client after context was resolved in `build:internal` command.

# Test Plan

### DIFF
--- a/packages/eas-cli/src/commandUtils/context/OptionalProjectConfigContextField.ts
+++ b/packages/eas-cli/src/commandUtils/context/OptionalProjectConfigContextField.ts
@@ -3,7 +3,7 @@ import { InvalidEasJsonError } from '@expo/eas-json/build/errors';
 
 import { getExpoConfig } from '../../project/expoConfig';
 import ContextField, { ContextOptions } from './ContextField';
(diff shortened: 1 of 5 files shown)


szdziedzic (approved):
Fix https://github.com/expo/eas-cli/actions/runs/4553250963/jobs/8029588104?pr=1760

dsokal (approved):

wkozyra95:
/changelog-entry chore Ignore `requireCommit` option in `eas build:internal`.

PR #1782, opened 2023-04-11
[eas-cli] always show logs from expo install

wkozyra95:
# Checklist

# Why

When reproducing unrelated issues I noticed that if global expo-cli is not installed then running `eas update` will trigger installing `expo-update` package but it will hang on a prompt that is never displayed because the output is not printed to a terminal.

The result of the above is that command hangs indefinitely unless users hits enter(but no prompt is displayed). It happens for SDK 45 and lower)

# How

Always show output install command

# Test Plan

before
20230411_11h26m54s_grim

after
20230411_11h25m16s_grim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


dsokal (approved):

wschurman (approved):

wschurman:
Doesn't this use the local CLI?

wkozyra95:
> Doesn't this use the local CLI?

It was for older SDK (45). Even if it's deprecated in Expo Go I think it's to early to not support it in EAS.

PR #1794, opened 2023-04-18
[eas-cli] [eas-json] [ENG-8120] Use `cache.paths` and deprecate `cache.customPaths`

radoslawkrzemien:
After disabling the default caching it was decided to rename `cache.customPaths` field to `cache.paths` since that name makes more sense. This PR adds the new field and deprecates the old one

# Checklist

# Why

After disabling the default caching it was decided to rename `cache.customPaths` field to `cache.paths` since that name makes more sense

See: https://linear.app/expo/issue/ENG-8120/rename-cachecustompaths-to-cachepaths-in-easjson-and-job-schema

# How

Added new field `cache.paths` to eas-json and deprecated the old `cache.customPaths` but still kept it operational for people having it set in their json files

# Test Plan

Added automated tests for handling both fields, checking for the deprecated field and for printing deprecation message

Manual testing:
- Build without `cache` settings
Screenshot 2023-04-21 at 14 08 03
- Build with `cache` settings, without `paths` or `customPaths`
Screenshot 2023-04-21 at 14 08 51
- Build with `cache` settings, with `paths`, without `customPaths`
Screenshot 2023-04-21 at 14 09 35
- Build with `cache` settings, with deprecated `customPaths`, without `paths`
Screenshot 2023-04-21 at 14 10 20
- Build with `cache` settings, with both `paths` and `customPaths`
Screenshot 2023-04-21 at 14 11 32


wkozyra95 (commented):
If I'm reading this correctly, this will send both cachePaths and paths to the server, but we only want to send the new value. deprecated filed should be respected, but sent under `path` field.

Why this is important? So, at some point we will need to drop support for customPath on an API level, but if cli is still sending that value (even if it's an empty list) would still fail even if they are aware of a deprecation.

 When removing customPaths make sure you check the actual final response, take a look at prepareJob.ts where we call sanitizeJob, which might apply defaults to the job object.

radoslawkrzemien:
Currently `sanitizeJob` applies defaults to both `paths` and `customPaths` so another update to `eas-build` is needed, if my thinking is correct

szdziedzic (commented):
>Currently sanitizeJob applies defaults to both paths and customPaths so another update to eas-build is needed, if my thinking is correct

Yeah, I think I agree with you 👍 It would be good to drop setting the default for `customPaths` there

radoslawkrzemien:
PR: https://github.com/expo/eas-build/pull/217

packages/eas-json/src/utils.ts
```diff
@@ -45,6 +45,17 @@ export class EasJsonUtils {
       });
     }
 
+    if (buildProfile.cache?.customPaths !== undefined) {
+      warnings.push({
+        message: [
+          `The "build.${
+            profileName ?? 'production'
+          }.cache.customPaths" field in eas.json is deprecated and will be removed in the future. Please use "build.production.cache.paths" instead.`,
```
szdziedzic, line 53:
```suggestion
          `The "build.${
            profileName ?? 'production'
          }.cache.customPaths" field in eas.json is deprecated and will be removed in the future. Please use "build.${
            profileName ?? 'production'
          }.cache.paths" instead.`,
```

szdziedzic (commented):
When manually testing `eas-json` deprecation warnings, please check how the warnings look like if there are multiple of them printed at the same time, to validate if it's visually pleasing for the users

packages/eas-cli/src/build/android/prepareJob.ts
```diff
@@ -106,3 +103,21 @@ export async function prepareJobAsync(
 
   return sanitizeJob(job);
 }
+
+function getCacheSettings(
```
wkozyra95, line 107: can you move that logic to eas-json package? The same behavior will be needed in www when we resolve to build a profile for builds using GitHub integration, so it would be good if the eas-json package already returned the profile without customPaths field.

I think it should be possible directly in joi schema to add some conditional logic that renames fields.

packages/eas-cli/src/build/android/prepareJob.ts
```diff
@@ -106,3 +103,21 @@ export async function prepareJobAsync(
...
+  ctx: BuildContext<Platform.ANDROID>
+): Cache {
+  const cacheSettings = {
+    ...cacheDefaults,
+    ...buildProfile.cache,
+    clear: ctx.clearCache,
+  };
+  if (cacheSettings.customPaths) {
+    if (cacheSettings.customPaths.length > 0) {
+      cacheSettings.paths.push(...cacheSettings.customPaths);
```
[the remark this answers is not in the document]
wkozyra95: agree, it should throw an error if both are defined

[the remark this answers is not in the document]
radoslawkrzemien: Changed appending into assignment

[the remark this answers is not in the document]
radoslawkrzemien: Also, throws an error when both fields are set

packages/eas-cli/src/build/android/prepareJob.ts
radoslawkrzemien: Moved it to `eas-json`'s `EasJsonUtils`

packages/eas-json/src/utils.ts
```diff
@@ -73,4 +93,36 @@ export class EasJsonUtils {
     const easJson = await accessor.readAsync();
     return resolveSubmitProfile({ easJson, platform, profileName });
   }
+
+  public static getCacheSettings<T extends Platform>(
+    buildProfile: BuildProfile<T>,
+    ctx: any
```
szdziedzic, line 99:
```suggestion
    ctx: BuildContext<T>
```
Please try not to type things as `any` in TypeScript.

szdziedzic (changes requested):
The change looks good overall at this point.

Could you please add screenshots of new warnings/errors in the PR description?

packages/eas-json/src/utils.ts
radoslawkrzemien: @szdziedzic The problem is `BuildContext` comes from `eas-cli`, so importing it causes a circular dependency. I would have to move it to `eas-json` as well

packages/eas-json/src/utils.ts
```diff
@@ -1,9 +1,11 @@
 import { Platform } from '@expo/eas-build-job';
+import { Cache } from '@expo/eas-build-job/dist/common';
```
wkozyra95, line 2: Do not import like this, if it's not exported from the entry point then make the change in a package to export it.

packages/eas-json/src/utils.ts
```diff
@@ -73,4 +93,36 @@ export class EasJsonUtils {
     const easJson = await accessor.readAsync();
     return resolveSubmitProfile({ easJson, platform, profileName });
   }
+
+  public static getCacheSettings<T extends Platform>(
```
wkozyra95, line 97: just moving this function here does not solve the issue I described in my other comment, want getBuildProfileAsync already returns the converted build profile. 

Also you should not need anything from eas-cli package here.

wkozyra95 (changes requested):

packages/eas-json/src/utils.ts
wkozyra95: As mentioned above, I recommend doing that using only Joi schema without adding any procedural code.

packages/eas-cli/src/build/android/__tests__/prepareJob-test.ts
```diff
@@ -0,0 +1,104 @@
...
+    const result = await prepareJobAsync(ctxMock, jobData);
+    expect(result?.cache).toBeDefined();
+    expect(result.cache.paths).toBeDefined();
+    expect(result.cache.paths).toEqual([]);
+    expect(result.cache.customPaths).toBeUndefined();
+  });
+  it('uses default paths when settings provided without either paths or customPaths', async () => {
+    const ctxMock = getBuildContextMock();
+    ctxMock.buildProfile = {
+      // @ts-ignore
```
dsokal, line 33: Why is it necessary?

packages/eas-cli/src/build/android/__tests__/prepareJob-test.ts
```diff
@@ -0,0 +1,104 @@
...
+      cache: {
+        disabled: false,
+      },
+      credentialsSource: CredentialsSource.LOCAL,
+      distribution: 'internal',
+    };
+    const result = await prepareJobAsync(ctxMock, jobData);
+    expect(result?.cache).toBeDefined();
+    expect(result.cache.paths).toBeDefined();
+    expect(result.cache.paths).toEqual([]);
```
dsokal, line 43:
```suggestion
    expect(result.cache.paths).toEqual([]);
```

packages/eas-cli/src/build/android/__tests__/prepareJob-test.ts
```diff
@@ -0,0 +1,104 @@
...
+      cache: {
+        disabled: false,
+        paths: ['index.ts'],
+      },
+      credentialsSource: CredentialsSource.LOCAL,
+      distribution: 'internal',
+    };
+    const result = await prepareJobAsync(ctxMock, jobData);
+    expect(result?.cache).toBeDefined();
+    expect(result.cache.paths).toBeDefined();
```
dsokal, line 58:
```suggestion
```

packages/eas-cli/src/build/android/__tests__/prepareJob-test.ts
```diff
@@ -0,0 +1,104 @@
...
+      // @ts-ignore
+      cache: {
+        disabled: false,
+        customPaths: ['index.ts'],
+      },
+      credentialsSource: CredentialsSource.LOCAL,
+      distribution: 'internal',
+    };
+    const result = await prepareJobAsync(ctxMock, jobData);
+    expect(result?.cache).toBeDefined();
```
dsokal, line 74: Is `?` required here? The same comment applies to other lines like this in the file.
```suggestion
    expect(result.cache).toBeDefined();
```

packages/eas-json/src/utils.ts
dsokal: If `ctx.clearCache` is the only property you need to access from `ctx`, you can change this line to:
```suggestion
    { clearCache }: { clearCache: TYPE }
```

packages/eas-json/src/utils.ts
```diff
@@ -73,4 +93,36 @@ export class EasJsonUtils {
...
+    if (
+      buildProfile?.cache?.paths &&
+      buildProfile.cache.paths.length > 0 &&
+      buildProfile?.cache?.customPaths &&
+      buildProfile.cache.customPaths.length > 0
+    ) {
+      throw new InvalidEasJsonError(
+        `Found invalid cache settings in ${chalk.bold('eas.json')}. Cannot use both ${chalk.bold(
+          'paths'
+        )} and ${chalk.bold('customPaths')} at the same time`
```
dsokal, line 110:
```suggestion
        )} and ${chalk.bold('customPaths')} at the same time.`
```

packages/eas-json/src/utils.ts
```diff
@@ -73,4 +93,36 @@ export class EasJsonUtils {
...
+      clear: ctx.clearCache,
+    };
+
+    if (cacheSettings.customPaths) {
+      if (cacheSettings.customPaths.length > 0) {
+        cacheSettings.paths = cacheSettings.customPaths;
+      }
+      delete cacheSettings.customPaths;
+    }
+    return cacheSettings;
```
dsokal, line 126: This saves one level of indentation.
```suggestion
    if (!cacheSettings.customPaths) {
      return cacheSettings;
    }
    if (cacheSettings.customPaths.length > 0) {
      cacheSettings.paths = cacheSettings.customPaths;
    }
    delete cacheSettings.customPaths;
    return cacheSettings;
```

radoslawkrzemien:
/changelog-entry chore Add `cache.paths` field and deprecate `cache.customPaths`

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -65,7 +65,8 @@ async function readProfileAsync<T extends ProfileType>({
       profileName
     );
 
-    maybePrintBuildProfileDeprecationWarnings(buildProfile, profileName);
+    const rawEasJson = await easJsonAccessor.readRawJsonAsync();
```
wkozyra95, line 68: Consider passing easJsonAccessor all the way and running readRawJsonAsync when you need the value, to avoid passing any through a bunch of functions.

packages/eas-json/src/utils.ts
```diff
@@ -30,21 +30,33 @@ export class EasJsonUtils {
...
-          }.cache.cacheDefaultPaths" field in eas.json is deprecated and will be removed in the future.`,
+          `The "build.${finalProfileName}.cache.cacheDefaultPaths" field in eas.json is deprecated and will be removed in the future.`,
         ],
         docsUrl: 'https://docs.expo.dev/build-reference/caching/#ios-dependencies',
       });
     }
 
+    if (
+      buildProfile.cache?.customPaths !== undefined ||
+      rawEasJson?.build?.[finalProfileName]?.cache?.customPaths !== undefined
```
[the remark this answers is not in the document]
radoslawkrzemien: It's a bit problematic here, since the raw JSON doesn't have the merged profiles as the buildProfile does. Also, the previous method implemented above also doesn't check for that 🤔 
What would be the best approach here? Looping through all the profiles (regardless of their name) and checking all of them?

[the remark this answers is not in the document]
radoslawkrzemien: I've changed the singular check into a recursive function going deeper through the `extends` key until the extensions finish or a deprecated field is found

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -79,15 +83,21 @@ async function readProfileAsync<T extends ProfileType>({
 
 let hasPrintedDeprecationWarnings = false;
 
-function maybePrintBuildProfileDeprecationWarnings(
+export function clearHasPrintedDeprecationWarnings(): void {
+  hasPrintedDeprecationWarnings = false;
+}
```
dsokal, line 88: This is only necessary for tests? Maybe add a comment it shouldn't be used elsewhere.

packages/eas-json/src/build/schema.ts
```diff
@@ -23,7 +23,11 @@ const CacheSchema = Joi.object({
   key: Joi.string().max(128),
   cacheDefaultPaths: Joi.boolean(),
   customPaths: Joi.array().items(Joi.string()),
-});
+  paths: Joi.array().items(Joi.string()),
+}).rename('customPaths', 'paths')
+  .messages({
+    'object.rename.override': 'Cannot provide both "cache.customPaths" and "cache.paths" - use "paths"'
```
dsokal, line 29:
```suggestion
    'object.rename.override': 'Cannot provide both "cache.customPaths" and "cache.paths" - use "cache.paths"'
```

packages/eas-cli/src/utils/__tests__/profiles-test.ts
```diff
@@ -94,3 +114,89 @@ describe(getProfilesAsync, () => {
     ).rejects.toThrowError(/eas.json is not valid/);
   });
 });
+describe(maybePrintBuildProfileDeprecationWarningsAsync, () => {
+  afterEach(() => {
+    clearHasPrintedDeprecationWarnings();
+    newLineSpy.mockClear();
+    warnSpy.mockClear();
```
dsokal, line 121: Can you please check whether this is necessary at all? I think our jest configuration could be clearing mocks by default after every test case.

packages/eas-cli/src/utils/__tests__/profiles-test.ts
```diff
@@ -94,3 +114,89 @@ describe(getProfilesAsync, () => {
...
+      expect(newLineSpy).toHaveBeenCalledTimes(2);
+      expect(warnSpy).toHaveBeenCalledTimes(3);
+      const warnCalls = warnSpy.mock.calls;
+      expect(warnCalls[0][0]).toEqual('Detected deprecated fields in eas.json:');
+      expect(warnCalls[1][0]).toEqual(
+        '\tThe "build.production.cache.customPaths" field in eas.json is deprecated and will be removed in the future. Please use "build.production.cache.paths" instead.'
+      );
+      const underlinedText = 'https://docs.expo.dev/build-reference/eas-json/#cache';
+      const dimmedText = `Learn more: ${chalk.underline(underlinedText)}`;
+      expect(warnCalls[2][0]).toEqual(`\t${chalk.dim(dimmedText)}`);
```
dsokal, line 159: Maybe it's worth considering using `.toMatchSnapshot` (https://jestjs.io/docs/snapshot-testing) instead of comparing every line of the output.

packages/eas-json/src/utils.ts
```diff
@@ -28,23 +28,52 @@ export class EasJsonUtils {
     return resolveBuildProfile({ easJson, platform, profileName });
   }
 
-  public static getBuildProfileDeprecationWarnings(
+  public static async getBuildProfileDeprecationWarningsAsync(
     buildProfile: BuildProfile,
```
dsokal, line 32:
```suggestion
```

packages/eas-json/src/utils.ts
```diff
@@ -28,23 +28,52 @@ export class EasJsonUtils {
...
+    easJsonAccessor: EasJsonAccessor,
     profileName?: string
```
dsokal, line 34:
```suggestion
    profileName: string
```

packages/eas-json/src/utils.ts
```diff
@@ -28,23 +28,52 @@ export class EasJsonUtils {
...
-  ): EasJsonDeprecationWarning[] {
+  ): Promise<EasJsonDeprecationWarning[]> {
     const warnings: EasJsonDeprecationWarning[] = [];
+    const rawEasJson = await easJsonAccessor.readRawJsonAsync();
+    const finalProfileName = profileName ?? 'production';
```
dsokal, line 38:
```suggestion
```

packages/eas-json/src/utils.ts
```diff
@@ -28,23 +28,52 @@ export class EasJsonUtils {
...
+      });
+    } else if (rawEasJson.build?.[profileName]?.extends !== undefined) {
+      warnings.push(
+        ...EasJsonUtils.getCustomPathsDeprecationWarnings(
+          rawEasJson,
+          buildProfileName,
+          rawEasJson.build?.[profileName].extends
+        )
+      );
+    }
```
wkozyra95, line 76:
```suggestion
    if (rawEasJson.build?.[profileName]?.extends !== undefined) {
      warnings.push(
        ...EasJsonUtils.getCustomPathsDeprecationWarnings(
          rawEasJson,
          buildProfileName,
          rawEasJson.build?.[profileName].extends
        )
      );
    }
```

wkozyra95 (approved):

packages/eas-cli/src/utils/__tests__/profiles-test.ts
radoslawkrzemien: It is needed, they don't get cleared by default

packages/eas-cli/src/utils/__tests__/profiles-test.ts
radoslawkrzemien: But here I check with which arguments the `warn()` function has been called, how do I get the entire output to match against the snapshot?

packages/eas-json/src/utils.ts
```diff
@@ -28,23 +28,57 @@ export class EasJsonUtils {
...
     return warnings;
   }
```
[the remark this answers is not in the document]
radoslawkrzemien: It will not result in the same message - my version will print the profile name of the profile being checked for deprecations, your version will print the name of the profile being extended that has the deprecated fields. I guess it's debatable which name is more useful to the user, but I would say the name of the final (extending) profile, not the base (extended) one

packages/eas-cli/src/utils/__tests__/profiles-test.ts
dsokal: Bump

packages/eas-json/src/utils.ts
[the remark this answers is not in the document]
radoslawkrzemien: Well, in that case I can make the change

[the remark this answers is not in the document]
radoslawkrzemien: Updated

dsokal (approved):

PR #1803, opened 2023-04-20
[eas-cli] Device delete not disabling on Apple side

keith-kurak:
# Why
For https://github.com/expo/eas-cli/pull/1787, I copied the Expo-to-Apple device cross-reference from `eas device:delete`. During testing, I discovered that it never found the device on Apple, and then when I went back to try out delete, it turned out that delete wasn't finding the device either, and thus never disabling it on the Apple Developer Portal.

# How
The previous `find` was comparing Expo GUID to an Apple 9-character alphanumeric ID. Updated to compare `udid` to `udid`.

Here's what the objects look like:

`AppleDevice` (Expo entity):
```
{
  id: 'f8108da4-d62b-4325-8f0d-b646133c2d16',
  identifier: '00008110-ABCDEFD12345678',
  name: 'keiths ipad',
  deviceClass: 'IPAD',
  enabled: true,
  model: null,
}
```

`Device` (from `@expo/apple-utils`):
```
{
  context: { providerId: 12345678, teamId: 'ABCDEF1234' },
  id: '12345678K6',
  attributes: {
    addedDate: '2023-04-20T14:33:24.000+00:00',
    serialNumber: null,
    inEligibilityDuration: 0,
    description: null,
    platform: 'IOS',
    name: 'keiths ipad',
    deviceClass: 'IPAD',
    model: 'iPad mini (6th generation)',
    udid: '00008110-ABCDEFD12345678',
    platformName: 'iOS',
    responseId: 'fd66367c-a738-4104-bd4b-41e69c41a9c2',
    status: 'ENABLED',
    devicePlatformLabel: 'iPad'
  }
}
```

Also added a bit more context when manually creating a device as to when the device is actually created on the Apple side when I realized that, in my haste when testing renames and deletes, I wasn't actually taking the steps to create the device on Apple.

# Test Plan
- [x] Tested `eas device:delete` with a device created on Apple, and it successfully disabled the device
- [x] Tested `eas device:delete` with a device not created on Apple, it it warned that it wasn't found on the Apple side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fix `device:delete` not disabling on Apple. ([#1803](https://github.com/expo/eas-cli/pull/1803) by [@keith-kurak](https://github.com/keith-kurak))
+
 ### 🧹 Chores
 
 - Throw an error if somebody tries to start iOS build with the `large` resource class selected using the deprecated `--resource-class` flag. ([#1795](https://github.com/expo/eas-cli/pull/1795) by [@szdziedzic](https://github.com/szdziedzic))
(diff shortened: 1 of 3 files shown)


dsokal (approved):
Thank you!

PR #1814, opened 2023-04-26
[eas-cli] Add account list to `eas whoami`/`eas account:view`

wschurman:
# Why

When looking into #1324, I realized there's nowhere in the CLI that we show all the accounts that the logged-in user has access to and their roles. 

I think it makes sense in this command, though I was hesitant at first to add it as there's a chance that people are programmatically parsing the output since it has historically been just a username. What do others think?

# How

Add output. Will add tests once others chime in about whether it makes sense here.

# Test Plan

Test manually:

```
$ neas whoami
fake_username

Accounts:
• fake_username (Role: Owner)
• fake_other_account (Role: Admin)
• test-org (Role: Developer)
```

### DIFF
--- a/packages/eas-cli/src/commands/account/view.ts
+++ b/packages/eas-cli/src/commands/account/view.ts
@@ -1,8 +1,10 @@
 import chalk from 'chalk';
+import nullthrows from 'nullthrows';
 
 import EasCommand from '../../commandUtils/EasCommand';
+import { Role } from '../../graphql/generated';
 import Log from '../../log';
-import { getActorDisplayName } from '../../user/User';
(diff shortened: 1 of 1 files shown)


wschurman:
/changelog-entry new-feature Add account list to `eas whoami`/`eas account:view`

wkozyra95:
> I was hesitant at first to add it as there's a chance that people are programmatically parsing the output since it has historically been just a username.
 
 I wouldn't worry about it, we provide json option in some commands for that purpose, but we can't worry about users parsing a regular output, otherwise, every change would be breaking change. I personally don't think it's needed but if you think someone might parse the output, then maybe add support for json flag.

Issue #1123, opened 2022-05-28
eas update missing --clear-cache option


jkk:
### Build/Submit details page URL

https://expo.dev/accounts/kavyar-org/projects/kavyar-app/updates/d58a32ad-7c5a-4122-8def-29c85401bb7d

### Summary

`eas update` seems to be missing a `--clear-cache` option. `expo publish` had this option and `eas build` has it. Not having it is causing problems with our app updates.

We are using a STAGE environment variable to auto-configure our app for the appropriate release environment (dev, staging, production). We inject it into the app using eas.json, app.config.js, and the `Constants.manifest2?.extra?.expoClient?.extra?.eas?.STAGE` technique recommended in the docs.

This works fine for the initial EAS build. However, when we do `eas update`, the generated bundle will often have a stale value (from a prior build) for the STAGE var that gets injected via `Constants`.

We used to use the `--clear` option with expo publish to work around this, but that option is not available for `eas update`.

This recently caused a staging config to be released into our production app, which created quite a mess.

### Managed or bare?

Managed

### Environment

```
  expo-env-info 1.0.3 environment info:
    System:
      OS: macOS 12.2.1
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 14.18.1 - ~/.nvm/versions/node/v14.18.1/bin/node
      Yarn: 1.22.18 - /opt/homebrew/bin/yarn
      npm: 6.14.15 - ~/.nvm/versions/node/v14.18.1/bin/npm
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
      Android SDK:
        API Levels: 31
        Build Tools: 31.0.0, 32.0.0
        System Images: android-31 | Google APIs ARM 64 v8a
    IDEs:
      Android Studio: 2020.3 AI-203.7717.56.2031.7784292
      Xcode: 13.4/13F17a - /usr/bin/xcodebuild
    npmPackages:
      babel-preset-expo: ~9.1.0 => 9.1.0 
      expo: ^45.0.0 => 45.0.4 
      react: 17.0.2 => 17.0.2 
      react-dom: 17.0.2 => 17.0.2 
      react-native: 0.68.2 => 0.68.2 
      react-native-web: 0.17.7 => 0.17.7 
    npmGlobalPackages:
      eas-cli: 0.52.0
      expo-cli: 5.4.7
    Expo Workflow: managed
```

[...]

PR #1839, opened 2023-05-14
[eas-cli] add clear cache option to updates

quinlanj:
# Why

The `clear-cache` option to remove the metro bundler cache exists for `eas build` and for the Classic Updates `expo publish` command, but it was overlooked for `eas update`. This PR adds the flag to `eas update` to bring it to parity with our other commands.

Fixes https://github.com/expo/eas-cli/issues/1123 . Also a bunch of other users reported issues with `eas update` that I suspect would be fixed by clearing the metro bundler too.

# How

`eas update` relies on the expo cli's `expo export` command to generate it's bundle. This PR plumb down the `clear-cache` flag to the `expo export` command. This should work with all released versions of `expo export` since support for the `clear` flag has always existed. 

# Test Plan

- [ ] tested manually `eas update --clear-cache`

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -59,6 +59,7 @@ type RawUpdateFlags = {
   platform: string;
   'input-dir': string;
   'skip-bundler': boolean;
+  'clear-cache': boolean;
   'private-key-path'?: string;
   'non-interactive': boolean;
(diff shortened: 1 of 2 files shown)


quinlanj:
/changelog-entry new-feature add clear cache flag to eas update

dsokal (approved):

wschurman (approved):

PR #1843, opened 2023-05-16
[eas-cli] remove asp prompt

quinlanj:
# Why

Back in 2020/2021 we made the EAS Submit service, and we used App Specific Password (ASP) to authenticate. There were 2 ways to provide an ASP - passing in an environment variable or through a prompt. 

We eventually moved away from ASPs to ASC Api Keys because they were easier to manage and Apple generally supported them better. When I added ASC Api Key support, we wanted to let people still authenticate with an ASP through an environment variable, but we no longer wanted to support the prompt. 

We figured that if we removed the ASP prompt right away, people would complain because the vast majority relied on authenticating via ASP. Instead, we decided to keep the prompt around for 'a little bit' and warn the user it was going away soon and to use the environment variable instead.

A couple years later, we are finally going to remove the prompt. It will simplify the submission credentials workflow quite a bit.

# How

Rebased https://github.com/expo/eas-cli/pull/743 with a couple notable changes (because it is so old 😅):
- @wkozyra95 added additional support for resolving a bundle identifier (https://github.com/expo/eas-cli/commit/b81b896bc61bfe513902c0d26bbf3402faefd44d and https://github.com/expo/eas-cli/commit/6bd62bac64ed5bef0c66a46361cae65b44b1373f) when we determine which source the credentials should be coming from (remote, local, etc). I've refactored this so the logic runs when the credentials are resolved, in `AscApiKeySource`

# Test Plan

- [x] new tests pass
- [x] manually tested `eas submit` and confirmed that the remote ASC Api Key is resolved correctly


quinlanj:
/changelog-entry chore Remove App Specific Password prompt

dsokal (approved):

wkozyra95 (approved):

PR #1851, opened 2023-05-19
[eas-cli] Improve types for update display

wschurman:
# Why

The main concern here was the any cast `update.actor as any`. This was causing us to miss cases that should've been handled.

This is related to https://github.com/expo/universe/pull/12383 in that I did these while investigating that issue which turned out to be a server-side issue. Figured I should commit these anyways since they're strictly safer.

# How

Remove casts, improve switch exhaustiveness, improve undefined label to say "unknown" instead of "anonymous".

# Test Plan

`yarn tsc`


wschurman:
/changelog-entry chore Improve typescript types for user display

quinlanj (approved):

dsokal (approved):

dsokal (approved):

PR #1855, opened 2023-05-23
[eas-cli] [ENG-8506] Handle pending cancel build status in cli

radoslawkrzemien:
# Why

Adding new build status `PENDING_CANCEL` which is supposed to give workers time to perform cleanup and upload logs when the build is canceled before the worker gets terminated by the launcher. This status is only relevant to EAS and from the user's perspective builds in this status should be treated the same as the ones that are already canceled. This PRs adds support for that

# How

In all the places where the behaviour is dictated by the build status a change is introduced to mirror the behaviour for canceled builds.

# Test Plan

Manually tested with modified response from `www` returning a build in `PENDING_CANCEL` status ⬇️ 
Screenshot 2023-05-24 at 14 01 34
Build `9ae0d32f-9ed5-4eb8-a396-bfbbd4ca95e7` is actually `PENDING_CANCEL` but is displayed as `CANCELED`

Testing with the version of CLI from main shows the build as in `unknown` status, which seems indicative for this change to not be breaking
Screenshot 2023-05-24 at 14 12 16

### DIFF
--- a/packages/eas-cli/src/build/utils/printBuildInfo.ts
+++ b/packages/eas-cli/src/build/utils/printBuildInfo.ts
@@ -80,7 +80,7 @@ function printBuildResult(build: BuildFragment): void {
     }
     return;
   }
-  if (build.status === BuildStatus.Canceled) {
+  if (build.status === BuildStatus.Canceled || build.status === BuildStatus.PendingCancel) {
     Log.addNewLineIfNone();
     Log.error(
       `${appPlatformEmojis[build.platform]} ${
--- a/packages/eas-cli/src/build/utils/printBuildInfo.ts
+++ b/packages/eas-cli/src/build/utils/printBuildInfo.ts
@@ -80,7 +80,7 @@ function printBuildResult(build: BuildFragment): void {
     }
     return;
   }
-  if (build.status === BuildStatus.Canceled || build.status === BuildStatus.PendingCancel) {
+  if ([BuildStatus.Canceled, BuildStatus.PendingCancel].includes(build.status)) {
     Log.addNewLineIfNone();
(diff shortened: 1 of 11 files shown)


packages/eas-cli/src/build/utils/printBuildInfo.ts
```diff
@@ -80,7 +80,7 @@ function printBuildResult(build: BuildFragment): void {
     }
     return;
   }
-  if (build.status === BuildStatus.Canceled) {
+  if (build.status === BuildStatus.Canceled || build.status === BuildStatus.PendingCancel) {
```
wkozyra95, line 83: nit: I would prefer pattern [status1, status2).includes(status)

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -174,7 +174,7 @@ export default class BuildCancel extends EasCommand {
     const spinner = ora().start('Canceling the build…');
     try {
       const { status } = await cancelBuildAsync(graphqlClient, buildId);
-      if (status === BuildStatus.Canceled) {
+      if (status === BuildStatus.Canceled || status === BuildStatus.PendingCancel) {
```
wkozyra95, line 177: the same

wkozyra95 (approved):

wschurman (approved):

radoslawkrzemien:
@szdziedzic The deployment plan is described in the PR mentioning this one, like https://github.com/expo/turtle-v2/pull/1306 :)

radoslawkrzemien:
/changelog-entry chore Add support for pending-cancel build status

PR #1857, opened 2023-05-24
[eas-cli] Refactor getExpoConfig to remove dangerous default

wschurman:
# Why

We had a bug where we were accidentally putting the private config in the public config due to a tough-to-catch refactoring error. This PR is an effort to make refactoring errors like this less likely in the future.

https://github.com/expo/sentry-expo/issues/321#issuecomment-1469862532

Closes ENG-7940.

# How

Separate out private and public config getters to make it clearer that there is a difference in name and type from the private and public expo configs. This should help code reviewers catch errors during refactors.

# Test Plan

Run all tests.


packages/eas-cli/src/commandUtils/context/OptionalProjectConfigContextField.ts
```diff
@@ -1,12 +1,12 @@
...
 import { InvalidEasJsonError } from '@expo/eas-json/build/errors';
 
-import { getExpoConfig } from '../../project/expoConfig';
+import { getPrivateExpoConfig } from '../../project/expoConfig';
 import ContextField, { ContextOptions } from './ContextField';
 import { findProjectDirAndVerifyProjectSetupAsync } from './contextUtils/findProjectDirAndVerifyProjectSetupAsync';
 import { getProjectIdAsync } from './contextUtils/getProjectIdAsync';
 
-export class OptionalProjectConfigContextField extends ContextField<
+export class OptionalPrivateProjectConfigContextField extends ContextField<
```
wkozyra95, line 9: maybe file also should be renamed?

packages/eas-cli/src/commandUtils/context/ProjectConfigContextField.ts
```diff
@@ -1,11 +1,11 @@
 import { ExpoConfig } from '@expo/config-types';
 
-import { getExpoConfig } from '../../project/expoConfig';
+import { getPrivateExpoConfig } from '../../project/expoConfig';
 import ContextField, { ContextOptions } from './ContextField';
 import { findProjectDirAndVerifyProjectSetupAsync } from './contextUtils/findProjectDirAndVerifyProjectSetupAsync';
 import { getProjectIdAsync } from './contextUtils/getProjectIdAsync';
 
-export default class ProjectConfigContextField extends ContextField<{
+export class PrivateProjectConfigContextField extends ContextField<{
```
wkozyra95, line 8: maybe file rename?

wkozyra95 (approved):

PR #1865, opened 2023-05-31
[ENG-8790][eas-cli] Fix updates synchronization of native files to include strings.xml

wschurman:
# Why

https://github.com/expo/expo/commit/166385e28b75d11684bacdcace0c4abb92434adb added this to the config plugin.

This code in eas-cli is supposed to emulate the config plugin, but for bare projects.

# How

Make the config plugin emulation also do the new synchronization of strings.xml.

# Test Plan

- npx create expo-app
- npx expo prebuild
- add runtime version to app.json
- install expo-updates
- commit everything 
- eas build (current prod version)
  - it added `<meta-data android:name="expo.modules.updates.EXPO_RUNTIME_VERSION" android:value="@string/expo_runtime_version"`
  - but there were no changes in strings.xml
- neas build (with this PR)
  - it added `<meta-data android:name="expo.modules.updates.EXPO_RUNTIME_VERSION" android:value="@string/expo_runtime_version"`
  - it added `<string name="expo_runtime_version">1.0</string>` to strings.xml

### DIFF
--- a/packages/eas-cli/src/update/android/UpdatesModule.ts
+++ b/packages/eas-cli/src/update/android/UpdatesModule.ts
@@ -18,8 +18,6 @@ export async function syncUpdatesConfigurationAsync(
   ensureValidVersions(exp, RequestedPlatform.Android);
   const accountName = (await getOwnerAccountForProjectIdAsync(graphqlClient, projectId)).name;
 
-  AndroidConfig.Updates.withUpdates(exp, { expoUsername: null });
-
   // sync AndroidManifest.xml
   const androidManifestPath = await AndroidConfig.Paths.getAndroidManifestAsync(projectDir);
   const androidManifest = await getAndroidManifestAsync(projectDir);
--- a/packages/eas-cli/src/update/android/UpdatesModule.ts
+++ b/packages/eas-cli/src/update/android/UpdatesModule.ts
@@ -1,11 +1,14 @@
 import { ExpoConfig } from '@expo/config';
-import { AndroidConfig, AndroidManifest } from '@expo/config-plugins';
+import { AndroidConfig, AndroidManifest, XML } from '@expo/config-plugins';
 
 import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
 import { RequestedPlatform } from '../../platform';
 import { getOwnerAccountForProjectIdAsync } from '../../project/projectUtils';
 import { ensureValidVersions } from '../utils';
 
+/**
+ * Synchronize updates configuration to native files. This needs to do essentially the same thing as `withUpdates`
+ */
 export async function syncUpdatesConfigurationAsync(
   graphqlClient: ExpoGraphqlClient,
   projectDir: string,
@@ -15,9 +18,11 @@ export async function syncUpdatesConfigurationAsync(
   ensureValidVersions(exp, RequestedPlatform.Android);
   const accountName = (await getOwnerAccountForProjectIdAsync(graphqlClient, projectId)).name;
 
+  AndroidConfig.Updates.withUpdates(exp, { expoUsername: null });
+
+  // sync AndroidManifest.xml
   const androidManifestPath = await AndroidConfig.Paths.getAndroidManifestAsync(projectDir);
   const androidManifest = await getAndroidManifestAsync(projectDir);
-
   const updatedAndroidManifest = AndroidConfig.Updates.setUpdatesConfig(
     projectDir,
     exp,
@@ -28,6 +33,17 @@ export async function syncUpdatesConfigurationAsync(
     androidManifestPath,
     updatedAndroidManifest
   );
+
+  // sync strings.xml
+  const stringsJSONPath = await AndroidConfig.Strings.getProjectStringsXMLPathAsync(projectDir);
+  const stringsResourceXML = await AndroidConfig.Resources.readResourcesXMLAsync({
+    path: stringsJSONPath,
+  });
+  const updatedStringsResourceXML = AndroidConfig.Updates.applyRuntimeVersionFromConfig(
+    exp,
+    stringsResourceXML
+  );
+  await XML.writeXMLAsync({ path: stringsJSONPath, xml: updatedStringsResourceXML });
 }
 
 export async function readReleaseChannelSafelyAsync(projectDir: string): Promise<string | null> {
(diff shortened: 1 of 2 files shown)


wschurman:
After figuring out why this code is necessary (i.e. why config plugins aren't being run), my next questions will be:
a) How can we make them run? Why don't they run? How do we expect people to do this?
b) Can we just call the config plugin here? Why is that necessary?

I think I'm missing a ton of context about why this sync even needs to happen in the first place.

packages/eas-cli/src/update/android/UpdatesModule.ts
```diff
@@ -15,9 +18,11 @@ export async function syncUpdatesConfigurationAsync(
   ensureValidVersions(exp, RequestedPlatform.Android);
   const accountName = (await getOwnerAccountForProjectIdAsync(graphqlClient, projectId)).name;
 
+  AndroidConfig.Updates.withUpdates(exp, { expoUsername: null });
```
wkozyra95, line 21: what is this line doing?

I might not understand how plugin stuff works, but when see what is inside withUpdates it leads me to think that it's either is doing the same stuff as the code bellow, so it does the same thing twice, or it only evaluates config with plugin which would mean it does not do anything because plugins were already evaluated in exp that is passed to the function.

wkozyra95:
> This code in eas-cli I guess is supposed to emulate the config plugin?

It's intended to sync values with native code in a bare project.

> After figuring out why this code is necessary (i.e. why config plugins aren't being run), my next questions will be:
a) How can we make them run? Why don't they run? How do we expect people to do this?
b) Can we just call the config plugin here? Why is that necessary?

I think answer for all of this is because this is used in bare projects, so we can't expect that prebuild will or even could be run.

> Can we just call the config plugin here? Why is that necessary?

And, specifically to that point. Maybe we can here, but not in general. Code in eas-cli needs to handle most (ideally all) native projects that EAS supports, so it's a higher standard than is expected in prebuild.

e.g. prebuild assumes that there will be no variant/flavors and when we interact with any with projects we need to be aware of that. For example instead of writing to main strings.xml you could write to runtime version to resource specific to a variant that is going to be built. (just an example I don't think it's need here)

packages/eas-cli/src/update/android/UpdatesModule.ts
wschurman: Oh, sorry, this was just me testing. I actually have no idea how to call a config plugin.

quinlanj (changes requested):
Thinking about this problem a bit more, how about colocating all the workflow agnostic 'modding code' in `@expo/config-plugins`? The problem is, we have 4 places which we 'mod' the Updates code:
- `expo prebuild`, managed workflow. Code in `@expo/config-plugins`, called in `npx expo prebuild` and Turtle Builders
- `eas updates:configure`, bare workflow. Calls out to a subroutine in `@expo/config-plugins`
- `eas build`, bare workflow. Calls out to a subroutine in `@expo/config-plugins`
- Turtle Builders, workflow agnostic. Code in [build-tools](https://github.com/expo/eas-build/blob/main/packages/build-tools/src/android/expoUpdates.ts#L35)

When doug added bug fixes to `@expo/config-plugins`, all the other call sites for the bare/agnostic workflows fell out of sync. They either broke completely by running a subset of the methods, or were flawed because they never picked up the bug fixes. What if we export a method called `modUpdatesBareOrAgnosticWorkflow ` in `@expo/config-plugins` that mods a bare/agnostic workflow, and have it be called by `eas-cli` and `build-tools`? That way, if we make any bug changes in the future, all our fixes can be made in `@expo/config-plugins` and we won't have to reimplement the same thing in 4 different places.

@wkozyra95 mentioned there is a higher standard in `eas-cli` and `build-tools`, where we cannot make assumptions about the variant/flavour of a project. However, if someone were to install `expo-updates`, we are requiring them to have the correct keys in AndroidManifest.xml and strings.xml. We can assert all our preconditions in the `@expo/config-plugins` for the bare and agnostic workflows if they are different from managed. So something like this:

```
export const withUpdates = [...]

// run by build-tools and eas-cli
export function modUpdatesBareOrAgnosticWorkflow () {

// assert preconditions for string.xml, if any

// run the same subroutines as withUpdates

}
```

wkozyra95:
> What if we export a method called modUpdatesBareOrAgnosticWorkflow  in @expo/config-plugins that mods a bare/agnostic workflow, and have it be called by eas-cli and build-tools?

The problem is that at different places we are doing slightly different things. if we had some method to sync configuration only for updates then we could use it in eas-cli and prebuild, but in turtle code does not sync values with app.json. It syncs one value passed as an argument.

We could write this method (e.g. setUpdatedConfiguration) in a way that takes everything as an argument and only updates stuff based on the passed object, but in that case, it wouldn't be full solution, just a helper method, so there still would be a risk that someone adds code to plugin and not this helper method. At this point, we are getting back to a problem what is a public API of config-plugins?

> where we cannot make assumptions about the variant/flavour of a project

That was just an example, there other alternatives could be 

The other approaches that we might consider:
- Using only well-defined utils and reimplementing top level stuff. It sounds like a lot of work, but after switch to versioned cli using config-plugins code becomes a lot more complicated, so overall it could be beneficial to just avoid reusing that code. In the current workflow when fixing sth in config-plugins you need to make a change there, publish it (which is complicated in itself) and update it everywhere. With flow like that, I think fixing sth in 2-3 places might be less work.
- Tagging with some comments about which functions are used in cli and turtle.

wkozyra95 (approved):

quinlanj (approved):
@wkozyra95 makes sense. At the very least, let's add a comment in `config-plugins` for anyone who tries to make changes in the future. That way, we can do our best to avoid this type of situation from happening again and it's not a lot of extra work.

wschurman:
Still need to test this before landing. I'm assuming the repro is something like: create bare project using old CLIs, prebuild using old CLI, build using the CLI in this PR?

wschurman:
/changelog-entry bug-fix Fix updates synchronization of native files to include strings.xml for bare projects

sparga:
Since was-cli 3.13.3, which include this PR, the XML prolog of my `strings.xml` is removed every time I start a build with `eas build --no-wait --non-interactive -p all --auto-submit`. 

And then it fails because I'm using `--non-interactive` and the git state is not clean anymore. Is it expected?

Here is the diff after trying to build:

```diff
diff --git a/android/app/src/main/res/values/strings.xml b/android/app/src/main/res/values/strings.xml
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -1,4 +0,3 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <resources>
   <string name="app_name">MyApp</string>
   <string name="expo_runtime_version">10</string>

```

PR #1881, opened 2023-06-15
[eas-cli] Initialize app.json if it does not exist, add expo key if missing

brentvatne:
# Why

It's unfortunate to have to tell people to manually create this file, we should do it automatically with `eas init` or `eas update:configure` or `eas build:configure`. Additionally, the current error is.. not very helpful: 

```
easd build:configure
EAS project not configured.
✔ Existing EAS project found for @brents/quin (id = 0f084c7f-d209-491d-97ac-c8c5c4c2c122). Configure this project? … yes
✖ Linking local project to EAS project 0f084c7f-d209-491d-97ac-c8c5c4c2c122
No config exists
    Error: build:configure command failed.
```

"No config exists"

# How

Wrap `modifyConfigAsync` and add some ensure calls where we read the app config.

# Test Plan

I tried this in a project without app.json, and with an app.json that does not have an `expo` key. Need to write actual unit tests to cover this behavior still

### DIFF
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -74,7 +74,7 @@ export function ensureExpoConfigExists(projectDir: string): void {
 
 async function ensureStaticExpoConfigIsValidAsync(projectDir: string): Promise<void> {
   const paths = getConfigFilePaths(projectDir);
-  if (paths?.staticConfigPath?.endsWith('app.json')) {
+  if (paths?.staticConfigPath?.endsWith('app.json') && !paths?.dynamicConfigPath) {
     const staticConfig = await JsonFile.readAsync(paths.staticConfigPath);
 
     // Add the "expo" key if it doesn't exist on app.json yet, such as in
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -25,7 +25,6 @@ interface ExpoConfigOptionsInternal extends ExpoConfigOptions {
 export async function createOrModifyExpoConfigAsync(
   projectDir: string,
   exp: Partial<ExpoConfig>,
-  // options is the same type as the third argument to modifyConfigAsync
   readOptions?: { skipSDKVersionRequirement?: boolean }
 ): ReturnType<typeof modifyConfigAsync> {
   ensureExpoConfigExists(projectDir);
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -1,5 +1,8 @@
-import { ExpoConfig, getConfig } from '@expo/config';
+import { ExpoConfig, getConfig, getConfigFilePaths, modifyConfigAsync } from '@expo/config';
 import { Env } from '@expo/eas-build-job';
+import JsonFile from '@expo/json-file';
+import fs from 'fs-extra';
+import path from 'path';
 
 export type PublicExpoConfig = Omit<
   ExpoConfig,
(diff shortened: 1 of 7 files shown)


brentvatne:
/changelog-entry chore Create app.json or add the "expo" key if either are missing, before modifying or reading the file

packages/eas-cli/src/project/expoConfig.ts
```diff
@@ -40,12 +59,38 @@ function getExpoConfigInternal(
...
+  ensureExpoConfigExists(projectDir);
+
   return getExpoConfigInternal(projectDir, { ...opts, isPublicConfig: false });
 }
 
+export function ensureExpoConfigExists(projectDir: string): void {
+  const paths = getConfigFilePaths(projectDir);
+  if (!paths?.staticConfigPath && !paths?.dynamicConfigPath) {
+    // eslint-disable-next-line node/no-sync
+    fs.writeFileSync(path.join(projectDir, 'app.json'), JSON.stringify({ expo: {} }, null, 2));
```
wschurman, line 71: I think `slug` and `name` are required to be considered valid (based on xdl schema)?

wschurman:
Also curious in what cases this wouldn't exist?

packages/eas-cli/src/project/expoConfig.ts
```diff
@@ -40,12 +59,38 @@ function getExpoConfigInternal(
...
+  }
+}
+
+async function ensureStaticExpoConfigIsValidAsync(projectDir: string): Promise<void> {
+  const paths = getConfigFilePaths(projectDir);
+  if (paths?.staticConfigPath?.endsWith('app.json')) {
```
quinlanj, line 77: Is it possible for someone to not have a static config path at all, and opt to do all their configuration with a dynamic config path? If so, then we will need to do something like this:
```suggestion
  if (paths?.staticConfigPath?.endsWith('app.json') && !paths?.dynamicConfigPath) {
```

packages/eas-cli/src/project/expoConfig.ts
quinlanj: fwiw `getConfig`from '@expo/config' will populate `slug` and `name` by default. For example, if I had an empty `app.json`, `getConfig()` will return smth like this:

```
{
  name: 'AwesomeProject2',
  slug: 'awesomeproject2',
...
}
```

As long as there is an `expo` field in `app.json`, any calls to `modifyConfig` should generate a correct `app.json` (there are some legacy reasons why this is the case)

quinlanj (approved):

quinlanj:
@wschurman we are optimizing for the case someone with a bare React Native app tries to use EAS Update. So for example, someone who generated their project with `npx react-native@latest init AwesomeProject` instead of what was in our docs

wkozyra95 (approved):

szdziedzic (approved):

packages/eas-cli/src/project/expoConfig.ts
wschurman: Interesting. I'm assuming that those fields are never written to the file, and instead just returned as part of the config from `getConfig`. This could have odd consequences if the user changes their directory name (it defaults to `path.basename(projectRoot)`); their project ID would now mismatch. Also we need to be sure that we only ever use `getConfig` to access the config and never upload it directly as there are definitely tools that require these fields to exist.

brentvatne: yeah - i noticed t

> fwiw `getConfig`from '@expo/config' will populate `slug` and `name` by default. For example, if I had an empty `app.json`, `getConfig()` will return smth like this:
> 
> ```
> {
>   name: 'AwesomeProject2',
>   slug: 'awesomeproject2',
> ...
> }
> ```
> 
> As long as there is an `expo` field in `app.json`, any calls to `modifyConfig` should generate a correct `app.json` (there are some legacy reasons why this is the case)

yeah, i originally had implemented this by checking if `getConfig` returned a truthy value then i remembered this 😅 so i changed to getting config file paths instead.

> This could have odd consequences if the user changes their directory name (it defaults to path.basename(projectRoot)); their project ID would now mismatch

definitely! it's been like this for a while though so not going to try to solve it in this pr

PR #1882, opened 2023-06-15
[eas-cli] Ensure that re-thrown error includes original stack

brentvatne:
# Why

Without this diff, all stack traces that you can see with `DEBUG=* eas [cmd]` will point to where we catch and sanitize the error `EasCommand.ts`, and not give you any indication of the original source of the error.

# How

Point the `stack` on the new error object to the original error stack.

# Test Plan

- Add an error somewhere inside a command
- Run `DEBUG=* easd [command]`
- You should see the appropriate stack

### DIFF
--- a/packages/eas-cli/src/commandUtils/EasCommand.ts
+++ b/packages/eas-cli/src/commandUtils/EasCommand.ts
@@ -203,6 +203,8 @@ export default abstract class EasCommand extends Command {
       Log.error(err.message);
     }
     Log.debug(err);
-    throw new Error(baseMessage);
+    const sanitizedError = new Error(baseMessage);
+    sanitizedError.stack = err.stack;
+    throw sanitizedError;
   }
 }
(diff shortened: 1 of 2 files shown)


ide (approved):

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -203,6 +203,8 @@ export default abstract class EasCommand extends Command {
       Log.error(err.message);
     }
     Log.debug(err);
```
wkozyra95, line 205: When you run with EXPO_DEBUG=1 this log will be printed and it will contain a real stacktrace.

I think the problem here is more caused by the fact that we have to independent debug varaible, so may be instead of that fix we could make debug logs to be printed when either EXPO_DEBUG=1 or DEBUG=* is set

wkozyra95 (approved):

packages/eas-cli/src/commandUtils/EasCommand.ts
brentvatne: ah good point @wkozyra95! agreed that it is confusing that we have both env vars for debugging purposes here. in either case i think that it is still not a bad idea to ensure the rethrown error includes the original stack

brentvatne:
/changelog-entry chore Include the original stack in re-thrown errors thrown from EAS CLI commands

PR #1887, opened 2023-06-15
[eas-cli] build:configure add channels to eas.json if using eas updates

quinlanj:
# Why

If someone is using EAS Update, and they run `eas build:configure`, modify their `eas.json` with the proper channels instead of telling them to go back and run `eas update:configure` again. 

# How you can help

Was there a reason why we chose not to do this in the past? Our docs either say to run `build:configure` then modify your `eas.json` for channels manually [here](https://docs.expo.dev/eas-update/getting-started/#configure-your-project) or to go back and run `update:configure` again after `build:configure`. 

# Test Plan

- [ ] tests still pass


quinlanj:
/changelog-entry chore build:configure -- add channels to eas.json if using eas updates

packages/eas-cli/src/commands/build/configure.ts
```diff
@@ -77,6 +79,11 @@ export default class BuildConfigure extends EasCommand {
           await syncIosUpdatesConfigurationAsync(graphqlClient, projectDir, exp, projectId);
         }
       }
+
+      // configure for EAS Update
+      if (exp.updates?.url === getEASUpdateURL(projectId)) {
```
wschurman, line 84: move this into a method called something like `isProjectUsingEASUpdate` to ensure we use the same definition of what it means to be using EAS Update everywhere.

wschurman (approved):

wkozyra95 (approved):

szdziedzic (approved):

PR #1901, opened 2023-06-27
Show build profile on `eas build:run`

keith-kurak:
# Why
I love using `eas build:run` to quickly install a build on an emulator/ simulator, but found I had to go back to my project page to check the build ID, because I was occasionally building non-development builds for various test cases (especially when it comes to APK files), and I could't tell which one was which.

Adding build profile to the selected list item lets you match your build with something easily at-hand in your local dev environment (your **eas.json** file), so it's easy to tell if you're grabbing the latest development build, or a standalone internal test version, etc.

# How

Added build profile to the list of data that shows when you arrow down to a build in the list.

Also slightly compressed the not-available warning based on available information..

# Test Plan

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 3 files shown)


keith-kurak:
/changelog new-feature show build profile when selecting a build for `eas build:run`

keith-kurak:
/changelog-entry new-feature show build profile when selecting a build for eas build:run

szdziedzic (approved):

PR #1906, opened 2023-06-28
[eas-cli] Ignore the existing entitlements when ios is gitignored

byCedric:
# Why

Fixes ENG-8481
Supersedes #1852

# How

This change depends on https://github.com/expo/expo/pull/23165, and uses the new `ignoreExistingNativeFiles` when using CNG/managed.

# Todos

There are still a couple of things to do before we can merge this PR.

- [x] Wait for new `@expo/config-plugins` release with feature from https://github.com/expo/expo/pull/23165
- [ ] How do we communicate this "change" in behavior to our users?
- [ ] How can we let users opt-out of this new behavior?

Optionally:

- [ ] Define in clear terms when EAS CLI considers projects to be managed, or bare/generic

# Test Plan

TBD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - Add styling to SSO auth redirect completion page. ([#1929](https://github.com/expo/eas-cli/pull/1929) by [@wschurman](https://github.com/wschurman))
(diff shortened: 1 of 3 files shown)


brentvatne (approved):

brentvatne:
- bumped deps in https://github.com/expo/eas-cli/commit/50e695b5a3d38ef091f7d18f09609fbd67c6e622

> How do we communicate this "change" in behavior to our users?
> How can we let users opt-out of this new behavior?

i don't think this is necessary - i don't expect that anybody would be depending on this behavior

byCedric:
Since the config plugins are published, I'll update this PR tomorrow and make it merge-ready.

szdziedzic (approved):

PR #1909, opened 2023-06-30
[ENG-9165][eas-cli] print the list of environment variables that we load in EAS CLI when evaluating app config for a build

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-9165/print-the-list-of-environment-variables-that-we-load-in-eas-cli-when

# How

Print list of env vars we load in EAS CLI and which can be used when evaluating app config for a build.

**Note**: This log will be always printed not only when evaluating `app.config.js`.

# Test Plan

Tested manually
Screenshot 2023-06-30 at 11 31 37
Screenshot 2023-06-30 at 11 30 26
Screenshot 2023-06-30 at 11 29 20

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -24,7 +24,7 @@ import {
 } from '../graphql/generated';
 import { BuildQuery } from '../graphql/queries/BuildQuery';
 import { toAppPlatform, toPlatform } from '../graphql/types/AppPlatform';
-import Log from '../log';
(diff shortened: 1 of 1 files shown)


szdziedzic (commented):
@brentvatne 
>Note: This log will be always printed, not only when evaluating app.config.js.

Is this correct?

brentvatne:
> @brentvatne
> 
> > Note: This log will be always printed, not only when evaluating app.config.js.
> 
> Is this correct?

yeah, I think that is fine :)

brentvatne (approved):

Issue #1689, opened 2022-12-10
Error: Runtime version policies are only supported in the managed workflow


huy-lv:
### Summary

Hi, I ran command `eas update --auto` and got this error. Please help :D
```
Error: Runtime version policies are only supported in the managed 
    workflow. In the bare workflow, runtime version needs to be set manually.
```

Runtime version policies are only supported
<img width="568" alt="image" src="https://user-images.githubusercontent.com/7688070/206857926-026e4a6b-a914-474a-a893-94aa9f5d3a24.png">

### What platform(s) does this occur on?

Android

### SDK Version

46

### Environment

 expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 13.0
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 14.19.3 - /usr/local/bin/node
      Yarn: 1.22.19 - ~/.nvm/versions/node/v14.19.3/bin/yarn
      npm: 6.14.17 - ~/.nvm/versions/node/v14.19.3/bin/npm
      Watchman: 2022.11.28.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.11.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    IDEs:
      Android Studio: 2021.2 AI-212.5712.43.2112.8609683
      Xcode: 14.1/14B47b - /usr/bin/xcodebuild
    npmPackages:
      expo: ~46.0.17 => 46.0.17 
      react: 18.0.0 => 18.0.0 
      react-dom: 18.0.0 => 18.0.0 
      react-native: 0.69.6 => 0.69.6 
      react-native-web: ~0.18.7 => 0.18.10 
    npmGlobalPackages:
      eas-cli: 3.0.0
      expo-cli: 6.0.5
    Expo Workflow: bare

### Minimal reproducible example

1. Build internal apk
2. update some code
3. run `eas update --auto`

ethanj20:
[...]

Issue #1899, opened 2023-06-26
[EAS Update] Documentation is not clear enough


ildaneta:
### Summary

The first point is about the inclusion of the `runTimeVersion` value within the `app.config.ts file`, which in the documentation does not specify that you can only include policy if the project is managed, because I was having an error because my project was bare and I was only able to solve it later to do as follows:

` runtimeVersion: "exposdk:47.0.0",`

Described in this issue that is open and where other people are going through the same situation.
https://github.com/expo/eas-cli/issues/1689

The error it gave me was:
```
Runtime version policies are only supported in the managed workflow. In the bare workflow, runtime version needs to be set manually.
```

However, my main doubt in this issue is about how EAS Update works in an OTA sense. I've already used Microsoft's code push with React Native and for me the operation was clear, that is, I published an interface update, ran the command, the publication was available on their website and when I closed my application in production, or generate a build with variant release on my device and opening it again, it already showed the update.

But how does it work with EAS Update? Is it exactly the same?
I've read the documentation a few times, in addition to having already done the whole process of configuring builds, configuring the `eas update` and performing an update, but this part of the update being available was not clear to me.

For example, I published an update with the EAS Update for my production branch, from that moment on, will my users who already have the settings that I made in the `eas update` have these updates automatically? (when closing and opening the phone again)

### Managed or bare?

Bare

### Environment

```
expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 12.5.1
      Shell: 5.8.1 - /bin/zsh
    Binaries:
      Node: 14.20.1 - ~/.nvm/versions/node/v14.20.1/bin/node
      Yarn: 1.22.19 - ~/.nvm/versions/node/v14.20.1/bin/yarn
      npm: 9.7.1 - ~/.nvm/versions/node/v14.20.1/bin/npm
      Watchman: 2022.10.10.00 - /usr/local/bin/watchman
    Managers:
      CocoaPods: 1.11.3 - /usr/local/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
      Android SDK:
        API Levels: 30, 31, 33
        Build Tools: 29.0.2, 30.0.3, 31.0.0, 33.0.0
[...]

PR #1910, opened 2023-06-30
[eas-cli] better runtime version error for bare workflow

quinlanj:
# Why

People on the bare workflow are running `eas update` with a runtimeVersion policy. When they receive our original error message, they don't know what to do. This PR changes the error message to be more actionable.

Fixes https://github.com/expo/eas-cli/issues/1689 and https://github.com/expo/eas-cli/issues/1899

# Test Plan

- [ ] Manually tested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Bump Expo package dependencies. ([#1911](https://github.com/expo/eas-cli/pull/1911) by [@brentvatne](https://github.com/brentvatne))
+- Better bare workflow runtimeVersion error. ([#1910](https://github.com/expo/eas-cli/pull/1910) by [@quinlanj](https://github.com/quinlanj))
 
 ## [3.15.0](https://github.com/expo/eas-cli/releases/tag/v3.15.0) - 2023-06-30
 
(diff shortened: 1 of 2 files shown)


quinlanj:
/changelog-entry chore better bare workflow runtimeVersion error

ildaneta:
@quinlanj When I try to put ```sdkVersion: "47.0.0", "runtimeVersion: "1.0.0"``` in the `app.config.ts` file it doesn't work when I have to test the EAS update with the QR Code generated inside the expo.dev.

Screenshot 2023-07-02 at 12 38 42

It only works when I set:
```
runtimeVersion: "exposdk:47.0.0",
sdkVersion: "47.0.0",
```

https://github.com/expo/eas-cli/assets/21963291/5806cbce-1166-4cdd-b2a4-b6eea2ba9bc9

So I don't know if just changing the error is the solution.

I think you might have to test this situation using a simple project, create a new update and test this scenario.
Also, I think the documentation should consider this case.

szdziedzic (approved):

quinlanj:
@ildaneta we've fixed this issue and have relaxed the runtime constraints on Expo Go. As long as your project has a compatible `expo` npm package installed, Expo Go will load your project. Please note that the same challenges will apply in the bare workflow -- if a native module is not bundled in Expo Go but is referenced in the JS layer of your app, [an error will be thrown](https://docs.expo.dev/bare/using-expo-client/#native-module-detection). 

The changes are expected to be released very soon within the coming weeks for the SDK 49 release. It will officially be available when the next version of Expo Go is released into the respective App Stores.

PR #1911, opened 2023-06-30
[eas-cli] Bump dependencies

brentvatne:
# Why

Routine task but also because we need to use a [new feature](https://github.com/expo/expo/commit/42d6b312c4eed14faf684a720eb5e55ac0f920cb)

# How

Bump versions

# Test Plan

Let CI run


szdziedzic (approved):
Looks good

It would be nice to run a few commands locally as a smoke test, to double check, that everything is fine

PR #1914, opened 2023-06-30
[ENG-9093][eas-cli] remove useClassicUpdates upon configuration

quinlanj:
# Why

Make sure a user isn't using Classic Updates and EAS Update at the same time. If a user has Classic Updates, set `useClassicUpdates` to false if a user runs `eas update:configure` or `eas update`.

# How

on `eas update:configure`: 
```
quins-MacBook-Pro:updates-test-4 quin$ ~/Documents/eas-cli/packages/eas-cli/bin/run update:configure
💡 The following process will configure your project to run EAS Update. These changes only apply to your local project files and you can safely revert them at any time.
✔ Configured updates.useClassicUpdates to "false". You cannot use EAS Update and Classic Updates at the same time.
```

on `eas update`:
```
quins-MacBook-Pro:updates-test-4 quin$ ~/Documents/eas-cli/packages/eas-cli/bin/run update
✔ Configured updates.useClassicUpdates to "false". You cannot be using both EAS Update and Classic Updates.
```

# Test Plan

- [ ] manually tested

### DIFF
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -84,6 +84,17 @@ export function isExpoUpdatesInstalledOrAvailable(
   return isExpoUpdatesInstalled(projectDir);
 }
 
+/**
+ * Projects on SDK 49+ are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
+ */
+export function isDefinitelyUsingClassicUpdates(exp: ExpoConfig): boolean {
+  return !!exp.updates?.useClassicUpdates;
+}
+
+export function isUsingClassicUpdates(exp: ExpoConfig, projectId: string): boolean {
+  return exp.updates?.url === getEASUpdateURL(projectId);
+}
+
 export function isUsingEASUpdate(exp: ExpoConfig, projectId: string): boolean {
   return exp.updates?.url === getEASUpdateURL(projectId);
 }
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -85,7 +85,7 @@ export function isExpoUpdatesInstalledOrAvailable(
 }
 
 /**
- * Projects on SDK 49+ are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
+ * Projects on SDK 49 are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
  */
 export function isDefinitelyUsingClassicUpdates(exp: ExpoConfig): boolean {
   return !!exp.updates?.useClassicUpdates;
--- a/packages/eas-cli/src/project/projectUtils.ts
+++ b/packages/eas-cli/src/project/projectUtils.ts
@@ -91,10 +91,6 @@ export function isDefinitelyUsingClassicUpdates(exp: ExpoConfig): boolean {
   return !!exp.updates?.useClassicUpdates;
 }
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -84,6 +84,17 @@ export function isExpoUpdatesInstalledOrAvailable(
...
+/**
+ * Projects on SDK 49+ are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
+ */
+export function isDefinitelyUsingClassicUpdates(exp: ExpoConfig): boolean {
+  return !!exp.updates?.useClassicUpdates;
+}
+
+export function isUsingClassicUpdates(exp: ExpoConfig, projectId: string): boolean {
+  return exp.updates?.url === getEASUpdateURL(projectId);
+}
```
brentvatne, line 96: it looks like this is the same as `isUsingEASUpdate`

quinlanj: My bad. The function I wanted to expose is `isDefinitelyUsingClassicUpdates`.  I've removed `isUsingClassicUpdates`

quinlanj:
/changelog-entry chore set useClassicUpdates to false on `eas update` and `eas update:configure`

packages/eas-cli/src/project/projectUtils.ts
```diff
@@ -84,6 +84,13 @@ export function isExpoUpdatesInstalledOrAvailable(
   return isExpoUpdatesInstalled(projectDir);
 }
 
+/**
+ * Projects on SDK 49+ are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
```
wschurman, line 88:
```suggestion
 * Projects on SDK 49 are required to set `updates.useClassicUpdates` to `true` to use Classic Updates.
```

packages/eas-cli/src/update/configure.ts
```diff
@@ -96,6 +97,10 @@ async function ensureEASUpdatesIsConfiguredInExpoConfigAsync({
 }): Promise<{ projectChanged: boolean; exp: ExpoConfig }> {
   const modifyConfig: Partial<ExpoConfig> = {};
 
+  if (isDefinitelyUsingClassicUpdates(exp)) {
+    modifyConfig.updates = { useClassicUpdates: false };
```
wschurman, line 101: I think we want to be careful about modifying it for them, as they could forget to change it back in the case that they miss the message that we changed it for them. Maybe I'm overthinking this though.

The main case I can think of people needing both is to publish hotfixes. In this case though, they would probably need to create a sort of hotfix branch and have a diverged config, so I guess modifying it for them is okay?

Fine to leave this as-is I think, just want to make sure we're thinking through the edge cases as far as using both goes.

ide: I put up an alternative PR here that doesn't modify the app config: https://github.com/expo/eas-cli/pull/1915. Regardless of the approach I think it is better to remove "useClassicUpdates" than to set it to false since the field is going away in SDK 50.

quinlanj: The other major point of configuration is `updates.url`. `update:configure` and `update` both set the `updates.url` to be the EAS Update one. In the case of using both Classic and EAS Update, the user will need to maintain a diverged config.

quinlanj:
Closing in favour of https://github.com/expo/eas-cli/pull/1915

PR #1915, opened 2023-06-30
[ENG-9093][eas-cli] Throw if useClassicUpdates is set when the developer is trying to use EAS Update

ide:
Why
---
Fail fast with a clear message if the app config contradicts itself or the developer's behavior.

How
---
Throw if useClassicUpdates is set in the app config and the developer is running an EAS Update-related command or if the app config uses an EAS Update URL.

Test Plan
---
Added a unit test for ensureEASUpdateIsConfiguredAsync

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - Show build profile when selecting a build for eas build:run. ([#1901](https://github.com/expo/eas-cli/pull/1901) by [@keith-kurak](https://github.com/keith-kurak))
-
 - Adds -m alias to --message in update/republish and removed README comment. ([#1905](https://github.com/expo/eas-cli/pull/1905) by [@pusongqi](https://github.com/pusongqi))
 
-
 ### 🐛 Bug fixes
+
+- Ensure useClassicUpdates is not set when using EAS Update commands. ([#1915](https://github.com/expo/eas-cli/pull/1915) by [@ide](https://github.com/ide))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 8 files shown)


quinlanj:
How about a middleground between both our PR's where you tell the user their configuration is incompatible and offer to change it for them (ie) `remove `useClassicUpdates` from your app config?`, since the solution is quite straightforward, it will help the user complete their workflow faster.

khamilowicz (approved):

szdziedzic (commented):
The code change looks good to me

Leaving an assessment of whether it can be done better to the others (https://github.com/expo/eas-cli/pull/1915#issuecomment-1615514584)

brentvatne (approved):
i believe this runs when you run `eas update` (and not just for `eas update:configure`) so this should be good

ide:
@quinlanj I'll look to see if there is a way to remove config options. I'm not sure it is worth it, though, since this affects just one SDK version.

ide:
/changelog-entry [bug-fix] Ensure useClassicUpdates is not set when using EAS Update commands

ide:
Added context-aware prompts that ask the developer if they would like to remove updates.useClassicUpdates. It will not offer to automatically remove the field if the project doesn't use app.json. Confirmed in a test project that the field is actually deleted, not set to false.

Choosing no:
Screenshot 2023-07-04 at 4 57 39 PM

Choosing yes:
Screenshot 2023-07-04 at 4 57 55 PM

With app.config.js instead of app.json:
Screenshot 2023-07-04 at 5 06 26 PM

ide:
/changelog-entry bug-fix Ensure useClassicUpdates is not set when using EAS Update commands

quinlanj (approved):
thank you!

PR #1921, opened 2023-07-03
[eas-cli] Handle free user limit error messages

sundeeppeswani:
# Why

Adds support for a new class of server-side errors related to build limits for free accounts.

# How

This uses existing support for server-side errors to render the error in the console: 
Screenshot 2023-06-25 at 10 26 24 AM

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -42,6 +42,8 @@ import {
   EasBuildFreeTierDisabledAndroidError,
   EasBuildFreeTierDisabledError,
   EasBuildFreeTierDisabledIOSError,
+  EasBuildFreeTierIosLimitExceededError,
+  EasBuildFreeTierLimitExceededError,
   EasBuildProjectArchiveUploadError,
   EasBuildResourceClassNotAvailableInFreeTierError,
   EasBuildTooManyPendingBuildsError,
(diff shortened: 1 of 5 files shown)


brentvatne:
> We've encountered a transient error. Try again shortly.
> Build request failed. Make sure ...
> Request ID: ...

I think it'd be better if we don't print any of this for this particular error message

szdziedzic (commented):
I think that changing https://github.com/expo/eas-cli/blob/f3577deac4b8c6aa644d4f657588812bc344ef99/packages/eas-cli/src/graphql/client.ts#L8-L13 to 
```ts
if (error) {
    if (
      error.graphQLErrors.some(
        e =>
          e?.extensions?.isTransient &&
          ![
            'EAS_BUILD_FREE_TIER_LIMIT_EXCEEDED_ERROR',
            'EAS_BUILD_FREE_TIER_IOS_LIMIT_EXCEEDED_ERROR',
          ].includes(e?.extensions?.errorCode as string)
      )
    ) {
      Log.error(`We've encountered a transient error. Try again shortly.`);
    }
    throw error;
  }
```
should do the job

packages/eas-cli/src/build/build.ts
```diff
@@ -183,6 +185,8 @@ const SERVER_SIDE_DEFINED_ERRORS: Record<string, typeof EasCommandError> = {
   EAS_BUILD_FREE_TIER_DISABLED: EasBuildFreeTierDisabledError,
   EAS_BUILD_FREE_TIER_DISABLED_IOS: EasBuildFreeTierDisabledIOSError,
   EAS_BUILD_FREE_TIER_DISABLED_ANDROID: EasBuildFreeTierDisabledAndroidError,
+  EAS_BUILD_FREE_TIER_LIMIT_EXCEEDED: EasBuildFreeTierLimitExceededError,
+  EAS_BUILD_FREE_TIER_IOS_LIMIT_EXCEEDED: EasBuildFreeTierIosLimitExceededError,
```
szdziedzic, line 189: Keys of this object must match error codes, so they can be handled as a server-side defined error

Applying the https://github.com/expo/universe/pull/10819/files#r1251610405 suggestion should make it work.

szdziedzic (approved):

PR #1924, opened 2023-07-07
[eas-cli] [ENG-8957] Handle multiple GraphQLErrors in CombinedError

radoslawkrzemien:
CombinedError technically can have more than 1 GraphQLError inside it, even though for now we only use it with 1. In such a case it would be good for the eas-cli to be able to handle and properly log/display all the multiple error messages

See: https://linear.app/expo/issue/ENG-8957/improve-error-handling-in-eas-cli-to-support-case-when-combined-error

# Why

CombinedError technically can have more than 1 GraphQLError inside it, even though for now we only use it with 1. In such a case it would be good for the eas-cli to be able to handle and properly log/display all the multiple error messages

# How

Use all error messages from GraphQLErrors in the final message logged/displayed

# Test Plan

Tested manually as well as added automated tests

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -4,6 +4,7 @@ import assert from 'assert';
 import chalk from 'chalk';
 import cliProgress from 'cli-progress';
 import fs from 'fs-extra';
+import { GraphQLError } from 'graphql/error';
 import nullthrows from 'nullthrows';
 
 import { BuildEvent } from '../analytics/AnalyticsManager';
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/build/build.ts
```diff
@@ -206,15 +207,19 @@ export function handleBuildRequestError(error: any, platform: Platform): never {
...
   } else if (error?.graphQLErrors) {
-    const graphQLError = error.graphQLErrors[0];
-    const requestIdLine = graphQLError?.extensions?.requestId
-      ? `\nRequest ID: ${graphQLError.extensions.requestId}`
-      : '';
-    const originalErrorMessage = graphQLError?.message
-      ? `\nError message: ${graphQLError.message}`
-      : '';
+    const errorMessage = error.graphQLErrors
+      .map((graphQLError: GraphQLError) => {
```
szdziedzic, line 211: Are we sure that items on this list will always be of the `GraphQLError` type?

Previously they were treated as `any` type and access to their fields was conditional. After your changes, it is no longer conditional and I wonder if this is 100% safe.

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -193,10 +194,15 @@ export default abstract class EasCommand extends Command {
     if (err instanceof EasCommandError) {
       Log.error(err.message);
     } else if (err instanceof CombinedError && err?.graphQLErrors) {
-      let cleanMessage = err.message.replace('[GraphQL] ', '');
-      cleanMessage += err?.graphQLErrors?.[0]?.extensions?.requestId
-        ? `\nRequest ID: ${err.graphQLErrors[0].extensions.requestId}`
-        : '';
+      const cleanMessage = err?.graphQLErrors
+        .map((graphQLError: GraphQLError) => {
+          const messageLine = graphQLError.message.replace('[GraphQL] ', '');
```
szdziedzic, line 199: Previously we removed the `[GraphQL]` part from the `CombinedError` message, not from a single one `GraphQLError`. Did you verify that a single `GraphQLError` can contain this part and if this logic makes sense in this context?

radoslawkrzemien:
Sorry Szymon, had not refreshed site and requested review before I addressed the comments

packages/eas-cli/src/commandUtils/EasCommand.ts
radoslawkrzemien: From what I saw, the `err.message` property was a concatenation of `err.graphQLErrors.[*].message` properties, therefore I assumed it those messages on graphQLErrors that have this `[GraphQL]` label in the beginning. I may be wrong, but even so would there be a problem with replacing it if it's not there? The `replace()` function wouldn't do anything in that case

packages/eas-cli/src/build/build.ts
radoslawkrzemien: From `urql` site:
```The graphQLErrors property may be an array that contains [normalized GraphQLErrors as they were received in the errors array from a GraphQL API.](https://graphql.org/graphql-js/error/)```
Hence I believe they are always GraphQLErrors

szdziedzic (approved):

radoslawkrzemien:
/changelog-entry bug-fix Handle multiple GraphQLErrors when receiving a CombinedError

PR #1925, opened 2023-07-09
[eas-cli] fix runtime logs

quinlanj:
# Why
There is a bug when printing out runtime versions:

When it is an object:
```
✔ Configured runtimeVersion for Android and iOS with "[object Object]"
```

When it is a string
```
✔ Configured runtimeVersion for Android and iOS with "1.0.0"
```

# How

This PR checks if a runtimeVersion is an object or string, then formats it accordingly.

When it is an object:
```
✔ Configured runtimeVersion for Android and iOS with "{"policy":"sdkVersion"}"
```

When it is a string:
```
✔ Configured runtimeVersion for Android and iOS with "1.0.0"
```

# Test Plan

- [ ] manually tested `eas update:configure` on managed and bare projects

### DIFF
--- a/packages/eas-cli/src/update/configure.ts
+++ b/packages/eas-cli/src/update/configure.ts
@@ -155,7 +155,9 @@ async function ensureEASUpdatesIsConfiguredInExpoConfigAsync({
   }
 }
 
-function printRuntimeVersion(runtimeVersion: NonNullable<ExpoConfig['runtimeVersion']>): string {
+function serializeRuntimeVersionToString(
+  runtimeVersion: NonNullable<ExpoConfig['runtimeVersion']>
+): string {
   if (typeof runtimeVersion === 'object') {
     return JSON.stringify(runtimeVersion);
   } else {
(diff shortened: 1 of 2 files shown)


quinlanj:
/changelog-entry chore fix runtime version print logs

douglowder (approved):

szdziedzic (approved):

packages/eas-cli/src/update/configure.ts
```diff
@@ -155,6 +155,14 @@ async function ensureEASUpdatesIsConfiguredInExpoConfigAsync({
   }
 }
 
+function printRuntimeVersion(runtimeVersion: NonNullable<ExpoConfig['runtimeVersion']>): string {
```
wschurman, line 158:
```suggestion
function serializeRuntimeVersionToString(runtimeVersion: NonNullable<ExpoConfig['runtimeVersion']>): string {
```

wschurman (approved):

PR #1931, opened 2023-07-13
[ENG-8956][Change] Change tsconfig.json options for easier development

khamilowicz:
# Why

https://exponent-internal.slack.com/archives/C017N0N99RA/p1686787194341989

# How

Changed `noUnusedLocals` and `noUnusedParameters` to false for easier local development and experimenting.
These checks do not run during local development. Instead, eslint catches these cases on github.

# Test Plan

1. Comment out code, import unused method, define unused variable
2. Run `yarn eas` command that would normally trigger error
3. Command should pass
4. Run `yarn lint`
5. It should throw error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Limit project file upload size to 2GB. ([#1928](https://github.com/expo/eas-cli/pull/1928) by [@khamilowicz](https://github.com/khamilowicz))
+
 ## [3.15.1](https://github.com/expo/eas-cli/releases/tag/v3.15.1) - 2023-07-11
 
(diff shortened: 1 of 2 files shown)


khamilowicz:
/changelog chore Skip `noUnusedLocals` and `noUnusedParameters` during local development

khamilowicz:
/changelog-entry chore Skip noUnusedLocals and noUnusedParameters during local development

szdziedzic (commented):
I would create 2 separate `tsconfigs` (1 for local dev and 1 for ci) instead of just making our main config less strict. The local config would be less strict and CI one would be more strict. So build command on CI fails when using unused imports, but the local one doesn't.

I think that in such case it would be reasonable to have 2 commands - `yarn build` and `yarn build-ci`

We probably want to apply the same `tsconfig` rules to the `eas-json` package as well, not only `eas-cli`

PR #1933, opened 2023-07-17
[ENG-8956][eas-build] Change tsconfig.json options for easier development

khamilowicz:
# Why

https://exponent-internal.slack.com/archives/C017N0N99RA/p1686787194341989

# How

Changed `noUnusedLocals` and `noUnusedParameters` to false for easier local development and experimenting.
These checks do not run during local development. Instead, eslint catches these cases on github.

# Test Plan

1. Comment out code, import unused method, define unused variable
2. Run `yarn eas` command that would normally trigger error
3. Command should pass
4. Run `yarn lint`
5. It should throw error


khamilowicz:
/changelog-entry chore Skip noUnusedLocals and noUnusedParameters during local development

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Limit project file upload size to 2GB. ([#1928](https://github.com/expo/eas-cli/pull/1928) by [@khamilowicz](https://github.com/khamilowicz))
+- Skip noUnusedLocals and noUnusedParameters during local development. ([#1933](https://github.com/expo/eas-cli/pull/1933) by [@khamilowicz](https://github.com/khamilowicz))
```
szdziedzic, line 20: I think that there's no need to add this to the changelog. Changelog should inform users about what changed in the given version. This is just for local dev.

packages/eas-json/tsconfig.build.json
```diff
@@ -1,4 +1,8 @@
 {
   "extends": "./tsconfig",
+  "compilerOptions": {
+    "noUnusedLocals": true,
+    "noUnusedParameters": true
+  },
```
szdziedzic, line 6: Most people use `yarn watch` and `yarn build` commands for local development. These commands use tsconfig.build.json, so, unfortunately, we don't resolve the issue here.
Screenshot 2023-07-17 at 16 22 04

What I meant was to create 2 different sets of config `local` and `production/ci` with different `tsconfig` and `tsconfig.build`

szdziedzic (approved):

wschurman:
One piece of feedback on this: While I understand the reasoning behind this, I find that finding out about unused locals after a CI run slows down my development a bit as it requires a round trip to the PR/CI and more thrash/emails for reviewers. Could just be me though. FWIW I think we've settled on not having separate tsconfigs in all of our other repos for this reason.

PR #1934, opened 2023-07-18
[eas-cli] fix incorrect handling of valid `inProgress` release status

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C017N0N99RA/p1689605257077099

Previously the `capitalize` function made the first letter upper case and other letters of the string lower case. So if we run `capitalize('inProgress')` we will get `'Inprogress'`.

This is the code that checks if the release status is supported
https://github.com/expo/eas-cli/blob/57dd372957d5f18da232c3efca5716a691e1f903/packages/eas-cli/src/submit/android/AndroidSubmitCommand.ts#L97-L112
https://github.com/expo/eas-cli/blob/57dd372957d5f18da232c3efca5716a691e1f903/packages/eas-cli/src/graphql/generated.ts#L4618-L4623

As you can see `'Inprogress'` is not a key of the `SubmissionAndroidReleaseStatus` enum, therefore the error will be thrown, which is incorrect behavior because it is a valid option.

# How

Change the `capitalize` function to `capitalizeFirstLetter` and make it work like `capitalizeFirstLetter('inProgress') => 'InProgress'`.

# Test Plan

Tested manually and added unit tests.

## Before
Screenshot 2023-07-18 at 10 06 38

## After
Screenshot 2023-07-18 at 10 07 15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fix incorrect handling of valid `inProgress` Android submission release status. ([#1934](https://github.com/expo/eas-cli/pull/1934) by [@szdziedzic](https://github.com/szdziedzic))
+
 ### 🧹 Chores
 
 - Limit project file upload size to 2GB. ([#1928](https://github.com/expo/eas-cli/pull/1928) by [@khamilowicz](https://github.com/khamilowicz))
--- a/packages/eas-cli/src/submit/android/AndroidSubmitCommand.ts
+++ b/packages/eas-cli/src/submit/android/AndroidSubmitCommand.ts
@@ -12,7 +12,7 @@ import {
   AmbiguousApplicationIdError,
   getApplicationIdAsync,
 } from '../../project/android/applicationId';
-import capitalize from '../../utils/expodash/capitalize';
+import capitalizeFirstLetter from '../../utils/expodash/capitalize';
 import { ArchiveSource } from '../ArchiveSource';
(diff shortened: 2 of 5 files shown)


szdziedzic:
/changelog-entry bug-fix Fix incorrect handling of valid `inProgress` Android submission release status

szdziedzic:
@kbrandwijk We are using Joi to validate our `eas.json` schema so any option other than exactly one of `completed`, `draft`, `halted, or `inProgress` will result in an error. So this should be the case.
Screenshot 2023-07-18 at 14 59 59
Screenshot 2023-07-18 at 15 00 06

https://github.com/expo/eas-cli/blob/57dd372957d5f18da232c3efca5716a691e1f903/packages/eas-json/src/submit/schema.ts#L5-L15

kbrandwijk:
> @kbrandwijk We use Joi to validate our `eas.json` schema so any option other than exactly one of `completed`, `draft`, `halted`, or `inProgress` will result in an error. So this shouldn't be the case. Screenshot 2023-07-18 at 14 59 59 Screenshot 2023-07-18 at 15 00 06
> 
> https://github.com/expo/eas-cli/blob/57dd372957d5f18da232c3efca5716a691e1f903/packages/eas-json/src/submit/schema.ts#L5-L15

Ah, that is good to know. That should be a non-issue then.

radoslawkrzemien (approved):
LGTM 🚀 
Nitpick: Would be nice to have a full test (not just the capitalizing function) in https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/submit/android/__tests__/AndroidSubmitCommand-test.ts

PR #1939, opened 2023-07-19
fix(eas-cli): implement variadic `--platform` flag when creating EAS updates

byCedric:
# Why

Fixes ENG-9308

> This PR assumes that https://github.com/expo/expo/pull/23621 is going to be merged and released.

# How

- Added warning when users create EAS update bundles with `expo.web.bundler = "metro"`
- Added version detection to use the new variadic `--platform` flag
- Added explicit `expo export --platform android --platform ios` flags when creating EAS update bundles

# Test Plan

- `$ yarn create expo ./test-updates -t tabs@49` (comes with Metro web)
- `$ cd ./test-updates`
- `$ eas update`
  → should create exports for `android` and `ios`, **_not_** `web`

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -25,7 +25,11 @@ import { promptAsync } from '../prompts';
 import { getBranchNameFromChannelNameAsync } from '../update/getBranchNameFromChannelNameAsync';
 import { formatUpdateMessage, truncateString as truncateUpdateMessage } from '../update/utils';
 import { PresignedPost, uploadWithPresignedPostWithRetryAsync } from '../uploads';
-import { expoCommandAsync, shouldUseVersionedExpoCLI } from '../utils/expoCli';
+import {
+  expoCommandAsync,
+  shouldUseVersionedExpoCLI,
+  shouldUseVersionedExpoCLIWithExplicitPlatforms,
+} from '../utils/expoCli';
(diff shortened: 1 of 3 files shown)


hasanaktas:
Any progress? in updates i have to cancel web and send ios and android at the same time. this pr seems to solve it.

packages/eas-cli/src/project/publish.ts
```diff
@@ -193,35 +197,60 @@ export async function buildBundlesAsync({
...
+    return await expoCommandAsync(projectDir, [
       'export',
       '--output-dir',
       inputDir,
+      '--experimental-bundle',
+      '--non-interactive',
       '--dump-sourcemap',
       '--dump-assetmap',
-      ...platformArgs,
+      ...platformFlag,
```
szdziedzic, line 210: Don't we need
```suggestion
      '--platform',
      ...platformFlag,
```
here?

byCedric: Yep, you are right. I just made it a bit more verbose, where all `platformFlag` instances are now `--platform=${platformFlag}`. So it's better spottable where we need to set it like this.

byCedric: I can change it back, but there were 2 instances where we needed to do this.

szdziedzic (approved):
Looks good 🚀 Thanks for working on it

PR #1944, opened 2023-07-20
[eas-cli] branch mapping utility functions

quinlanj:
# Why

This PR:
- adds branch mapping utility functions and updates their definitions to be the same as www. The previous definitions were a rough approximation of what was declared in www and were not correct.
- does not change the behaviour of our commands does not change. However, some assertions were added in order to get our cli code to play well with the new branch mapping definitions.
- adds rollout utility functions for use later

# Test Plan

- [ ] new tests pass
- [ ] current tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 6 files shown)


quinlanj:
/changelog-entry chore  added branch mapping utility functions

wschurman (approved):

PR #1949, opened 2023-07-25
[eas-cli][1/2] unify channel graphql query types

quinlanj:
# Why

For graphql queries that have the same fields, unify them with fragments if possible. There is one exception:

```
  query FooQuery($appId: String!, $offset: Int!, $limit: Int!) {
    app {
      byId(appId: $appId) {
        id
        updateChannels(offset: $offset, limit: $limit) {
          updateBranches(offset: 0, limit: 5) { <---- We plan to parameterize these arguments instead of keeping it constant
            id
          }
        }
      }
    }
  }
```

Unfortunately, graphql does not allow for [parameterized fragments ](https://github.com/graphql/graphql-spec/issues/204) so I create an intersection type to unify the channel fields: `UpdateChannelObject = ViewUpdateChannelsOnAppObject & UpdateChannelByNameObject;`

# How

Create fragments and intersection types 

# Test Plan

- [ ] current tests pass

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -2573,18 +2573,6 @@
         "description": "Auth configuration data for an SSO account.",
         "fields": [
           {
-            "name": "authEndpoint",
-            "description": null,
-            "args": [],
-            "type": {
-              "kind": "SCALAR",
-              "name": "String",
-              "ofType": null
-            },
-            "isDeprecated": false,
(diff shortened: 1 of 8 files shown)


quinlanj:
/changelog-entry chore unify channel graphql query types

khamilowicz (approved):

szdziedzic (approved):

wschurman (approved):

PR #1953, opened 2023-07-26
[eas-cli] fetch entire relay compliant dataset

quinlanj:
# Why

A utility function to fetch a complete dataset from a Relay compliant endpoint in a batched way. See a concrete usage of it in `getChannelsDatasetAsync`.

# Test Plan

- [ ] new tests pass

### DIFF
--- a/packages/eas-cli/src/channel/queries.ts
+++ b/packages/eas-cli/src/channel/queries.ts
@@ -8,6 +8,8 @@ import { withErrorHandlingAsync } from '../graphql/client';
 import {
   CreateUpdateChannelOnAppMutation,
   CreateUpdateChannelOnAppMutationVariables,
+  PageInfo,
+  UpdateChannelBasicInfoFragment,
   ViewBranchesOnUpdateChannelQueryVariables,
   ViewUpdateChannelsOnAppQueryVariables,
 } from '../graphql/generated';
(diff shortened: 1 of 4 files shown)


wschurman (commented):
Did you explore using the `relayStylePagination` utility? I'm not completely clear on what it does but it might be worth exploring if you haven't already. https://www.apollographql.com/docs/react/pagination/cursor-based/#relay-style-cursor-pagination. https://github.com/apollographql/apollo-client/blob/50004385b3a3f99db37390ff723239b43b375ca9/src/utilities/policies/pagination.ts#L95

quinlanj:
@wschurman yes I did. We use an `urql` client instead of Apollo, but they also have support for [relay pagination](https://formidable.com/open-source/urql/docs/graphcache/local-resolvers/#relay-pagination). The reason why I didn't convert to using the helper is because:
- I'd have to convert our urql client from the default[ Document caching](https://formidable.com/open-source/urql/docs/basics/document-caching/) policy to their normalized Graphcache one. The default Document caching is a conservative policy that nukes all Foo query items cached if there is a Foo mutation. We have graphql mutations in our codebase which will require additional logic for to ensure our cached queries are still correct under the new graphcache scheme
- IIUC, their relay helper's value proposition is that it merges data when you fetch another page. This is more compelling in a website setting where we want to track the dataset in one variable and rely on a context provider to trigger the rerenders and update components when the dataset changes.
- In a cli setting, there are less moving parts (no react components, no expensive renders, etc). I think it'll be easier to write the helper/utility functions ourselves to merge the paginated data.

packages/eas-cli/src/channel/queries.ts
```diff
@@ -264,3 +268,46 @@ export async function ensureChannelExistsAsync(
...
+      assetSpinner.text = `Fetched ${totalNodesFetched} channels`;
+    }
+  };
+  const dataset = await getPaginatedDatasetAsync({
+    queryAsync,
+    afterEachQuery,
+    filterPredicate,
+    batchSize,
+  });
+  assetSpinner.succeed(`Fetched all channels`);
```
wschurman, line 311: this message could be incorrect if `maxNodesFetched` was less than the total, but I think it likely isn't an issue

wschurman (approved):

PR #1954, opened 2023-07-26
[ENG-9422][eas-cli] Use .nvmrc to set node version

khamilowicz:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1689882938014309

# How

Read .nvmrc content and set `node` value in `buildProfile`.

# Test Plan

1. Create valid .nvmrc file
2. Build project
3. Check in build logs if non-default node version is installed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- .nvmrc support for setting node version. ([#1954](https://github.com/expo/eas-cli/pull/1954) by [@khamilowicz](https://github.com/khamilowicz))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
--- a/packages/eas-cli/src/utils/__tests__/profiles-test.ts
+++ b/packages/eas-cli/src/utils/__tests__/profiles-test.ts
@@ -1,5 +1,8 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile, EasJsonAccessor, EasJsonUtils, errors } from '@expo/eas-json';
+import fs from 'fs-extra';
+import { vol } from 'memfs';
+import os from 'os';
 
 import Log from '../../log';
 import { selectAsync } from '../../prompts';
@@ -9,6 +12,7 @@ import {
   maybePrintBuildProfileDeprecationWarningsAsync,
 } from '../profiles';
 
+jest.mock('fs');
 jest.mock('../../prompts');
 jest.mock('@expo/eas-json', () => {
   const actual = jest.requireActual('@expo/eas-json');
@@ -32,6 +36,7 @@ const getBuildProfileDeprecationWarningsAsync = jest.spyOn(
 );
 const newLineSpy = jest.spyOn(Log, 'newLine');
 const warnSpy = jest.spyOn(Log, 'warn');
+const projectDir = '/app';
 
 describe(getProfilesAsync, () => {
   afterEach(() => {
@@ -51,6 +56,7 @@ describe(getProfilesAsync, () => {
       platforms: [Platform.ANDROID, Platform.IOS],
       profileName: undefined,
       type: 'build',
+      projectDir,
     });
 
     expect(result[0].profileName).toBe('production');
@@ -71,6 +77,7 @@ describe(getProfilesAsync, () => {
         platforms: [Platform.ANDROID],
         profileName: undefined,
         type: 'build',
+        projectDir,
       })
     ).rejects.toThrowError(errors.MissingProfileError);
   });
@@ -86,6 +93,7 @@ describe(getProfilesAsync, () => {
       platforms: [Platform.ANDROID, Platform.IOS],
       profileName: 'custom-profile',
       type: 'build',
+      projectDir,
     });
 
     expect(result[0].profileName).toBe('custom-profile');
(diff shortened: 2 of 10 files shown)


khamilowicz:
/change-log feature `.nvmrc` support for setting node version

khamilowicz:
/changelog-entry new-feature .nvmrc support for setting node version

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- .nvmrc support for setting node version. ([#1954](https://github.com/expo/eas-cli/pull/1954) by [@khamilowicz](https://github.com/khamilowicz))
```
szdziedzic, line 11:
```suggestion
- `.nvmrc` support for setting node version. ([#1954](https://github.com/expo/eas-cli/pull/1954) by [@khamilowicz](https://github.com/khamilowicz))
```

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -37,6 +42,11 @@ export async function getProfilesAsync<T extends ProfileType>({
       type,
       profileName,
     });
+
+    if (isBuildProfile(profile, type)) {
+      await setNodeVersionFromFileAsync(projectDir, profile);
+    }
```
szdziedzic, line 48: Maybe it's better to move this logic to https://github.com/expo/eas-cli/blob/fd99365655872dc6d7ce33e762c9bd3a99e0696a/packages/eas-cli/src/utils/profiles.ts#L61-L71

We already have some conditional logic only for building profiles there

(not a strong opinion just a suggestion)

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,6 +57,45 @@ export async function getProfilesAsync<T extends ProfileType>({
...
+  projectDir: string,
+  profile: BuildProfile<Platform>
+): Promise<void> {
+  const nodeVersion = await getNodeVersionFromFileAsync(projectDir);
+  if (nodeVersion) {
+    if (profile.node && profile.node !== nodeVersion) {
+      throw new Error(
+        `Node version defined in eas.json (${profile.node}) does not match the version defined in .nvmrc (${nodeVersion})`
+      );
+    }
```
szdziedzic, line 74: I don't think that we want to throw errors here.

I believe that the logic here should be as follows:
1. If the `node` version is specified in `eas.json` just use this value (even though somebody has a `.nvmrc` file)
2. If the `node` version is not specified in `eas.json` and somebody has a `.nvmrc` file use version from the `.nvmrc` file as an override (be vocal about it)

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,6 +57,45 @@ export async function getProfilesAsync<T extends ProfileType>({
...
+  try {
+    const nodeVersion = (await fs.readFile(nvmrcPath, 'utf8')).toString().trim();
+    if (!semver.valid(nodeVersion)) {
+      throw new Error(`Invalid node version in .nvmrc: ${nodeVersion}`);
+    }
+    return nodeVersion;
+  } catch (error: any) {
+    if (error.code !== 'ENOENT') {
+      Log.warn(`Failed to read node version from ${nvmrcPath}: ${error.message}`);
+    }
```
szdziedzic, line 94: I think it's better to check if a file exists before accessing it using `fs.exists` function

szdziedzic (commented):
Generally looks good, few suggestions/comments:

packages/eas-cli/src/utils/__tests__/profiles-test.ts
```diff
@@ -98,6 +104,53 @@ describe(getProfilesAsync, () => {
...
+        type: 'build',
+        projectDir,
+      });
+
+      expect(result[0].profile.node).toBe(nodeVersion);
+    });
+
+    describe('with .nvmrc', () => {
+      beforeEach(async () => {
+        await fs.writeFile(path.join(projectDir, '.nvmrc'), nodeVersion);
```
szdziedzic, line 128: Is it using some kind of mock filesystem or temporary dir?

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,16 +54,49 @@ export async function getProfilesAsync<T extends ProfileType>({
...
+    profile.node = nodeVersion;
+  }
+}
+
+async function getNodeVersionFromFileAsync(projectDir: string): Promise<string | undefined> {
+  const nvmrcPath = path.join(projectDir, '.nvmrc');
+  // eslint-disable-next-line node/no-sync
+  if (!fs.existsSync(nvmrcPath)) {
+    return;
+  }
```
szdziedzic, line 79:
```suggestion
  if (!(await fs.exists(nvmrcPath))) {
    return;
  }
```

don't use sync 😄

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,16 +54,49 @@ export async function getProfilesAsync<T extends ProfileType>({
...
+
+  const nodeVersion = (await fs.readFile(nvmrcPath, 'utf8')).toString().trim();
```
szdziedzic, line 81:
```suggestion
  let nodeVersion: string;
  try {
    nodeVersion = (await fs.readFile(nvmrcPath, 'utf8')).toString().trim();
  } catch {
    return undefined;
  }
```

packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,16 +54,49 @@ export async function getProfilesAsync<T extends ProfileType>({
...
+  if (!semver.valid(nodeVersion)) {
+    throw new Error(`Invalid node version in .nvmrc: ${nodeVersion}`);
+  }
+  return nodeVersion;
+}
+
 async function readProfileAsync<T extends ProfileType>({
```
szdziedzic, line 88:
```suggestion
async function readProfileWithOverridesAsync<T extends ProfileType>({
```

I think this name better describes what this function is actually doing

packages/eas-cli/src/utils/__tests__/profiles-test.ts
khamilowicz: it writes the file to fixtures directory

szdziedzic: Ok, so I think it would be good to mock `fs`. We already have a mock for it using `fs` from `memfs` in this repo.

Here's how you can do this using this mock (just do `jest.mock('fs')`) and `vol` from the `memfs` package https://github.com/expo/eas-cli/blob/fd99365655872dc6d7ce33e762c9bd3a99e0696a/packages/eas-cli/src/project/ios/__tests__/bundleIdentifier-test.ts

szdziedzic (approved):
🚀

PR #1957, opened 2023-07-27
[eas-cli] more branch mapping utility

quinlanj:
# Why

This PR adds support for:
- Rollout convenience objects
- more branch mapping utilities
- printing functions for rollout state

# Test Plan

- [x] new tests pass

### DIFF
--- a/packages/eas-cli/src/rollout/branch-mapping.ts
+++ b/packages/eas-cli/src/rollout/branch-mapping.ts
@@ -1,0 +1,265 @@
+import assert from 'assert';
+
+import {
+  BranchMapping,
+  BranchMappingNode,
+  assertNodeObject,
+  assertNumber,
+  assertStatement,
+  assertString,
+  getBranchMapping,
+  getNodesFromStatement,
+  isAlwaysTrue,
+  isAndStatement,
+  isStatement,
+} from '../channel/branch-mapping';
+import { getUpdateBranch } from '../channel/utils';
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/rollout/branch-mapping.ts
```diff
@@ -0,0 +1,265 @@
...
+  return 'runtimeVersion' in rollout;
+}
+
+export function isConstrainedRollout(rollout: Rollout): rollout is ConstrainedRollout {
+  return isConstrainedRolloutInfo(rollout);
+}
+
+export function getRolloutInfo(basicChannelInfo: UpdateChannelBasicInfoFragment): RolloutInfo {
+  const branchMapping = getBranchMapping(basicChannelInfo.branchMapping);
+  assertRollout(branchMapping);
```
wschurman, line 55: it might be better to either a) combine these (so that `getBranchMapping` never returns an invalid branch mapping) or b) rename `getBranchMapping` to `parseBranchMappingJSON` and have it return an `object` rather than a `BranchMapping`.

wschurman (approved):

packages/eas-cli/src/rollout/branch-mapping.ts
quinlanj: so the issue here is that all rollouts are branch mappings but not all branch mappings are rollouts. `assertRollout` ensures the branch mapping has the rolloutToken among other things, as opposed to the standard 'alwaysTrue' `channel->branch` mapping.

PR #1958, opened 2023-07-28
[eas-cli] select existing rollouts and channels

quinlanj:
# Why

This PR does the following:
- Defines our `Action` model in the `eas-update` folder.  Every action implements the `EASUpdateAction` interface. This  allows us to build our commands with these modular Actions.
- Makes utility classes to select existing channels and rollouts. `SelectChannel` and `SelectRollout` are a concrete example of these Actions

# Test Plan

- [ ] new tests pass


quinlanj:
/changelog-entry chore utility classes to select existing rollouts and channels

szdziedzic (approved):

PR #1968, opened 2023-08-01
[ENG-8616][eas-cli] change update default runtime policy

quinlanj:
# Why

Whenever someone tries to setup EAS Update, we used to default them to the `sdkVersion` runtime policy because Expo Go for 48 and below would not load any other policy. However, we've changed this in 49 and we can now switch folks over to the more expressive `appVersion`. 

# How

Detect if someone is setting up EAS Update on sdk 49.0.0 or higher. If they are, default them to `appVersion` runtime policy.

# Test Plan

- [x] new unit tests pass
- [x] manually tested


quinlanj:
/changelog-entry chore Change default runtime version policy for EAS Update to appVersion

wschurman (approved):

PR #1971, opened 2023-08-02
[eas-cli] fix update:republish group id

quinlanj:
# Why

`update:republish` is broken, and as of https://github.com/expo/eas-cli/commit/8d57828f7e40c0be2ef6de74ee92857a3ceb9db9 it throws. I added some assertions to `republishAsync` that uncovered the bug where we were confusing `update.id` with its group ID (`update.group`).

# How

Use the correct group and rename some variables to make code clearer

# Test Plan

- [ ] manually tested `update:republish` with channel, branch and group flags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


quinlanj:
/changelog-entry bug-fix Pass correct group into `update:republish`

wschurman (approved):

PR #1973, opened 2023-08-02
[eas-cli] Fix republishing with code signing

wschurman:
# Why

This fixes the following issue:
- A republished update will have the same manifest info group but different generated metadata from the server (different groupId and potentially different branch name).
- Previously, this was using the same signature, but for the different blob. Therefore it would fail.

Closes ENG-9596.

# How

This fixes it by requiring providing the private key to republish when code signing is involved.

# Test Plan

1. Set up code signing: https://docs.expo.dev/eas-update/code-signing/
1. `eas update`, see it is signed
1. `eas update:republish`, see it signs correctly now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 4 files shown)


wschurman:
/changelog-entry bug-fix Fix republishing with code signing

quinlanj (approved):
thank you! lgtm 🙏

Issue #1975, opened 2023-08-03
Invalid node version in .nvmrc: vXX


IjzerenHein:
### Summary

As of `eas-cli@3.18.0`, the following command has started failing in our environment:

```
eas build -p ios --profile test --non-interactive --no-wait
```

With the following error:

```
Invalid node version in .nvmrc: v16
```

Contents of our `.nvmrc` file:

```
v16
```

### Managed or bare?

Managed

### Environment

```
expo-env-info 1.0.5 environment info:
    System:
      OS: macOS 13.4.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 16.20.0 - ~/.nvm/versions/node/v16.20.0/bin/node
      Yarn: 1.22.19 - ~/.nvm/versions/node/v16.20.0/bin/yarn
      npm: 8.19.4 - ~/.nvm/versions/node/v16.20.0/bin/npm
      Watchman: 2023.05.22.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.12.1 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 22.4, iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4
    IDEs:
      Android Studio: 2022.2 AI-222.4459.24.2221.10121639
      Xcode: 14.3.1/14E300c - /usr/bin/xcodebuild
    npmPackages:
      expo: ^48.0.4 => 48.0.4
      react: 18.2.0 => 18.2.0
      react-dom: 18.2.0 => 18.2.0
      react-native: 0.71.3 => 0.71.3
      react-native-web: ~0.18.9 => 0.18.10
```

### Error output

Invalid node version in .nvmrc: v16

### Reproducible demo or steps to reproduce from a blank project

- Add a `.nvmrc` file to your project with the following contents:

```
v16
```

- Make sure eas.config does not contain a node version number

- Run EAS build:

```
eas build -p ios
```

IjzerenHein:
It seems like this is the offending line: https://github.com/expo/eas-cli/blob/47968a1b8b2cf80e19bd9fb695a05e851d7d6326/packages/eas-cli/src/utils/profiles.ts#L86

semver is used to validate the version, however `.nvmrc` has a more liberal format and the value should probably be coerced using `semver.coerce` first.

szdziedzic:
Thank you very much for reporting this!

[...]

PR #1976, opened 2023-08-03
Revert "[ENG-9422][eas-cli] Use .nvmrc to set node version (#1954)"

szdziedzic:
# Why

https://github.com/expo/eas-cli/issues/1975

# How

Revert https://github.com/expo/eas-cli/pull/1954

# Test Plan

Tests

Run build locally and see that we don't read `.nvmrc`


szdziedzic:
/changelog-entry bug-fix Revert adding `.nvmrc` support for setting node version

PR #1979, opened 2023-08-03
[eas-cli] use 'roll out' instead of rollout for verbs

quinlanj:
# Why

grammar feedback

# Test Plan

- [ ] tests still pass


quinlanj:
/changelog-entry chore use 'roll out' instead of rollout for verbs

szdziedzic (approved):

wschurman (approved):

PR #1980, opened 2023-08-04
[eas-cli] more grammar fixes

quinlanj:
# Why

more grammar feedback from dogfooding

# Test Plan

- [ ] tests still pass

### DIFF
--- a/packages/eas-cli/src/commands/update/republish.ts
+++ b/packages/eas-cli/src/commands/update/republish.ts
@@ -44,7 +44,7 @@ export default class UpdateRepublish extends EasCommand {
 
   static override flags = {
     channel: Flags.string({
-      description: 'Channel name to select an update to republish from',
+      description: 'Channel name to select an update gto republish from',
       exclusive: ['branch', 'group'],
     }),
     branch: Flags.string({
@@ -57,7 +57,7 @@ export default class UpdateRepublish extends EasCommand {
     }),
     message: Flags.string({
       char: 'm',
-      description: 'Short message describing the republished update',
+      description: 'Short message describing the republished update group',
       required: false,
     }),
     platform: Flags.enum({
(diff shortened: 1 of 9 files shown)


packages/eas-cli/src/commands/update/republish.ts
```diff
@@ -44,7 +44,7 @@ export default class UpdateRepublish extends EasCommand {
 
   static override flags = {
     channel: Flags.string({
-      description: 'Channel name to select an update to republish from',
+      description: 'Channel name to select an update gto republish from',
```
szdziedzic, line 47: Was the `gto` added on purpose?

szdziedzic (approved):

packages/eas-cli/src/rollout/actions/EndRollout.ts
```diff
@@ -200,21 +192,21 @@ export class EndRollout implements EASUpdateAction<UpdateChannelBasicInfoFragmen
...
       const actions = formatFields([
         {
           label: '1.',
-          value: `🔁 Republish the Update Group from ${chalk.bold(
+          value: `🔁 Republish the update group from ${chalk.bold(
             rolledOutBranch.name
           )} onto ${chalk.bold(defaultBranch.name)}`,
         },
-        { label: '2.', value: `⬅️  Route all users back to ${chalk.bold(defaultBranch.name)}` },
+        { label: '2.', value: `⬅️  Send all users back to ${chalk.bold(defaultBranch.name)}` },
```
wschurman, line 204: I'm having trouble trying to figure out words. I definitely prefer "send" over "route", but here's some other ones I like:
- `Put all users back on main`
- `Revert all users back to main` (this one would need an opposite of revert for the other sentence, i.e. `What percent of users should <branch_name> be rolled out to?`)

I think I prefer revert/roll out.

packages/eas-cli/src/commands/update/republish.ts
quinlanj: good catch, thanks!

packages/eas-cli/src/rollout/actions/EndRollout.ts
quinlanj: ok. I've gone with revert/roll out where it makes sense, and send otherwise. For example, it makes sense to say `revert users back to main` or `roll out to staging`. But here's a case where it makes sense to use 'send': `10% of users are rolled out to main and 90% of users are sent to staging`

wschurman: `10% of users remain on main and 90% of users are rolled out on staging` (or vice versa, not sure which direction the roll out was)

wschurman (approved):

PR #1981, opened 2023-08-04
[ENG-9603][eas-cli] rollout edit edge case

quinlanj:
# Why

warn the user if they edit their rollout to 0 or 100 that it wont end the rollout. They'll need to end the rollout from the main menu

# Test Plan

- [ ] tests still pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Handle rollout edit edge case. ([#1981](https://github.com/expo/eas-cli/pull/1981) by [@quinlanj](https://github.com/quinlanj))
+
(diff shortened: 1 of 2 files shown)


quinlanj:
/changelog-entry chore handle rollout edit edge case

szdziedzic (approved):

wschurman (commented):
I think we may want to wait to do this as it will make testing a lot more difficult. Because the product is nondeterministic by design, it's helpful to coerce it to be deterministic by faking 0 and 100 percent as a method of simulating a device being selected into the control/rollout group in a real random trial.

quinlanj:
@wschurman i changed editing to 0 or 100 to just produce a warning log that tells people if they want to end the rollout, to do it from the main menu instead. But otherwise let's them go ahead and perform the edit

wschurman (approved):

PR #1983, opened 2023-08-04
Added --non-interactive and --force support when --id is not passed to eas init command

mymattcarroll:
# Why

We have recently updated to version `3.17.0` of `eas-cli` (from `1.2.0`). The `eas init` command previously supported running without prompts to either: 
- create a new project within expo if the project did not exist
- or if the project did exist in expo, add the associated id to the `app.json` file as `extra.eas.projectId`.

The version `3.17.0` `eas init` command does support `--non-interactive` and `--force` but not for the two use cases above.

# How

I added support the --force and --non-interactive to the following two prompts (ids and names are just examples):

```
Existing project found: accountname/projectname (ID: 73b72c6f-76d5-4728-9dc7-d9f6633c0d4e). Link this project?
```

```
Would you like to create a project for accountname/projectname?
```

# Test Plan

Untested, open to feedback before continuing with implementation.


szdziedzic (commented):
Hi 👋 

Thank you very much for contributing 👏 ! Sorry that you waited so long for the response 🙏.

It generally looks good 🚀! I just tested it locally and it seems like it won't work in its current form because
https://github.com/expo/eas-cli/blob/04049a8d1f4f6506c5a5ff311f30b4930daad33b/packages/eas-cli/src/commands/project/init.ts#L35-L43
the `force` and `non-interactive` flags depend on the `id` flag. So currently you won't be able to use `force` and `non-interactive` without the `id` flag specified.

Are you still interested in working on this PR? If so, I can provide you with some guidance:
- https://github.com/expo/eas-cli/blob/main/CONTRIBUTING.md - Here are some instructions regarding how you can run EAS CLI locally to test it, it will be easier for you if you can actually test it 😄 
- The next thing to do is to delete this flag dependency. I hope that after this is done everything will work fine! 🚀 

Let me know if you have any more questions or need some help!

Thanks again 🙏!

mymattcarroll:
@szdziedzic , do you have some time to have another look at this? The flag dependency has been removed.

packages/eas-cli/src/commands/project/init.ts
```diff
@@ -219,7 +215,8 @@ export default class ProjectInit extends EasCommand {
   private static async initializeWithInteractiveSelectionAsync(
```
szdziedzic, line 215:
```suggestion
  private static async initializeWithoutExplicitIDAsync(
```

packages/eas-cli/src/commands/project/init.ts
```diff
@@ -34,13 +35,8 @@ export default class ProjectInit extends EasCommand {
     }),
     force: Flags.boolean({
       description: 'Whether to overwrite any existing project ID',
```
szdziedzic, line 37:
```suggestion
      description: 'Whether to create a new project/link an existing project without additional prompts or overwrite any existing project ID when running with --id flag',
```

szdziedzic (changes requested):
Looks great! I tested it and I thing I noticed is that
https://github.com/oneblink/eas-cli/blob/main/packages/eas-cli/src/commands/project/init.ts#L240-L264
does not respect `--non-interactive` mode

I think we can change it to
```ts
 // if no owner field, ask the user which account they want to use to create/link the project
    let accountName = exp.owner;
    if (!accountName) {
      if (allAccounts.length === 1) {
        accountName = allAccounts[0].name;
      } else if (nonInteractive) {
        accountName = allAccounts[0].name;
        Log.log(`Using default account ${accountName} for non-interactive mode`);
      } else {
        const choices = ProjectInit.getAccountChoices(
          actor,
          accountNamesWhereUserHasSufficientPermissionsToCreateApp
        );

        accountName = (
          await promptAsync({
            type: 'select',
            name: 'account',
            message: 'Which account should own this project?',
            choices,
          })
        ).account.name;
      }
    }

    if (!accountName) {
      throw new Error('No account selected for project. Canceling.');
    }
```

mymattcarroll:
In regards to `accountName` no supporting non-interactive mode, I though it might be better to require the force flag to use the default account if a user has access to more than one. I'm happy to change to your suggestion if you believe the force flag requirement is overkill.

szdziedzic (changes requested):
When it comes to `accountName`:

So if one uses `nonInteractive` without force this error will be thrown https://github.com/expo/eas-cli/blob/752ff42570824e0cea9731931e5922b31a73352e/packages/eas-cli/src/commands/project/init.ts#L266-L268, right?

I believe that this is a confusing error message in such a case. I believe we should make it more actionable:
```ts
if (!accountName) { 
   if (nonInteractive) {
       throw new Error(`There are multiple accounts that you have access to: ${allAccounts.map(a => a.name).join(', ')}. Explicitly set the owner property in your app config or run this command with the --force flag to proceed with a default ${allAccounts[0].name} account.`)
   }
   throw new Error('No account selected for project. Canceling.'); 
 } 
```

mymattcarroll:
💯 agree! Ill update now to achieve the desired logic.

szdziedzic (commented):
Amazing 🎉 LGTM!

Can you add a changelog entry for this PR by editing the `CHANGELOG.md` file, please 🙏?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Added `--non-interactive` and `--force` support when `--id` is not passed to `eas init` command by [@mymattcarroll](https://github.com/mymattcarroll))
```
szdziedzic, line 11:
```suggestion
- Add `--non-interactive` and `--force` support when `--id` is not passed to the `eas init` command. ([#1983](https://github.com/expo/eas-cli/pull/1983) by [@mymattcarroll](https://github.com/mymattcarroll))
```

szdziedzic (commented):
one last tweak to make changelog linter happy

szdziedzic (approved):

PR #1984, opened 2023-08-05
[ENG-9614][eas-cli] rollout ux improvements

quinlanj:
# Why

UX improvements from dogfooding feedback:
- Better runtime selection: remove the option to manually input a runtime in the case where there is no shared runtime on the two rollout branches. Instead, let the user choose a runtime from an existing branch or use the runtime specified in their project config.
- Remove automatic selection if only 1 is available: Automatic selection in this case is the most 'efficient' but causes confusion without human prompting. For example, when determining a channel or branch for a rollout, the selectors used to return the only eligible item without prompting, but this ended up causing confusion. This PR changes the selectors to always require human interaction to select an option (even if there is only 1 available). 
- Print better logs if no eligible item: In cases where there isn't a branch available for a rollout, instead of saying the user has no branches, say tell the user they need to make another branch.

# Test Plan

- [ ] manually tested


szdziedzic (approved):

szdziedzic:
Tests are failing https://github.com/expo/eas-cli/actions/runs/5773134340/job/15648770217?pr=1984

PR #1994, opened 2023-08-08
[eas-json][eas-cli] add global `withoutCredentials` property to `eas.json`

szdziedzic:
# Why

Add global `withoutCredentials` property to `eas.json` (it currently only exists for Android). We discussed it with Brent and decided that it would be handy for custom builds.

# How

Add global `withoutCredentials` property to `eas.json`

# Test Plan

Add automated tests


szdziedzic:
/changelog-entry new-feature Add the `withoutCredentials` option as a common build profile field in `eas.json`

khamilowicz (approved):

PR #1998, opened 2023-08-09
[ENG-9422]  nvmrc support for setting node version

khamilowicz:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1689882938014309

# How

Read .nvmrc content and set `node` value in `buildProfile`.

# Test Plan

1. Create valid .nvmrc file
2. Build project
3. Check in build logs if non-default node version is installed.

# Deploy plan

1. https://github.com/expo/turtle-v2/pull/1393
2. https://github.com/expo/eas-cli/pull/1998

### DIFF
--- a/packages/eas-cli/src/utils/__tests__/profiles-test.ts
+++ b/packages/eas-cli/src/utils/__tests__/profiles-test.ts
@@ -1,5 +1,8 @@
 import { Platform } from '@expo/eas-build-job';
 import { BuildProfile, EasJsonAccessor, EasJsonUtils, errors } from '@expo/eas-json';
+import fs from 'fs-extra';
+import { vol } from 'memfs';
+import os from 'os';
 
 import Log from '../../log';
 import { selectAsync } from '../../prompts';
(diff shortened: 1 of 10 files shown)


packages/eas-cli/src/utils/profiles.ts
```diff
@@ -47,16 +54,53 @@ export async function getProfilesAsync<T extends ProfileType>({
...
 }
 
-async function readProfileAsync<T extends ProfileType>({
+async function setNodeVersionFromFileAsync(
+  projectDir: string,
+  profile: BuildProfile<Platform>
+): Promise<void> {
+  if (profile?.node) {
+    return;
+  }
```
radoslawkrzemien, line 63:
```suggestion
async function maybeSetNodeVersionFromFileAsync(
  projectDir: string,
  profile: BuildProfile<Platform>
): Promise<void> {
  if (profile?.node) {
    return;
  }
```
If `profile?.node` is set then it is not set from file

radoslawkrzemien (approved):
LGTM 🚀

packages/eas-cli/src/utils/__tests__/profiles-test.ts
```diff
@@ -98,6 +106,52 @@ describe(getProfilesAsync, () => {
     expect(getBuildProfileAsync).toBeCalledWith(easJsonAccessor, Platform.IOS, 'custom-profile');
   });
 
+  describe('node version', () => {
+    const nodeVersion = '14.17.1';
```
szdziedzic, line 110: I think it might be good to add the test case for values like `v16`

PR #2001, opened 2023-08-10
[eas-cli] more branch map utility functions

quinlanj:
# Why

extra utility function for branch mapping

# Test Plan

- [ ] new tests pass


quinlanj:
/changelog-entry chore more branch map utility functions

packages/eas-cli/src/rollout/branch-mapping.ts
```diff
@@ -257,6 +257,17 @@ export function isRollout(channelInfo: ChannelBasicInfo): boolean {
   return isRolloutBranchMapping(branchMapping);
 }
 
+export function targetsRollout(
```
wschurman, line 260:
```suggestion
export function doesTargetRollout(
```

wschurman (approved):
One naming nit, otherwise looks good.

PR #2002, opened 2023-08-11
[eas-cli][update] Only export at most ios and android dist

wschurman:
# Why

Sometime in SDK 49 web was added as a default exported platform when calling `npx expo export -p all`. EAS CLI is unversioned, so it was always just passing in `all` as the flag when `all` was requested. But EAS update doesn't really support web yet so exporting and uploading the web update doesn't make sense yet. 

So only export at most ios and android.

Closes ENG-9300.
Closes ENG-9723.

# Test Plan

`neas update` on a project configured for metro web, see it no longer publishes web by default.


wschurman:
/changelog-entry breaking-change Only export at most ios and android dist for EAS updates

quinlanj (approved):
thanks, lgtm!

PR #2005, opened 2023-08-14
[eas-cli][eas-update] Support republishing roll back to embedded updates

wschurman:
# Why

https://github.com/expo/eas-cli/pull/2004#issuecomment-1677356680

When a roll back to embedded type update was republished, the code failed. This fixes it.

# How

Support republishing signed roll-back-to-embedded updates.

# Test Plan

$ neas update:rollback --private-key-path keys/private-key.pem
✔ Which type of update would you like to roll back to? › Published Update
✔ Find update by branch or channel? › Channel
✔ Select a channel to view › main
✔ Load more update groups? › [Aug 14 09:38 by wschurman, runtimeVersion: 1.0.0] rb
✔ The republished update group will appear only on: all
✔ Provide an update message. … Republish "rb" - group: c59d7422-0064-4fee-b14e-259517a223f0
✔ The republished update group will be signed
- Republishing...
🔒 Signing republished update group
- Republishing...
✔ Republished update group

Branch             main
Runtime version    1.0.0
Platform           android, ios
Update group ID    0849f86f-5808-4c4d-8c08-f257998e235d
Android update ID  e6266b29-07ff-4470-91e9-b8352b10dd3c
iOS update ID      894ba423-53a0-4564-afa7-d6362ebda3fb
Message            Republish "rb" - group: c59d7422-0064-4fee-b14e-259517a223f0
Website link       https://expo.dev/accounts/wschurman/projects/watwathuh/updates/0849f86f-5808-4c4d-8c08-f257998e235d

PR #2006, opened 2023-08-15
[eas-cli][eas-update] Support republishing roll back to embedded updates

wschurman:
Accidentally merged https://github.com/expo/eas-cli/pull/2005 into its parent PR rather than waiting for it to land and rebasing on main. This rebases on main and lands there.


wschurman:
/changelog-entry bug-fix Support republishing roll back to embedded updates

PR #2020, opened 2023-08-25
[eas-cli] [ENG-7536] Stop app config error repeating

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-7536/make-app-config-error-on-credentials-command-not-repeat-indefinitely

# How

If app config error is thrown it is re-thrown with modified message instead of printing it to the as a "cryptic error" and repeating with 0% chances of success (since the config is incorrect)

# Test Plan

Added automated test

Tested manually:
Screenshot 2023-08-25 at 16 54 36

The link from the message leads to the docs:
Screenshot 2023-08-25 at 16 54 57

### DIFF
--- a/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
+++ b/packages/eas-cli/src/credentials/manager/ManageAndroid.ts
@@ -4,6 +4,7 @@ import assert from 'assert';
 
 import Log, { learnMore } from '../../log';
 import { GradleBuildContext, resolveGradleBuildContextAsync } from '../../project/android/gradle';
+import { getProjectConfigDescription } from '../../project/projectUtils';
 import { promptAsync } from '../../prompts';
 import { AssignFcm } from '../android/actions/AssignFcm';
 import { AssignGoogleServiceAccountKey } from '../android/actions/AssignGoogleServiceAccountKey';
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/credentials/manager/ManageAndroid.ts
```diff
@@ -131,6 +132,18 @@ export class ManageAndroid {
 
         await this.runProjectSpecificActionAsync(ctx, chosenAction, gradleContext);
       } catch (err) {
+        if (
+          err instanceof Error &&
+          err.message ===
+            `Specify "android.package" in ${getProjectConfigDescription(
+              ctx.projectDir
+            )} and run this command again.`
```
szdziedzic, line 140: Is there any better way to determine if this is this specific error (not relying on a message)? 🤔 Ideally, it would be good to use an error code or sth like this.

What do you think?

radoslawkrzemien: Thought about it, I can make a custom error class for that, probably the best option

radoslawkrzemien:
/changelog-entry bug-fix Make app config error not repeat indefinitely

szdziedzic (approved):

PR #2036, opened 2023-09-01
[eas-cli] only print channel-branch pairing if we created a channel

quinlanj:
# Why

When someone runs `eas update` to publish an update to a branch (let's call it Branch X for our example), we always print `Channel X points to Branch X`. However, this is not always true. 

# How
I suspect the statement was meant to be put in the adjacent `if` clause, so I've moved it there.

In the event where Branch X does not exist, we create it. Then, we also create a channel with the same name (X). If Channel X can be created, we link it to Branch X. In the case we can create Channel X, then the statement `Channel X points to Branch X` will always be true.

# Test Plan

- [ ] manually tested


quinlanj:
/changelog-entry bug-fix update: only print channel-branch pairing if we created a channel

wschurman (approved):

PR #2047, opened 2023-09-06
[eas-cli] rollouts: more robust printing function

quinlanj:
# Why

More robust printing function for rollouts

# Test Plan

- [ ] manually tested


quinlanj:
/changelog-entry chore rollouts: more robust printing function

wschurman (approved):

PR #2050, opened 2023-09-07
feat(eas-cli): expose expo export dev flag as an option in eas update

nderscore:
# Why

When `__DEV__` is unset or `false`, [expo-dev-menu's `registerDevMenuItems` option is a no-op](https://github.com/expo/expo/blob/main/packages/expo-dev-menu/src/DevMenu.ts#L38-L40). 

In order to be able to use custom expo-dev-menu menu items in updates/previews for a debug build, it is necessary for the `__DEV__` global variable to be set to `true` in the generated updates and preview bundles.

EAS does not currently expose the `--dev` option from Expo CLI's `export` command, which is what determines whether or not `__DEV__` gets stripped during bundling.

# How

Adds a new CLI flag (`--dev`) to the `eas update` command.

When set, an additional CLI argument `--dev` is passed to `expo update`.

# Test Plan

I tested this in my own project and it successfully deployed an update where `__DEV__` is set to true and my custom dev menu items are being registered correctly.

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -56,6 +56,7 @@ import { getVcsClient } from '../../vcs';
 
 type RawUpdateFlags = {
   auto: boolean;
+  dev: boolean;
   branch?: string;
   channel?: string;
   message?: string;
@@ -74,6 +75,7 @@ type RawUpdateFlags = {
 
 type UpdateFlags = {
   auto: boolean;
+  dev: boolean;
   platform: ExpoCLIExportPlatformFlag;
   branchName?: string;
   channelName?: string;
@@ -134,6 +136,10 @@ export default class UpdatePublish extends EasCommand {
       default: 'all',
       required: false,
     }),
+    dev: Flags.boolean({
+      description: 'Publish an unminified dev bundle without stripping __DEV__ global',
+      default: false,
+    }),
     auto: Flags.boolean({
       description:
         'Use the current git branch and commit message for the EAS branch and update message',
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -208,6 +210,7 @@ export async function buildBundlesAsync({
       '--dump-sourcemap',
       '--dump-assetmap',
       `--platform=${platformFlag}`,
+      ...(dev ? ['--dev'] : []),
```
nderscore, line 213: I'm not sure if this flag exists on the legacy Expo CLI

douglowder (approved):

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -134,6 +136,10 @@ export default class UpdatePublish extends EasCommand {
       default: 'all',
       required: false,
     }),
+    dev: Flags.boolean({
+      description: 'Publish an unminified dev bundle without stripping __DEV__ global',
```
quinlanj, line 140: `__DEV__` is an implementation detail used under the hood to enable debugging features. best to not mention it in the high level description.
```suggestion
      description: 'Publish a development bundle',
```

packages/eas-cli/src/project/publish.ts
quinlanj: Confirming it does! https://github.com/expo/expo-cli/blob/main/packages/expo-cli/src/commands/export/export.ts#L26

quinlanj (approved):
some nits, otherwise lgtm!

wschurman:
@nderscore - can you manually add an entry to https://github.com/expo/eas-cli/blob/main/CHANGELOG.md

Then we can go ahead an merge!

PR #2052, opened 2023-09-13
[FIX] Remove semver validation from node version read from .nvmrc

khamilowicz:
# Why

https://github.com/expo/eas-cli/issues/1975#issuecomment-1702986526
Node version can be specified as an alias (such as `iojs`, `lts/hydrogen` etc), so validating it as semver may cause errors for some users.

# How

Remove validation

# Test Plan

Specify node version in `.nvmrc` as something like `lts/gallium` and run build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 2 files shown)


khamilowicz:
/changelog-entry bug-fix Support node aliases in `.nvmrc`.

khamilowicz:
/changelog-entry bug-fix Support node aliases in .nvmrc.

szdziedzic (approved):

khamilowicz:
@szdziedzic yep 
image

PR #2067, opened 2023-09-27
[eas-cli] [ENG-10225] add requiredPackageManager to build metadata

kadikraman:
# Why

We want to know what the user's package manager is going to be at the "install custom tools" step of the build process, so we're determining it based on the user's lockfiles and adding it to build metadata.

1. `eas-build-job` https://github.com/expo/eas-build/pull/281
2. `www` https://github.com/expo/universe/pull/13522
3. 👉 `eas-cli` https://github.com/expo/eas-cli/pull/2067
4. `turtle-v2` https://github.com/expo/turtle-v2/pull/1460

# How

Added the field.

# Test Plan

Unit tests and building locally.

### DIFF
--- a/packages/eas-cli/src/build/metadata.ts
+++ b/packages/eas-cli/src/build/metadata.ts
@@ -64,6 +64,8 @@ export async function collectMetadataAsync<T extends Platform>(
     buildMode: ctx.buildProfile.config ? BuildMode.CUSTOM : BuildMode.BUILD,
     customWorkflowName: ctx.customBuildConfigMetadata?.workflowName,
     developmentClient: ctx.developmentClient,
+    // @ts-expect-error TODO needs @expo/eas-build-job to be relesed
+    requiredPackageManager: ctx.requiredPackageManager,
   };
   return sanitizeMetadata(metadata);
 }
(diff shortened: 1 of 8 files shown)


packages/eas-cli/src/build/metadata.ts
```diff
@@ -64,6 +64,8 @@ export async function collectMetadataAsync<T extends Platform>(
     buildMode: ctx.buildProfile.config ? BuildMode.CUSTOM : BuildMode.BUILD,
     customWorkflowName: ctx.customBuildConfigMetadata?.workflowName,
     developmentClient: ctx.developmentClient,
+    // @ts-expect-error TODO needs @expo/eas-build-job to be relesed
+    requiredPackageManager: ctx.requiredPackageManager,
```
szdziedzic, line 68: I think this type will require it to be ~ `string | undefined`, but the type of it is `string | null`

szdziedzic (commented):
It looks good, one comment:

Also one reminder: remember to update dependencies in `turtle-v2` before deploying it

packages/eas-cli/src/build/metadata.ts
kadikraman: Good spot, I'll add `?? undefined` to match the others!

szdziedzic (approved):
Nice. Thanks!

PR #2068, opened 2023-09-27
[ENG-10176][eas-json] update images info in `eas.schema.json`

szdziedzic:
# Why

We've added a new `macos-ventura-13.6-xcode-15.0` image. It's also a good opportunity to clean up a bit by deprecating some images and deleting images only available for Intel workers which we will delete soon (29th September).

# How

Updated `eas.schema.json`. Use @byCedric suggestion to mark some images as deprecated by using the new syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


szdziedzic:
/changelog-entry chore Update EAS Build images description in our VSCode plugin. Add new `macos-ventura-13.6-xcode-15.0` image

brentvatne (approved):

PR #2069, opened 2023-09-28
feat: add tap to pay on iphone entitlement

fobos531:
# Why

Seems like it's missing - but not sure if that was intentional or there's any other hindrance to it? 

# How

Add the entitlement object exactly like the others. It doesn't seem like there's anything else to do? 

# Test Plan

@EvanBacon It seems `@expo/apple-utils` is a private package, so I cannot contribute a change to add a `TAP_TO_PAY_ON_IPHONE` constant for the `CapabilityType` - for now I just force casted this to keep TS happy. 

Thoughts?

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
@@ -734,6 +734,13 @@ export const CapabilityMapping: CapabilityClassifier[] = [
     getOptions: getBooleanOptions,
   },
   {
+    entitlement: 'com.apple.developer.proximity-reader.payment.acceptance',
+    name: 'Tap to Pay on iPhone',
+    capability: CapabilityType.TAP_TO_DISPLAY_ID,
+    validateOptions: validateBooleanOptions,
+    getOptions: getBooleanOptions,
+  },
+  {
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -733,6 +733,13 @@ export const CapabilityMapping: CapabilityClassifier[] = [
     validateOptions: validateBooleanOptions,
     getOptions: getBooleanOptions,
   },
+  {
+    entitlement: 'com.apple.developer.proximity-reader.payment.acceptance',
+    name: 'Tap to Pay on iPhone',
+    capability: 'TAP_TO_PAY_ON_IPHONE' as CapabilityType,
```
EvanBacon, line 739: I've published `@expo/apple-utils@1.3.2` with `TAP_TO_PAY_ON_IPHONE`, `SUSTAINED_EXECUTION`, and `NETWORK_SLICING`.

fobos531: Thank you! The PR is updated to reflect this change.

brentvatne (approved):

PR #2073, opened 2023-10-02
[eas-cli] [ENG-9957] Don't overwrite distribution for simulator builds

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-9957/fix-internal-distribution-pages-for-simulator-builds
Companion for: https://github.com/expo/eas-build/pull/283, https://github.com/expo/universe/pull/13524, and https://github.com/expo/universe/pull/13523

# How

- distribution value is kept instead of hardcoded to `simulator` if simulator flag is set, keeping its original value
- simulator field has been added to metadata to retain its value

# Test Plan

Existing tests pass

# Screens
#### Internal build
Screenshot 2024-01-31 at 17 03 36
#### Internal simulator build
Screenshot 2024-01-31 at 17 03 39
#### Store build
Screenshot 2024-01-31 at 17 03 42
#### Store simulator build
Screenshot 2024-01-31 at 17 03 47
#### Internal build
Screenshot 2024-01-31 at 17 04 01
#### Internal simulator build
Screenshot 2024-01-31 at 17 04 05
#### Store build
Screenshot 2024-01-31 at 17 04 08
#### Store simulator build
Screenshot 2024-01-31 at 17 04 13
#### Internal build
Screenshot 2024-01-31 at 16 52 09
#### Internal simulator build
Screenshot 2024-01-31 at 16 52 17
#### Store build
Screenshot 2024-01-31 at 16 52 21
#### Store simulator build
Screenshot 2024-01-31 at 16 52 26
#### Internal build
Screenshot 2024-01-31 at 16 52 45
#### Internal simulator build
Screenshot 2024-01-31 at 16 52 49
#### Store build
Screenshot 2024-01-31 at 16 52 53
#### Store simulator build
Screenshot 2024-01-31 at 16 52 57

# Deployment
1. www
2. website
3. cli


packages/eas-cli/src/build/metadata.ts
```diff
@@ -26,10 +26,7 @@ export async function collectMetadataAsync<T extends Platform>(
 ): Promise<Metadata> {
   const vcsClient = getVcsClient();
   const channelOrReleaseChannel = await resolveChannelOrReleaseChannelAsync(ctx);
-  const distribution =
-    ('simulator' in ctx.buildProfile && ctx.buildProfile.simulator
-      ? 'simulator'
-      : ctx.buildProfile.distribution) ?? 'store';
+  const distribution = ctx.buildProfile.distribution ?? 'store';
```
szdziedzic, line 29: This change will break stuff like `eas build:run` 🤔 

I'm wondering if altering the `distributionType` behavior is necessary 🤔.

 Maybe we can add some new field to Metadata to keep the raw `buildProfile.distribution` value there? I mean it's probably not optimal but I feel that this change can introduce some regressions in some parts of the codebase.
 
 What do you think?

radoslawkrzemien: Can you elaborate on why `eas build:run` would be broken by this?
As for another field for "original distribution" or whatever we want to call it, I'm open for discussion but it feels to me like hacking around an existing hack (overwriting the distribution). I feel like we should remove the reliance on this hardcoded `simulator` value for `distribution` instead of building upon it, wdyt?

radoslawkrzemien: I see, it runs a simulator build and fetches it by filtering by distribution = simulator.
Maybe it should handle both options (distribution = simulator or simulator = true) for a while instead of introducing another field?

sjchmiela: I concur, we should list builds with both `simulator: true` and `distribution: simulator`.

sjchmiela (approved):
:+1:, but I'm with @szdziedzic, I think we should not "break" listing simulator builds for now.

packages/eas-cli/src/build/metadata.ts
radoslawkrzemien: It's handled in https://github.com/expo/universe/pull/13524

packages/eas-cli/src/run/utils.ts
```diff
@@ -13,6 +13,7 @@ export function isRunnableOnSimulatorOrEmulator(build: BuildFragment): boolean {
     build.status === BuildStatus.Finished &&
     !!build.artifacts?.applicationArchiveUrl &&
     ((build.platform === AppPlatform.Ios && build.distribution === DistributionType.Simulator) ||
```
sjchmiela, line 15:
```suggestion
```
I guess the `isForSimulator` already takes care of it?

sjchmiela (approved):

packages/eas-cli/src/build/metadata.ts
```diff
@@ -24,10 +24,7 @@ export async function collectMetadataAsync<T extends Platform>(
   ctx: BuildContext<T>
 ): Promise<Metadata> {
   const channelOrReleaseChannel = await resolveChannelOrReleaseChannelAsync(ctx);
-  const distribution =
-    ('simulator' in ctx.buildProfile && ctx.buildProfile.simulator
-      ? 'simulator'
-      : ctx.buildProfile.distribution) ?? 'store';
+  const distribution = ctx.buildProfile.distribution ?? 'store';
```
szdziedzic, line 27:
```suggestion
  const distribution = ctx.buildProfile.distribution ?? BuildDistributionType.STORE;
```

packages/eas-cli/src/commands/build/list.ts
```diff
@@ -69,6 +71,11 @@ export default class BuildList extends EasCommand {
       distribution: buildDistribution,
       'non-interactive': nonInteractive,
     } = flags;
+    if (buildDistribution === BuildDistributionType.SIMULATOR) {
+      Log.warn(
+        `Using "distribution" flag with "simulator" value is deprecated - use "simulator" flag instead`
+      );
```
szdziedzic, line 77: Nice 👏 
```suggestion
      Log.warn(
        `Using --distribution flag with "simulator" value is deprecated - use --simulator flag instead`
      );
```

packages/eas-cli/src/commands/build/list.ts
```diff
@@ -51,6 +52,7 @@ export default class BuildList extends EasCommand {
     ...EasPaginatedQueryFlags,
     limit: getLimitFlagWithCustomValues({ defaultTo: 10, limit: BUILDS_LIMIT }),
     ...EasNonInteractiveAndJsonFlags,
+    simulator: Flags.boolean(),
```
szdziedzic, line 55: Maybe it would be good to add a description to this flag, putting emphasis on the fact that it will only filter the **iOS** simulator builds. People tend to confuse Android emulator with iOS simulator and one could think that this will also query for Android emulator builds.

szdziedzic, line 55: Maybe we should also enforce that this is only applicable when using the `ios` value for the `--platform` flag? What do you think?

szdziedzic (commented):
Looks good, few small comments

szdziedzic (approved):
Looks good

Remember to regenerate the GQL schema if there are any changes to it in WWW PR

I would also do one test run of the `eas build` command with the simulator target and see if you are prompted correctly to install it on your device afterward and run the `eas build:run` command just to double check

radoslawkrzemien:
/changelog-entry chore Add simulator flag to metadata

radoslawkrzemien:
Added screens for comparison

PR #2086, opened 2023-10-13
[eas-cli] VCS Context Fields to replace helper functions

Josh-McFarlin:
# Why

Context fields make it easier for eas-cli commands to express their dependencies.

# How

Created a new context field for `getVcsClient` and `ensureRepoExistsAsync` and replaced usages of their original helper function calls with references to EAS Command contexts.

# Test Plan

Run `yarn test` locally and all tests pass on CI.

image


packages/eas-cli/src/commandUtils/context/VcsGitRepoExistsContextField.ts
```diff
@@ -0,0 +1,26 @@
+import { getVcsClient } from '../../vcs';
+import ContextField from './ContextField';
+
+export default class VcsRepoExistsContextField extends ContextField<{
```
wschurman, line 4: remove this until we have a use case

packages/eas-cli/src/commandUtils/context/VcsClientContextField.ts
```diff
@@ -0,0 +1,9 @@
+import { getVcsClient } from '../../vcs';
+import { Client } from '../../vcs/vcs';
+import ContextField from './ContextField';
+
+export default class VcsClientContextField extends ContextField<Client> {
+  async getValueAsync(): Promise<Client> {
+    return getVcsClient();
```
wschurman, line 7: Ideally, in a follow-up PR, we could get rid of this function so that future command authors aren't tempted to just call this and instead are led to the "pit of success" by only being able to access this via a context. Of course, to do this will require more mocking in tests, so that's why I suggest doing this in a separate PR after this one.

For context (no pun intended), one of the original reasons for having context fields was to prevent command authors from calling getProjectAsync directly without calling a method to set it up. Now, all accesses to the project run the same code that ensures the project is set up correctly.

wschurman (approved):
This looks great and is a great step towards injecting more dependencies! See inline comment about a potential follow-up PR.

Josh-McFarlin:
/changelog-entry chore Move `getVcsClient` and `ensureRepoExistsAsync` into command context. ([#2086](https://github.com/expo/eas-cli/pull/2086) by [@Josh-McFarlin](https://github.com/Josh-McFarlin))

PR #2093, opened 2023-10-18
[eas-cli][eas-json] remove Intel resource class related code

szdziedzic:
# Why

We no longer have Intel Macs in our fleet.

# How

Clean up Intel resource class related code

# Test Plan

Tests, test manually


radoslawkrzemien (approved):

packages/eas-json/src/build/types.ts
```diff
@@ -8,19 +8,10 @@ export enum CredentialsSource {
...
   DEFAULT = 'default',
   LARGE = 'large',
-  /**
-   * @deprecated use M_MEDIUM instead
-   */
-  M1_EXPERIMENTAL = 'm1-experimental',
   /**
    * @deprecated use M_MEDIUM instead
    */
   M1_MEDIUM = 'm1-medium',
```
sjchmiela, line 14: How about… you know what I'm thinking?

szdziedzic: Hmm, I think that we can't simply remove it, because when we originally introduced the M1 resource class, the way to use it was to set `m1-medium` in their `eas.json` (currently we recommend `m-medium` so you don't have to distinguish between M1, M2 and so on). That's why many people still have this value in their `eas.json` to this day. If we remove it, this change will break their builds. 

I mean it's been deprecated for a while and we are printing a deprecation warning if you use it. So maybe we can do it? I don't know, it's a hard call. What do you think?

kbrandwijk: I wouldn't remove it (yet), unless that CLI warning includes a 'do you want us to update your eas.json for you' step.

szdziedzic:
/changelog-entry chore Clean up Intel resource classes code after their deletion

PR #2095, opened 2023-10-18
[ci] Try to fix changelog-bot usage of hub CLI

wschurman:
# Why

`hub` was removed: https://github.com/mislav/hub/issues/3338

our action is failing: https://github.com/expo/eas-cli/actions/runs/6565215869

# How

Use `gh` CLI instead.

# Test Plan

It is preinstalled so it should be good to go.


ide (approved):
LGTM, `gh pr` is what GitHub tells humans to run, too

PR #2101, opened 2023-10-25
[eas-cli] [ENG-8973] Use selected builds profile name if not specified

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-8973/eas-submit-doesnt-use-asc-configuration-from-selected-build-profile

# How

Moved the archive selection before any submit options are determined, and if the archive is selected from EAS then the profile in the submission context is refreshed to use the submit profile with same name as the name of the build profile in the archive (if exists)

# Test Plan

Added automated tests and checked parts of the process manually

When corresponding build profile does not exist
Screenshot 2023-10-25 at 19 11 06
Screenshot 2023-10-25 at 19 10 07
Screenshot 2023-10-25 at 19 10 23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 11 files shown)


radoslawkrzemien:
/changelog-entry new-feature Use corresponding submit profile when selecting build from EAS

szdziedzic (approved):
Looks good!

PR #2119, opened 2023-11-10
[eas-cli] [ENG-10555] Fix device provsioning

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-10555/ios-devices-get-stuck-in-failed-to-provision-state-until-the-developer
https://www.reddit.com/r/expo/comments/17kr7rr/failed_to_provision/?sort=new

# How

The problem happened when the provisioning profile already existed and the user wanted to add new devices to it. In such a case, a missing device was added to the Apple Developer Portal if needed, but the provisioning profile from EAS was used unchanged. This meant that no matter what the devices list in the provisioning profile used was the same as before and thus resulted in a warning message and a failure.
Now another check has been added, if the id of the profile from EAS matches the one from Apple the list of devices is compared. If it's the same then the profile is reused as before. If the list is different then the profile in EAS gets updated first with the current devices data from Apple

# Test Plan

Tested manually by reproducing the issue mentioned by the users, applying fix and confirming that the issue went away.
Also added an automated test to cover the new case and adjusted existing tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fixed provisioning of new devices into an existing profile. ([#2119](https://github.com/expo/eas-cli/pull/2119) by [@radoslawkrzemien](https://github.com/radoslawkrzemien))
+
 ### 🧹 Chores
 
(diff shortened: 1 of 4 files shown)


radoslawkrzemien:
/changelog-entry bug-fix Fixed provisioning of new devices into an existing profile

szdziedzic (approved):
Nice! Looks reasonable

PR #2123, opened 2023-11-16
[ENG-10464][eas-cli] throw error if custom build config is gitignored

szdziedzic:
# Why

If you create a custom build with git ignored config file build on a server will fail even though local validation passed. 

# How

Use the VCS client to check if the file is ignored. Throw an error if it is.

# Test Plan

Test locally. Git ignore the config. Make sure the build runs fine without git ignored config.

Screenshot 2023-11-16 at 17 07 44

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -143,7 +143,11 @@ export async function runBuildAndSubmitAsync(
     {};
   for (const buildProfile of buildProfiles) {
     validateBuildProfileVersionSettings(buildProfile, easJsonCliConfig);
-    const maybeMetadata = await validateCustomBuildConfigAsync(projectDir, buildProfile.profile);
+    const maybeMetadata = await validateCustomBuildConfigAsync({
+      projectDir,
+      profile: buildProfile.profile,
+      vcsClient,
+    });
     if (maybeMetadata) {
       customBuildConfigMetadataByPlatform[toAppPlatform(buildProfile.platform)] = maybeMetadata;
     }
--- a/packages/eas-cli/src/project/customBuildConfig.ts
+++ b/packages/eas-cli/src/project/customBuildConfig.ts
@@ -5,14 +5,21 @@ import chalk from 'chalk';
 import fs from 'fs-extra';
 import path from 'path';
 
+import { Client } from '../vcs/vcs';
+
 export interface CustomBuildConfigMetadata {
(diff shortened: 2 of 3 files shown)


szdziedzic:
/changelog-entry chore Throw error if custom build config is gitignored

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -143,7 +143,11 @@ export async function runBuildAndSubmitAsync(
     {};
   for (const buildProfile of buildProfiles) {
     validateBuildProfileVersionSettings(buildProfile, easJsonCliConfig);
-    const maybeMetadata = await validateCustomBuildConfigAsync(projectDir, buildProfile.profile);
+    const maybeMetadata = await validateCustomBuildConfigAsync({
+      projectDir,
+      profile: buildProfile.profile,
+      vcsClient,
+    });
```
sjchmiela, line 150: ❤️

packages/eas-cli/src/project/customBuildConfig.ts
```diff
@@ -24,6 +31,13 @@ export async function validateCustomBuildConfigAsync(
       `Custom build configuration file ${chalk.bold(relativeConfigPath)} does not exist.`
     );
   }
+  if (await vcsClient.isFileIgnoredAsync(relativeConfigPath)) {
```
sjchmiela, line 34: May a config file reference other files? Maybe not now, but in the future? Or TypeScript functions? Should we make sure whole `.eas` is committed instead (instead of just the config file)?

szdziedzic: >May a config file reference other files?

https://docs.expo.dev/custom-builds/schema/#import

>Or TypeScript functions?

https://docs.expo.dev/custom-builds/schema/#functionsfunction_namepath

>Should we make sure whole .eas is committed instead (instead of just the config file)?

Good question 🤔 

I believe that we should fail if the currently used config file is git ignored, but maybe display a warning if other files in `.eas` are git ignored? Ideally, I feel like we can try to resolve which custom functions will be needed for a given build and which config files will be needed and validate that they are not git ignored, but I think it will require some more changes, and I think that we can merge it as v1 to make it better and not wait with it for other stuff to come along. WDYT?

brentvatne (approved):
s/custom build/job for future proofing, but otherwise looks good

szdziedzic:
/changelog-entry chore Throw error if custom build config is gitignored

PR #2124, opened 2023-11-16
[ENG-9772][eas-cli] don't ask user to install dev client if running in non-interactive mode

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-9772/prevent-interactive-cli-commands-in-build-configuration-step

We don't want to prompt users to install the dev client in non-interactive mode.

# How

Instead of displaying the confirm prompt in non-interactive mode fail with a helpful error message.

# Test Plan

Test locally 
Screenshot 2023-11-16 at 17 38 56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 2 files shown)


szdziedzic:
/changelog-entry bug-fix Don't ask a user to install the dev client if running in non-interactive mode

PR #2125, opened 2023-11-17
[eas-cli] Always refresh provisioning profile

radoslawkrzemien:
# Why

https://exponent-internal.slack.com/archives/C9PRD479V/p1700236895524379
https://exponent-internal.slack.com/archives/C9PRD479V/p1700229660146479

# How

There are other cases when the provisioning profile needs to be refreshed before use and which cannot be easily validated at runtime without bigger changes in the code. It is simpler (at least for the time being) to just refresh existing profile each time

# Test Plan

Run automated tests


kadikraman (approved):
Looks good to me! Thanks for jumping on this 🙌

szdziedzic (approved):

PR #2147, opened 2023-12-12
[ci] Only lint changelog in Node 16

sjchmiela:
# Why

`lint-changelog` tends to sometimes fail on Node 18 (see [this](https://github.com/expo/eas-cli/actions/runs/7184124974/job/19571620519)). I don't want to debug why.

# How

After noticing `scripts/tsconfig.json` mentions "node16", I figured maybe scripts are supposed to only run on Node 16. Adjusted matrix and job spec.

https://github.com/expo/eas-cli/blob/04a8b08f5981615373bca680530e5c1a0bb904e4/scripts/tsconfig.json#L4-L5

# Test Plan

I mean, it's linting changelog…

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,8 @@ jobs:
       matrix:
         node: ['14', '16', '18']
         include:
+          - node: '16'
+            lint-changelog: true
           - node: '18'
             coverage: true
     name: Test with Node ${{ matrix.node }}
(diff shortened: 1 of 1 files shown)


szdziedzic (approved):
I think this is good for now, but in my opinion, we should try to make it work with Node 18 because Node 16 is no longer supported. I believe we should also add support for Node 20 run our test workflows on Node 16,18,20, and drop support for Node 14. WDYT?

Feel free to merge this as it is. We can figure it out later on.

szdziedzic:
Ok, I think that it broke in Node `18.19.0`. I was able to replicate it locally.

szdziedzic:
WDYT about https://github.com/expo/eas-cli/pull/2153?

sjchmiela:
Superseded by https://github.com/expo/eas-cli/pull/2153

PR #2148, opened 2023-12-13
[eas-cli][eas-update] Allow undefined update message

wschurman:
# Why

This was requested for cases where there isn't a VCS and also no update message is manually supplied.

Closes ENG-9633.

# How

It is already optional in the GraphQL mutation, so this is just a product decision that we need to deliberately make.

# For Reviewers

I think it's worth having a discussion on the UX benefits/costs of this product decision. This PR is meant to just start the discussion, not convey a strong intent to make the change as I'm probably oblivious to potential downsides.

# Test Plan

Create an update and don't specify message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Allow undefined update message for EAS Update publishing when no VCS. ([#2148](https://github.com/expo/eas-cli/pull/2148) by [@wschurman](https://github.com/wschurman))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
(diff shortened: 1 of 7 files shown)


wschurman:
/changelog-entry new-feature Allow undefined update message for EAS Update publishing

quinlanj (approved):

quinlanj:
Most of our UI's and tooling rely on people to provide a message to be usable, and we assume most people are using some kind of VCS. No VCS would be a bad user experience, but in the absence of a VCS, we should not be blocking people from publishing.

wschurman:
That's a good point. Maybe we should only allow this to be empty when there's no VCS?

wschurman:
/changelog-entry new-feature Allow undefined update message for EAS Update publishing when no VCS

PR #2149, opened 2023-12-13
[ENG-10412] Upload projectMetadata file

khamilowicz:
# Why

[ENG-10412](https://linear.app/expo/issue/ENG-10412/ui-to-inspect-the-contents-of-eas-build-archive-for-a-build) We want to allow user to easily check the content of build archive. This should streamline debugging for certain cases.

# How

CLI generates file containing a list of files present in project archive. This file is uploaded together with project archive to the same bucket.

The reason for generating this file at this step is that:
- GCP does not provide means to peek into archive content
- The only place on the backend where the project archive is downloaded is in the worker, which is well isolated from other parts of the app, including DBs and GCP, so it would required creating some new service just for the purpose of downloading and listing files in this archive
- it is much easier to do it on user's computer where the archive is created
- data is available immediately, regardless of worker availability

Uploading the file to GCP (as opposed to, for example, storing it in the DB) has following advantages:
- doesn't pollute DB with unnecessary, read-only data, which would be used by a small minority of users
- makes it automatically follow project archive TTL policy - obsolete data is automatically deleted

# Test Plan

Create new build. 
image


khamilowicz:
/changelog-entry new-feature Generate metadata file for project archive

khamilowicz:
/changelog-entry new-feature Generate metadata file for project archive

packages/eas-cli/src/build/build.ts
```diff
@@ -133,9 +133,11 @@ export async function prepareBuildRequestForPlatformAsync<
 
   let projectArchive: ArchiveSource | undefined;
   if (ctx.localBuildOptions.localBuildMode === LocalBuildMode.LOCAL_BUILD_PLUGIN) {
+    const projectPath = (await makeProjectTarballAsync(ctx.vcsClient)).path;
     projectArchive = {
       type: ArchiveSourceType.PATH,
-      path: (await makeProjectTarballAsync(ctx.vcsClient)).path,
+      path: projectPath,
+      metadataLocation: (await makeProjectMetadataFileAsync(projectPath)).path,
```
szdziedzic, line 140: Do we need metadata file for local builds?

packages/eas-cli/src/build/build.ts
```diff
@@ -274,7 +279,24 @@ async function uploadProjectAsync<TPlatform extends Platform>(
...
+          projectMetadataFile.path,
+          createProgressTracker({
+            total: projectMetadataFile.size,
+            message: ratio =>
+              `Uploading metadata to EAS Build (${formatBytes(projectMetadataFile.size * ratio)} / ${formatBytes(
+                projectMetadataFile.size
+              )})`,
+            completedMessage: (duration: string) => `Uploaded to EAS ${chalk.dim(duration)}`,
+          })
+        );
```
szdziedzic, line 297: What does an example path to a metadata file look like?

How long do we want to keep this file around? What is the lifecycle rule for this bucket? 2hrs?

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,6 +83,35 @@ export async function commitPromptAsync(
...
+    onentry: (entry: tar.ReadEntry) => {
+      if (entry.type === 'File') {
+        archiveContent.push(entry.path);
+      }
+    }
+  });
+
+  await fs.writeJSON(metadataLocation, {
+    archiveContent
+  });
```
szdziedzic, line 105: What does the example metadata file look like?

szdziedzic (commented):
It looks reasonable, I just have few more questions before merging it

packages/eas-cli/src/build/build.ts
sjchmiela, line 140: How about we extract it to `const metadataPath` and use in `projectArchive` like this?

packages/eas-cli/src/build/build.ts
```diff
@@ -274,7 +279,24 @@ async function uploadProjectAsync<TPlatform extends Platform>(
             completedMessage: (duration: string) => `Uploaded to EAS ${chalk.dim(duration)}`,
           })
         );
-        return bucketKey;
+
+        const projectMetadataFile = await makeProjectMetadataFileAsync(projectTarball.path);
```
sjchmiela, line 283: Can we make it failsafe? Like, it's not crucial to run the build, so maybe if it fails it's ok and we should only log this error in Sentry or somewhere to know what should we fix?

packages/eas-cli/src/build/build.ts
sjchmiela: Objects in `turtle-v2-projects` are stored for [up to 60 days](https://github.com/expo/universe/blob/6b9314d9b34c6af2f912bd7d544660a0c98df383/terraform/turtle_v2_gcs.tf#L100-L117), but usually they're deleted [right after `retryExpiresAt` by `buildReaper`](https://github.com/expo/turtle-v2/blob/69f901feebc8b3a47063dc62165bf387a5efb72f/src/services/turtle-api/src/storage/db.ts#L114-L124).

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,6 +83,35 @@ export async function commitPromptAsync(
   });
 }
 
+export async function makeProjectMetadataFileAsync(archivePath: string): Promise<{ path: string; size: number }> {
```
sjchmiela, line 86: I think "[project] archive listing" would be a better name than "project metadata". Unless you're thinking of making it reused for other things (ugh, another metadata…), then I would nest the listing under a key of the file, not at the top level.

packages/eas-cli/src/build/utils/repository.ts
khamilowicz:
```
{
    "archiveContent": [
        "paths/to/files"
    ]
}
```

packages/eas-cli/src/build/utils/repository.ts
sjchmiela: Oh, only when you commented I saw it's indeed nested under `archiveContent`, sorry!

packages/eas-cli/src/build/build.ts
khamilowicz: Do you mean:
```
const metadataPath = (await makeProjectMetadataFileAsync(projectPath)).path;
// ...
metadataLocation: metadataPath
```
?

packages/eas-cli/src/build/build.ts
khamilowicz: I can put it in `try`, but I don't see any way to log it to sentry in eas-cli

packages/eas-cli/src/build/build.ts
khamilowicz: @szdziedzic nope, it only affects `projectArchive` files

szdziedzic: @khamilowicz So metadata file will be kept around for 60 days (default bucket policy)?

packages/eas-cli/src/build/build.ts
szdziedzic: Yeah, I think we don't have Sentry configured

packages/eas-cli/src/build/build.ts
```diff
@@ -274,7 +282,28 @@ async function uploadProjectAsync<TPlatform extends Platform>(
...
+                )} / ${formatBytes(projectMetadataFile.size)})`,
+              completedMessage: (duration: string) => `Uploaded to EAS ${chalk.dim(duration)}`,
+            })
+          );
+          return { bucketKey, metadataLocation };
+        } catch (e) {
+          Log.warn('Failed to upload metadata to EAS Build');
+          Log.debug(e);
+          return { bucketKey };
+        }
```
szdziedzic, line 306: Maybe we want also to delete the metadata file the same way we do here https://github.com/expo/eas-cli/pull/2149/files#diff-a19f6ae31352620fe175af46b2a62d16c12d93f2b1014d59ccbf97fe45ac08baR322-R324

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,6 +83,36 @@ export async function commitPromptAsync(
...
+    onentry: (entry: tar.ReadEntry) => {
+      if (entry.type === 'File' && !entry.path.includes('.git')) {
+        archiveContent.push(entry.path);
+      }
+    },
+  });
+
+  await fs.writeJSON(metadataLocation, {
+    archiveContent,
+  });
```
szdziedzic, line 107: maybe we want to wrap it in try-catch and add logic similar to https://github.com/expo/eas-cli/pull/2149/files#diff-877970adc4c19b7884f2246d873bbfd7a36bd66a803ee839566585499c45a77fR144-R149

packages/eas-cli/src/build/build.ts
khamilowicz: @szdziedzic right

packages/eas-cli/src/build/build.ts
```diff
@@ -274,7 +282,33 @@ async function uploadProjectAsync<TPlatform extends Platform>(
...
+          return { bucketKey, metadataLocation };
+        } catch (e) {
+          Log.warn('Failed to upload metadata to EAS Build');
+          Log.debug(e);
+          return { bucketKey };
+        } finally {
+          if (projectMetadataFile) {
+            await fs.remove(projectMetadataFile.path);
+          }
+        }
```
szdziedzic, line 311:
```suggestion
```

szdziedzic: ...and below top-level `try-catch` block:
```ts
        let projectMetadataFile: any;
        try {
          projectMetadataFile = await makeProjectMetadataFileAsync(projectTarball.path);
          const metadataLocation = await uploadFileAtPathToGCSAsync(
            ctx.graphqlClient,
            UploadSessionType.EasBuildGcsProjectSources,
            projectMetadataFile.path,
            createProgressTracker({
              total: projectMetadataFile.size,
              message: ratio =>
                `Uploading metadata to EAS Build (${formatBytes(
                  projectMetadataFile.size * ratio
                )} / ${formatBytes(projectMetadataFile.size)})`,
              completedMessage: (duration: string) => `Uploaded to EAS ${chalk.dim(duration)}`,
            })
          );
          return { bucketKey, metadataLocation };
        } catch (e) {
          Log.warn('Failed to upload metadata to EAS Build');
          Log.debug(e);
          return { bucketKey };
        } finally {
          if (projectMetadataFile) {
            await fs.remove(projectMetadataFile.path);
          }
        }
 ```
 so the `try-catch` aren't nested making it easier to read

File: packages/eas-cli/src/build/graphql.ts [hunk 9]
```diff
@@ -25,6 +25,7 @@ export function transformProjectArchive(archiveSource: ArchiveSource): ProjectAr
     return {
       type: ProjectArchiveSourceType.S3,
       bucketKey: archiveSource.bucketKey,
+      metadataLocation: archiveSource.metadataLocation,
     };
   } else if (archiveSource.type === ArchiveSourceType.GCS) {
     return {
```
szdziedzic, line 31: Add the `metadataLocation` to GCS as well please

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,6 +83,50 @@ export async function commitPromptAsync(
...
+    await fs.writeJSON(metadataLocation, {
+      archiveContent,
+    });
+  } catch (e) {
+    clearTimeout(timer);
+    if (spinner.isSpinning) {
+      spinner.fail();
+    }
+    throw e;
+  }
```
szdziedzic, line 121: Nice! Thanks!

szdziedzic (commented):
It might be good to add deployment order to each PR description, so it is clear in which order the changes should be deployed.

Last few comments:

packages/eas-cli/src/build/graphql.ts
```diff
@@ -25,11 +25,13 @@ export function transformProjectArchive(archiveSource: ArchiveSource): ProjectAr
...
       type: ProjectArchiveSourceType.S3,
       bucketKey: archiveSource.bucketKey,
+      metadataLocation: archiveSource.metadataLocation,
```
szdziedzic, line 28:
```suggestion
```
We are no longer uploading archive sources to S3 AFAIK, so I believe we can delete it

szdziedzic (commented):
I think that it looks good at the moment. Nice! 👏 

Thanks for applying my suggestions 🙇 

My final suggestion is to specify a deployment order, so it is clear what should land when

packages/eas-cli/src/build/build.ts
```diff
@@ -284,15 +311,26 @@ async function uploadProjectAsync<TPlatform extends Platform>(
       }
     );
   } catch (err: any) {
-    let errMessage = 'Failed to upload the project tarball to EAS Build';
+    let errMessage = bucketKey
+      ? 'Failed to upload metadata to EAS Build'
+      : 'Failed to upload the project tarball to EAS Build';
```
sjchmiela, line 316: To me, managing state by existence of `bucketKey` feels too clever.

I would suggest to extract metadata upload to a separate function `uploadMetadataAsync`. This would also remove the inclination to have the same `metadataLocation` in all upload types, which, [as discussed](https://github.com/expo/eas-build/pull/301#discussion_r1426579106), is a little confusing next to `bucketKey`. `bucketKey` is clearly to-be-appended-to-bucket-URL. `metadataLocation` is… what?

sjchmiela: Maybe `metadataBucketKey`? What do you think?

sjchmiela: @khamilowicz what did you mean by resolving this conversation? Have you talked to @szdziedzic and convinced him using `metadataLocation` is better?

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,6 +83,50 @@ export async function commitPromptAsync(
...
+  startTimer(timerLabel);
+
+  const metadataLocation = path.join(getTmpDirectory(), `${uuidv4()}-eas-build-metadata.json`);
+  const archiveContent: string[] = [];
+
+  try {
+    await tar.list({
+      file: archivePath,
+      onentry: (entry: tar.ReadEntry) => {
+        if (entry.type === 'File' && !entry.path.includes('.git')) {
```
sjchmiela, line 106: Should we switch to `startsWith('.git/')` or `startsWith('/.git/')` (I don't know which would be the correct option)? This way we'll also show `.gitignore`, `.github/`.

packages/eas-cli/src/build/build.ts
```diff
@@ -296,6 +318,44 @@ async function uploadProjectAsync<TPlatform extends Platform>(
...
+    if (err.message) {
+      errMessage += `\n\nReason: ${err.message}`;
+    }
+
+    Log.warn(errMessage);
+    return { metadataLocation: null };
+  } finally {
+    if (projectMetadataFile) {
+      await fs.remove(projectMetadataFile);
+    }
```
sjchmiela, line 355: Thanks for making sure this is cleaned up.

packages/eas-cli/src/build/build.ts
```diff
@@ -229,9 +234,16 @@ export function handleBuildRequestError(error: any, platform: Platform): never {
   throw error;
 }
 
-async function uploadProjectAsync<TPlatform extends Platform>(
-  ctx: BuildContext<TPlatform>
-): Promise<string> {
+async function uploadProjectAsync<TPlatform extends Platform>({
+  ctx,
+  generateMetadataFile = true,
```
sjchmiela, line 239: When do we not want to generate metadata file? I can see no `generateMetadataFile: false` anywhere. If that's so, let's remove this option altogether.

Also, a side suggestion — I think `should[Action]` reads better for boolean options than `[action]`. `generateMetadataFile` could be more than just a boolean. With `should…` it's clearer.

packages/eas-cli/src/build/build.ts
sjchmiela: Yup. I still think something like
```ts
} else if (!ctx.localBuildOptions.localBuildMode) {
  const bucketKey = await uploadProjectAsync(ctx);
  const metadataBucketKey = await uploadMetadataFileAsync({ ctx, projectTarball }); // string | null
  projectArchive = {
    type: 'GCS',
    bucketKey,
    metadataBucketKey,
  };
```

would read better, but it's only my opinion. To me, metadata file feels like something separate from the "project", so it shouldn't live in `uploadProject`. On the other hand it already is part of `projectArchive` type, so one could argue that it _is_ part of the `project`.

Either way, my suggestion would be to make keys of `projectArchive` explicit, not implicit (replace spreading for explicit `const { bucketKey, ... } = await` + `{ ..., bucketKey, ... }`.

sjchmiela (commented):
Looks better. A few comments!

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Generate metadata file for project archive ([#2149](https://github.com/expo/eas-cli/pull/2149) by [@khamilowicz](https://github.com/khamilowicz))
```
szdziedzic, line 11: It would be good to rebase this PR and ensure that this changelog entry is not under some old CLI version published a while ago, but under the `main` section, because it's easy to forget about it when rebasing

packages/eas-cli/src/build/build.ts
```diff
@@ -229,9 +235,15 @@ export function handleBuildRequestError(error: any, platform: Platform): never {
...
 }
 
-async function uploadProjectAsync<TPlatform extends Platform>(
-  ctx: BuildContext<TPlatform>
-): Promise<string> {
+async function uploadProjectAsync<TPlatform extends Platform>({
+  ctx,
+}: {
+  ctx: BuildContext<TPlatform>;
+  generateMetadataFile: boolean;
```
szdziedzic, line 242: We should either use `generateMetadataFile` or remove it from function arguments

packages/eas-cli/graphql.schema.json
szdziedzic, line 1: It would be also probably nice to regenerate GQL schema after merging WWW PR

szdziedzic (approved):
Few minor comments, code looks reasonable to me 👍 If the manual testing works fine after merging WWW it should be good to go 👏 

Please rebase/merge main into this PR

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -83,9 +83,54 @@ export async function commitPromptAsync(
...
+  const timerLabel = 'makeProjectMetadataFileAsync';
+  const timer = setTimeout(
+    () => {
+      spinner.start();
+    },
+    Log.isDebug ? 1 : 1000
+  );
+  startTimer(timerLabel);
+
+  const metadataLocation = path.join(getTmpDirectory(), `${uuidv4()}-eas-build-metadata.json`);
```
sjchmiela, line 102: Is the filename preserved — do GCS files have the same name? Maybe it would be better to call them `eas-build-metadata-${uuidv4()}.json` if that's the case?

packages/eas-cli/src/build/build.ts
```diff
@@ -274,7 +282,11 @@ async function uploadProjectAsync<TPlatform extends Platform>(
             completedMessage: (duration: string) => `Uploaded to EAS ${chalk.dim(duration)}`,
           })
         );
-        return bucketKey;
+        const { metadataLocation } = await uploadMetadataFileAsync<TPlatform>(projectTarball, ctx);
+        if (metadataLocation) {
+          return { bucketKey, metadataLocation };
```
sjchmiela, line 287: I remain on the position the fact we have different names (key, location) for the same thing in the same object and the newer name is of less specific meaning is wrong. This comment has been left unanswered for a couple iterations now (first mentioned [here](https://github.com/expo/eas-cli/pull/2149#discussion_r1432764804)) and the conversation marked as resolved twice which is confusing for me.

sjchmiela (approved):

packages/eas-cli/src/graphql/generated.ts
szdziedzic, line 1: Did you regenerate the schema once again by running `yarn generate-graphql-code` 🤔? I'm asking because there are 6k lines of code changed in this file and when I run the command on main I only get 30 lines of change.

Screenshot 2024-02-12 at 23 32 29

khamilowicz: Strange, yesterday I regenerated it and there was a ton of changes, and today it is just 3 lines.

PR #2151, opened 2023-12-13
fix(eas-cli): execute `expo` through a package manager to support isolated modules

byCedric:
# Why

This is a POC for #2150, to test how we can support isolated modules in a good way.

# How

- Removed the existing "resolve expo binary" + "execute expo binary directly".
- Added execute `expo` through the used package manager.

# TODO

- Add back the error handling when `expo` can't be found or executed

# Test Plan

See #2150


szdziedzic (commented):
These changes look reasonable and seem like a good direction 👍. I know that this is a draft but I think we can iterate on it and merge it soon.

Thank you for reporting this issue and working on it 🙏

PR #2153, opened 2023-12-14
[ci] fix `yarn lint-changelog` and use node `18.19.0` by default

szdziedzic:
# Why

`yarn lint-changelog` is failing on main

It broke since our tests started to use the newly published Node `18.19.0`

Seems like an issue with `ts-node`.

# How

Bump default Node used across EAS CLI to `18.19.0`.

Change scripts to use
https://github.com/TypeStrong/ts-node/issues/1997#issuecomment-1537111441

# Test Plan

Tests


sjchmiela (approved):
👌

PR #2157, opened 2023-12-18
[0/1][eas-cli] rename getUpdateGroupJsonInfo

quinlanj:
# Why

Rename `getUpdateGroupJsonInfo` because it is not correct. 

The output of the our `publishUpdateGroups` mutation (which can publish multiple update groups) is an array of updates, belonging to 1 or more groups. With the outputted `Update[]`, we generate some json info with `getUpdateGroupJsonInfo`

# Test Plan

- [ ] typecheck passes

### DIFF
--- a/packages/eas-cli/src/update/utils.ts
+++ b/packages/eas-cli/src/update/utils.ts
@@ -201,8 +201,8 @@ export function formatUpdateTitle(update: UpdateFragment): string {
   )} by ${actorName}, runtimeVersion: ${runtimeVersion}] ${message}`;
 }
 
-export function getUpdateGroupJsonInfo(updateGroups: UpdateFragment[]): UpdateJsonInfo[] {
-  return updateGroups.map(update => ({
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/update/utils.ts
```diff
@@ -201,8 +201,8 @@ export function formatUpdateTitle(update: UpdateFragment): string {
   )} by ${actorName}, runtimeVersion: ${runtimeVersion}] ${message}`;
 }
 
-export function getUpdateGroupJsonInfo(updateGroups: UpdateFragment[]): UpdateJsonInfo[] {
-  return updateGroups.map(update => ({
+export function getUpdatesJsonInfo(updates: UpdateFragment[]): UpdateJsonInfo[] {
```
wschurman, line 204: optional, though I find it clearer

```suggestion
export function getUpdateJsonInfosForUpdates(updates: UpdateFragment[]): UpdateJsonInfo[] {
```

wschurman (approved):

PR #2158, opened 2023-12-18
[1/1][eas-cli] generate eas-update-metadata.json  on publish

quinlanj:
# Why

When we publish an update, a `metadata.json` file gets generated in the `dist` folder. This PR patches the `metadata.json` with some `eas` metadata. 

# motivation
It would be nice to have this for sentry-expo integration. While it is possible for users to specify the `json` flag for the update command to output the same information, it will be nicer if we can tell people to just run `eas update` and then run another script that will read from the `metadata.json` file

# Test Plan

manually tested

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -29,13 +29,13 @@ import {
   collectAssetsAsync,
   defaultPublishPlatforms,
   filterExportedPlatformsByFlag,
+  generateEasMetadataAsync,
   getBranchNameForCommandAsync,
   getRequestedPlatform,
   getRuntimeToPlatformMappingFromRuntimeVersions,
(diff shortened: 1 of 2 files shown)


quinlanj:
/changelog-entry new-feature patch metadata.json with eas info on publish

wschurman (commented):
What do you think about putting this in a separate file? Something like `eas-update-metadata.json` or `eas-update-info.json`?

I worry that putting it in the same file could cause nondeterministic issues for other tooling that depends on `metadata.json` being just the original.

quinlanj:
@wschurman i've amended the PR to generate `eas-update-metadata.json` instead

wschurman (approved):

szdziedzic (commented):
It looks reasonable to me, but I'm leaving the review for the others, more familiar with the matter

PR #2166, opened 2024-01-05
[ci] Cancel in progress jobs when new are run

sjchmiela:
# Why

I've noticed tests ran for an old commit even though I pushed a new one.

# How

Added `concurrency` where I thought it was applicable.

# Test Plan

None.


szdziedzic (approved):
Looks reasonable

PR #2175, opened 2024-01-11
[ENG-10799][ci][eas-cli][eas-json] test against node 20 and drop support for node 14

szdziedzic:
# Why

Node 14 is no longer maintained, Node 20 is the current active

Similar to https://github.com/expo/eas-cli/pull/1098

# How

Test against Node 20, make it the default Node version used to work with EAS CLI.

Drop support for Node 14, it's not maintained for a longer while (even 16 is not maintained anymore). Changing the `engines` value to `14.0.0` will make it fail for the people trying to install EAS CLI when using Node 14.

# Questions

Do we want to announce this change publicly or there is no need for additional communication (Maybe because of the reasons similar to https://github.com/expo/eas-cli/pull/1098#pullrequestreview-977382502?) and we can just release it? It will make people using Node 14 unable to install new EAS versions. 

End of life date for Node 14 was April 30, 2023.

# Test Plan

Tests, test manually using Node 20


sjchmiela (approved):
Do we let users know that there's a newer `eas-cli` available if they're running an older version? If so, I think I'm ok with bumping major version, since it's a breaking change, and not doing any more communication. I'm also open to other solutions.

szdziedzic:
>Do we let users know that there's a newer eas-cli available if they're running an older version? 

yes

szdziedzic:
/changelog-entry breaking-change Drop support for Node 14

brentvatne (approved):

PR #2176, opened 2024-01-12
[ENG-11071][eas-json][eas-cli] bump dependencies and fix errors/warnings

szdziedzic:
# Why

Bump dependencies

# How

Bump dependencies. Fix errors caused by bumping. Fix `jest` warnings, make `jest` configs use TS.

# Test Plan

Tests + run some commands manually


szdziedzic:
/changelog-entry chore Update `eas-cli` and `@expo/eas-json` dependencies

szdziedzic:
Oh well...

After the `eslint`, `prettier`, and eslint rules update I was seeing errors due to incorrect order of imports and missing whitespaces between import groups. I ran the `yarn lint --fix` command to fix these but it edited 160 files making it slightly harder to review. From what I see all of these are import reorders or adding whitespaces between imports.

This is the commit that edited these files https://github.com/expo/eas-cli/pull/2176/commits/2b0a99fad047e8b20ffc121d68ca1271f7ada986

I hope it isn't a big problem when it comes to reviewing this PR.

szdziedzic:
https://github.com/expo/eas-cli/actions/runs/7501659185/job/20422721322?pr=2176 - it fails because it checkouts `main` version of the codebase and runs it's against the `lerna` version from this branch, which is newer and incompatible with `useWorkspaces` option used in `lerna.json` option on `main` branch. I think we can safely ignore it.

packages/eas-cli/src/build/ios/__tests__/version-test.ts
```diff
@@ -62,7 +62,7 @@ describe(evaluateTemplateString, () => {
 // bare workflow
 describe(bumpVersionAsync, () => {
   beforeEach(() => {
-    getPbxproj.mockImplementation(() => ({} as any));
+    getPbxproj.mockImplementation(() => ({}) as any);
```
radoslawkrzemien, line 65: `({})` is a bit weird

radoslawkrzemien (approved):
Noice 🚀

wschurman (approved):

PR #2178, opened 2024-01-12
[eas-cli] [ENG-11026] Add build:delete command

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-11026/add-eas-builddelete-command

# How

Added a new `build:delete` command. It shares some similarities with `build:cancel` command, can be used with build id provided or prompt the user for id if not (not available in non-interactive mode). The difference is that `build:cancel` has restrictions on build status, and `build:delete` does not. The shared code has been extracted to a helper file under `src/commandUtils`. Additionally, 2 filter flags has been added for the case when the user does not provide the ID of the build to be deleted and a list of builds is printed, it can be filtered by `platform` and/or `profile`. The same has been applied to `build:cancel` as well

### With build id
1
Screenshot 2024-01-12 at 15 30 48
2
Screenshot 2024-01-12 at 15 31 30

### With prompt
1
Screenshot 2024-01-12 at 15 32 05
2
Screenshot 2024-01-12 at 15 32 16
3
Screenshot 2024-01-12 at 15 32 44

### Non-interactive with id
1
Screenshot 2024-01-12 at 15 45 44

### Non-interactive with prompt (not allowed)
1
Screenshot 2024-01-12 at 15 47 04

### No flags
Screenshot 2024-01-16 at 13 49 46

### `build:delete` with `platform/p` flag
1
Screenshot 2024-01-16 at 13 50 08
2
Screenshot 2024-01-16 at 13 50 19

### `build:delete` with `profile/e` flag
1
Screenshot 2024-01-16 at 13 50 50
2
Screenshot 2024-01-16 at 13 51 38

### Same for `build:cancel` command
1
Screenshot 2024-01-16 at 13 52 03
2
Screenshot 2024-01-16 at 13 52 14
3
Screenshot 2024-01-16 at 13 52 24

### Using both build ID and filter flags is not allowed
1
Screenshot 2024-01-16 at 15 25 39
2
Screenshot 2024-01-16 at 15 27 06

# Test Plan

Added tests similar to those for `build:cancel`
Tested manually

# Deployment Plan

No dependency on other services, just release a new version of CLI when needed

### DIFF
--- a/packages/eas-cli/src/commandUtils/builds.ts
+++ b/packages/eas-cli/src/commandUtils/builds.ts
@@ -14,7 +14,7 @@ import { ExpoGraphqlClient } from './context/contextUtils/createGraphqlClient';
 
 const platformToAppPlatform: Record<
   Exclude<RequestedPlatform, RequestedPlatform.All>,
-  AppPlatform | undefined
+  AppPlatform
 > = {
   android: AppPlatform.Android,
   ios: AppPlatform.Ios,


packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+  } else {
+    for (const status of statuses) {
+      const buildsForStatus = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
+        appId: projectId,
+        offset: 0,
+        limit: 10,
+        filter: { status },
+      });
+      builds.push(...buildsForStatus);
```
sjchmiela, line 39: Should we order them by `createdAt`?

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,68 @@
...
+    await BuildQuery.byIdAsync(graphqlClient, buildId);
+  } catch {
+    throw new Error(`Couldn't find a build matching the id ${buildId}`);
+  }
+}
+
+export async function fetchBuildsAsync(
+  graphqlClient: ExpoGraphqlClient,
+  projectId: string,
+  statuses?: BuildStatus[]
```
sjchmiela, line 22: I would use a `{ graphqlClient, projectId, statuses }` argument. This way, if we add more filters (like `platform: Platform[]`) it's much easier and readable.

packages/eas-cli/src/commands/build/delete.ts
```diff
@@ -0,0 +1,122 @@
...
+    builds = await fetchBuildsAsync(graphqlClient, projectId);
+    spinner.stop();
+  } catch (error) {
+    spinner.fail(
+      `Something went wrong and we couldn't fetch the builds for the project ${projectDisplayName}.`
+    );
+    throw error;
+  }
+  if (builds.length === 0) {
+    Log.warn(`There aren't any builds for the project ${projectDisplayName}.`);
```
sjchmiela, line 55:
```suggestion
    Log.warn(`There are no builds for the project ${projectDisplayName}.`);
```
Or maybe something like "we couldn't find any builds"? I think it feels more human?

packages/eas-cli/src/commands/build/delete.ts
```diff
@@ -0,0 +1,122 @@
...
+    const displayName = await getDisplayNameForProjectIdAsync(graphqlClient, projectId);
+
+    if (buildIdFromArg) {
+      await ensureBuildExistsAsync(graphqlClient, buildIdFromArg);
+    }
+
+    let buildId: string | null = buildIdFromArg;
+    if (!buildId) {
+      if (nonInteractive) {
+        throw new Error('BUILD_ID must not be empty in non-interactive mode');
```
sjchmiela, line 104: Do users know the argument is called `BUILD_ID`? If not, maybe we should changes this to say "The build ID argument"?

sjchmiela (approved):
Cool!

packages/eas-cli/src/commands/build/delete.ts
```diff
@@ -0,0 +1,122 @@
...
+    }))
+      ? buildId
+      : null;
+  }
+}
+
+export default class BuildDelete extends EasCommand {
+  static override description = 'delete a build';
+  static override args = [{ name: 'BUILD_ID' }];
+  static override flags = {
```
szdziedzic, line 77: Maybe we can allow users to filter by platform and profile as well?

szdziedzic (commented):
It looks really good 🙌

packages/eas-cli/src/commandUtils/builds.ts
radoslawkrzemien: I think they are sorted like that implicitly, but need to double check

packages/eas-cli/src/commands/build/delete.ts
radoslawkrzemien: The message was copied from an existing command, but I can update both 😄

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,68 @@
...
+      limit: 10,
+    });
+  } else {
+    for (const status of statuses) {
+      const buildsForStatus = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
+        appId: projectId,
+        offset: 0,
+        limit: 10,
+        filter: { status },
+      });
```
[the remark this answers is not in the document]
radoslawkrzemien: It's so obvious, and yet I always forget about that 😬

packages/eas-cli/src/commands/build/delete.ts
radoslawkrzemien: WDYM? If they provide platform to delete all builds from a platform? That seems rather... prone to human errors

packages/eas-cli/src/commandUtils/builds.ts
radoslawkrzemien: They were, but it fell apart when combining results for different statuses, as this would group the builds by status too. Added explicit sorting before returning

packages/eas-cli/src/commandUtils/builds.ts
radoslawkrzemien: Updated

packages/eas-cli/src/commands/build/delete.ts
radoslawkrzemien: Updated

packages/eas-cli/src/commands/build/delete.ts
radoslawkrzemien: Good point

radoslawkrzemien: Updated

packages/eas-cli/src/commands/build/delete.ts
szdziedzic: Oh, I meant, when displaying the list of builds. So the people can decide what filters should be applied to this list.

packages/eas-cli/src/commandUtils/builds.ts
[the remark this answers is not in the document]
radoslawkrzemien: Updated

khamilowicz (approved):

packages/eas-cli/src/commands/build/delete.ts
radoslawkrzemien: Added the filter flags for `platform` and `profile`

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,91 @@
...
+  filters,
+}: {
+  graphqlClient: ExpoGraphqlClient;
+  projectId: string;
+  filters?: { statuses?: BuildStatus[]; platform?: string; profile?: string };
+}): Promise<BuildFragment[]> {
+  let builds: BuildFragment[];
+  const queryFilters: InputMaybe<BuildFilter> = {};
+  if (filters?.platform && filters?.platform !== 'all') {
+    queryFilters['platform'] = filters?.platform.toUpperCase() as AppPlatform;
```
szdziedzic, line 38: What do you think about using some mapping from one type to the other type here? By doing this TS will warn us if something type-related changes and it will be more explicit (for example we will see if due to some changes the `toUpperCase` isn't enough)

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (!filters?.statuses) {
+    builds = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
+      appId: projectId,
+      offset: 0,
+      limit: 10,
+      filter: queryFilters ? queryFilters : undefined,
+    });
+  } else {
+    builds = (
+      await Promise.all(
```
szdziedzic, line 52: Thanks!

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -116,24 +89,24 @@ export async function selectBuildToCancelAsync(
...
+      char: 'p',
+      description: 'Platform for which to list builds when ID not provided',
+      options: ['android', 'ios', 'all'],
+    }),
+    profile: Flags.string({
+      char: 'e',
+      description:
+        'Name of the build profile from eas.json. Defaults to "production" if defined in eas.json.',
+      helpValue: 'PROFILE_NAME',
+    }),
```
szdziedzic, line 109: This is probably a nitpick, but can we throw an error if the combination of build ID and these flags is provided? I want to avoid any confusion for the users here. What I mean is a situation where users believe this flag will only delete a build with such id if it is of a given platform and profile.

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -116,24 +89,24 @@ export async function selectBuildToCancelAsync(
...
 export default class BuildCancel extends EasCommand {
   static override description = 'cancel a build';
 
   static override args = [{ name: 'BUILD_ID' }];
 
   static override flags = {
     ...EASNonInteractiveFlag,
+    platform: Flags.string({
+      char: 'p',
+      description: 'Platform for which to list builds when ID not provided',
```
szdziedzic, line 101:
```suggestion
      description: 'Filter builds by the platform if BUILD_ID is not provided',
```
WDYT?

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -116,24 +89,24 @@ export async function selectBuildToCancelAsync(
...
+      options: ['android', 'ios', 'all'],
+    }),
+    profile: Flags.string({
+      char: 'e',
+      description:
+        'Name of the build profile from eas.json. Defaults to "production" if defined in eas.json.',
```
szdziedzic, line 107:
```suggestion
        'Filter builds by profile if BUILD_ID is not provided',
```

szdziedzic (commented):
Nice! I have a few last suggestions 🙌

packages/eas-cli/src/commands/build/cancel.ts
radoslawkrzemien: That's a valid point

radoslawkrzemien: Updated

packages/eas-cli/src/commands/build/cancel.ts
radoslawkrzemien: Updated

packages/eas-cli/src/commands/build/cancel.ts
radoslawkrzemien: Updated

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  const status = chalk.blue(statusText);
+  return `${platform} Started at: ${startTime}, Status: ${status}, Id: ${build.id}`;
+}
+
+function toAppPlatform(platform: string): AppPlatform {
+  const capitalizedPlatform = (platform[0].toUpperCase() +
+    platform.substring(1)) as keyof typeof AppPlatform;
+  return AppPlatform[capitalizedPlatform];
+}
```
sjchmiela, line 97: We could also do
```
const platformToAppPlatform: Record<string, AppPlatform | undefined> = {
  android: AppPlatform.Android,
  ios: AppPlatform.Ios,
};
```

A bit simpler and safer.

radoslawkrzemien: I like that

packages/eas-cli/src/commandUtils/builds.ts
```diff
@@ -0,0 +1,96 @@
...
+  InputMaybe,
+} from '../graphql/generated';
+import { BuildQuery } from '../graphql/queries/BuildQuery';
+import { appPlatformEmojis } from '../platform';
+import { ExpoGraphqlClient } from './context/contextUtils/createGraphqlClient';
+
+const platformToAppPlatform: Record<string, AppPlatform | undefined> = {
+  android: AppPlatform.Android,
+  ios: AppPlatform.Ios,
+};
```
szdziedzic, line 18:
```suggestion
const platformToAppPlatform: Record<
  Exclude<RequestedPlatform, RequestedPlatform.All>,
  AppPlatform
> = {
  android: AppPlatform.Android,
  ios: AppPlatform.Ios,
};
```

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -116,24 +89,23 @@ export async function selectBuildToCancelAsync(
...
 
   static override args = [{ name: 'BUILD_ID' }];
 
   static override flags = {
     ...EASNonInteractiveFlag,
+    platform: Flags.string({
+      char: 'p',
+      description: 'Filter builds by the platform if build ID is not provided',
+      options: ['android', 'ios', 'all'],
+    }),
```
szdziedzic, line 103:
```suggestion
    platform: Flags.enum({
      char: 'p',
      description: 'Filter builds by the platform if build ID is not provided',
      options: Object.values(RequestedPlatform),
    }),
```

packages/eas-cli/src/commands/build/cancel.ts
```diff
@@ -145,8 +117,15 @@ export default class BuildCancel extends EasCommand {
...
       args: { BUILD_ID: buildIdFromArg },
-      flags: { 'non-interactive': nonInteractive },
+      flags: { 'non-interactive': nonInteractive, platform, profile },
     } = await this.parse(BuildCancel);
+
+    if (buildIdFromArg && (platform || profile)) {
+      throw new Error(
+        'Build ID cannot be used together with platform and profile flags. They are used to filter the list of builds when not providing the build ID'
+      );
+    }
```
szdziedzic, line 127: ❤️

packages/eas-cli/src/commands/build/delete.ts
```diff
@@ -0,0 +1,147 @@
...
+    ...EASNonInteractiveFlag,
+    platform: Flags.string({
+      char: 'p',
+      description: 'Filter builds by the platform if build ID is not provided',
+      options: ['android', 'ios', 'all'],
+    }),
```
szdziedzic, line 88:
```suggestion
    platform: platform: Flags.enum({
      char: 'p',
      description: 'Filter builds by the platform if build ID is not provided',
      options: Object.values(RequestedPlatform),
    }),
```

szdziedzic (approved):
Please don't kill me, these are the last 3 comments that can allow you to type everything better 🙈

PR #2179, opened 2024-01-12
[eas-cli] Remove bare-specific eas.json template

sjchmiela:
# Why

Simplification, unification.

# How

Removed bare-specific template.

# Test Plan

None.


sjchmiela:
/changelog-entry chore Remove "bare"-specific **eas.json** template.

szdziedzic (approved):
👍

PR #2184, opened 2024-01-17
[ENG-11092][eas-json] update `eas.schema.json` to add support for Xcode 15.2 image

szdziedzic:
# Why

Update `eas.schema.json` after adding Xcode 15.2 image

# How

Update `eas.schema.json` after adding Xcode 15.2 image


szdziedzic:
/changelog-entry chore Update `eas.schema.json` to after adding Xcode 15.2 image

radoslawkrzemien (approved):

PR #2185, opened 2024-01-18
[ENG-11076][eas-cli] prompt user to map channel on update if unmapped

quinlanj:
# Why
When an update is created with the `channel` flag, it will error if the channel is not mapped to a branch. 

```
> eas update --channel main
Channel has no branches associated with it. Run 'eas channel:edit' to map a branch
```

This PR allows the user to select the branch to map in this case:
```
>  eas-dev update --channel blah
Channel has no branches associated with it.
✔ Which branch would you like blah to point at? › gh/quinlanj/3/orig
```

# Test Plan

- [ ] amended test to expect new behaviour

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This is the log of notable changes to EAS CLI and related packages.
 - Remove "bare"-specific **eas.json** template. ([#2179](https://github.com/expo/eas-cli/pull/2179) by [@sjchmiela](https://github.com/sjchmiela))
 - Prompt users if they want to continue if EAS CLI fails to provision the devices. ([#2181](https://github.com/expo/eas-cli/pull/2181) by [@szdziedzic](https://github.com/szdziedzic))
 - Update `eas-cli` and `@expo/eas-json` dependencies. ([#2176](https://github.com/expo/eas-cli/pull/2176) by [@szdziedzic](https://github.com/szdziedzic))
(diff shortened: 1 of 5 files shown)


quinlanj:
/changelog-entry chore  prompt user to map channel on update if unmapped

szdziedzic (approved):

wschurman:
It feels awkward to me that we allow creating an update with just a channel flag since updates are associated with branches instead of channels. I assume the channel flag was added as a convenience flag for singly-mapped channel-branch combinations? 

PR that added it makes it seem like this behavior might have been intentional: https://github.com/expo/eas-cli/pull/1567 
cc @jonsamp

wschurman:
I guess my other worry is that this might be too heavy of an operation for people to comprehend what is being done. By pointing an existing channel at an existing branch, any other/existing builds with that channel will now receive compatible updates on that branch. I worry that people running a simple `eas update` command may not comprehend the potential magnitude of such an operation and may just answer "yes" out of habit, which could have unintended consequences in production.

wschurman (changes requested):
Requesting changes for now to put it back in your queue to assess.

quinlanj:
@wschurman you raise a really good point in https://github.com/expo/eas-cli/pull/2185#issuecomment-1899035520 . People will have to select a branch to map to a channel (they don't reply `y`), but I agree someone could just be pressing the Return key out of habit, not expecting that some variant of `eas update` would change the mapping of their channel. 

Closing this PR to retain the original behaviour.

PR #2191, opened 2024-01-19
Use credentials types from eas-build-job where possible

sjchmiela:
# Why

I've noticed we have two very similar types in `eas-build` and `eas-cli`. The types in `eas-cli` are just intermediate — in the end, when we hand off the build request to `www` we change them into the `eas-build` ones.

Wouldn't it be nice if we used the target types from the beginning?

# How

Removed `TargetCredentials` in favor of `Ios.TargetCredentials`. `IosCredentials` in favor of `Ios.BuildCredentials`. Basically, that's it!

Also made some minor change to `credentialsJson/types` so that less types are exported so we are forced to try to use the `eas-build-job` ones.

# Test Plan

_TODO_


sjchmiela:
/changelog-entry chore Migrate to using credentials types from `eas-build-job` where possible

PR #2197, opened 2024-01-23
[fcmv1] Manage FCM V1 Google Service Account Key in CLI

christopherwalter:
# Why
NOTE: DO NOT LAND THIS UNTIL GRAPHQL CHANGES LAND IN WWW

We're adding support for FCM V1 credentials, as Google is shutting down the FCM Legacy API for sending Android notifications in June.

# How
Add new prompts and refactor existing prompts to match this schematic:
eas-cli

# Test Plan
Verified all functionality e2e:
fcm_v1_eas_cli_manage_creds


packages/eas-cli/graphql-codegen.yml
```diff
@@ -16,7 +14,7 @@ generates:
       dedupeOperationSuffix: true
     hooks:
       afterOneFileWrite:
-        - 'node ./scripts/annotate-graphql-codegen.js'
+        - ./annotate-graphql-codegen.sh
```
quinlanj, line 17: Confirming this is an intentional change? I see `./scripts/annotate-graphql-codegen.js` in the repo but not `./annotate-graphql-codegen.sh`.

packages/eas-cli/src/credentials/android/actions/AssignGoogleServiceAccountKeyForFcmV1.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  CommonAndroidAppCredentialsFragment,
+  GoogleServiceAccountKeyFragment,
+} from '../../../graphql/generated';
+import Log from '../../../log';
+import { CredentialsContext } from '../../context';
+import { AppLookupParams } from '../api/GraphqlClient';
+
+export class AssignGoogleServiceAccountKeyForFcmV1 {
```
quinlanj, line 9: Let's refactor this into the existing [AssignGoogleServiceAccountKey](https://github.com/expo/eas-cli/blob/eb9152aa9c9590ca8720496b22c6085776217428/packages/eas-cli/src/credentials/android/actions/AssignGoogleServiceAccountKey.ts#L9). We can have a parameter for the service we want to assign the GSAK to (submissions vs FCM V1)

packages/eas-cli/src/credentials/android/actions/CreateGoogleServiceAccountKeyForFcmV1.ts
```diff
@@ -0,0 +1,76 @@
...
+import Log, { learnMore } from '../../../log';
+import { promptAsync } from '../../../prompts';
+import { CredentialsContext } from '../../context';
+import { GoogleServiceAccountKey } from '../credentials';
+import {
+  detectGoogleServiceAccountKeyPathAsync,
+  readAndValidateServiceAccountKey,
+} from '../utils/googleServiceAccountKey';
+
+export class CreateGoogleServiceAccountKeyForFcmV1 {
```
quinlanj, line 14: This does the same thing as [CreateGoogleServiceAccountKey](https://github.com/expo/eas-cli/blob/eb9152aa9c9590ca8720496b22c6085776217428/packages/eas-cli/src/credentials/android/actions/CreateGoogleServiceAccountKey.ts#L14), so it is not necessary

packages/eas-cli/src/credentials/android/actions/SetUpGoogleServiceAccountKeyForFcmV1.ts
```diff
@@ -0,0 +1,87 @@
...
+  CommonAndroidAppCredentialsFragment,
+  GoogleServiceAccountKeyFragment,
+} from '../../../graphql/generated';
+import Log from '../../../log';
+import { promptAsync } from '../../../prompts';
+import { CredentialsContext } from '../../context';
+import { MissingCredentialsNonInteractiveError } from '../../errors';
+import { AppLookupParams } from '../api/GraphqlClient';
+
+export class SetUpGoogleServiceAccountKeyForFcmV1 {
```
quinlanj, line 16: Let's refactor this in a similar way to the AssignGSAK class. We'll have a single `SetUpGoogleServiceAccountKey` that takes a parameter as input to determine whether we will set up the GSAK for submissions or fcmv1

packages/eas-cli/src/credentials/android/utils/googleServiceAccountKey.ts
```diff
@@ -96,7 +96,7 @@ export async function detectGoogleServiceAccountKeyPathAsync(
 ): Promise<string | null> {
   const foundFilePaths = await glob('**/*.json', {
     cwd: projectDir,
-    ignore: ['app.json', 'package*.json', 'tsconfig.json', 'node_modules'],
+    ignore: ['app.json', 'package*.json', 'tsconfig.json', 'node_modules', 'google-services.json'],
```
quinlanj, line 99: 💯

quinlanj (changes requested):
I ran this version of `eas-cli` against your changes in `www` and the UX looks good. Thank you! Requesting changes for deduping classes 🙏

packages/eas-cli/graphql-codegen.yml
christopherwalter: nope not intentional, will remove

packages/eas-cli/src/credentials/android/actions/CreateGoogleServiceAccountKeyForFcmV1.ts
christopherwalter: I left this as its own class because of the logging/erroring that's specific to FCM V1. I could put some conditional logic in `CreateGoogleServiceAccountKey` to render FCM-specific logs or not based on args from the caller, but this seemed cleaner to me.

packages/eas-cli/src/credentials/android/actions/SetUpGoogleServiceAccountKeyForFcmV1.ts
christopherwalter: Similar reasoning to above, it seems premature/not super necessary to abstract this given there's only two use cases for it. There's definitely functionality overlap but it seems like a wash given there's only 2 use cases.

quinlanj:
### Assigning actions
I agree with you here -- these actions makes sense to keep the submission and fcmv1 separate.
AssignGoogleServiceAccountKey -  Let's rename this class to `AssignGoogleServiceAccountKeyForSubmissions`: https://github.com/expo/eas-cli/blob/eb9152aa9c9590ca8720496b22c6085776217428/packages/eas-cli/src/credentials/android/actions/AssignGoogleServiceAccountKey.ts#L9

### Create GSAK actions
There will be 3 'contexts' for creation:
1. generic `Upload a GSAK` option
2. setting up a submission GSAK (creates and assigns for submissions)
3. setting up an FCM v1 GSAK (creates and assigns for FCMV1)

IMO the most important bit of logging here is this [fyi bit](https://github.com/expo/eas-cli/blob/eb9152aa9c9590ca8720496b22c6085776217428/packages/eas-cli/src/credentials/android/actions/CreateGoogleServiceAccountKey.ts#L52). For the generic case, we'd want the fyi to link a generic page which lists the pages for submissions and fcmv1. For the specialized FCM and Submission cases, we'd want the fyi to lead to their respective instructions for the permissions the user needs to assign to that GSAK. 

My vote is to refactor this one into a single class since there are 3 logging contexts, but the underlying logic is the same

### SetUp GSAK actions
I'm ok with what you have here 👍

szdziedzic (approved):
It looks reasonable to me 👍 

Thanks for sharing the video from e2e testing the change and thanks for taking care of it 🙌

christopherwalter:
> IMO the most important bit of logging here is this [fyi bit](https://github.com/expo/eas-cli/blob/eb9152aa9c9590ca8720496b22c6085776217428/packages/eas-cli/src/credentials/android/actions/CreateGoogleServiceAccountKey.ts#L52). For the generic case, we'd want the fyi to link a generic page which lists the pages for submissions and fcmv1. For the specialized FCM and Submission cases, we'd want the fyi to lead to their respective instructions for the permissions the user needs to assign to that GSAK.

@quinlanj I removed CreateGoogleServiceAccountKeyForFcmV1 and updated the logging in `CreateGoogleServiceAccountKey` to include language about both the submission use case as well as the notifications use case. I also have this [PR](https://github.com/expo/fyi/pull/146) up to update the FYI docs for the two use cases. 

Let me know what you think.

packages/eas-cli/src/credentials/manager/AndroidActions.ts
```diff
@@ -92,25 +92,61 @@ export const fcmActions: ActionInfo[] = [
...
+    scope: Scope.Manager,
+  },
+  {
+    value: AndroidActionType.ManageGoogleServiceAccountKeyForFcmV1,
+    title: 'Manage your Google Service Account Key for Push Notifications (FCM V1)',
+    scope: Scope.Manager,
+  },
+  {
+    value: AndroidActionType.CreateGsaKey,
+    title: 'Upload a Google Service Account Key to the Keystore',
```
quinlanj, line 144: or maybe `Upload a Google Service Account Key to Expo`? A keystore is a type of Android build credential, so best not to confuse folks with that term.
```suggestion
    title: 'Upload a Google Service Account Key',
```

packages/eas-cli/src/credentials/manager/AndroidActions.ts
```diff
@@ -92,25 +92,61 @@ export const fcmActions: ActionInfo[] = [
...
     scope: Scope.Project,
   },
   {
     value: AndroidActionType.RemoveGsaKey,
-    title: 'Delete a Google Service Account Key',
+    title: 'Delete a Google Service Account Key from the Keystore',
```
quinlanj, line 149: ditto
```suggestion
    title: 'Delete a Google Service Account Key',
```

quinlanj (approved):

christopherwalter:
/changelog-entry new-feature Support configuring a Google Service Account Key via `eas credentials`, for sending Android Notifications via FCM V1

christopherwalter:
/changelog-entry new-feature Support configuring a Google Service Account Key via eas credentials, for sending Android Notifications via FCM V1

PR #2214, opened 2024-02-06
Revert "feat(eas-cli): expose expo export dev flag as an option in eas update (#2050)"

wschurman:
# Why

This reverts commit 8751a2835f38e4be5079c7f2ffe7129f9a060301.

The reason we're reverting this is that it is causing some confusion around whether dev bundles work with EAS update. The answer is that they probably shouldn't, or at least aren't officially supported.

> "Dev" bundles are quite different from production bundles (no minification, possibly different Babel presets running, different NODE_ENV is set, and `__DEV__` value is different). The value of `__DEV__` is used in many places to change behavior to fit a development environment, for example in [expo-asset](https://github.com/expo/expo/blob/dd5f37fc1a8df22b7ae1da98b085b9c5ad06dcd1/packages/expo-asset/src/AssetSourceResolver.ts#L13-L23). If we want to support loading updates that are published in dev mode, we should ensure we have adequate test coverage and that this is something we are consciously supporting.

Instead, to address the original concern in https://github.com/expo/eas-cli/pull/2050 (accessing the dev menu), we'll address that issue in a more targeted manner: by making accessing the dev menu even in places when `__DEV__` isn't set.

Alternatively, one could theoretically do the export step manually during an `eas update` by running the `npx expo export` command manually and then running `eas update --skip-bundler`.

Closes ENG-11290.

# How

Revert.

# Test Plan

CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Revert expose expo export dev flag as an option in eas update. ([#2214](https://github.com/expo/eas-cli/pull/2214) by [@wschurman](https://github.com/wschurman))
+
 ### 🧹 Chores
 
 - Remove duplicated log message when creating ASC API key. ([#2208](https://github.com/expo/eas-cli/pull/2208) by [@radoslawkrzemien](https://github.com/radoslawkrzemien))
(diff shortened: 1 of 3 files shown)


brentvatne (approved):

quinlanj (approved):

wschurman:
/changelog-entry bug-fix Revert expose expo export dev flag as an option in eas update

PR #2216, opened 2024-02-06
Add workflow for bumping eas-build dependencies

sjchmiela:
# Why

It is currently not possible to use `if: ${ always() }` in Custom Build workflows due to outdated `@expo/steps` dependency.

# How

Added a GitHub workflow, similar to the one we have in `turtle-v2`, which will automatically upgrade all dependencies from the `eas-build` repo.

# Test Plan

https://github.com/expo/eas-cli/pull/2220


szdziedzic (approved):
Looks good

Issue #1757, opened 2023-03-26
Request for documentation on .easignore and requireCommit in eas.json


andy-HM:
Hello,

I am writing to request better documentation on the usage of `.easignore` and the `requireCommit` property in the `eas.json`.  I am finding that documentation in these areas are severely lacking in the Expo documentation.  

At the time of writing this, there are 0 results returned from searching "easignore" in the[ Expo documentation site](https://docs.expo.dev/). 

Searching for "requireCommit" does return results in the search but those are mostly instances where `requireCommit` appears in examples of `eas.json`.

What I am looking for is:

1. How does one properly use `.easignore`
2. How does `requireCommit` behave when set to `false`? When set to false in my case, the uploaded files are half the size as when set to true.  Then the build fails because files are missing.
3. Do these two functions work together?  Are they separate?
4. Any other best practices in these areas

Thank you.

wkozyra95:
https://expo.fyi/eas-build-archive

1. In the root of a repository (it replaces all dockerfiles), but .easignore is subdirectories do not do anything(.dockerignore works the same way)
2. false is a default
3. No, `.easignore` does not work with with `requireCommit: true`. `requireCommit: true` is making shallow clone using git(so it can't respect .easignore), and `requireCommit: false` is packaging files using our implementations that slightly diverges from git specs. 
4. Don't use .easignore, we added it because it was quite common request, but there is always a better approach. Also `requireCommit: true` would also be consider a best practice, but it was to annoying for some workflows, so we changed the default to false.

You can use `eas build:inspect --stage archive to check what will be uploaded)

andy-HM:
hello @wkozyra95 thank you for the reply!  This is incredibly helpful.  I was not aware of this level of documentation on Github for Expo.

dtx:
@wkozyra95's reply should be a part of the documentation. lol.

macksal:
I was bitten by this - previously using `.easignore` but it stopped applying after setting `requireCommit: true`. This is a little confusing: `requireCommit` feels like it would a safety feature to make sure the working tree isn't dirty, not a fundamental change to the archiving method.

Would it be appropriate to print a warning if `requireCommit: true` is specified while a `.easignore` file is also present?

bj97301:
if i understand it correctly, if theres one file i want to be included in my build(google-services.json) i should duplicate my .gitignore file and then remove the file(s) i want to be included in the build.  So for me just the google-services.json file. Sounds about right?

PR #2225, opened 2024-02-10
[eas-cli] Add support for updated fingerprint policy to eas update

wschurman:
# Why

Corresponding eas-cli change to https://github.com/expo/expo/pull/26901 for the updates part.

This does fingerprint runtime version policy evaluation during `eas update`.

# How

Intercept the calls to get the runtime version and, if fingerprint, run the fingerprint.

I still need to figure out how to enforce that the native directories are in a good state (have been prebuilt if necessary). TBH I'm not sure how we'll do this since we can't tell whether the project uses prebuild here. I guess we could just check for the platform native directory (`ios/` or `android/` respectively) and if it isn't there throw?

# Test Plan

This is just an RFC for now to go along with https://github.com/expo/expo/pull/26901, so it hasn't been tested fully end-to-end yet.

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -1,6 +1,7 @@
 import { ExpoConfig, Platform } from '@expo/config';
 import { Updates } from '@expo/config-plugins';
 import { Platform as EASBuildJobPlatform, Workflow } from '@expo/eas-build-job';
+import * as Fingerprint from '@expo/fingerprint';
 import JsonFile from '@expo/json-file';
 import assert from 'assert';
 import chalk from 'chalk';
@@ -8,7 +9,6 @@ import crypto from 'crypto';
 import fs from 'fs-extra';
 import Joi from 'joi';
 import mime from 'mime';
-import nullthrows from 'nullthrows';
 import path from 'path';
 import promiseLimit from 'promise-limit';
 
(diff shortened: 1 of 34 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -681,43 +681,73 @@ export async function getRuntimeVersionObjectAsync(
...
+  if (platform === 'web') {
+    return 'UNVERSIONED';
+  }
+
+  const runtimeVersion = exp[platform]?.runtimeVersion ?? exp.runtimeVersion;
+  if (typeof runtimeVersion === 'object') {
+    const policy = runtimeVersion.policy;
+
+    if ((policy as any) === 'fingerprintNativeExperimental') {
+      // TODO(wschurman): ensure prebuild has been run
```
wschurman, line 714: this is the main concern we need to address. how do we ensure that the native directories are in a "clean" state, i.e. the same state they are in which the fingerprint is generated at build time.

wschurman:
Just a note: will need to modify the `Ignore` class to not use `.easignore` for determining Managed vs Generic I think since we can't use `.easignore` in `expo-updates` (`expo-updates` is backend-agnostic).

wschurman:
FWIW the `.easignore` file was added in https://github.com/expo/eas-cli/pull/462 and it seems to have been specifically for making shallow copies? Before this PR, it superseded `.gitignore` for all VCS ignore detection and shallow copies (this PR adds an explicit arg to instruct it to supersede gitignores). My guess is that a lot of functionality has been built on top of this without considering that the functionality may be different depending on whether `.easignore` is used.

I also see that we've run into issue with it conflicting with `requireCommit`: https://github.com/expo/fyi/blob/main/eas-build-archive.md, https://github.com/expo/eas-cli/issues/1757. So I think it's in a pretty rough state as it is (GitClient vcs ignores it entirely, GitNoCommitClient and NoVcsClient use it, so workflow detection amongst other things could differ based on whether there's a vcs present?).

Any reviewers know if `.easignore` should only be used for copying files to be built? Or does it make sense to keep using it for all ignore detection like workflow detection?

wschurman:
One other potential issue is that the version of the `@expo/fingerprint` package could differ between the dependencies of `expo-updates` and the dependencies of `eas-cli`. We probably need to create a CLI command in expo-updates to generate the fingerprint or something so that the same version is guaranteed.

wschurman:
Closing to incorporate https://github.com/expo/expo/pull/27119

PR #2243, opened 2024-02-21
[eas-cli] Fix expo-updates package version detection for canaries

wschurman:
# Why

Currently canaries fail these version checks since their versions are of the form `0.0.1-canary-20240109-93608d8` (semver 0.0.1). We've been investigating changing the canary versioning scheme, but it's a bit more complex than anticipated. Instead, just handle the current canary versioning scheme.

# How

If canary is in the version, assume that it has features (for now).

# Test Plan

Inspect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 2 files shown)


wschurman:
/changelog-entry bug-fix Fix expo-updates package version detection for canaries

tsapeta (approved):

PR #2248, opened 2024-02-23
Add `config` property to `eas.schema.json`

sjchmiela:
# Why

We want users to feel like they're adding something we support.

# How

Added `config` property to build profile level and platform-specific profile properties levels.

# Test Plan

None.


sjchmiela:
/changelog-entry bug-fix Add missing `config` property to `eas.json` schema.

szdziedzic (approved):

PR #2250, opened 2024-02-24
[vscode] Include src/**/build directories in search and replace

wschurman:
# Why

Directories named `build` are a special case in the default vscode search excludes since most projects don't want to search the build folder since it is compiled code (including us in this repo, we build to the `build` folder as destination and don't want the compiled js files to be included in search results).

But, we also have some folders named `build` that need to be named build to appease oclif that we do want to appear in search results and be respected in find/replace.

# How

Add vscode setting to re-include certain `build` folders.

Note that usually I don't recommend checking in this .vscode folder or any editor settings folder (they're excluded by default and even this file is still excluded in this repo, it was force added to the commit), but in this instance I think it's warranted.

# Test Plan

Search for "getRuntimeVersion" and now see the file in the build folder show up.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,10 @@
+{
+  "eslint.enable": true,
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/bower_components": true,
+    "**/*.code-search": true,
+    "**/src/**/build/**": false,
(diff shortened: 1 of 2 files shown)


szdziedzic (approved):
thanks!

sebm123:
1. [](url)

PR #2251, opened 2024-02-24
[eas-cli] Use expo-updates runtime version CLI to generate runtime versions

wschurman:
# Why

This makes use of https://github.com/expo/expo/pull/27263 to generate the correct runtime version for build metadata and updates.

# How

Call the command in the package to use the versioned config-plugins for the app. This way the runtime version is consistent no matter what version of eas-cli is used (and for canary versions so we don't have to update eas-cli to use canaries).

# Test Plan

Run tests.

Manual test:
1. `neas build` with a bare-app in expo/expo
2. `neas update` with same app, including with `EXPO_DEBUG=1` to see fingerprint sources.


wschurman:
/changelog-entry bug-fix Use expo-updates runtime version CLI to generate runtime versions

quinlanj (approved):

szdziedzic (approved):

PR #2269, opened 2024-03-07
[eas-cli] Use new expo-updates configuration:syncnative for versioned native sync

wschurman:
# Why

This makes use of the new command added in https://github.com/expo/expo/pull/27511 to ensure that the runtime version resolution and synchronization are both versioned.

# How

Call the CLI instead of using the version of config plugins depended on by eas-cli when possible.

# Test Plan

In a generic project using fingerprint and `requireCommit: true`:
```
yarn link expo-updates
neas build -p android
```

and ensure the native sync is no longer the old version (meaning it called the CLI to do the native sync).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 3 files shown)


wschurman:
/changelog-entry new-feature Use new expo-updates configuration:syncnative for versioned native sync

quinlanj (approved):

PR #2271, opened 2024-03-08
[ENG-11416][eas-cli] add `--auto-submit` flag to `eas build:internal` command

szdziedzic:
# Why

We want to add support for `--auto-submit` for our GitHub integration. Yet the only place where we can automagically resolve the credentials for GH builds is EAS CLI if don't want to ask the user for any input like bundle identifier and so on and I believe that this is what we decided to do. We already run `eas build:internal` during the build process to resolve build credentials for the same reasons as mentioned above, so we can just add the auto-submit flag to trigger the submission from the build worker.

It's hacky but at the same time, it's not very different from running `eas build:internal` which we already do.

# How

Add `--auto-submit` flags to the `eas build:internal` command.

# Test Plan

Tested manually. Checked if a submission was created.

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -19,7 +19,7 @@ import { ensureProjectConfiguredAsync } from './configure';
 import { BuildContext } from './context';
 import { createBuildContextAsync } from './createContext';
 import { prepareIosBuildAsync } from './ios/build';
-import { LocalBuildOptions } from './local';
+import { LocalBuildMode, LocalBuildOptions } from './local';
 import { ensureExpoDevClientInstalledForDevClientBuildsAsync } from './utils/devClient';
 import { printBuildResults, printLogsUrls } from './utils/printBuildInfo';
 import { ensureRepoIsCleanAsync } from './utils/repository';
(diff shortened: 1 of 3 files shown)


szdziedzic:
/changelog-entry new-feature Add `--auto-submit` option to `eas build:internal` command

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -180,13 +180,23 @@ export async function runBuildAndSubmitAsync(
...
 
-  Log.newLine();
-  printLogsUrls(startedBuilds.map(startedBuild => startedBuild.build));
-  Log.newLine();
+  if (flags.localBuildOptions.localBuildMode === LocalBuildMode.INTERNAL) {
+    const startedBuild = await BuildQuery.byIdAsync(
+      graphqlClient,
+      nullthrows(process.env.EAS_BUILD_ID, 'EAS_BUILD_ID is not defined')
+    );
+    startedBuilds.push({ build: startedBuild, buildProfile: buildProfiles[0] });
```
sjchmiela, line 192: Is `buildProfiles[0]` safe? Should we use `startedBuild.buildProfile.profile` instead?

sjchmiela (approved):

packages/eas-cli/src/build/runBuildAndSubmit.ts
szdziedzic: It's safe, this array can have a max length of 2 (one build profile for iOS and one for Android) and a min length of 1. In this case, it will always have length 1 for the `internal` command because we are concerned about the single build for a given platform with a given ID.

`startedBuild.buildProfile.profile` - it won't work, it's the data from GQL, and `startedBuild.buildProfile` is a string. We don't store build profiles in WWW.

PR #2278, opened 2024-03-12
[eas-json] mark Xcode 15 image as `default`

szdziedzic:
# Why

Companion to https://github.com/expo/turtle-v2/pull/1733

# How

mark Xcode 15 image as `default`


szdziedzic:
/changelog-entry Mark the `macos-ventura-13.6-xcode-15.2` image as `default` in the `eas.json` schema

PR #2282, opened 2024-03-14
[eas-cli] add a `credentials:configure-build` subcommand that prepares a project for EAS Build without executing the build from the CLI

FiberJW:
# Why

To help a developer get configured for GitHub builds, it would be nice to let them use an EAS CLI command locally that sets up credentials for a build without actually executing it. In theory it would be awesome to be able to do this work from the browser/server, but as we know, some constraints prevent us from that.

# How

I added a new command that largely inherits most of its behavior from `ManageIos/ManageAndroid`, but goes right to the `SetUpBuildCredentials` step for each platform, skipping printing out existing credentials and the 4+ step prompt flow that takes you to this action.

# Test Plan

With `neas` as an alias pointed to my local build of EAS CLI,

`neas credentials:configure-build` should prompt you for both parameters and execute the build credentials setup for the selected platform.

`neas credentials:configure-build --platform <android/ios>` should prompt you for a build profile listed inside your **eas.json** and execute the build credentials setup for the selected platform.

`neas credentials:configure-build --platform <android/ios> --profile PROFILE_NAME` should immediately execute the build credentials setup for the selected platform.

### DIFF
--- a/packages/eas-cli/src/credentials/manager/SetUpAndroidBuildCredentials.ts
+++ b/packages/eas-cli/src/credentials/manager/SetUpAndroidBuildCredentials.ts
@@ -1,0 +1,155 @@
+import { Platform } from '@expo/eas-build-job';
+import { BuildProfile } from '@expo/eas-json';
+import assert from 'assert';
+
+import { AndroidActionType } from './Actions';
+import { CheckBuildProfileFlagAgainstEasJson } from './CheckBuildProfileFlagAgainstEasJson';
+import { CreateAndroidBuildCredentials } from './CreateAndroidBuildCredentials';
+import { Action } from './HelperActions';
+import { SelectExistingAndroidBuildCredentials } from './SelectAndroidBuildCredentials';
+import { SetDefaultAndroidKeystore } from './SetDefaultAndroidKeystore';
+import { GradleBuildContext, resolveGradleBuildContextAsync } from '../../project/android/gradle';
+import { AssignFcm } from '../android/actions/AssignFcm';
+import { AssignGoogleServiceAccountKeyForFcmV1 } from '../android/actions/AssignGoogleServiceAccountKeyForFcmV1';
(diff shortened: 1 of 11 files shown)


packages/eas-cli/src/credentials/manager/SetUpAndroidBuildCredentials.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (this.setUpBuildCredentialsWithProfileNameFromFlag) {
+      await this.runProjectSpecificActionAsync(
+        ctx,
+        AndroidActionType.CreateKeystore, // Directly proceed to CreateKeystore
+        gradleContext
+      );
+    }
+  }
+}
```
quinlanj, line 60: If someone has already set up their keystore, we don't want to make it again. 

instead of calling the `CreateKeystore` action, I'd consider calling the `SetUpBuildCredentials` action instead because it detects if credentials are already set up, then creates the keystore if not. It is the logic that runs on `eas build` so it will be great if we could reuse that https://github.com/expo/eas-cli/blob/94b217c16ed7fc540c439591ca9fd07b8ac1869e/packages/eas-cli/src/credentials/android/actions/SetUpBuildCredentials.ts#L31-L56

quinlanj (approved):
lgtm, thx!

szdziedzic (approved):
it's awesome, thanks! it looks good to me! I gave it a try locally and everything works well.

One follow-up thing we can do is to add `credentials:configure-submit`. It would be useful to also set up people using GH builds with `--auto-submit` this way, so we make sure that they have an ASC API key set on our servers. Or even make it a part of this command? What do you think about it?

FiberJW:
Setting up auto-submit credentials is a good idea! I think making that a separate command is a good way to start and allow people to run what they want individually, but we can create a super credentials command that sets up everything for build, submit, notifications, and anything else later

Issue #2160, opened 2023-12-20
Custom EAS Build incorrectly evaluates path on Native Windows (WSL is fine)


OlliePugh:
### Build/Submit details page URL

https://expo.dev/accounts/opugh/projects/rwg-client/builds/e6a39185-c5f7-4824-8ab0-a268cf3c659b

### Summary

The path uses `\` when on windows native, which escapes the characters in the path, so the pipeline can't find the build file.

### Managed or bare?

Bare

### Environment

expo-env-info 1.0.5 environment info:
    System:
      OS: Windows 10 10.0.19045
    Binaries:
      Node: 18.18.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
      npm: 9.6.6 - ~\workspace\real-world-games\node_modules\.bin\npm.CMD
    IDEs:
      Android Studio: AI-222.4459.24.2221.9971841
    npmPackages:
      expo: ~49.0.13 => 49.0.20 
      react: 18.2.0 => 18.2.0 
      react-dom: 18.2.0 => 18.2.0 
      react-native: 0.72.6 => 0.72.6
      react-native-web: ~0.19.6 => 0.19.9
    Expo Workflow: bare
    
    ✔ Check Expo config for common issues
✔ Check package.json for common issues
✔ Check dependencies for packages that should not be installed directly        
✔ Check for common project setup issues
✔ Check npm/ yarn versions
✔ Check Expo config (app.json/ app.config.js) schema
✔ Check that native modules do not use incompatible support packages
✔ Check for legacy global CLI installed locally
✔ Check that native modules use compatible support package versions for installed Expo SDK
✔ Check that packages match versions required by installed Expo SDK
 

### Error output

Failed to parse the custom build config file.
ENOENT: no such file or directory, open '/home/expo/workingdir/temporary-custom-build/client/.eas\build\development-build-android.yaml'

### Reproducible demo or steps to reproduce from a blank project

create-expo-app on windows, then create `.eas/build/build-name.yaml` and execute it `eas build -p android -e build-name`

szdziedzic:
Hi 👋 ,

Yeah, it seems like the issue. I can take a look at it and try to make it better 😄.

szdziedzic:
This should be fixed

OlliePugh:
Awesome! Thanks for taking a look at this!

PR #2285, opened 2024-03-15
[eas-cli] send a path to custom build config to EAS Build server with posix separator for Windows

szdziedzic:
# Why

Our server expects to get this path with a POSIX separator

Closes https://github.com/expo/eas-cli/issues/2160

# How

If a build is not local use the posix separator for Windows.

# Test Plan

Tests

Comment out `process.platform === 'win32'` and see if it works on my Mac machine

I don't have a Windows machine to test it

### DIFF
--- a/packages/eas-cli/src/build/ios/prepareJob.ts
+++ b/packages/eas-cli/src/build/ios/prepareJob.ts
@@ -51,7 +51,8 @@ export async function prepareJobAsync(
     : undefined;
   if (
     maybeCustomBuildConfigPath &&
-    ctx.localBuildOptions.localBuildMode !== LocalBuildMode.LOCAL_BUILD_PLUGIN
+    ctx.localBuildOptions.localBuildMode !== LocalBuildMode.LOCAL_BUILD_PLUGIN &&
+    process.platform === 'win32'
   ) {
     maybeCustomBuildConfigPath = path.posix.join(...maybeCustomBuildConfigPath.split(path.sep));
   }
--- a/packages/eas-cli/src/build/ios/prepareJob.ts
+++ b/packages/eas-cli/src/build/ios/prepareJob.ts
@@ -46,10 +46,7 @@ export async function prepareJobAsync(
   }
 
   const maybeCustomBuildConfigPath = buildProfile.config
-    ? getCustomBuildConfigPathForJob({
-        configFilename: buildProfile.config,
-        localBuildMode: ctx.localBuildOptions.localBuildMode,
-      })
+    ? getCustomBuildConfigPathForJob(buildProfile.config)
     : undefined;
 
   const job: Ios.Job = {
--- a/packages/eas-cli/src/build/ios/prepareJob.ts
+++ b/packages/eas-cli/src/build/ios/prepareJob.ts
@@ -14,10 +14,9 @@ import slash from 'slash';
 
 import { IosCredentials, TargetCredentials } from '../../credentials/ios/types';
 import { IosJobSecretsInput } from '../../graphql/generated';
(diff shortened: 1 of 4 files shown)


szdziedzic:
/changelog-entry bug-fix Use a custom build config path with POSIX separator when sending data to the EAS Build server

sjchmiela (approved):

packages/eas-cli/src/build/ios/prepareJob.ts
```diff
@@ -45,9 +46,16 @@ export async function prepareJobAsync(
...
-  const maybeCustomBuildConfigPath = buildProfile.config
+  let maybeCustomBuildConfigPath = buildProfile.config
     ? getCustomBuildConfigPath(buildProfile.config)
     : undefined;
+  if (
+    maybeCustomBuildConfigPath &&
+    ctx.localBuildOptions.localBuildMode !== LocalBuildMode.LOCAL_BUILD_PLUGIN &&
+    process.platform === 'win32'
+  ) {
+    maybeCustomBuildConfigPath = path.posix.join(...maybeCustomBuildConfigPath.split(path.sep));
```
sjchmiela, line 57:
```suggestion
    maybeCustomBuildConfigPath = path.posix.normalize(maybeCustomBuildConfigPath);
```
would this have the same effect?

packages/eas-cli/src/build/ios/prepareJob.ts
```diff
@@ -45,9 +46,16 @@ export async function prepareJobAsync(
     }
   }
 
-  const maybeCustomBuildConfigPath = buildProfile.config
```
radoslawkrzemien, line 48: If the logic here is the same for ios and android maybe in `getCustomBuildConfigPath`?

radoslawkrzemien (approved):

packages/eas-cli/src/build/ios/prepareJob.ts
sjchmiela: This crossed my mind too, I figured we may want to keep `getCustomBuildConfigPath` locale to current environment, e.g. for `eas build --local`. I guess `prepareJob` is the borderline between "we're running on this computer" and "sending this to the server".

szdziedzic: Yeah, you are right, we should keep this logic in a single place. Since `getCustomBuildConfigPath` can be also used in other context than sending the data to the EAS server I created a new function just for preparing this data for the Job object 👍

packages/eas-cli/src/build/ios/prepareJob.ts
sjchmiela: Pity. I noticed it works the other way

```
> path.win32.normalize("d/b/a.yaml")
'd\\b\\a.yaml'
```
and I hoped it would go the other way too, but it doesn't.

```
> path.posix.normalize("d\\b\\a.yaml")
'd\\b\\a.yaml'
```

packages/eas-cli/src/project/customBuildConfig.ts
```diff
@@ -68,3 +69,21 @@ export async function validateCustomBuildConfigAsync({
 export function getCustomBuildConfigPath(configFilename: string): string {
   return path.join('.eas/build', configFilename);
 }
+
+export function getCustomBuildConfigPathForJob({
```
sjchmiela, line 73: With a separate function we could even do:
```
getCustomBuildConfigPathForJob {
  path.posix.join(...)
}

getCustomBuildConfigPath {
  path.join(...)
}
```

szdziedzic: you are right! thanks!

PR #2295, opened 2024-03-21
[eas-cli] use `GitNoCommitClient` for `eas build:internal` command

szdziedzic:
# Why

We don't want to fail builds when running `eas build:internal` for people who have `requiredCommit: true` and have modified their project files before the `CONFIGURE_PROJECT` build phase is run. We want to always override the `vcsClient` for the `build:internal` command to `GitNoCommitClient`.

# How

While working on it I noticed that resolving the `vcsClient` works correctly when running `getContextAsync` only because it is always specified as a last field in the command `contextDefinition`. If the order was different `getVcsClient` (used in the `vcsClient` field resolver) might have returned an incorrect value because `setVcsClient` is run separately in other places and it might not be well synchronized.

I deleted the `getVcsClient` and `setVcsClient` functions so `getContextAsync` is the only source of `vcsClient`.

I added the `vcsClientOverride` option to the `getContextAsync` function so we can manually override it if needed like in the `build:internal` command case.

# Test Plan

Use a project with uncommitted changes with `requireCommit: true` and run the `build:internal` command.

Before:
Screenshot 2024-03-21 at 08 56 58

After:
Screenshot 2024-03-21 at 08 57 31

Tests

Run the `eas build` command and check that `vcsClient` behaves correctly depending on whether there are any uncommitted changes and `requireCommit` option value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Use a custom build config path with POSIX separator when sending data to the EAS Build server. ([#2285](https://github.com/expo/eas-cli/pull/2285) by [@szdziedzic](https://github.com/szdziedzic))
+- Improve resolving vcsClient as a part of the project context. ([#2295](https://github.com/expo/eas-cli/pull/2295) by [@szdziedzic](https://github.com/szdziedzic))
 
 ### 🧹 Chores
 
--- a/packages/eas-cli/src/build/android/__tests__/version-test.ts
+++ b/packages/eas-cli/src/build/android/__tests__/version-test.ts
@@ -8,7 +8,7 @@ import os from 'os';
 import path from 'path';
 
 import { getAppBuildGradleAsync, resolveConfigValue } from '../../../project/android/gradleUtils';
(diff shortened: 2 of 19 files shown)


sjchmiela (approved):
Should we use vcsClient from context everywhere and instead override the vcsClient the context uses? Having two ways to get a vcsClient seems strange? When should I use which?

sjchmiela:
Also, once this is released, what will we need to do to get this fixed in Turtle?

szdziedzic:
It will be fixed automatically because we use the npm tags to set which version is used on staging and prod
https://github.com/expo/eas-build/blob/0a3819f36c1244a859c731571116c2f9e6c19112/packages/build-tools/src/common/easBuildInternal.ts#L12-L14

szdziedzic:
>Should we use vcsClient from context everywhere and instead override the vcsClient the context uses? Having two ways to get a vcsClient seems strange? When should I use which?

You should always use the context one unless you know what you are doing and you want to force override something

szdziedzic:
/changelog-entry bug-fix Improve resolving vcsClient as a part of the project context

wschurman (approved):

sjchmiela (approved):

PR #2311, opened 2024-04-03
Fix project metadata spinner persisting after upload

khamilowicz:
# Why

Fixes bug where `eas build` would not exit after build is finished

# How

Clear spinner timer

### DIFF
--- a/packages/eas-cli/src/build/utils/repository.ts
+++ b/packages/eas-cli/src/build/utils/repository.ts
@@ -133,6 +133,7 @@ export async function makeProjectMetadataFileAsync(archivePath: string): Promise
     }
     throw e;
   }
+  clearTimeout(timer);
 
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -133,6 +133,7 @@ export async function makeProjectMetadataFileAsync(archivePath: string): Promise
     }
     throw e;
   }
+  clearTimeout(timer);
```
sjchmiela, line 136: Would it make sense to move it to `finally`? And remove the `catch`'s `clearTimeout` then?

khamilowicz: It would, on line 180 in `makeProjectTarballAsync` it is outside `finally` though.

PR #2317, opened 2024-04-10
fix: readme source code path

gabimoncha:
# Why

The README source code link is broken

# How

Updated the path of the source code links

# Test Plan

Tested by clicking on the updated link path

PR #2318, opened 2024-04-10
[eas-cli] fix `README.md` command files URLs

szdziedzic:
# Why

URLs to our command source files are broken in [README](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/README.md).

@gabimoncha caught it in https://github.com/expo/eas-cli/pull/2317

Instead of pointing to `packages/eas-cli/src/commands/xyz` URLs point to `src/commands/xyz` (incorrect, nonexistent location)

Example: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/README.md#eas-accountlogin

Screenshot 2024-04-10 at 13 58 13

# How

Since our `README` generation process is automated by the oclif tool we should fix the `patch-readme` script we have to actually do its job and patch `README` correctly and change `src/commands` to `package/eas-cli/src/commands`. The `README` generation process is run every time we do a new release.

https://github.com/expo/eas-cli/blob/dbb070f9a4dcb9f2868715cf314119173242d6a9/packages/eas-cli/package.json#L218

# Test Plan

Test locally by running `yarn version` in `packages/eas-cli` dir and inspecting the generated `README`.

Do a new release after merging it to see that it works on `main`.


szdziedzic:
/changelog-entry bug-fix Fix command source files URLs in autogenerated `README`

sjchmiela (approved):

PR #2320, opened 2024-04-11
[eas-cli] stop querying `Build.resourceClass`

szdziedzic:
# Why

This field is not used anywhere and we want to delete it from WWW in the future.

# How

stop querying `Build.resourceClass`

# Test Plan

Tests


sjchmiela (approved):
Cool. If these are the only changes required I think we should be able to simplify `Build.resourceClass` in `www`.

szdziedzic:
/changelog-entry chore Stop querying `Build.resourceClass` field

PR #2324, opened 2024-04-12
[eas-cli] Fix display of errors when expo-updates CLI command fails

wschurman:
# Why

It seems that the error thrown by the expo-updates CLI is a normal JS Error, and due to the way the CLI is called, the JS error doesn't print super well when it is thrown.

# How

This PR wraps all stderr output in a new error type that at least shows the full error to the user. Still room for improvement on the expo-updates CLI end to make the stderr cleaner.

# Test Plan

```
neas update
...
/Users/wschurman/temp/my-app/node_modules/expo-updates/utils/build/resolveRuntimeVersionAsync.js:24
        throw new Error(`You're currently using the bare workflow, where runtime version policies are not supported. You must set your runtime version manually. For example, define your runtime version as "1.0.0", not {"policy": "appVersion"} in your app config. https://docs.expo.dev/eas-update/runtime-versions`);
              ^

Error: You're currently using the bare workflow, where runtime version policies are not supported. You must set your runtime version manually. For example, define your runtime version as "1.0.0", not {"policy": "appVersion"} in your app config. https://docs.expo.dev/eas-update/runtime-versions
    at resolveRuntimeVersionAsync (/Users/wschurman/temp/my-app/node_modules/expo-updates/utils/build/resolveRuntimeVersionAsync.js:24:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async resolveRuntimeVersion (/Users/wschurman/temp/my-app/node_modules/expo-updates/cli/build/resolveRuntimeVersion.js:63:32)

Node.js v18.18.2

    Error: update command failed.
```

### DIFF
--- a/packages/eas-cli/src/utils/expoUpdatesCli.ts
+++ b/packages/eas-cli/src/utils/expoUpdatesCli.ts
@@ -5,6 +5,7 @@ import { link } from '../log';
 
 export class ExpoUpdatesCLIModuleNotFoundError extends Error {}
 export class ExpoUpdatesCLIInvalidCommandError extends Error {}
+export class ExpoUpdatesCLICommandFailedError extends Error {}
 
 export async function expoUpdatesCommandAsync(projectDir: string, args: string[]): Promise<string> {
   let expoUpdatesCli;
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/utils/expoUpdatesCli.ts
```diff
@@ -24,13 +25,18 @@ export async function expoUpdatesCommandAsync(projectDir: string, args: string[]
   }
 
   try {
-    return (await spawnAsync(expoUpdatesCli, args)).stdout;
+    return (await spawnAsync(expoUpdatesCli, args, { stdio: 'pipe' })).stdout;
```
sjchmiela, line 28: This technically is not required here, because this `spawn` should default to `pipe`, right? Asking to confirm.

We're not using the `turtle-spawn` here that defaults to `inherit`:
https://github.com/expo/eas-build/blob/997d6e65c7fea16184d038fdf831352346c1685d/packages/turtle-spawn/src/index.ts#L16-L19

sjchmiela (approved):

packages/eas-cli/src/utils/expoUpdatesCli.ts
wschurman: Correct. I added it just to keep the code reasonably consistent between this and `build-tools`.

wschurman:
/changelog-entry bug-fix Fix display of errors when expo-updates CLI command fails

PR #2326, opened 2024-04-12
[1/2][eas-cli] paginated progress bar

quinlanj:
### background
There are a couple flows in `eas-cli` where we fetch the user's entire dataset (ie) dist certs, push keys, etc. We don't expect the user to have many, but sometimes there are outlier users who end up having way more items than expected. I'm moving these fetches to use our paginated endpoints. We currently attempt to fetch the entire dataset in one request, and for the outlier users with many items, it causes the request to fail/timeout. 

# Why
This PR makes a generic pagination fetcher to get an entire dataset. 
- Happy case: most users will have <= 1 page of data (20 items), so in this case we don't bother showing a progress bar
- Medium outlier case: If a user has many items, a progress bar will be displayed. Because we use relay pagination, you cannot know the total number of items in the dataset until you fetch them all, so I've set the progress bar to become full at around 10 pages

https://github.com/expo/eas-cli/assets/6380927/29f4d3bd-5e65-492f-a2a7-80e312f049d0
- Super outlier case: If a user has so many items that it fills up the progress bar, we update the total estimated work, so the user can still see progress is being made:

https://github.com/expo/eas-cli/assets/6380927/49265393-2994-4e39-b0f2-b0bc90e5c711

# Test Plan

- [ ] new tests on full and empty datasets pass


quinlanj:
/changelog-entry chore add progress bar for fetching paginated datasets

khamilowicz (approved):

szdziedzic (approved):

PR #2334, opened 2024-04-19
[eas-cli] amend credential removal wording

quinlanj:
# Why

I recently capped the number of Entities we load from www on the ApplePushKey `iosAppCredentialsList` edge and the AppleDistributionCertificate `iosAppBuildCredentialsList` edge because we had some outlier users that had thousands of entities on those edges. We use the edges to tell the user which apps still use the credential when users try to delete it. 

# How

Amend the wording to tell the user that they may have more apps relying on the credential if the `www` limit has been hit. 

Screenshot 2024-04-18 at 8 49 19 PM

# Test Plan

- [x] manually tested apple push key removal on <20 apps and 20+ apps
- [x] manually tested apple dist cert removal on <20 apps and 20+ apps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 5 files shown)


szdziedzic (approved):
awesome, thanks!

PR #2336, opened 2024-04-22
[ENG-12057][eas-cli] don't prompt user to set `android.package` and `ios.bundleIdentifier` value when runing in non-interactive mode

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-12057/fail-with-a-nice-and-meaningful-error-message-if-iosbundleidentifier

# How

Throw meaningful error instead

# Test Plan

Tests

### DIFF
--- a/packages/eas-cli/src/project/ios/__tests__/bundleIdentifier-test.ts
+++ b/packages/eas-cli/src/project/ios/__tests__/bundleIdentifier-test.ts
@@ -111,8 +111,28 @@ describe(ensureBundleIdentifierIsDefinedForManagedProjectAsync, () => {
           projectId: '1234',
           exp: {} as any,
           vcsClient,
+          nonInteractive: false,
         })
       ).rejects.toThrowError(/we can't update this file programmatically/);
+    });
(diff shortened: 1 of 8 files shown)


szdziedzic:
/changelog-entry bug-fix Don't prompt to set `android.package` and `ios.bundleIdentifier` values when running in non-interactive mode

packages/eas-cli/src/project/ios/__tests__/bundleIdentifier-test.ts
```diff
@@ -111,9 +111,29 @@ describe(ensureBundleIdentifierIsDefinedForManagedProjectAsync, () => {
           projectId: '1234',
           exp: {} as any,
           vcsClient,
+          nonInteractive: false,
         })
       ).rejects.toThrowError(/we can't update this file programmatically/);
     });
+    it('throws an error if using app.config.js', async () => {
```
radoslawkrzemien, line 118: Duplicated test name

radoslawkrzemien (approved):

PR #2342, opened 2024-04-24
[ENG-12120][eas-cli] make multi-select for revoking distribution certificates more readable

szdziedzic:
# Why

https://linear.app/expo/issue/ENG-12120/make-multi-select-for-revoking-distribution-certificates-more-readable

# How

Make it look better

# Test Plan

test manually


szdziedzic:
/changelog-entry chore Make multi-select for revoking distribution certificates more readable

PR #2339, opened 2024-04-23
[ENG-11658][eas-cli] Add --no-channel flag for update and rollback to embedded

quinlanj:
# Why

When users create an update, we make sure that the branch they publish to is linked to a channel of the same name. However, this behaviour is not always desirable. For example, in the development usecase where someone is trying to load their update in the dev-client, a channel is not necesary. Users will either navigate to the update they want in the dev client menu or via a preview QR code. 

When we've interviewed users who use the dev client/qr code flow, they are always running `eas delete:channel` after the `eas update` command. 

# How

This PR adds a flag, `--no-channel` for the development usecase. This prevents a channel from being automatically created. 

# Test Plan

- [ ] new tests pass

PR #2346, opened 2024-04-25
[ENG-11658][eas-cli] dont generate channel on `eas update` without flag

quinlanj:
# Why
An alternative solution to https://github.com/expo/eas-cli/pull/2339 , after talking to @jonsamp and @wschurman 

When users create an update, we make sure that the branch they publish to is linked to a channel of the same name. However, this behaviour is not always desirable. For example, in the development usecase where someone is trying to load their update in the dev-client, a channel is not necesary. Users will either navigate to the update they want in the dev client menu or via a preview QR code.

# How

- Only create a channel and link it to a branch when the `--channel` flag is passed

# Test Plan

- [x] ensure `eas update --channel bar` creates channel `bar`, links it to branch `bar`, then publishes an update to the branch
- [x] ensure `eas update:roll-back-to-embedded --channel bar` creates channel `bar`, links it to branch `bar`, then publishes an update to the branch
- [x] ensure `eas update --branch bar` and `eas update --auto` no longer create a channel
- [x] - [ ] ensure `eas update:roll-back-to-embedded --branch bar` and `eas update:roll-back-to-embedded --auto` no longer create a channel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- Stop creating a channel on `eas update` and `eas update:roll-back-to-embedded` unless the `--channel` flag is specified. ([#2346](https://github.com/expo/eas-cli/pull/2346) by [@quinlanj](https://github.com/quinlanj))
+
 ### 🎉 New features
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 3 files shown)


quinlanj:
/changelog-entry breaking-change Stop creating a channel on `eas update` and `eas update:roll-back-to-embedded` unless the `--channel` flag is specified

wschurman (approved):

wschurman:
Related consequence of this: https://github.com/expo/expo/issues/29205. I think this is still the right call, but we'll definitely need to audit the docs for local builds to ensure channels are manually created since they're no longer generated here.

PR #2347, opened 2024-04-26
[ENG-8801][eas-cli] allow modification of provisioning profile in CI, add `--freeze-credentials` flag

quinlanj:
# Why

We currently support passing in an ASC Api Key to eas build through environment variables. However, we've barred modifying iOS credentials automatically in CI. For many credentials like a Push Key, revoking/generating them in CI is a bad idea since an account has a very low limit on how many keys they are allowed to generate and revoking a push key can have a large impact on production. However, modifying provisioning profiles are relatively safe and needed on a semi-regular basis (ie) annual profile expiration

# How

This PR allows for the creation and modification of app store and enterprise Provisioning Profiles in CI. Apple resigns app store provisioning profiles when it is distributed, so this should be relatively safe. For users that want to ensure their credentials don't get modified in CI, I've added a `--freeze-credentials` flag. 

# Test Plan

- [ ] added + amended a bunch of tests

### Manual sanity tests
- [ ] `--non-interactive`, wrong authentication method, expect to fail with helpful error message
- [ ] ASC api key auth, `--freeze-credentials`, expect to fail with helpful error message
- [ ] `--non-interactive`, ASC api key auth, succeed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- Allow modification of provisioning profile in CI, add --freeze-credentials flag. ([#2347](https://github.com/expo/eas-cli/pull/2347) by [@quinlanj](https://github.com/quinlanj))
+
 ### 🎉 New features
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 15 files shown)


quinlanj:
/changelog-entry breaking-change  Allow modification of provisioning profile in CI, add --freeze-credentials flag

sjchmiela (approved):
This looks great to me!!

szdziedzic (approved):

PR #2349, opened 2024-04-29
fix parsing EXPO_APPLE_PROVIER_ID env var

louix:
# Why

`getenv.int(process.env.EXPO_APPLE_PROVIER_ID)` will lookup the value of the `process.env.EXPO_APPLE_PROVIER_ID` in `process.env` (i.e. `process.env[process.env.EXPO_APPLE_PROVIER_ID]`) and try to parse that (undefined value) as an int.

If at runtime `process.env.EXPO_APPLE_PROVIER_ID` is `12345`, the code will fail with:

```
GetEnv.Nonexistent: 12345 does not exist and no fallback value provided.
```

Where process.env["12345"] does not exist.

# Test Plan

Simply try to set EXPO_APPLE_PROVIER_ID to anything on the current eas-cli build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Add explicit workflow arg to expo-update CLI calls. ([#2340](https://github.com/expo/eas-cli/pull/2340) by [@wschurman](https://github.com/wschurman))
+- Correctly parse the EXPO_APPLE_PROVIER_ID environment variable. ([#2349](https://github.com/expo/eas-cli/pull/2349) by [@louix](https://github.com/louix))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


louix:
/changelog-entry bug-fix fix parsing EXPO_APPLE_PROVIER_ID env var

louix:
Sure, yeah sorry I should have run prettier.

Is the /changelog-entry thing broken? Manually added it anyway.

szdziedzic:
>Is the /changelog-entry thing broken? Manually added it anyway.

It seems to be broken for external contributors 🤔 I will take a look at it later on. Meanwhile, feel free to ignore this failed changelog action and merge the PR

louix:
I cannot merge this PR by the way, so I assume someone on your side needs to.

> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

louix:
@sjchmiela @szdziedzic could you merge the PR since I do not have write access

szdziedzic:
merged ✅

PR #2350, opened 2024-04-29
Implement `eas env` commands

khamilowicz:
# Why

Implements the following commands:

- `eas env:create` ENG-11913
- `eas env:get` ENG-11915
- `eas env:delete` ENG-11914
- `eas env:list` ENG-11917
- `eas env:update` ENG-11918 (Currently only delegates to `env:create`)

Also, as discussed on Linear, implements `eas env:shared:create/update/get/delete/link/unlink` as `--scope shared`  flag on mentioned commands. (ENG-11927, ENG-11926, ENG-11925, ENG-11921, ENG-11922, ENG-11923).

# How

Created commands and mutations using updated backend (https://github.com/expo/universe/pull/15153).

# Deployment

(We'll wait with merging when everything from https://linear.app/expo/project/eas-environment-variables-2813a6f4c62e is done)

1. https://github.com/expo/universe/pull/15153
2. https://github.com/expo/eas-cli/pull/2350

# Test Plan

Each command accepts arguments both in interactive mode (user is prompted for input) and non-interactive (arguments are passed inline).

The following flow will create both project and shared variables, then it will link and unlink shared variable from the project, and finally delete both of them. Commands are executed both with inline arguments and prompts.
From an existing app directory:

1. Create project variable interactively - `eas env:create` and provide name, value and select environment 
3. Create shared variable non-interactively - `env:create --scope SHARED --name  sharedVariable --value someValue`
4. `eas env:list` and select environment - displays only project var
5. `eas env:list --format long --scope SHARED` - displays only shared var in long format
6. `eas env:link` - select shared var and environment
7. `eas env:list` - shared variable should be added to selected env
8. `eas env:unlink --environment SELECTEDENV --name sharedVariable`
9. `eas env:list` - shared variable should be removed
10. `eas env:delete` - delete project variable
11. `eas env:delete --scope SHARED` - delete shared environment


packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -1,5 +1,15 @@
 import { Flags } from '@oclif/core';
 
+import { EnvironmentVariableEnvironment, EnvironmentVariableScope } from '../graphql/generated';
+
+async function upperCaseAsync(input: string): Promise<string> {
+  return input.toUpperCase();
+}
```
szdziedzic, line 7: It doesn't like it need to be `async`

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -10,6 +20,42 @@ export const EasNonInteractiveAndJsonFlags = {
...
 
+export const EASEnvironmentFlag = {
+  environment: Flags.enum({
+    description: 'Environment to create the secret in',
+    parse: upperCaseAsync,
+    options: mapToLowercase([
+      EnvironmentVariableEnvironment.Development,
+      EnvironmentVariableEnvironment.Preview,
+      EnvironmentVariableEnvironment.Production,
+    ]),
```
szdziedzic, line 31:
```suggestion
    options: [
      EnvironmentVariableEnvironment.Development,
      EnvironmentVariableEnvironment.Preview,
      EnvironmentVariableEnvironment.Production,
    ].map(e => e.toLowerCase()),
```

nit: do we really need the `mapToLowerCase` function?

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,189 @@
...
+            graphqlClient,
+            existingVariable.id,
+            projectId,
+            environment
+          );
+          Log.withTick(
+            `Unlinking shared variable ${chalk.bold(name)} on project ${chalk.bold(
+              projectDisplayName
+            )}.`
+          );
```
szdziedzic, line 112: I believe we should ask user if they want to do it before we do so

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,189 @@
...
+        } else {
+          await EnvironmentVariableMutation.deleteAsync(graphqlClient, existingVariable.id);
+
+          Log.withTick(
+            `Deleting existing variable ${chalk.bold(name)} on project ${chalk.bold(
+              projectDisplayName
+            )}.`
```
szdziedzic, line 119: ditto

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,189 @@
...
+        )} on project ${chalk.bold(projectDisplayName)}.`
+      );
+    } else if (scope === EnvironmentVariableScope.Shared) {
+      const sharedVariables = await EnvironmentVariablesQuery.sharedAsync(graphqlClient, projectId);
+      const existingVariable = sharedVariables.find(variable => variable.name === name);
+      if (existingVariable) {
+        Log.error(
+          'Variable with this name already exists on this account. Please use a different name .'
+        );
+        return;
```
szdziedzic, line 147:
```suggestion
        throw new Error(
          'Variable with this name already exists on this account. Please use a different name .'
        );
```

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,110 @@
...
+      flags: { name, environment, 'non-interactive': nonInteractive, scope },
+    } = await this.parse(EnvironmentVariableDelete);
+    const {
+      privateProjectConfig: { projectId },
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(EnvironmentVariableDelete, {
+      nonInteractive,
+    });
+    let variable;
+    let variables;
```
szdziedzic, line 46: Can we explicitly type these?

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,110 @@
...
+      ({ variable } = await promptAsync({
+        type: 'autocomplete',
+        name: 'variable',
+        message: 'Pick the variable to be deleted:',
+        choices: variables.map(variable => ({
+          title: variable.name,
+          value: variable,
+        })),
+      }));
+
```
szdziedzic, line 75:
```suggestion
      const { selectedVariable } = await promptAsync({
        type: 'autocomplete',
        name: 'selectedVariable',
        message: 'Pick the variable to be deleted:',
        choices: variables.map(variable => ({
          title: variable.name,
          value: variable,
        })),
      });
      variable = createdVariable;

```

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (!name) {
+      const validationMessage = 'Variable name to delete may not be empty.';
+      if (nonInteractive) {
+        throw new Error(validationMessage);
+      }
+
+      ({ variable } = await promptAsync({
+        type: 'autocomplete',
```
szdziedzic, line 67: Why not 
```suggestion
        type: 'select',
```
?

packages/eas-cli/src/commands/env/delete.ts
szdziedzic, line 46: Using `let`s is generally an antipattern so we should avoid using them if possible.

I believe that if move logic to query variables and select a variable to a function you should be able to do something like this:
```suggestion
    const variables = getVariablesForScopeAndEnvironmentAsync({ scope, environment });
    const variable = selectVariableToBeDeleted({ variables });
```

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (!name) {
+      name = await promptVariableNameAsync(nonInteractive);
+    }
+
+    if (!environment && scope === EnvironmentVariableScope.Project) {
+      environment = await promptVariableEnvironmentAsync(nonInteractive);
+    }
+
+    let variable;
```
szdziedzic, line 57: ditto

packages/eas-cli/src/commands/env/link.ts
```diff
@@ -0,0 +1,81 @@
...
+    } = await this.parse(EnvironmentVariableLink);
+    const {
+      privateProjectConfig: { projectId },
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(EnvironmentVariableLink, {
+      nonInteractive,
+    });
+
+    const projectDisplayName = await getDisplayNameForProjectIdAsync(graphqlClient, projectId);
+    let variable;
```
szdziedzic, line 41: ditto

packages/eas-cli/src/commands/env/list.ts
```diff
@@ -0,0 +1,81 @@
...
+      flags: { environment, format, scope },
+    } = await this.parse(EnvironmentValueList);
+    const {
+      privateProjectConfig: { projectId },
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(EnvironmentValueList, {
+      nonInteractive: true,
+    });
+
+    let variables;
```
szdziedzic, line 43: ditto

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,91 @@
...
+    } = await this.getContextAsync(EnvironmentVariableUnlink, {
+      nonInteractive,
+    });
+
+    if (!environment) {
+      environment = await promptVariableEnvironmentAsync(nonInteractive);
+    }
+
+    const projectDisplayName = await getDisplayNameForProjectIdAsync(graphqlClient, projectId);
+    let variable;
```
szdziedzic, line 45: ditto

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,91 @@
...
+      graphqlClient,
+      variable.id,
+      projectId,
+      environment
+    );
+    if (!variable) {
+      throw new Error(
+        `Could not unlink variable with name ${unlinkedVariable.name} to project with id ${projectId}`
+      );
+    }
```
szdziedzic, line 83:
```suggestion
    if (!unlinkedVariable) {
      throw new Error(
        `Could not unlink variable with name ${variable.name} to project with id ${projectId}`
      );
    }
```

packages/eas-cli/src/utils/prompts.ts
szdziedzic, line 1: nit: `packages/eas-cli/src/utils/prompts.ts` -> `packages/eas-cli/src/environment-variables/prompts.ts`

szdziedzic (changes requested):

packages/eas-cli/src/commandUtils/flags.ts
khamilowicz: We don't need it, but we use it in two places and this looks IMO nicer. But it's just a preference.

packages/eas-cli/src/commands/env/create.ts
khamilowicz: We do, but we don't specify that it will be unlinked.
image
image

packages/eas-cli/src/commands/env/create.ts
khamilowicz: vide ultra

packages/eas-cli/src/commands/env/link.ts
```diff
@@ -0,0 +1,81 @@
...
+      graphqlClient,
+      selectedVariable.id,
+      projectId,
+      environment
+    );
+    if (!selectedVariable) {
+      throw new Error(
+        `Could not link variable with name ${linkedVariable.name} to project with id ${projectId}`
+      );
+    }
```
szdziedzic, line 73: Why is this condition based on `selectedVariable` and not `linkedVariable`?

packages/eas-cli/src/commands/env/list.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    } else {
+      if (scope === EnvironmentVariableScope.Shared) {
+        Log.log(chalk`{bold Shared variables for this account:}`);
+      } else {
+        Log.log(chalk`{bold Variables for this project for environment ${environment}:}`);
+      }
+      Log.log(
+        variables.map(variable => formatVariable(variable)).join(`\n\n${chalk.dim('———')}\n\n`)
+      );
```
szdziedzic, line 65: How does it look like when displayed in the terminal?

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    if (!name) {
+      name = await selectAsync(
+        'Select shared variable',
+        linkedVariables.map(variable => ({
+          title: variable.name,
+          value: variable.name,
+        }))
+      );
```
szdziedzic, line 66: How do these selects look like when somebody has 100 env vars?

packages/eas-cli/src/environment-variables/prompts.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (nonInteractive) {
+    throw new Error(validationMessage);
+  }
+
+  const { variableValue } = await promptAsync({
+    type: 'text',
+    name: 'variableValue',
+    message: 'Variable value:',
+    // eslint-disable-next-line async-protect/async-suffix
+    validate: async variableValue => {
```
szdziedzic, line 27: Is it needed?
```suggestion
    validate: variableValue => {
```

packages/eas-cli/src/graphql/queries/EnvironmentVariablesQuery.ts
```diff
@@ -0,0 +1,103 @@
...
+    projectId: string,
+    environment: string
+  ): Promise<EnvironmentVariableFragment[]> {
+    const { sharedVariables, appVariables } = await this.byAppIdAsync(
+      graphqlClient,
+      projectId,
+      environment
+    );
+    return [...appVariables, ...sharedVariables];
+  },
```
szdziedzic, line 102: I don't think it is used anywhere
```suggestion
```

szdziedzic (changes requested):

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -10,6 +20,42 @@ export const EasNonInteractiveAndJsonFlags = {
...
+  sensitive: Flags.boolean({
+    description: 'Encrypt variable value at rest',
+    default: false,
+  }),
+};
+
+export const EASVariableScopeFlag = {
+  scope: Flags.enum({
+    description: 'Scope for the variable',
+    options: mapToLowercase([EnvironmentVariableScope.Shared, EnvironmentVariableScope.Project]),
```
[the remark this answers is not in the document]
khamilowicz: Up till now we were using `shared` for this kind of variables, for example: https://linear.app/expo/issue/ENG-11921/add-eas-envsharedcreate-command#comment-53b40cb2

I'm not a fan of hardcoding readable values in this way, maybe something like:

```
const userReadableScope = {
  [EnvironmentVariableScope.Shared]: 'account'
}
```

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,112 @@
...
+      const confirmed = await toggleConfirmAsync({
+        message: `Are you sure you wish to proceed?${
+          selectedVariable.scope === EnvironmentVariableScope.Shared
+            ? ' This variable is applied across your whole account and may affect multiple apps.'
+            : ''
+        }`,
+      });
+      if (!confirmed) {
+        Log.error(`Canceled deletion of variable ${selectedVariable.name}.`);
+        process.exit(1);
```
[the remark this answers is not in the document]
khamilowicz: That's how it was done in the analogous command in secrets. https://github.com/expo/eas-cli/blob/a9c46336a2ac97024336ad18970bfc475e6589e4/packages/eas-cli/src/commands/secret/delete.ts#L84

packages/eas-cli/src/commands/env/list.ts
khamilowicz: image

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    const variable = await getVariableAsync(graphqlClient, scope, projectId, name, environment);
+
+    if (!variable) {
+      Log.error(`Variable with name "${name}" not found`);
+      return;
+    }
+
+    if (format === 'short') {
+      Log.log(chalk`{bold ${variable.name}}=${variable.value || '*****'}`);
```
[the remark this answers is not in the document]
khamilowicz: Yes, it will be made available in the following PR

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,188 @@
...
+    ...EASVariableSensitiveFlag,
+    ...EASVariableScopeFlag,
+    ...EASEnvironmentFlag,
+    ...EASNonInteractiveFlag,
+  };
+
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectConfig,
+    ...this.ContextOptions.LoggedIn,
+  };
```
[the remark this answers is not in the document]
khamilowicz: I'm not sure about it, we already do it on the backend

packages/eas-cli/src/commands/env/unlink.ts
khamilowicz: https://github.com/expo/eas-cli/assets/1774298/bbcc5a8d-aee3-447b-ba16-cf55430b0907

szdziedzic: awesome!

szdziedzic (changes requested):
I believe this PR now requires changes due to changes in the available env var types (added secret type)

Also, preferably it would be good to split the PR into smaller chunks to make reviews easier (if possible)

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -10,6 +20,42 @@ export const EasNonInteractiveAndJsonFlags = {
...
+    default: 'short',
+  }),
+};
+
+export const EASVariableSensitiveFlag = {
+  sensitive: Flags.boolean({
+    description: 'Encrypt variable value at rest',
+    default: false,
+  }),
+};
```
szdziedzic, line 48: how can one create a secret variable? Maybe this flag should be `--type public|sensitive|secret`?

szdziedzic (changes requested):

packages/eas-cli/src/commandUtils/flags.ts
khamilowicz: I added `--secret` and `--public` flags.

packages/eas-cli/src/commands/env/list.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const variables =
+      scope === EnvironmentVariableScope.Project && environment
+        ? (await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, projectId, environment))
+            .appVariables
+        : await EnvironmentVariablesQuery.sharedAsync(graphqlClient, projectId);
+
+    if (format === 'short') {
+      for (const variable of variables) {
+        Log.log(chalk`{bold ${variable.name}}=${variable.value || '*****'}`);
```
szdziedzic, line 55:
```suggestion
        Log.log(`${chalk.bold(variable.name)}=${variable.value || '***** (This is a secret env variable that can only be accessed on EAS builder and can't be read in any UI. Learn more.)'}`);
```

`Learn more` would need to link to the docs that we don't have at the moment

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,96 @@
...
+    const linkedVariables = appVariables.filter(({ id }) => sharedVariablesIds.includes(id));
+
+    if (linkedVariables.length === 0) {
+      Log.fail('No variables to unlink');
+      return;
+    }
+
+    if (!name) {
+      name = await selectAsync(
+        'Select shared variable',
```
szdziedzic, line 63:
```suggestion
        'Select shared variable to unlink',
```

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,96 @@
...
+      projectId,
+      environment
+    );
+    const sharedVariablesIds = sharedVariables.map(v => v.id);
+    const linkedVariables = appVariables.filter(({ id }) => sharedVariablesIds.includes(id));
+
+    if (linkedVariables.length === 0) {
+      Log.fail('No variables to unlink');
+      return;
+    }
```
szdziedzic, line 59:
```suggestion
    if (linkedVariables.length === 0) {
      throw new Error(`There are no linked shared env variables for project ${projectDisplayName}`);
    }
```

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const projectDisplayName = await getDisplayNameForProjectIdAsync(graphqlClient, projectId);
+
+    const { appVariables, sharedVariables } = await EnvironmentVariablesQuery.byAppIdAsync(
+      graphqlClient,
+      projectId,
+      environment
+    );
+    const sharedVariablesIds = sharedVariables.map(v => v.id);
+    const linkedVariables = appVariables.filter(({ id }) => sharedVariablesIds.includes(id));
```
szdziedzic, line 54: Would it be possible to do sth like:
```suggestion
    const { appVariables } = await EnvironmentVariablesQuery.byAppIdAsync(
      graphqlClient,
      projectId,
      environment
    );
    const linkedVariables = appVariables.filter(({ scope }) => scope === EnvironmentVariableScope.Shared);
```
?

This way we could only do 1 query instead of 2 (assuming that we would also modify queries).

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,96 @@
...
+      throw new Error(
+        `Could not unlink variable with name ${selectedVariable.name} from project ${projectDisplayName}`
+      );
+    }
+
+    Log.withTick(
+      `Unlinked variable ${chalk.bold(unlinkedVariable.name)} with value ${chalk.bold(
+        unlinkedVariable.value
+      )} to project ${chalk.bold(projectDisplayName)}.`
+    );
```
szdziedzic, line 94: ditto

packages/eas-cli/src/graphql/queries/EnvironmentSecretsQuery.ts
```diff
@@ -27,18 +28,18 @@ export const EnvironmentSecretsQuery = {
...
                   ownerAccount {
                     id
-                    environmentSecrets {
+                    environmentSecrets(filterNames: $filterNames) {
                       id
                       ...EnvironmentSecretFragment
                     }
                   }
-                  environmentSecrets {
+                  environmentSecrets(filterNames: $filterNames) {
```
szdziedzic, line 42: Why is this needed? It doesn't look like we use it anywhere.

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import { EnvironmentVariableEnvironment } from '../graphql/generated';
+import { promptAsync, selectAsync } from '../prompts';
+
+export async function promptVariableEnvironmentAsync(
+  nonInteractive: boolean
+): Promise<EnvironmentVariableEnvironment> {
+  if (nonInteractive) {
+    throw new Error('Environment may not be empty.');
+  }
```
szdziedzic, line 11:
```suggestion
  if (nonInteractive) {
    throw new Error('The `--environment` flag must be set when running in `--non-interactive` mode.');
  }
```

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -0,0 +1,90 @@
...
+  return await selectAsync('Select environment:', [
+    { title: 'Development', value: EnvironmentVariableEnvironment.Development },
+    { title: 'Preview', value: EnvironmentVariableEnvironment.Preview },
+    { title: 'Production', value: EnvironmentVariableEnvironment.Production },
+  ]);
```
szdziedzic, line 16:
```suggestion
  return await selectAsync('Select environment:', Object.keys(EnvironmentVariableEnvironment).map(key => ({
        title: key,
        value: EnvironmentVariableEnvironment[key]
    }));
```

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (nonInteractive && required) {
+    throw new Error(
+      `Environment variable needs 'value' to be specified when running in non-interactive mode. Run the command with ${chalk.bold(
+        '--value VARIABLE_VALUE'
+      )} flag to fix the issue`
+    );
+  }
+
+  const { variableValue } = await promptAsync({
+    type: 'text',
```
szdziedzic, line 36:
```suggestion
    type: [EnvironmentVariableVisibility.Sensitive, EnvironmentVariableVisibility.Secret].includes(visibility) ? 'password' : 'text',
```

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -0,0 +1,90 @@
...
+  ]);
+}
+export async function promptVariableValueAsync({
+  nonInteractive,
+  required = true,
+  initial,
+}: {
+  nonInteractive: boolean;
+  required?: boolean;
+  initial?: string | null;
```
szdziedzic, line 25: these 2 options don't seem to be used

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -10,6 +20,53 @@ export const EasNonInteractiveAndJsonFlags = {
...
+    description: 'Encrypt variable value at rest and protect it from being displayed',
+    default: false,
+    exclusive: ['sensitive', 'public'],
+  }),
+  public: Flags.boolean({
+    description: 'Make variable value public',
+    default: false,
+    exclusive: ['sensitive', 'secret'],
+  }),
+};
```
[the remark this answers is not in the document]
khamilowicz: In my opinion this approach is clearer. As an user I don't have to rememente whether it was `type` or `visibility` or some otherr name, I can just add `--secret` or `--sensitive` flag.

packages/eas-cli/src/commands/env/link.ts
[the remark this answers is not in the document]
khamilowicz: This is not in scope.

packages/eas-cli/src/graphql/queries/EnvironmentSecretsQuery.ts
khamilowicz: We keep environment secrets for compatibility, so you still have commands like `eas secret:create`.

packages/eas-cli/src/utils/prompts.ts
khamilowicz: Will be used in the following PRs.

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,117 @@
...
+
+function formatVariable(variable: EnvironmentVariableFragment): string {
+  return formatFields([
+    { label: 'ID', value: variable.id },
+    { label: 'Name', value: variable.name },
+    { label: 'Value', value: variable.value || '*****' },
+    { label: 'Scope', value: variable.scope },
+    { label: 'Created at', value: dateFormat(variable.createdAt, 'mmm dd HH:MM:ss') },
+  ]);
+}
```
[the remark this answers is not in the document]
khamilowicz: I have logged this issue in Linear.

packages/eas-cli/src/utils/prompts.ts
khamilowicz: Ditto, IMO current version is clearer, you know what it does at first sight. Meanwhile, the approach with extracting keys and values from EnvironmentVariableEnvironment requires some decoding.

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (variable.value === null) {
+      Log.log(
+        chalk`{bold ${variable.name}} is a secret variable and cannot be displayed once it has been created.`
+      );
+      return;
+    }
+
+    if (format === 'short') {
+      Log.log(chalk`{bold ${variable.name}}=${variable.value}`);
```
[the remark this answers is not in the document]
khamilowicz: When is `chalk.bold` preferred over `chalk\`{bold}`?

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,219 @@
...
+              Log.log('Aborting');
+              return;
+            }
+          }
+          await EnvironmentVariableMutation.unlinkSharedEnvironmentVariableAsync(
+            graphqlClient,
+            existingVariable.id,
+            projectId,
+            environment
+          );
```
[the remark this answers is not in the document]
khamilowicz: I created an issue for this in linear.

packages/eas-cli/src/commandUtils/flags.ts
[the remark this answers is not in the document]
khamilowicz: "Shared" will be used on the website: https://www.figma.com/design/759buSR2O3iV4P6lyZhowc/Environment-variables?node-id=5-2580&t=ujCz6lxKi4HMHR7S-1

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (!variable) {
+      Log.error(`Variable with name "${name}" not found`);
+      return;
+    }
+    if (variable.value === null) {
+      Log.log(
+        chalk`{bold ${variable.name}} is a secret variable and cannot be displayed once it has been created.`
+      );
+      return;
+    }
```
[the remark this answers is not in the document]
khamilowicz: It will be changed after we merge pull/push PRs (sudo mode is implemented there, so user will be able to peek sensitive variables)

packages/eas-cli/src/commands/env/get.ts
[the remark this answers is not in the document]
szdziedzic: TBH it doesn't really matter, I just wanted us to stick to one of these two conventions. I believe that `chalk.bold` is used more often across the codebase then `chalk{bold...`, but this is a nit

packages/eas-cli/src/commandUtils/flags.ts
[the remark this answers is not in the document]
szdziedzic: Yeah, that's a good point, but at the same time, this is an atypical approach across our codebase. For example we do `--platform android` vs `--android`.

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -10,6 +20,53 @@ export const EasNonInteractiveAndJsonFlags = {
...
```
szdziedzic, line 59: While I agree that your explanation makes sense and when using --sensitive` you don't need to remember about the visibility concept, it's not a standard practice across the EAS CLI codebase.

I mean we don't do `eas build --android`, but `eas build --platform android`.

In my opinion, it would be better to do `--visibility sensitive|secret|public`, but if you strongly disagree, that's not a hill I'm going to die on.

CC: @radoslawkrzemien what do you think?

szdziedzic: Using these 3 flags instead of 1 `visibility` flag makes it harder to describe what the default is. It's not clear looking at these 3 flags that the default behavior is `public`, while if you do the `visibility` flag with `default: 'public'` it's entirely clear.

szdziedzic: We also have `--scope shared|project` vs `--shared` and `--project` flags, it seems a little bit inconsistent

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,251 @@
...
+        if (existingVariable.scope === EnvironmentVariableScope.Shared) {
+          if (!nonInteractive) {
+            const confirmation = await confirmAsync({
+              message: `Shared variable ${name} already exists on this project. Do you want to unlink it first?`,
+            });
+
+            if (!confirmation) {
+              Log.log('Aborting');
+              return;
+            }
```
szdziedzic, line 128: ditto

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,251 @@
...
+        } else {
+          if (!nonInteractive) {
+            const confirmation = await confirmAsync({
+              message: `Variable ${name} already exists on this project. Do you want to overwrite it?`,
+            });
+
+            if (!confirmation) {
+              Log.log('Aborting');
+              return;
+            }
```
szdziedzic, line 155: ditto

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,251 @@
...
+
+      if (environment && !link) {
+        if (!nonInteractive) {
+          const confirmation = await confirmAsync({
+            message: `Unexpected argument: --environment can only be used with --link flag. Do you want to link the variable to the current project?`,
+          });
+
+          if (!confirmation) {
+            Log.log('Aborting');
+            return;
```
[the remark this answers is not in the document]
szdziedzic: I just checked it and I see that there are examples in the EAS CLI codebase where we do 
```
Log.log('Aborting');
return;
```
but it's mostly older 3/4 y.o. code.

In the newer code we usually stick to throwing an error
https://github.com/expo/eas-cli/blob/da9f949ea135c396cea8c0b8de3016feefd2f4a7/packages/eas-cli/src/rollout/actions/EndRollout.ts#L84-L86

The old `return` behavior should probably be changed by us as a part of some maintenance task.

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,251 @@
...
+          }
+        } else {
+          throw new Error('Unexpected argument: --environment can only be used with --link flag.');
+        }
```
szdziedzic, line 206: I believe that this could also be moved to the `validateFlags`/`sanitizeFlags` function, but this is more or less a nit

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,115 @@
...
+            .appVariables
+        : await EnvironmentVariablesQuery.sharedAsync(graphqlClient, projectId);
+
+    if (!name) {
+      if (nonInteractive) {
+        throw new Error(
+          `Environment variable needs 'name' to be specified when running in non-interactive mode. Run the command with ${chalk.bold(
+            '--name VARIABLE_NAME'
+          )} flag to fix the issue`
+        );
```
szdziedzic, line 65: ditto

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,115 @@
...
+      const confirmed = await toggleConfirmAsync({
+        message: `Are you sure you wish to proceed?${
+          selectedVariable.scope === EnvironmentVariableScope.Shared
+            ? ' This variable is applied across your whole account and may affect multiple apps.'
+            : ''
+        }`,
+      });
+      if (!confirmed) {
+        Log.error(`Canceled deletion of variable ${selectedVariable.name}.`);
+        return;
```
szdziedzic, line 107: ditto

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,110 @@
...
+      nonInteractive,
+    });
+
+    if (!name) {
+      name = await promptVariableNameAsync(nonInteractive);
+    }
+
+    if (environment && scope === EnvironmentVariableScope.Shared) {
+      throw new Error(`Unexpected argument: --environment can only be used with project variables`);
+    }
```
szdziedzic, line 55: ditto

packages/eas-cli/src/commands/env/get.ts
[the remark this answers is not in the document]
szdziedzic: Let's use one of these 2 ways for env vars and let's stick to it if possible

packages/eas-cli/src/utils/formatVariable.ts
```diff
@@ -0,0 +1,14 @@
+import dateFormat from 'dateformat';
+
+import formatFields from './formatFields';
+import { EnvironmentVariableFragment } from '../graphql/generated';
+
+export function formatVariable(variable: EnvironmentVariableFragment): string {
+  return formatFields([
+    { label: 'ID', value: variable.id },
+    { label: 'Name', value: variable.name },
+    { label: 'Value', value: variable.value || '' },
```
szdziedzic, line 10:
```suggestion
    { label: 'Value', value: variable.value ?? '' },
```
nit: prefer `??` over `||`

szdziedzic, line 10:
```suggestion
    { label: 'Value', value: variable.value ?? '(secret)' },
```

szdziedzic (changes requested):
Thanks for applying all of these changes, we are really close here! Good job! 🙏 

Sorry if these comments seem like nit-picking, let me know if you don't agree with some of them, butI try to serve my duty as a codeowner well and try to keep our codebase standardized:

packages/eas-cli/src/commands/env/delete.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (scope === EnvironmentVariableScope.Project) {
+      if (!environment) {
+        environment = await promptVariableEnvironmentAsync(nonInteractive);
+      }
+    }
+
+    const variables =
+      scope === EnvironmentVariableScope.Project && environment
+        ? (await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, projectId, environment))
+            .appVariables
```
szdziedzic, line 62: I believe that these can also include `shared` variables, right?

szdziedzic: One issue I believe the users will run into is a scenario when they do `eas env:delete --scope project` and select a shared variable then we delete it for them in `--non-interactive` mode or prompt them to delete it in standard mode. It can lead to some frustration among users because I believe if you are a new user and you don't understand our concepts and you do `eas env:delete --scope project` you probably want to delete the shared env variable only from this project (unlink it) rather then delete it globally. You probably would like to do so when doing `--scope shared`.

Maybe one should be unable to delete `shared` variables when doing a `--scope project`?

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,150 @@
...
+    } = await this.getContextAsync(EnvironmentVariableGet, {
+      nonInteractive,
+    });
+    if (!name) {
+      name = await promptVariableNameAsync(nonInteractive);
+    }
+
+    if (environment && scope === EnvironmentVariableScope.Shared) {
+      throw new Error(`Unexpected argument: --environment can only be used with project variables`);
+    }
```
szdziedzic, line 73: I think we already check it in `validateFlags` function

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    if (!environment && scope === EnvironmentVariableScope.Project) {
+      environment = await promptVariableEnvironmentAsync(nonInteractive);
+    }
+
+    assert(scope);
```
szdziedzic, line 79: Is it possible for `scope` to be `undefined`?

packages/eas-cli/src/commands/env/get.ts
```diff
@@ -0,0 +1,150 @@
...
+      return;
+    }
+    if (!variable.value) {
+      Log.log(
+        `${chalk.bold(
+          variable.name
+        )} is a secret variable and cannot be displayed once it has been created.`
+      );
+      return;
+    }
```
szdziedzic, line 94: Maybe this should be an error, what do you think?

packages/eas-cli/src/commands/env/link.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectConfig,
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    let {
+      flags: { name, 'non-interactive': nonInteractive, environment },
+    } = await this.parse(EnvironmentVariableLink);
```
szdziedzic, line 34: we probably should validate that `name` and `environment` are specified in `non-interactive` mode

packages/eas-cli/src/commands/env/unlink.ts
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectConfig,
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    let {
+      flags: { name, 'non-interactive': nonInteractive, environment },
```
szdziedzic, line 34: we should verify that name and environment are specified for non-interactive mode

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+      flags: {
+        name,
+        value,
+        scope,
+        'non-interactive': nonInteractive,
+        environment,
+        link,
+        visibility,
+      },
+    } = await this.parse(EnvironmentVariableUpdate);
```
szdziedzic, line 61: we should validate that `name`, `environemnt`, ... are specified when in non-interactive mode

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+      if (!variable) {
+        throw new Error(
+          `Could not update variable with name ${name} on project ${projectDisplayName}`
+        );
+      }
+
+      Log.withTick(
+        `Updated ${chalk.bold(name)} with value ${chalk.bold(value)} on project ${chalk.bold(
+          projectDisplayName
+        )}.`
```
szdziedzic, line 131:
```suggestion
      Log.withTick(
        `Updated env variable ${chalk.bold(name)} on project ${chalk.bold(
          projectDisplayName
        )}.`
```

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+            title: variable.name,
+            value: variable.name,
+          }))
+        );
+      }
+
+      const existingVariable = sharedVariables.find(variable => variable.name === name);
+      if (!existingVariable) {
+        throw new Error(
+          'Variable with this name already exists on this account. Please use a different name.'
```
szdziedzic, line 149:
```suggestion
          'Variable with this name doesn't exist on this account. Please use a different name.'
```

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+      const variable = await EnvironmentVariableMutation.createSharedVariableAsync(
+        graphqlClient,
+        {
+          name,
+          value,
+          visibility,
+          overwrite: true,
+        },
+        ownerAccount.id
+      );
```
szdziedzic, line 177: Sanity check: if we overwrite the shared env var, it is still linked to places it was linked to, right?

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      if (!variable) {
+        throw new Error(
+          `Could not update variable with name ${name} on account ${ownerAccount.name}`
+        );
+      }
+
+      Log.withTick(
+        `Updated ${chalk.bold(name)} with value ${chalk.bold(value)} on account ${chalk.bold(
```
szdziedzic, line 186:
```suggestion
        `Updated shared variable ${chalk.bold(name)} on account ${chalk.bold(
```

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -0,0 +1,209 @@
...
+        await EnvironmentVariableMutation.linkSharedEnvironmentVariableAsync(
+          graphqlClient,
+          variable.id,
+          projectId,
+          environment
+        );
+        Log.withTick(
+          `Linked shared variable ${chalk.bold(name)} to project ${chalk.bold(projectDisplayName)}.`
+        );
+      }
```
szdziedzic, line 206: It makes sense to make `link` a part of the create command, but I'm not sure if it's necessary for update

szdziedzic (changes requested):

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -1,5 +1,21 @@
...
+  EnvironmentVariableEnvironment,
+  EnvironmentVariableScope,
+  EnvironmentVariableVisibility,
+} from '../graphql/generated';
+
+// NOTE: not exactly true, but, provided mapToLowercase and upperCaseAsync
+// are used in tandem, it saves on unnecessary typying in commands
+async function upperCaseAsync<T>(input: string): Promise<T> {
+  return input.toUpperCase() as T;
+}
```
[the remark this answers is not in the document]
khamilowicz: Typescript uses this flag implicitly and won't accept `string` when `enum` is expected.

packages/eas-cli/src/commands/env/create.ts
```diff
@@ -0,0 +1,268 @@
...
+          environment
+        );
+        Log.withTick(
+          `Linked shared variable ${chalk.bold(name)} to project ${chalk.bold(projectDisplayName)}.`
+        );
+      }
+    }
+  }
+
+  private validateFlags(flags: CreateFlags): CreateFlags {
```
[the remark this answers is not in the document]
khamilowicz: I can imaging using `--force` flag in interactive mode

packages/eas-cli/src/commands/env/update.ts
khamilowicz: yes, we perform update on EnvironmentVariableEntity, without destroying it like previously.

packages/eas-cli/src/graphql/queries/EnvironmentVariablesQuery.ts
```diff
@@ -0,0 +1,130 @@
...
+      environment,
+      filterNames,
+    }: {
+      appId: string;
+      environment: EnvironmentVariableEnvironment;
+      filterNames?: string[];
+    }
+  ): Promise<{
+    appVariables: EnvironmentVariableFragment[];
+  }> {
```
szdziedzic, line 27: It's not used anywhere, is it? If it's not and we are dropping sudo requirement, we can delete this I guess and just use standard env vars query?

szdziedzic (approved):
Awesome, thanks for applying all of the review suggestions 👏 🚀

packages/eas-cli/src/graphql/queries/EnvironmentVariablesQuery.ts
```diff
@@ -0,0 +1,130 @@
...
+              $filterNames: [String!]
+              $environment: EnvironmentVariableEnvironment!
+            ) {
+              app {
+                byId(appId: $appId) {
+                  id
+                  environmentVariablesIncludingSensitive(
+                    filterNames: $filterNames
+                    environment: $environment
+                  ) {
```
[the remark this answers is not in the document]
khamilowicz: If we get rid of sudo entirely then yes, but if we decide to implement it anyway it will be useful to have `environmentVariablesIncludingSensitive` and `environmentVariables` separate.

packages/eas-cli/src/graphql/queries/EnvironmentVariablesQuery.ts
khamilowicz: ditto

szdziedzic (approved):

PR #2359, opened 2024-05-03
[eas-cli] Pass env from process.env and build profile to expo-updates CLI calls where applicable

wschurman:
# Why

This is the eas-cli equivalent of https://github.com/expo/eas-build/pull/388.

Essentially, when getting the project config in expo-updates CLI, the env must be supplied.

# How

Pass the env to the `spawnAsync`. Note that this passes the build profile env when possible as well (only in build command).

# Test Plan

Still need to come up with a test plan here. I think it will be:
1. Create a test app with an app.config.ts similar to https://github.com/expo/expo/blob/main/apps/eas-expo-go/app.config.ts and expo-updates installed (beta). (`yarn create expo-app --template default@beta`)
2. `eas init`, `eas update:configure`, `eas build:configure`
3. Add environment variable to eas.json env key in each profile.
4. Run `eas build -p android`, see it fails (current prod version of eas-cli)
5. Run `neas build -p android`, see it succeeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Pass env from process.env and build profile to expo-updates CLI calls where applicable. ([#2359](https://github.com/expo/eas-cli/pull/2359) by [@wschurman](https://github.com/wschurman))
+
 ### 🧹 Chores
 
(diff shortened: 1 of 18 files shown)


wschurman:
/changelog-entry bug-fix Pass env from process.env and build profile to expo-updates CLI calls where applicable

szdziedzic (approved):
looks reasonable, thanks!

PR #2363, opened 2024-05-08
[eas-json] update image tags in `eas.schema.json`

szdziedzic:
# Why

Update image tags in `eas.schema.json`

# How

Update image tags in `eas.schema.json`

# Test Plan

Tests


sjchmiela (approved):
Does it make sense to mention what is the image for a given SDK? We already show the mapping in docs, maybe there's little upside to having it here too? Not too strong opinion though.

szdziedzic:
>Does it make sense to mention what is the image for a given SDK? We already show the mapping in docs, maybe there's little upside to having it here too? Not too strong opinion though.

The benefit of having it here is that it shows in VSCode Expo plugin, so it's slightly more convenient for the people

sjchmiela:
I know. Still, if I'm on SDK XX do I care about the actual image doing my build as long as it is successful? (And if it isn't, the image is printed in the worker description.) To rephrase it differently — ideally no one using Expo wants/needs to use anything other than `sdk-XX` aliases.

szdziedzic:
I mean I don't have a strong opinion, I can just delete it

szdziedzic:
should be better now

PR #2365, opened 2024-05-09
[eas-cli] fix onboarding mutation

tchayen:
# Why

There's a bug in how we update onboarding data. UserPreferences is a key-value store with shallow comparison so we also need to pass the other info that was stored there for it to work.

# How

By spreading previous preferences.

# Test Plan

Not sure how to test it. @szdziedzic can you help me?

### DIFF
--- a/packages/eas-cli/src/commands/project/onboarding.ts
+++ b/packages/eas-cli/src/commands/project/onboarding.ts
@@ -174,8 +174,9 @@ export default class Onboarding extends EasCommand {
       });
     }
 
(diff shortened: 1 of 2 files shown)


szdziedzic (approved):
awesome, good catch!

how I would test it:
1. run `yarn start` in repo root
2. add `easd` to your path https://github.com/expo/eas-cli/blob/main/CONTRIBUTING.md#development
3. logout from your expo account `easd logout`
4. start new onboarding process
5. navigate to `configure:credentials` screen
6. run `easd init:onboarding`

PR #2368, opened 2024-05-09
[eas-cli] make sure that spinner is not used for interactive `git clone` command in `inint:onboarding`

szdziedzic:
# Why

Before:
screenshot_2024-05-09_at_18 56 01_720

spinner was printed on top of the logs which contained a prompt to accept something

# How

make sure that it's not used for `git clone`

# Test Plan

After:
screenshot_2024-05-09_at_19 04 12_720


tchayen (approved):
Thanks for changing it so quickly!

PR #2371, opened 2024-05-10
[eas-cli] improve prompt asking user to specify target clone dir in `eas init:onboarding`

szdziedzic:
# Why

improve prompt asking the user to specify target clone dir in `eas init:onboarding`

# How

Show the target path more clearly

# Test Plan

Screenshot 2024-05-10 at 14 00 17


tchayen (approved):
I like this idea

PR #2373, opened 2024-05-10
[eas-cli] add `init:onboarding` flow for no gh repo connected

szdziedzic:
# Why

add `init:onboarding` flow for no gh repo connected

# How

- clone default expo template
- reset git repo
- generate needed app.json and eas.json
- install expo libs (updates, dev client, insights)
- run a build

# Test Plan

test manually


tchayen (approved):
The onboarding-related parts look good to me.

Ideally in future we would be able to avoid repeating the backend code but for now I don't see any obvious way to do it.

PR #2378, opened 2024-05-13
[eas-cli] diagnostics: add loader and progress information

Simek:
# Why

Let's give users an information that something is happening when `diagnostics` take a while to gather data.

# How

Add Ora spinner with simple process progress information + small formatting tweaks.

# Test Plan

The changes have been tested locally.

# Preview

https://github.com/expo/eas-cli/assets/719641/0f9fa5aa-4d5f-4b97-9c48-b09c443ab94b

### DIFF
--- a/packages/eas-cli/src/commands/diagnostics.ts
+++ b/packages/eas-cli/src/commands/diagnostics.ts
@@ -1,8 +1,10 @@
 import { Platform } from '@expo/eas-build-job';
+import chalk from 'chalk';
 import envinfo from 'envinfo';
 
 import EasCommand from '../commandUtils/EasCommand';
 import Log from '../log';
+import { ora } from '../ora';
 import { resolveWorkflowAsync } from '../project/workflow';
 import { easCliVersion } from '../utils/easCli';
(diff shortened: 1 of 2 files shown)


Simek:
/changelog-entry [chore] [Add loader and progress information to `diagnostics` command]

packages/eas-cli/src/commands/diagnostics.ts
```diff
@@ -20,6 +22,7 @@ export default class Diagnostics extends EasCommand {
       nonInteractive: true,
     });
 
+    const spinner = ora().start(`Gathering diagnostic information...`);
     const info = await envinfo.run(
       {
         System: ['OS', 'Shell'],
```
szdziedzic, line 28: Maybe we should `try / catch` it and call `spinner.failed` when there is an error?

Simek: AFAIK `envinfo` always "succeedes" and only prints error messages instead of throwing on read failures.
* https://github.com/tabrindle/envinfo

szdziedzic (approved):
awesome

PR #2397, opened 2024-05-22
[eas-cli] use correct app config for no gh `init:onboarding` flow

szdziedzic:
# Why

Instead of extending and overriding some parts of the base config we just created a new, incomplete one

# How

extend and override base config from `expo-template-default` instead of creating a new incomplete one

# Test Plan

test manually


Simek (approved):

szdziedzic:
/changelog-entry bug-fix Use the correct app config for no GitHub flow in `init:onboarding`

PR #2406, opened 2024-06-03
Fix expiration date check for build list source

sjchmiela:
# Why

The check for build expiration was wrong.

# How

Instead of relying on an old hardcoded value, use `expirationDate` fetched from GraphQL.

# Test Plan

Adjusted tests. Also tested manually:
Zrzut ekranu 2024-06-4 o 11 33 06

```
43f278b2-3257-49c6-9c01-615a3f958f27	2023-07-20 14:36:18.021724+00
5167ab23-3913-4741-8e08-db75c9c420ab	2024-07-26 19:54:01.386308+00
f73eec2a-3936-40e3-9504-1cb4ebb29e12	2024-04-27 15:25:24.338+00
```


sjchmiela:
/changelog-entry bug-fix Disallow picking expired builds as submit archive source

szdziedzic (approved):

Issue #2404, opened 2024-05-30
Resign command doesn't filter builds by profile


d4rky-pl:
### Summary

When running `eas build:resign --platform ios --profile testing` the interactive list of builds includes builds from all profiles, not just from `testing`

### Managed or bare?

managed

### Environment

```
  expo-env-info 1.2.0 environment info:
    System:
      OS: macOS 14.4
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 16.15.0 - ~/.asdf/installs/nodejs/16.15.0/bin/node
      Yarn: 1.22.15 - ~/.asdf/installs/nodejs/16.15.0/bin/yarn
      npm: 8.5.5 - ~/.asdf/plugins/nodejs/shims/npm
      Watchman: 2024.05.06.00 - /opt/homebrew/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: DriverKit 23.5, iOS 17.5, macOS 14.5, tvOS 17.5, visionOS 1.2, watchOS 10.5
      Android SDK:
        API Levels: 33, 33, 34
        Build Tools: 30.0.3, 33.0.0, 33.0.2, 34.0.0, 34.0.0
        System Images: android-33 | Google APIs ARM 64 v8a, android-34 | Google APIs ARM 64 v8a
    IDEs:
      Xcode: 15.4/15F31d - /usr/bin/xcodebuild
    npmPackages:
      @expo/webpack-config: ^18.0.1 => 18.0.2
      babel-preset-expo: ^9.3.0 => 9.3.1
      expo: ^48.0.0 => 48.0.9
      react: 18.2.0 => 18.2.0
      react-dom: 18.2.0 => 18.2.0
      react-native: 0.71.4 => 0.71.4
      react-native-web: ~0.18.11 => 0.18.12
    npmGlobalPackages:
      eas-cli: 9.1.0
    Expo Workflow: managed
```

### Reproducible demo or steps to reproduce from a blank project

1. Create a project with `profileA` and `profileB` in eas.json
2. Build iOS apps for both profiles
3. Run `eas build --platform ios --profile profileA`

d4rky-pl:
@szdziedzic I mean `build:resign`, my bad. There's no --buildProfile flag according to the documentation

(sorry for closing the ticket, butterfingers 😅)

szdziedzic:
Oh, sorry, now I get it. Yeah, you are right, thanks for report!

szdziedzic:
[...]

PR #2408, opened 2024-06-05
[eas-cli] display build profile in `eas build:list` command details

szdziedzic:
# Why

When looking at https://github.com/expo/eas-cli/issues/2404 I noticed that we don't display build profile info in this list, yet we allow people to filter by it which makes it confusing

# How

Display build profile info as well

# Test Plan

test manually


szdziedzic:
/changelog-entry new-feature Display build profile in the output of `eas build:list`

radoslawkrzemien (approved):

PR #2409, opened 2024-06-05
[eas-cli] make flags for `eas build:list` command more aligned with flags for rest of the commands

szdziedzic:
# Why

We use `--this-convention` to name our flags in the rest of the EAS CLI, meanwhile `eas build:list` has been using flags with `--thisConvention`.

Let's unify it.

# How

Unify flag naming convention for `eas build:list`. Add `-p` and `-e` aliases for `--platform` and `--profile` flags as we do for `eas build`.

This is not a breaking change because we still keep the aliases around.

# Test Plan

test manually, tests


szdziedzic:
/changelog-entry chore Make flags for `eas build:list` command more aligned with flags for rest of the commands

radoslawkrzemien (approved):
Noice

PR #2410, opened 2024-06-05
[eas-cli] add `target-profile` and `source-profile` flags to `eas build:resign`

szdziedzic:
# Why

https://github.com/expo/eas-cli/issues/2404

# How

Add 2 new flags:
- `source-profile`
- `target-profile`

The previous `profile` flag was really a `target-profile`, so it can be an alias for it so we don't introduce a breaking change.

See https://github.com/expo/eas-cli/issues/2404 for more context

# Test Plan

Test manually


szdziedzic:
@d4rky-pl do you want to take a look?

szdziedzic:
/changelog-entry new-feature Add `target-profile` and `source-profile` flags to the `eas build:resign` command

packages/eas-cli/src/commands/build/resign.ts
```diff
@@ -65,10 +67,16 @@ export default class BuildResign extends EasCommand {
...
       char: 'e',
       description:
-        'Name of the build profile from eas.json. Defaults to "production" if defined in eas.json.',
+        'Name of the target build profile from eas.json. Used to resolve credentials for re-signing. Defaults to "production" if defined in eas.json.',
+      helpValue: 'PROFILE_NAME',
+      aliases: ['profile'],
+    }),
+    'source-profile': Flags.string({
+      description:
+        'Name of the source build profile from eas.json. Used to filter builds eligible for re-signing.',
```
khamilowicz, line 79: I don't think descriptions are really helpful. Will users know what these refer to?

khamilowicz (approved):

packages/eas-cli/src/commands/build/resign.ts
szdziedzic: I hope so, I mean I understood it and I assumed that everybody else would also understand 😅. But maybe that's not the case 😂.  Do you have any suggestions about how I can improve it?

d4rky-pl: The `source-profile` description seems clear enough to me but I agree the `target-profile` description may not be immediately obvious if you don't have the context (that I now have after reading your explanation in #2404)

I'm afraid I can't come up with a better suggestion right now (traveling and not in the right space to think carefully) but I can take a second look next week if nobody else steps in to help :)

radoslawkrzemien: Maybe `Name of the target build profile from eas.json. Credentials and environment variables from this profile will be used when re-signing. Defaults to "production" if defined in eas.json.` ?

radoslawkrzemien (approved):

PR #2411, opened 2024-06-06
[eas-cli] [ENG-11843] Change default `appVersionSource` to `REMOTE`

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-11843/change-default-appversionsource-to-remote
Companion to: https://github.com/expo/expo/pull/29501

# How

When `appVersionSource` is `undefined`, the user is now prompted to select what they want to do:
- update their `eas.json` to set `appVersionSource` to `remote`, which is described as the default behavior
- update their `eas.json` to set `appVersionSource` to `local`, which is described as the previous default behavior
- not update their `eas.json`, abort the command, and configure manually

This way:
- users who had `appVersionSource` explicitly configured will not be affected by the change
- users who used the `local` as a default knowingly will be prompted once to automatically set it explicitly
- users who used the `local` as a default unknowingly, but still want to use it will be prompted once to automatically set it explicitly
- users who used the `local` as a default unknowingly, but would prefer to use `remote` instead will be prompted once to switch and automatically set it
- users who don't know what this setting does, will have a chance to abort the command for now and will get a link to the docs page to learn more

# Test Plan
## Automated tests
Updated the automated tests, and added new cases for different user selections

### `appVersionSource` not set and user chooses to set it to `remote`
Screenshot 2024-06-12 at 11 18 12
Screenshot 2024-06-12 at 11 18 35

### `appVersionSource` not set and user chooses to set it to `local`
Screenshot 2024-06-12 at 11 18 59
Screenshot 2024-06-12 at 11 19 13

### `appVersionSource` not set and the user chooses to abort and configure manually
Screenshot 2024-06-12 at 11 19 35
Screenshot 2024-06-12 at 11 19 51

### Edit - updated message ☝️ 
Screenshot 2024-06-12 at 13 44 09


radoslawkrzemien:
/changelog-entry chore Treat undefined appVersionSource as REMOTE as a default

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -12,21 +12,25 @@ export async function ensureVersionSourceIsRemoteAsync(
...
     return;
   }
   if (nonInteractive) {
     throw new Error(
       `This project is not configured for using remote version source. Add ${chalk.bold(
         '{"cli": { "appVersionSource": "remote" }}'
-      )} in eas.json or re-run this command without "--non-interactive" flag.`
+      )} in eas.json (or remove ${chalk.bold(
+        '{"cli": { "appVersionSource": "local" }}'
+      )} to use "remote" as a default) or re-run this command without "--non-interactive" flag.`
```
szdziedzic, line 24:
```suggestion
      )} in eas.json or re-run this command without the "--non-interactive" flag.`
```

It doesn't seem necessary. Overwhelming users with too much information might confuse them.

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -12,21 +12,25 @@ export async function ensureVersionSourceIsRemoteAsync(
...
     );
   }
 
   Log.log(
     `The app version source defines whether the app version is stored locally in your project source (e.g. in app.json, Info.plist, or build.gradle) or remotely on EAS servers and only applied to the project at build time. To use this command, you will need to enable the remote version policy by adding  ${chalk.bold(
       '{"cli": { "appVersionSource": "remote" }}'
-    )} in eas.json.`
+    )} in eas.json or removing ${chalk.bold(
+      '{"cli": { "appVersionSource": "local" }}'
+    )} to use "remote" as a default.`
```
szdziedzic, line 33:
```suggestion
    )} in eas.json`
```

szdziedzic (changes requested):
>Maybe query the user with undefined appVersionSource if they want to update eas.json with appVersionSource="local" or keep it as-is

I fear that it can be confusing for people. I mean they are on a happy path using some default settings and out of nowhere we are prompting them if they want to keep using some policy they see for the first time in their life.  On the other hand, prompting probably would be a better option for more aware and advanced users, our communication would also be more explicit. As a counter-argument to the previous sentence, we can say that this is a breaking change for a reason, and breaking changes are supposed to be breaking 🤔.

What do you think @brentvatne @kadikraman @sjchmiela?

One more Q: 
Did you test manually how it would work in the case when:
1. you are using a current default (local)
2. you have some build number 123 in your app.json
3. you then update the CLI to the version with a new default behavior

? 
Will we initialize the remote build number with the value from app.json? Will the transition be smooth and not confusing for a user?

radoslawkrzemien:
@szdziedzic I don't think it will necessary be confusing if we word the message well. We can explain what is happening and why and then give them a choice
- change eas.json to local (keep the same behaviour)
- change eas.json to remote (change appVersionSource explicitly)
- keep eas.json unchanged (change appVersionSource implicitly)

kadikraman:
@szdziedzic From the user's point of view it might make more sense to have the field mandatory (instead of changing the default) 🤔
Then, if someone had this field defined already - nothing changes
If they didn't - prompt them to add `<current default>`
And new users would have to choose.

radoslawkrzemien:
@kadikraman please have another look whenever you have the time. WDYT about the flow now?

radoslawkrzemien:
@sjchmiela @szdziedzic Thoughts/opinions on how it's done now?

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -102,3 +127,62 @@ export function getBuildVersionName(platform: Platform): string {
     return 'buildNumber';
   }
 }
+
+export async function ensureAppVersionSourceIsSetAsync(
+  easJsonAccessor: EasJsonAccessor
+): Promise<AppVersionSourceUpdateOption | undefined> {
+  Log.log(
+    `This project is using the default app version source, which currently is the "remote" version source. Prior to version 10.0.0 of CLI the default used to be the "local" version source. App version source can be set for you automatically, or you can configure it yourself - if you wish to use the default "remote" version source add ${chalk.bold(
```
kadikraman, line 135: I like it a lot, I think this is the clearest option! Could I suggest a small wording change? I think it would help to briefly explain what the version source means.


**Please select your app version source. With "local" the build android.versionCode / ios.buildNumber needs to be incremented manually in app.json / app.config.js, and with "remote" this value is stored in EAS and will be incremented automatically with each build. Until now, this project has been using the "local" version source (which was previously the default). App version source can be set for you automatically, or you can configure it yourself - if you wish to use the default "remote" version source add**

kadikraman (approved):
I think this is great! Just a small comment on explaining what the version source does, but other than that, looks good to me!

I think it's worth explaining, because the field it called "version source", but it's actually incrementing what most people think of as the "build number" so it confuses people.

packages/eas-cli/src/project/remoteVersionSource.ts
radoslawkrzemien: I updated the message, thank you for the feedback 🙇

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -45,11 +60,21 @@ export async function ensureVersionSourceIsRemoteAsync(
...
+  projectDir: string
+): Promise<void> {
+  if (cliConfig?.appVersionSource === undefined) {
+    const easJsonAccessor = EasJsonAccessor.fromProjectPath(projectDir);
+    const selection = await ensureAppVersionSourceIsSetAsync(easJsonAccessor);
+    if (selection === AppVersionSourceUpdateOption.SET_TO_LOCAL && cliConfig) {
+      cliConfig.appVersionSource = AppVersionSource.LOCAL;
+    } else if (selection === AppVersionSourceUpdateOption.SET_TO_REMOTE && cliConfig) {
+      cliConfig.appVersionSource = AppVersionSource.REMOTE;
+    }
```
szdziedzic, line 75: maybe we wouldn't need to duplicate this logic in multiple places if `await ensureAppVersionSourceIsSetAsync(easJsonAccessor)` returned the updated `cliConfig`?

szdziedzic (changes requested):
I believe we should also update 
https://github.com/expo/eas-cli/blob/7139e4f57689ec1ab1e9acd85924e2a5a6bc949f/packages/eas-cli/src/build/configure.ts#L55-L72

https://github.com/expo/eas-cli/blob/7139e4f57689ec1ab1e9acd85924e2a5a6bc949f/packages/eas-cli/src/commands/project/onboarding.ts#L411-L440

and

https://github.com/expo/universe/blob/794a217a3480a8e44773a1f7b02500197a5dfecf/server/www/src/utils/onboarding/configureEasInRepositoryAsync.ts#L303-L332

packages/eas-cli/src/build/android/build.ts
```diff
@@ -76,7 +76,7 @@ This means that it will most likely produce an AAB and you will not be able to i
     gradleContext
   );
   const versionCodeOverride =
-    ctx.easJsonCliConfig?.appVersionSource === AppVersionSource.REMOTE
+    ctx.easJsonCliConfig?.appVersionSource !== AppVersionSource.LOCAL
```
szdziedzic, line 79: I would stick to 
```suggestion
    ctx.easJsonCliConfig?.appVersionSource === AppVersionSource.REMOTE
```
because of the `--non-interactive` issues I highlighted in the previous comment

packages/eas-cli/src/commands/project/onboarding.ts
```diff
@@ -411,6 +411,7 @@ async function configureProjectFromBareDefaultExpoTemplateAsync({
   const easJson: EasJson = {
     cli: {
       version: `>= ${easCliVersion}`,
+      appVersionSource: AppVersionSource.REMOTE,
```
szdziedzic, line 414: ditto

szdziedzic (changes requested):
I left some comments. If they end up being implemented I believe it would be good to request the review from Brent/Kadi as well.

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +123,84 @@ export function getBuildVersionName(platform: Platform): string {
...
+    Errors.error('"appVersionSource" must be configured in non-interactive mode', { exit: 1 });
+  }
+
+  Log.log(
+    `Please select your app version source. With "local" the build android.versionCode / ios.buildNumber needs to be incremented manually in app.json / app.config.js, and with "remote" this value is stored in EAS and will be incremented automatically with each build. Until now, this project has been using the "local" version source (which was previously the default). App version source can be set for you automatically, or you can configure it yourself - if you wish to use the default "remote" version source add ${chalk.bold(
+      '{"cli": { "appVersionSource": "remote" }}'
+    )} to your eas.json or if you want to use the "local" version source add ${chalk.bold(
+      '{"cli": { "appVersionSource": "local" }}'
+    )} to your eas.json.`
+  );
```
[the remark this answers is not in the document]
radoslawkrzemien: So what would you do if they don't have autoIncrement:true / have runtimeVersion.policy: nativeVersion / have autoIncrement: version ?
Don't prompt them and just set it to local, plus maybe display a message that it has been done?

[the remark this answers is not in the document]
szdziedzic: > if they don't have autoIncrement:true

I think we can just print a warning that this field will be required in the future, but not stop them from doing what they do to select the version source in a prompt? Does it make sense? Most of our users don't know what the version source is, so it would be confusing to them to be forced to understand it and set it, while they never relied on it.

>have runtimeVersion.policy: nativeVersion / have autoIncrement: version

Set the source  to `local`

@kadikraman what do you think?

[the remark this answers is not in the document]
radoslawkrzemien: Hm, I'm wondering how best to approach it - the `appVersionSource` setting is profile-agnostic, while autoIncrement is not. If the user runs a non-production build then almost certainly `autoIncrement` is not set, but it could be for a `production` profile 🤔

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +123,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+
+export async function ensureAppVersionSourceIsSetAsync(
+  easJsonAccessor: EasJsonAccessor,
+  easJsonCliConfig: EasJson['cli'] | undefined,
+  nonInteractive: boolean
+): Promise<EasJson['cli'] | undefined> {
+  let selectOption;
+  if (nonInteractive) {
+    Log.warn(
+      '"appVersionSource is not specified. The previous default value for "appVersionSource" used to be "local", but it has been changed to "remote".'
```
amandeepmittal, line 135:
```suggestion
      '"appVersionSource" is not specified. The previous default value for "appVersionSource" used to be "local", but it has been changed to "remote".'
```

Let's add that missing end quote on the first word.

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +123,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+    );
+    Log.warn(
+      'Cannot prompt to select the value of "appVersionSource" in non-interactive mode. Setting it to "local" to keep the backwards compatibility. If you wish to use the "remote" "appVersionSource" update the "eas.json" manually.'
```
amandeepmittal, line 138:
```suggestion
      'Cannot prompt to select the value of "appVersionSource" in non-interactive mode. Setting it to "local" to keep the backwards compatibility. If you wish to use the "remote" "appVersionSource", update the "eas.json" manually.'
```

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +123,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+      learnMore('https://docs.expo.dev/build-reference/app-versions/', {
+        learnMoreMessage: 'See the docs to learn more.',
+        dim: false,
+      })
+    );
+    selectOption = AppVersionSourceUpdateOption.SET_TO_LOCAL;
+    process.env.EAS_BUILD_AUTOCOMMIT = '1'; // the user can't commit changes to eas.json in non-interactive mode
+  } else {
+    Log.log(
+      `Please select your app version source. With "local" the build android.versionCode / ios.buildNumber are taken from app.json / app.config.js files and need to be incremented manually or automatically by setting autoIncrement: true in eas.json. With "remote" this value is handled remotely by EAS and can also be incremented automatically with each build when setting autoIncrement: true in eas.json. Until now, this project has been using the "local" version source (which was previously the default when the app version source was not specified). App version source can be set for you automatically, or you can configure it yourself - if you wish to use the default "remote" version source add ${chalk.bold(
```
amandeepmittal, line 150:
```suggestion
      `Please select your app version source. With "local", the build android.versionCode / ios.buildNumber are taken from app.json / app.config.js files and need to be incremented manually or automatically by setting autoIncrement: true in eas.json. With "remote", this value is handled remotely by EAS and can also be incremented automatically with each build by setting autoIncrement: true in eas.json. Until now, this project has been using the "local" version source (which was previously the default when the app version source was not specified). App version source can be set for you automatically, or you can configure it yourself - if you wish to use the default "remote" version source add ${chalk.bold(
```

amandeepmittal (commented):
Left a few suggestions for the copy of the CLI message.

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +136,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+      },
+      {
+        title: 'Update eas.json to use "local" version source',
+        value: AppVersionSourceUpdateOption.SET_TO_LOCAL,
+      },
+      {
+        title: "Don't update eas.json, abort command and configure manually",
+        value: AppVersionSourceUpdateOption.ABORT,
+      },
+    ]);
```
szdziedzic, line 183:
```suggestion
    selectOption = await selectAsync(`What would you like to do?`, [
      {
        title: 'Update eas.json to use the default "remote" version source (recommended)',
        value: AppVersionSourceUpdateOption.SET_TO_REMOTE,
      },
      {
        title: 'Update eas.json to use "local" version source (old behavior)',
        value: AppVersionSourceUpdateOption.SET_TO_LOCAL,
      },
      {
        title: "Don't update eas.json, abort command and configure manually",
        value: AppVersionSourceUpdateOption.ABORT,
      },
    ]);
```

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +136,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+    }
+    Log.withTick('Updated eas.json');
+  } else {
+    Log.warn(`You'll need to configure ${chalk.bold('appVersionSource')} manually.`);
+    Log.warn(
+      learnMore('https://docs.expo.dev/build-reference/app-versions/', {
+        learnMoreMessage: 'See the docs on what are the options and how to do it.',
+        dim: false,
+      })
+    );
```
szdziedzic, line 218:
```suggestion
    Log.warn(`You'll need to configure ${chalk.bold('appVersionSource')} manually. ${learnMore('https://docs.expo.dev/build-reference/app-versions/')}`);
```

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -1,18 +1,33 @@
...
+  if (easJsonCliConfig?.appVersionSource === undefined) {
+    easJsonCliConfig = await ensureAppVersionSourceIsSetAsync(
+      easJsonAccessor,
+      easJsonCliConfig ?? undefined,
+      nonInteractive
+    );
+  }
+  if (easJsonCliConfig?.appVersionSource !== AppVersionSource.LOCAL) {
     return;
   }
```
szdziedzic, line 32:
```suggestion
  if (easJsonCliConfig?.appVersionSource === AppVersionSource.REMOTE) {
    return;
  }
```

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -45,10 +60,33 @@ export async function ensureVersionSourceIsRemoteAsync(
...
+    profileInfo.profile.autoIncrement !== 'version'
+  ) {
+    if (profileInfo.profile.autoIncrement !== true) {
+      Log.log('The field "cli.appVersionSource" is not set, but it will be required in the future');
+      Log.log(
+        learnMore('https://docs.expo.dev/build-reference/app-versions/', {
+          learnMoreMessage: 'See the docs to learn more.',
+          dim: false,
+        })
+      );
```
szdziedzic, line 80:
```suggestion
      Log.warn(`The field "cli.appVersionSource" is not set, but it will be required in the future. ${learnMore('https://docs.expo.dev/build-reference/app-versions/')}`);
```

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -412,7 +400,28 @@ async function prepareAndStartBuildAsync({
...
+    } else {
+      const easJsonAccessor = EasJsonAccessor.fromProjectPath(projectDir);
+      easJsonCliConfig = await ensureAppVersionSourceIsSetAsync(
+        easJsonAccessor,
+        easJsonCliConfig,
+        flags.nonInteractive
+      );
+    }
+  }
+  if (easJsonCliConfig?.appVersionSource !== AppVersionSource.LOCAL) {
```
szdziedzic, line 424:
```suggestion
  if (easJsonCliConfig?.appVersionSource === AppVersionSource.REMOTE) {
```

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +136,88 @@ export function getBuildVersionName(platform: Platform): string {
...
+    );
+    Log.warn(
+      'Cannot prompt to select the value of "appVersionSource" in non-interactive mode. Setting it to "local" to keep the backwards compatibility. If you wish to use the "remote" "appVersionSource", update the "eas.json" manually.'
+    );
+    Log.warn(
+      learnMore('https://docs.expo.dev/build-reference/app-versions/', {
+        learnMoreMessage: 'See the docs to learn more.',
+        dim: false,
+      })
+    );
```
szdziedzic, line 158:
```suggestion
    Log.warn(`The field "cli.appVersionSource" is not set, but it will be required in the future. ${learnMore('https://docs.expo.dev/build-reference/app-versions/')}`);
```

sjchmiela (approved):
If we're going to require an explicit value here, does it make sense to change the default? My thinking — existing projects probably (?) already have _some_ way of managing versions (or at least understanding of what's happening). Let's fill the value to `"local"` in `eas.json` without asking the user for confirmation, instead printing a warning and a suggestion to move to `"remote"`? New projects will use the recommended way.

Also, are we sure that moving from "local" to "remote" source does not do anything funny (like, the version maintains its current value and only goes up)?

Just my 2 cents.

radoslawkrzemien:
@sjchmiela After the previous review discussions the default is actually unchanged - if it is not set and the user isn't prompted to choose the value it is still treated as `local`. And the user is only prompted to choose when `appVersionSource` is not set and `autoIncrement` is set to `true` (and without `non-interactive`)

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +134,96 @@ export function getBuildVersionName(platform: Platform): string {
...
+    Log.warn(
+      `The field "cli.appVersionSource" is not set, but it will be required in the future. ${learnMore(
+        'https://docs.expo.dev/build-reference/app-versions/'
+      )}`
+    );
+    selectOption = AppVersionSourceUpdateOption.SET_TO_LOCAL;
+    updateEasJson = false;
+  } else {
+    Log.log(
+      'Since EAS CLI version `11.0.0` explicitly specifying app version source is required. Please select your app version source:'
```
szdziedzic, line 154: We probably need to update the version here to `12.0.0`

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +134,96 @@ export function getBuildVersionName(platform: Platform): string {
...
+          ...easJsonRawObject?.cli,
+          appVersionSource: AppVersionSource.LOCAL,
+        };
+        return easJsonRawObject;
+      });
+      await easJsonAccessor.writeAsync();
+    }
+    if (easJsonCliConfig) {
+      easJsonCliConfig.appVersionSource = AppVersionSource.LOCAL;
+    }
```
szdziedzic, line 201: awesome 👏 

did you test that `local` will be respected here for `--non-interactive` even if `eas.json` file wasn't updated?

packages/eas-cli/src/project/remoteVersionSource.ts
```diff
@@ -98,3 +134,96 @@ export function getBuildVersionName(platform: Platform): string {
...
+
+export async function ensureAppVersionSourceIsSetAsync(
+  easJsonAccessor: EasJsonAccessor,
+  easJsonCliConfig: EasJson['cli'] | undefined,
+  nonInteractive: boolean
+): Promise<EasJson['cli'] | undefined> {
+  let selectOption, updateEasJson;
+  if (nonInteractive) {
+    Log.warn(
+      `The field "cli.appVersionSource" is not set, but it will be required in the future. ${learnMore(
```
szdziedzic, line 146:
```suggestion
    Log.warn(
      `The field "cli.appVersionSource" is not set, but it will be required in the future Proceeding with the default "local" value. ${learnMore(
```

szdziedzic (approved):
awesome 🚀

packages/eas-cli/src/project/remoteVersionSource.ts
radoslawkrzemien: Pretty sure I did, but I will again

radoslawkrzemien: Yes, it is

PR #2412, opened 2024-06-07
fix: update lockfile to only include `@types/node@20.11.0`

byCedric:
# Why

Running `yarn typecheck` in **packages/eas-cli** fails due to outdated `@types/node` being used. 

This is caused because our lockfile resolves to the outdated node types.
- Some libraries are using `dependencies: @types/node: *`, which is ok
- Our lockfile resolves `@types/node@*` to `14.0.27`, which is not ok

# How

- Update `@types/node` in our lockfile to only contain `20.11.0`

# Test Plan

Run this before changing anything:
- `$ node --print "require('@types/node/package.json').version"`
- Should print `14.0.27`, which is not correct

After pulling this PR:
- `$ yarn install --frozen-lockfile`
- `$ cd packages/eas-cli`
- `$ yarn typecheck` (should pass)
- `$ node --print "require('@types/node/package.json').version"` (should print `20.11.0`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Update lockfile to only include `@types/node@20.11.0`. ([#2412](https://github.com/expo/eas-cli/pull/2412) by [@byCedric](https://github.com/byCedric))
+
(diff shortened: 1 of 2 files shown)


byCedric:
/changelog-entry chore Update lockfile to only include `@types/node@20.11.0`

sjchmiela (approved):

PR #2413, opened 2024-06-07
chore!: drop support for Node 16

byCedric:
# Why

- Node 16 is End Of Life (EOL)
- Node 18 is currently in maintenance LTS
- Node 20 is currently in active LTS
- Node 22 is scheduled to take active LTS status on **2024-10-29**

> Source: https://nodejs.org/en/about/previous-releases

To support Node 22, we'd need to remove `node-fetch`. Starting from Node 18, we can use the built-in fetch which is based on `undici`. Unfortunately, we still need the `undici` package to maintain support for `https_proxy` functionality through `ProxyAgent`.

Dropping support for Node 16 helps in this transition, and matches our other tooling support (which is LTS+ - both maintenance and active LTS versions).

# How

- Removed Node 16 from workflow
- Added Node 22 to workflow

# Test Plan

See GitHub Actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['16', '18', '20']
+        node: ['18', '20', '22']
         include:
           - node: '20'
             coverage: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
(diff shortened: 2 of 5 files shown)


byCedric:
/changelog-entry chore Update test workflow Node versions to 18, 20, and 22

sjchmiela (changes requested):
Should we also bump `engines.node`s in `package.json`s?

byCedric:
> Should we also bump `engines.node`s in `package.json`s? x-ref: [expo/eas-build#406](https://github.com/expo/eas-build/pull/406)

Yeah, if we decide to drop Node 16 (which I'm suggesting here) we should update the `engines` too! Sorry, overlooked that one.

sjchmiela (approved):

szdziedzic (approved):

PR #2414, opened 2024-06-07
refactor(eas-cli): swap `node-fetch` for `undici` to support Node 22 better

byCedric:
> ⚠️ Warning, there seems to be an issue related to `FormData` and fetch. Right now, we are using `import 'form-data'`, which doesn't get handled properly with `undici` or Node fetch. https://github.com/expo/eas-cli/pull/2414#discussion_r1633244275

# Why

Fixes ENG-12411

It also brings us closer to Node's own `fetch` implementation, based on `undici` and is fetch-spec compliant. This is to improve Node 22 support, without warnings of deprecated system packages.

# How

- ~~Cherry-picked #2412~~, required for typechecks
  → **landed & rebased**
- ~~Cherry-picked #2413~~ required for tests (Node 18+)
  → **landed & rebased**
- Swapped `node-fetch` for `undici`
- Upgraded `nock` to `14.0.0-beta.7` for built-in fetch support
  - Worked around `undici.fetch` issue, `nock` is only attaching to `global.fetch` ([see here](https://github.com/nock/nock/issues/2183))

# Test Plan

See tests in GitHub Actions


packages/eas-cli/src/commandUtils/context/contextUtils/createGraphqlClient.ts
```diff
@@ -44,19 +43,17 @@ export function createGraphqlClient(authInfo: {
       }),
       fetchExchange,
     ],
-    // @ts-expect-error Type 'typeof fetch' is not assignable to type '(input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>'.
-    fetch,
-    fetchOptions: (): RequestInit => {
+    // @ts-expect-error - Type '(url: RequestInfo, init?: RequestInit) => Promise<Response>' is not assignable to type '{ (input: RequestInfo | URL, init?: RequestInit | undefined): Promise<Response>; (input: string | Request | URL, init?: RequestInit | undefined): Promise<...>; }'.
+    fetch: (url: RequestInfo, init?: RequestInit) =>
```
ide, line 47: Does this fix the type error?
```suggestion
    fetch: (url: RequestInfo | URL, init?: RequestInit) =>
```

packages/eas-cli/src/fetch.ts
```diff
@@ -13,23 +19,21 @@ export class RequestError extends Error {
...
+  const response = await fetch(url, init);
   if (response.status >= 400) {
     throw new RequestError(`Request failed: ${response.status} (${response.statusText})`, response);
   }
+
   return response;
 }
+
+function installProxyAgent(): void {
+  const httpsProxyUrl = process.env.https_proxy;
```
ide, line 34: Optional but I like to import `node:process` to be explicit.

packages/eas-cli/src/commandUtils/context/contextUtils/createGraphqlClient.ts
byCedric: Unfortunately not, it seems that typescript interprets the `@types/node` differently. 

TL;DR;
- in Node, they use `url: string | URL | globalThis.Request`
- in Undici, they use `url: RequestInfo`, where `type RequestInfo = string | URL | Request` ([see here](https://github.com/nodejs/undici/blob/250c863ce22a12223217c2d5cb2841b99c40b8c1/types/fetch.d.ts#L12))

These should not conflict, but I have a feeling that typescript is trying to load the web fetch type somehow.

byCedric: Since we might be dropping support for Node 16, we could just avoid importing `fetch` from undici though. That should resolve this typing issue as well.

byCedric: Opening a stacked PR for this, we need to refactor more than just removing the `import { fetch } from 'undici'`. This stacked PR may or may not succeed, depending on how much work still is required.

https://github.com/expo/eas-cli/pull/2419

byCedric: Ok, so there is an issue. The summary is:
- `FormData` is used from the `form-data` package, which is old and uses `util.isArray` (last update was 3y ago)
- `form-data` is different from `undici`'s `FormData` / Node built-in `FormData`
- `form-data` supports Node's `fs.createReadStream` instances (`ReadStream` types)
- `undici` does NOT support Node's `ReadStream` type, and only works with `File` (added in `18.13.0`) or Blob (added in `14.x.x`)
- Starting from Node `20.x.x`, there is a `fs.openAsBlob` method to stream files as blobs, and pass it to `FormData`, but we support Node 18 still

There are rather "dirty" hacks to work around this limitation, e.g. [this one](https://stackoverflow.com/a/76957447). I also have concerns with `undici`'s `FormData` not sending the `Content-Length` header.

Right now, `undici` also doesn't play nice with the `form-data` package. We'd need to add a `Readable.toWeb(form)` to make it work.

For now, I've reached my timebox. Happy to circle back to this later though.

byCedric: Here is an example request made, when using the workaround from SO:

image

While, originally, this is the request made:

image

ide: Node 18's end of life will be in April 2025. So it's not soon, but eventually we'll be able to rely on `fs.openAsBlob`. Could you document these findings in a task assigned to yourself so that it's easier to follow up on this after we've dropped Node 18?

byCedric: One workaround for Node 18 that [I've found to work well is this one](https://github.com/byCedric/test-expo-eas-cli-2414/blob/main/src/test.js#L73-L97):

```js
async function createBlobFromPath(filePath) {
  if (typeof fs.openAsBlob === 'function') {
    console.log('Using fs.openAsBlob:', filePath);
    return await fs.openAsBlob(filePath);
  }

  console.log('Falling back to File polyfill:', filePath);
  const { size } = await fs.promises.stat(filePath);

  return {
    [Symbol.toStringTag]: 'File', // Trick Node into thinking this is a File reference
    get type() { return undefined }, // Always undefined, it's the mime type
    get size() { return size }, // Ensure the `content-length` header is set through form data
    stream() { return Readable.toWeb(fs.createReadStream(filePath)) }, // The magic part that makes it work

    // Not necessary as we override the filename through formdata
    // get name() { return name },

    // Could add the full API, but it's not necessary for form data
    // arrayBuffer(...args) { return this.stream().arrayBuffer(...args) },
    // text(...args) { return this.stream().text(...args) },
    // slice(...args) { return this.stream().slice(...args) }
  };
}
```

This [seems to run fine on Node 18 for ubuntu, macos, and windows](https://github.com/byCedric/test-expo-eas-cli-2414/actions/runs/9464200632). It's still leveraging Node 20+ APIs, but with a weird fallback. It's not spec-compliant, but enough compliant for `FormData` itself.

PR #2422, opened 2024-06-11
[eas-cli] [ENG-12650] add fingerprint sources to build metadata

kadikraman:
### Related PRs
- https://github.com/expo/universe/pull/15517
- 👉 https://github.com/expo/eas-cli/pull/2422
- https://github.com/expo/eas-build/pull/407
- https://github.com/expo/eas-build/pull/421
- https://github.com/expo/turtle-v2/pull/1893
- https://github.com/expo/universe/pull/15592
-  https://github.com/expo/terraform-iam/pull/303

# Why

It's very difficult to debug why the local and remote fingerprint are different.

Currently the only way to do it would be manual: add step to print out the remote sources, calculate local fingerprint, diff the two. Ideally we'd like to show the diff in the build step instead.

# How

Save the local fingerprint in a google cloud bucket and show a diff if the fingerprints are different.

# Test Plan

Ran turtle and www locally with changes from the these packages manually linked to turtle.

Screenshot 2024-06-20 at 15 05 04

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -670,6 +670,10 @@ async function createAndUploadFingerprintAsync<T extends Platform>(
     cwd: ctx.projectDir,
   });
 
+  /**
+   * It's ok for fingerprintSources to be empty
+   * fingerprintSources only exist if the project is using runtimeVersion.policy: fingerprint
+   */
   if (!resolvedRuntimeVersion?.fingerprintSources) {
     return {
       runtimeVersion: resolvedRuntimeVersion?.runtimeVersion ?? undefined,
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -681,6 +681,7 @@ async function createAndMaybeUploadFingerprintAsync<T extends Platform>(
     };
   }
 
+  await fs.mkdirp(getTmpDirectory());
   const fingerprintLocation = path.join(getTmpDirectory(), `${uuidv4()}-runtime-fingerprint.json`);
 
   await fs.writeJSON(fingerprintLocation, {
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -2,6 +2,7 @@ import {
   ArchiveSource,
   ArchiveSourceType,
   BuildJob,
+  FingerprintSourceType,
   Metadata,
   Platform,
 } from '@expo/eas-build-job';
@@ -166,7 +167,7 @@ export async function prepareBuildRequestForPlatformAsync<
   }
   assert(projectArchive);
 
-  const runtimeMetadata = await createAndUploadFingerprintAsync(ctx);
+  const runtimeMetadata = await createAndMaybeUploadFingerprintAsync(ctx);
   const metadata = await collectMetadataAsync(ctx, runtimeMetadata);
   const buildParams = resolveBuildParamsInput(ctx, metadata);
   const job = await builder.prepareJobAsync(ctx, {
@@ -655,11 +656,11 @@ function formatAccountSubscriptionsUrl(accountName: string): string {
   ).toString();
 }
 
-async function createAndUploadFingerprintAsync<T extends Platform>(
+async function createAndMaybeUploadFingerprintAsync<T extends Platform>(
   ctx: BuildContext<T>
 ): Promise<{
   runtimeVersion?: string;
-  fingerprintGCSBucketKey?: string;
+  fingerprintSource?: Metadata['fingerprintSource'];
 }> {
   const resolvedRuntimeVersion = await resolveRuntimeVersionAsync({
     exp: ctx.exp,
@@ -687,6 +688,16 @@ async function createAndUploadFingerprintAsync<T extends Platform>(
     sources: resolvedRuntimeVersion.fingerprintSources,
   });
 
+  if (ctx.localBuildOptions.localBuildMode === LocalBuildMode.LOCAL_BUILD_PLUGIN) {
+    return {
+      runtimeVersion: resolvedRuntimeVersion?.runtimeVersion ?? undefined,
+      fingerprintSource: {
+        type: FingerprintSourceType.PATH,
+        path: fingerprintLocation,
+      },
+    };
+  }
+
   let fingerprintGCSBucketKey = undefined;
   try {
     fingerprintGCSBucketKey = await uploadFileAtPathToGCSAsync(
@@ -711,6 +722,9 @@ async function createAndUploadFingerprintAsync<T extends Platform>(
 
   return {
     runtimeVersion: resolvedRuntimeVersion?.runtimeVersion ?? undefined,
-    fingerprintGCSBucketKey,
+    fingerprintSource: {
+      type: FingerprintSourceType.GCS,
+      bucketKey: fingerprintGCSBucketKey,
+    },
   };
 }
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -12,6 +12,8 @@ import cliProgress from 'cli-progress';
 import fs from 'fs-extra';
 import { GraphQLError } from 'graphql/error';
 import nullthrows from 'nullthrows';
+import path from 'path';
+import { v4 as uuidv4 } from 'uuid';
 
 import { BuildContext } from './context';
 import {
@@ -60,9 +62,11 @@ import {
   appPlatformEmojis,
   requestedPlatformDisplayNames,
 } from '../platform';
+import { resolveRuntimeVersionAsync } from '../project/resolveRuntimeVersionAsync';
 import { uploadFileAtPathToGCSAsync } from '../uploads';
 import { formatBytes } from '../utils/files';
 import { printJsonOnlyOutput } from '../utils/json';
+import { getTmpDirectory } from '../utils/paths';
 import { createProgressTracker } from '../utils/progress';
 import { sleepAsync } from '../utils/promise';
 
@@ -162,7 +166,8 @@ export async function prepareBuildRequestForPlatformAsync<
   }
   assert(projectArchive);
 
-  const metadata = await collectMetadataAsync(ctx);
+  const runtimeMetadata = await createAndUploadFingerprintAsync(ctx);
+  const metadata = await collectMetadataAsync(ctx, runtimeMetadata);
   const buildParams = resolveBuildParamsInput(ctx, metadata);
   const job = await builder.prepareJobAsync(ctx, {
     projectArchive,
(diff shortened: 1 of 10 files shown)


wschurman (approved):
I think the build-tools PR will need to be deployed first (and maybe even the www one) but I'm not super familiar with the push safety between these repos as far as the build job validation goes.

kadikraman:
@wschurman you're right, this is what the package dependencies look like:

Screenshot 2024-06-12 at 09 35 22

It seems like actually I might have to break the `@expo/eas-build-job` into a separate PR so it could be merged separately since the dependency is used in `@expo/build-tools`.

So the order would be:
1. `@expo/eas-build-job` -> merge and publish
2. `@expo/build-tools` -> add dependency (`@expo/eas-build-job`), merge and release
3. `turtle worker` -> add dependencies (`@expo/eas-build-job` and `@expo/build-tools`), merge and deploy
4. `www` -> merge and deploy
5. `eas-cli` -> add dependency (`@expo/eas-build-job`), merge and release

Kudo (approved):

szdziedzic (approved):
>It seems like actually I might have to break the @expo/eas-build-job into a separate PR so it could be merged separately since the dependency is used in @expo/build-tools.
So the order would be:
@expo/eas-build-job -> merge and publish
@expo/build-tools -> add dependency (@expo/eas-build-job), merge and release
turtle worker -> add dependencies (@expo/eas-build-job and @expo/build-tools), merge and deploy
www -> merge and deploy
eas-cli -> add dependency (@expo/eas-build-job), merge and release

Re `1` and `2`: The `eas-build` repo is set up in a way that packages from `eas-build` use local versions of other packages from the same repo by default, so you just need to rebuild the `@expo/eas-build-job` package to be able to use it in the `@expo/build-tools`. You don't even need to split the `@expo/eas-build-job` and `@expo/build-tools` changes into 2 PRs. You can just merge it all together and once you run `yarn release` our release script will publish correct versions in correct order and bump everything in `package.json`.

szdziedzic (changes requested):
Oh, looking at https://github.com/expo/universe/pull/15517#pullrequestreview-2113659688 I believe that the fingerprint file upload to GCS should happen in CLI (same as we do for project archive and build metadata) and we should just pass a data necessary to link uploaded file with build to mutation

packages/eas-cli/src/build/build.ts
```diff
@@ -649,3 +654,59 @@ function formatAccountSubscriptionsUrl(accountName: string): string {
...
+    let errMessage = 'Failed to upload fingerprint to EAS';
+
+    if (err.message) {
+      errMessage += `\n\nReason: ${err.message}`;
+    }
+
+    Log.warn(errMessage);
+    return {
+      runtimeVersion: resolvedRuntimeVersion?.runtimeVersion ?? undefined,
+    };
```
wschurman, line 703: I think it makes sense to fail the build command if this isn't able to be uploaded?

packages/eas-cli/src/build/build.ts
```diff
@@ -649,3 +654,59 @@ function formatAccountSubscriptionsUrl(accountName: string): string {
...
+  const fingerprintLocation = path.join(getTmpDirectory(), `${uuidv4()}-runtime-fingerprint.json`);
+
+  await fs.writeJSON(fingerprintLocation, {
+    hash: resolvedRuntimeVersion.runtimeVersion,
+    sources: resolvedRuntimeVersion.fingerprintSources,
+  });
+
+  let fingerprintGCSLocation = undefined;
+  try {
+    fingerprintGCSLocation = await uploadFileAtPathToGCSAsync(
```
wschurman, line 688: would be good to name this bucket key both here and maybe even in build metadata

```suggestion
    fingerprintGCSBucketKey = await uploadFileAtPathToGCSAsync(
```

wschurman (changes requested):
One minor change to naming and a question, otherwise looks good.

packages/eas-cli/src/build/build.ts
kadikraman: I think we shouldn't since it's not necessary for the build to work. We [don't fail the build](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/build/build.ts#L336) if uploading metadata fails either.

packages/eas-cli/src/build/build.ts
kadikraman: Updated in [www](https://github.com/expo/universe/pull/15517/commits/ee28a4cbc7f878d6228ccc29bc452a39e27a8600), [eas-build-job](https://github.com/expo/eas-build/pull/407/commits/96eb10e8fec7eff8dbc1fa65e4f4c6b91689ac41) and [here](https://github.com/expo/eas-cli/pull/2422/commits/c413b410be5ac16c3fa58f957651feb6ca3d67d1) ✅

wschurman (approved):
Code looks good. Double check CI failures before landing. And also ensure the other PRs land ahead of this.

szdziedzic (changes requested):
If comments from https://github.com/expo/eas-build/pull/407#pullrequestreview-2135113278 end up being applicable we will probably also need to change metadata structure here

packages/eas-cli/src/build/build.ts
```diff
@@ -649,3 +654,59 @@ function formatAccountSubscriptionsUrl(accountName: string): string {
...
+    exp: ctx.exp,
+    platform: ctx.platform,
+    workflow: ctx.workflow,
+    projectDir: ctx.projectDir,
+    env: ctx.buildProfile.env,
+    cwd: ctx.projectDir,
+  });
+
+  if (!resolvedRuntimeVersion?.fingerprintSources) {
+    return {
```
Kudo, line 674: nit: add comment to explain the case here is when the `runtimeVersion.policy` is not fingerprint.

Kudo (approved):

packages/eas-cli/src/build/graphql.ts
```diff
@@ -42,6 +42,7 @@ export function transformProjectArchive(archiveSource: ArchiveSource): ProjectAr
 export function transformMetadata(metadata: Metadata): BuildMetadataInput {
   return {
     ...metadata,
+    fingerprintSource: metadata.fingerprintSource as BuildMetadataInput['fingerprintSource'],
```
szdziedzic, line 45: Can we do `assert(metadata.fingerprintSource.type === GCS)` instead of type casting here?

packages/eas-cli/src/build/metadata.ts
```diff
@@ -27,15 +30,13 @@ export async function collectMetadataAsync<T extends Platform>(
...
-        env: ctx.buildProfile.env,
-        cwd: ctx.projectDir,
-      })) ?? undefined,
+    runtimeVersion: runtimeMetadata?.runtimeVersion,
+    fingerprintSource: runtimeMetadata?.fingerprintGCSBucketKey
+      ? {
+          type: FingerprintSourceType.GCS,
+          bucketKey: runtimeMetadata?.fingerprintGCSBucketKey,
+        }
+      : undefined,
```
szdziedzic, line 39: I believe we should also try to add support for `fingerprintSource.type === PATH` in this PR. 

If you want to check if a build is run using local mode you can check if `buildCtx.localBuildOptions.localBuildMode` is defined.

szdziedzic (changes requested):

packages/eas-cli/src/build/build.ts
```diff
@@ -649,3 +654,63 @@ function formatAccountSubscriptionsUrl(accountName: string): string {
...
+   * It's ok for fingerprintSources to be empty
+   * fingerprintSources only exist if the project is using runtimeVersion.policy: fingerprint
+   */
+  if (!resolvedRuntimeVersion?.fingerprintSources) {
+    return {
+      runtimeVersion: resolvedRuntimeVersion?.runtimeVersion ?? undefined,
+    };
+  }
+
+  const fingerprintLocation = path.join(getTmpDirectory(), `${uuidv4()}-runtime-fingerprint.json`);
```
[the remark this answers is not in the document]
kadikraman: Hm. How should I do that? It doesn't seem like the [other places `getTmpDirectory` is used](https://github.com/search?q=repo:expo/eas-cli%20getTmpDirectory&type=code) do an additional check. Should this be added everywhere?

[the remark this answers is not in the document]
kadikraman: Can you explain what you mean by checking that it exists? That the path is non-empty or that the folder exists? I copied this implementation from [metadata](https://github.com/expo/eas-cli/blob/1c61465c5f7526770d3fc489828bef18803dd365/packages/eas-cli/src/build/utils/repository.ts#L103C9-L103C25) btw

packages/eas-cli/src/build/graphql.ts
kadikraman: Doesn't work
Screenshot 2024-06-28 at 10 46 22

If I use the enum from gql it also doesn't work
Screenshot 2024-06-28 at 10 47 54

This is the gql one:
```ts
export enum FingerprintSourceType {
  Gcs = 'GCS'
}
```

This is the `@expo/eas-build-job` one:
```ts
export declare enum FingerprintSourceType {
    'GCS' = "GCS",
    'PATH' = "PATH",
    'URL' = "URL"
}
```

I couldn't figure out how to make it work anyway.

packages/eas-cli/src/build/build.ts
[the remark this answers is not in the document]
szdziedzic: >That the path is non-empty or that the folder exists? 

I want us to check if the folder already exists.

>I copied this implementation from [metadata](https://github.com/expo/eas-cli/blob/1c61465c5f7526770d3fc489828bef18803dd365/packages/eas-cli/src/build/utils/repository.ts#L103C9-L103C25) btw

Oh okay, I believe it works for metadata, and here without this check because it is already created when uploading project archive tarball (`fs.mkdirp`)
https://github.com/expo/eas-cli/blob/0516b507bce06c136e38700e0d333675e719a8ba/packages/eas-cli/src/build/utils/repository.ts#L122-L126

I believe we can add the `fs.mkdirp` call here for extra safety so that refactor in the project tarball upload code won't break anything here. We can also fix it in metadata upload.

Do you think it makes sense? What do you think?

kadikraman:
I think this should do it @szdziedzic ! Tested both iOS and Android on both `--local` and EAS 👀 

## On EAS
| Android      | iOS |
| ----------- | ----------- |
| Screenshot 2024-06-28 at 15 40 39      | Screenshot 2024-06-28 at 15 40 22       |

## With `--local`
| Android      | iOS |
| ----------- | ----------- |
| Screenshot 2024-06-28 at 15 46 47      | Screenshot 2024-06-28 at 15 48 26       |

packages/eas-cli/src/build/build.ts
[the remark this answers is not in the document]
kadikraman: Makes sense to me! Added here for both metadata and fingerprint: https://github.com/expo/eas-cli/pull/2422/commits/005c0388c5b039b9dbbca4f96741c207c99dda15

szdziedzic (approved):
awesome! good job! 🙌 🚀

wschurman:
@giorgiofellipe - please file a new issue with steps to reproduce the issue you're seeing. Fingerprint is still in beta (appreciate you testing it out!), so we're still working out the kinks, and a step-by-step set of repro steps is the easiest way to communicate the issue.

tfcornerstone:
@kadikraman Doesn't work with `--non-interactive`?

wschurman:
@tfcornerstone - please file a new issue. See https://github.com/expo/eas-cli/pull/2422#issuecomment-2207383416

PR #2435, opened 2024-06-17
[eas-cli] make sure we filter out any comments in `build.gradle` file before passing to to `gradle-to-js` parser

szdziedzic:
# Why

I discovered that the `gradle-to-js` parser malfunctions when there are comments in the `build.gradle` file.

https://exponent-internal.slack.com/archives/C017N0N99RA/p1718618741696809?thread_ts=1718388095.340689&cid=C017N0N99RA

# How

Make sure we filter any comments in the `build.gradle` file before passing it to the `gradle-to-js` parser

# Test Plan

test manually


packages/eas-cli/src/project/android/gradleUtils.ts
```diff
@@ -30,7 +30,15 @@ export const DEFAULT_MODULE_NAME = 'app';
 export async function getAppBuildGradleAsync(projectDir: string): Promise<AppBuildGradle> {
   const buildGradlePath = AndroidConfig.Paths.getAppBuildGradleFilePath(projectDir);
   const rawBuildGradle = await fs.readFile(buildGradlePath, 'utf8');
-  return await g2js.parseText(rawBuildGradle);
+
+  // filter out any comments
+  // when comments are present, gradle-to-js fails to parse the file
+  const rawBuildGradleWithoutComments = rawBuildGradle
+    .split('\n')
+    .filter(line => !line.startsWith('//'))
```
sjchmiela, line 38: - do we need to `line.trim()`?
- does it also fail on multi-line comments (`/* comment */`?)

sjchmiela (changes requested):

szdziedzic:
After additional investigation, I noticed that the parser actually works well with comments and comment blocks.

It only seems to fail on an edge case - empty single-line comment
```
//
```

szdziedzic:
So we should focus on filtering out this edge case

packages/eas-cli/src/project/android/gradleUtils.ts
szdziedzic: >do we need to line.trim()?

yes 👍

szdziedzic: >does it also fail on multi-line comments (/* comment */?)

no

szdziedzic: >shall we add some simple test for this?

we can!

PR #2437, opened 2024-06-17
Add eas `env:push` and `env:pull` commands

khamilowicz:
# Why

User should be able to synchronise local environment variables with upstream.

- https://linear.app/expo/issue/ENG-11920/add-eas-envpush-command
- https://linear.app/expo/issue/ENG-11919/add-eas-envpull-command

# How

- added `eas env:pull` and `eas env:push` commands
- `eas env:pull` handles sudo access to request values of sensitive variables

# Test Plan

Tested manually

### DIFF
--- a/packages/eas-cli/src/authUtils.ts
+++ b/packages/eas-cli/src/authUtils.ts
@@ -1,0 +1,21 @@
+import SessionManager from './user/SessionManager';
+
+// Handle the case where a user needs to be in sudo mode to perform an action.
+export async function handleSudoCallAsync<T>(
+  sessionManager: SessionManager,
+  fn: () => Promise<T>
+): Promise<T> {
+  try {
+    return await fn();
+  } catch (error: any) {
+    const isSudoError = error.graphQLErrors?.some(
+      (err: { extensions: { errorCode: string } }) =>
+        err?.extensions?.errorCode === 'SUDO_MODE_REQUIRED'
+    );
+    if (!isSudoError) {
+      throw error;
+    }
+    await sessionManager.showSudoPromptAsync({ sso: false });
+    return await fn();
+  }
+}
--- a/packages/eas-cli/src/authUtils.ts
+++ b/packages/eas-cli/src/authUtils.ts
@@ -1,12 +1,12 @@
 import SessionManager from './user/SessionManager';
 
 // Handle the case where a user needs to be in sudo mode to perform an action.
-export async function handleSudoCallAsync<T>(
+export async function withSudoModeAsync<T>(
   sessionManager: SessionManager,
-  fn: () => Promise<T>
+  graphqlQuery: () => Promise<T>
 ): Promise<T> {
   try {
-    return await fn();
+    return await graphqlQuery();
   } catch (error: any) {
     const isSudoError = error.graphQLErrors?.some(
       (err: { extensions: { errorCode: string } }) =>
@@ -16,6 +16,6 @@ export async function handleSudoCallAsync<T>(
       throw error;
     }
     await sessionManager.showSudoPromptAsync({ sso: false });
-    return await fn();
+    return await graphqlQuery();
   }
 }
--- a/packages/eas-cli/src/commands/env/list.ts
+++ b/packages/eas-cli/src/commands/env/list.ts
@@ -45,9 +45,13 @@ export default class EnvironmentValueList extends EasCommand {
 
     const variables =
       scope === EnvironmentVariableScope.Project && environment
-        ? (await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, projectId, environment))
-            .appVariables
-        : await EnvironmentVariablesQuery.sharedAsync(graphqlClient, projectId);
+        ? (
+            await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, {
+              appId: projectId,
+              environment,
+            })
+          ).appVariables
+        : await EnvironmentVariablesQuery.sharedAsync(graphqlClient, { appId: projectId });
 
     if (format === 'short') {
       for (const variable of variables) {
--- a/packages/eas-cli/src/commands/env/pull.ts
+++ b/packages/eas-cli/src/commands/env/pull.ts
@@ -1,9 +1,9 @@
 import { Flags } from '@oclif/core';
 import fs from 'fs-extra';
 
-import { handleSudoCallAsync } from '../../authUtils';
+import { withSudoModeAsync } from '../../authUtils';
 import EasCommand from '../../commandUtils/EasCommand';
-import { EASEnvironmentFlag } from '../../commandUtils/flags';
+import { EASEnvironmentFlag, EASNonInteractiveFlag } from '../../commandUtils/flags';
 import { EnvironmentVariableFragment } from '../../graphql/generated';
 import { EnvironmentVariablesQuery } from '../../graphql/queries/EnvironmentVariablesQuery';
 import Log from '../../log';
@@ -11,6 +11,8 @@ import { confirmAsync } from '../../prompts';
 
 export default class EnvironmentValuePull extends EasCommand {
   static override description = 'pull env file';
+
+  static override hidden = true;
 
   static override contextDefinition = {
     ...this.ContextOptions.ProjectConfig,
@@ -20,36 +22,37 @@ export default class EnvironmentValuePull extends EasCommand {
 
   static override flags = {
     ...EASEnvironmentFlag,
+    ...EASNonInteractiveFlag,
     path: Flags.string({
-      description: 'Path to save the env file',
+      description: 'Path to the result `.env` file',
       default: '.env.local',
     }),
   };
 
   async runAsync(): Promise<void> {
     const {
-      flags: { environment, path: targetPath },
+      flags: { environment, path: targetPath, 'non-interactive': nonInteractive },
     } = await this.parse(EnvironmentValuePull);
     const {
       privateProjectConfig: { projectId },
       loggedIn: { graphqlClient },
       sessionManager,
     } = await this.getContextAsync(EnvironmentValuePull, {
-      nonInteractive: true,
+      nonInteractive,
     });
 
     if (!environment) {
       throw new Error('Please provide an environment to pull the env file from.');
     }
 
-    const { appVariables: environmentVariables } = await handleSudoCallAsync(sessionManager, () =>
+    const { appVariables: environmentVariables } = await withSudoModeAsync(sessionManager, () =>
       EnvironmentVariablesQuery.byAppIdWithSensitiveAsync(graphqlClient, {
         appId: projectId,
         environment,
       })
     );
 
-    if (await fs.exists(targetPath)) {
+    if (!nonInteractive && (await fs.exists(targetPath))) {
       const result = await confirmAsync({
         message: `File ${targetPath} already exists. Do you want to overwrite it?`,
       });
@@ -62,6 +65,7 @@ export default class EnvironmentValuePull extends EasCommand {
     const filePrefix = `# Environment: ${environment}\n\n`;
 
     const envFileContent = environmentVariables
+      .filter((variable: EnvironmentVariableFragment) => !!variable.value)
       .map((variable: EnvironmentVariableFragment) => {
         return `${variable.name}=${variable.value}`;
       })
@@ -69,6 +73,6 @@ export default class EnvironmentValuePull extends EasCommand {
 
     await fs.writeFile(targetPath, filePrefix + envFileContent);
 
-    Log.log(`Pull env file from ${environment} environment to ${targetPath}.`);
+    Log.log(`Pulled environment variables from ${environment} environment to ${targetPath}.`);
   }
 }
--- a/packages/eas-cli/src/commands/env/pull.ts
+++ b/packages/eas-cli/src/commands/env/pull.ts
@@ -1,5 +1,4 @@
 import { Flags } from '@oclif/core';
-import assert from 'assert';
 import fs from 'fs-extra';
 
 import { withSudoModeAsync } from '../../authUtils';
@@ -47,8 +46,6 @@ export default class EnvironmentValuePull extends EasCommand {
       'non-interactive': nonInteractive,
     } = this.validateFlags(flags);
 
-    assert(environment);
-
     const {
       privateProjectConfig: { projectId },
       loggedIn: { graphqlClient },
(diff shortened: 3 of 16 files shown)


packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,176 @@
...
+      graphqlClient,
+      { appId: projectId, environment, filterNames: variableNames }
+    );
+
+    // Check if any of the variables already exist in the environment. Prompt the user to overwrite them.
+    const existingSensitiveVariables = existingVariables.filter(
+      variable => variable.value === null
+    );
+
+    if (existingSensitiveVariables.length > 0) {
```
szdziedzic, line 65:
```suggestion
    if (existingDifferentSensitiveVariables.length > 0) {
```
Shouldn't we only print it for different variables?

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,176 @@
...
+      variable => variable.scope === EnvironmentVariableScope.Shared
+    );
+
+    if (existingDifferentSharedVariables.length > 0) {
+      const existingDifferentSharedVariablesNames = existingDifferentSharedVariables.map(
+        variable => variable.name
+      );
+      Log.error(
+        'Shared variables cannot be overwritten.\nPlease remove them from the env file or unlink them from the project:'
+      );
```
szdziedzic, line 100:
```suggestion
      Log.error(
        'Shared variables cannot be overwritten.'
      );
      Log.error(
        'Remove them from the env file or unlink them from the project to continue:'
      );
```

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,176 @@
...
+      existingDifferentSharedVariablesNames.forEach(name => Log.error(`- ${name}`));
+      return;
```
szdziedzic, line 102:
```suggestion
      throw new Error('Shared variables cannot be overwritten by eas env:push command');
```

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,176 @@
...
+      variable => variable.value === null
+    );
+
+    if (existingSensitiveVariables.length > 0) {
+      const existingSensitiveVariablesNames = existingSensitiveVariables.map(
+        variable => variable.name
+      );
+      Log.log(
+        'Sensitive variables cannot be overwritten with `push` command.\nUse `eas env:update` to change them:'
+      );
```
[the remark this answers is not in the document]
szdziedzic: Why do we restrict sensitive variable updates?

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,176 @@
...
+          variables[name] = {
+            name,
+            value,
+            environment,
+            visibility,
+          };
+        }
+      });
+    return variables;
+  }
```
szdziedzic, line 175: I would suggest using `dotenv.parse` for this

https://www.npmjs.com/package/dotenv#%EF%B8%8F-usage

szdziedzic (changes requested):

packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -205,6 +208,45 @@ export default class SessionManager {
     }
   }
 
+  public async showSudoPromptAsync({ sso = false } = {}): Promise<void> {
```
szdziedzic, line 211:
```suggestion
  public async showSudoPromptAsync({ sso = false } = { sso?: boolean; }): Promise<void> {
```

szdziedzic (changes requested):
I'm not an expert on our auth/sudo system. Let's ask for the @wschurman review as well.

packages/eas-cli/src/commands/env/push.ts
khamilowicz: Both `Log.error` and `throw`, correct?

packages/eas-cli/src/commands/env/push.ts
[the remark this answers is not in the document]
khamilowicz: IMO it may lead to user unintentionally overwritting important variables. But we can probably allow it (with user's confirmation).

packages/eas-cli/src/user/fetchSessionSecretAndUser.ts
```diff
@@ -26,3 +26,18 @@ export async function fetchSessionSecretAndUserAsync({
...
+    password,
+    otp,
+  }: {
+    password: string;
+    otp?: string;
+  }
+): Promise<void> {
+  await apiV2Client.postAsync('auth/upgradeSudo', {
+    body: { password, otp },
+  });
```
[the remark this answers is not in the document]
khamilowicz: Right

packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -229,6 +271,19 @@ export default class SessionManager {
...
+      accessToken: this.getAccessToken(),
+      sessionSecret: this.getSessionSecret(),
+    };
+    const apiV2Client = new ApiV2Client(authenticationInfo);
+    await upgradeSudoSessionAndAsync(apiV2Client, input);
+  }
+
+  private async ssoSudoUpgradeAsync(): Promise<void> {
+    throw Error('Sudo upgrade with SSO is not yet supported');
+  }
```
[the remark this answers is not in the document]
wschurman: Yep. This will need to do something similar to how SSO login is done in the CLI (using the website I believe). Can trace the SSO code in the CLI `ssoLoginAsync` to see how SSO auth works. Happy to dig deeper if necessary.

packages/eas-cli/src/user/fetchSessionSecretAndUser.ts
```diff
@@ -26,3 +26,18 @@ export async function fetchSessionSecretAndUserAsync({
     username: userData.username,
   };
 }
+
+export async function upgradeSudoSessionAndAsync(
```
wschurman, line 30:
```suggestion
export async function upgradeToSudoSessionAsync(
```

packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -229,6 +271,19 @@ export default class SessionManager {
     });
   }
 
+  private async sudoLoginAsync(input: { password: string; otp?: string }): Promise<void> {
```
wschurman, line 274:
```suggestion
  private async upgradeToSudoSessionAsync(input: { password: string; otp?: string }): Promise<void> {
```

packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -205,6 +208,45 @@ export default class SessionManager {
     }
   }
 
+  public async showSudoPromptAsync({ sso = false }: { sso?: boolean }): Promise<void> {
+    if (sso) {
```
wschurman, line 212: this should go below check for `sessionData`?

packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -337,7 +392,8 @@ export default class SessionManager {
     metadata: {
       secondFactorDevices?: SecondFactorDevice[];
       smsAutomaticallySent?: boolean;
-    }
+    },
+    sudo: boolean = false
```
wschurman, line 396: prefer destructured object arg for boolean arguments so that name is retained

packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -0,0 +1,78 @@
...
+  async runAsync(): Promise<void> {
+    const {
+      flags: { environment, path: targetPath, 'non-interactive': nonInteractive },
+    } = await this.parse(EnvironmentValuePull);
+    const {
+      privateProjectConfig: { projectId },
+      loggedIn: { graphqlClient },
+      sessionManager,
+    } = await this.getContextAsync(EnvironmentValuePull, {
+      nonInteractive: true,
```
wschurman, line 41:
```suggestion
      nonInteractive,
```

wschurman (changes requested):
Added some info on how to do SSO upgrade to an inline comment.

One other change is to expand upon the test plan to detail how this was tested.

packages/eas-cli/src/user/SessionManager.ts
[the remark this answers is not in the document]
wschurman: Would be good to stack a PR on top of this that adds SSO and then land them together.

wschurman (approved):
Since eas-cli doesn't have feature gating and does continuous deployment it'd probably be best to have the SSO functionality ironed out in a stacked PR before landing this. Otherwise we'd risk deploying a broken experience. Though I guess since the command is new and hidden it's okay for now, we can just fast-follow with the SSO fix.

packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!environment) {
+      throw new Error('Please provide an environment to pull the env file from.');
+    }
+
+    const { appVariables: environmentVariables } = await withSudoModeAsync(sessionManager, () =>
+      EnvironmentVariablesQuery.byAppIdWithSensitiveAsync(graphqlClient, {
+        appId: projectId,
+        environment,
+      })
```
szdziedzic, line 52: Shouldn't this be 
```suggestion
      await EnvironmentVariablesQuery.byAppIdWithSensitiveAsync(graphqlClient, {
        appId: projectId,
        environment,
      })
```
?

Or maybe it doesn't matter here?

packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    if (!nonInteractive && (await fs.exists(targetPath))) {
+      const result = await confirmAsync({
+        message: `File ${targetPath} already exists. Do you want to overwrite it?`,
+      });
+      if (!result) {
+        Log.log('Aborting...');
+        return;
```
szdziedzic, line 61: I would prefer throwing an error here vs returning, to have non-zero exit code when command fails

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,171 @@
...
+      Log.warn('Some variables already exist in the environment.');
+
+      const { variablesToOverwrite } = await promptAsync({
+        type: 'multiselect',
+        name: 'variablesToOverwrite',
+        message: 'Select variables to overwrite:',
+        // @ts-expect-error property missing from `@types/prompts`
+        optionsPerPage: 20,
+        choices: existingDifferentVariables.map(variable => ({
+          title: `${variable.name}: ${updateVariables[variable.name].value}`,
```
szdziedzic, line 116:
```suggestion
          title: `${variable.name}: ${updateVariables[variable.name].value} (was ${variable.value})`,
```

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,171 @@
...
+          value: variable.name,
+        })),
+      });
```
szdziedzic, line 119: I think that if one wants to push 50 env vars, this can be annoying. Maybe we want to first ask users
```
The ENV_VAR1, ENV_VAR2, ENV_VAR3 environment variables already exist in ${environment} environment. Do you want to override them all?
```
and if not fallback to multi-select.

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    const variables: Record<string, string> = dotenv.parse(await fs.readFile(envPath, 'utf8'));
+
+    for (const [name, value] of Object.entries(variables)) {
+      pushInput[name] = {
+        name,
+        value,
+        environment,
+        visibility: name.startsWith('EXPO_SECRET')
+          ? EnvironmentVariableVisibility.Sensitive
```
szdziedzic, line 164: This is a super cool idea! 👏 

I'm wondering whether it wouldn't be better to make it
```suggestion
        visibility: name.startsWith('EXPO_SENSITIVE')
          ? EnvironmentVariableVisibility.Sensitive
```
?

szdziedzic (changes requested):

packages/eas-cli/src/commands/env/push.ts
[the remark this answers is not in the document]
szdziedzic: >we can probably allow it (with user's confirmation).
Yeah I think we can do it

khamilowicz:
@wschurman @szdziedzic Ok, I'll work on SSO. I think the flow will be similar to that of login via SSO and I'll reuse some of its logic

wschurman (approved):

packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    const { appVariables: environmentVariables } = await withSudoModeAsync(
+      sessionManager,
+      async () =>
+        await EnvironmentVariablesQuery.byAppIdWithSensitiveAsync(graphqlClient, {
+          appId: projectId,
+          environment,
+        })
+    );
```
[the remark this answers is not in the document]
khamilowicz: Discussed offline - we remove requirement of sudo mode until such problems are resolved.

packages/eas-cli/src/authUtils.ts
```diff
@@ -0,0 +1,21 @@
+import SessionManager from './user/SessionManager';
+
+// Handle the case where a user needs to be in sudo mode to perform an action.
+export async function withSudoModeAsync<T>(
```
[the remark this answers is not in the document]
khamilowicz: No, it wouldn't make sense to automatically reauthorize in non-interactive mode.

packages/eas-cli/src/commands/env/list.ts
```diff
@@ -45,9 +45,13 @@ export default class EnvironmentValueList extends EasCommand {
 
     const variables =
       scope === EnvironmentVariableScope.Project && environment
-        ? (await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, projectId, environment))
-            .appVariables
-        : await EnvironmentVariablesQuery.sharedAsync(graphqlClient, projectId);
+        ? (
+            await EnvironmentVariablesQuery.byAppIdAsync(graphqlClient, {
```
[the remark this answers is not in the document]
khamilowicz: maybe with '--include-sensitive' flag? We'll save on decrypting values when user just want to know whether a variable is defined.

packages/eas-cli/src/authUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+      (err: { extensions: { errorCode: string } }) =>
+        err?.extensions?.errorCode === 'SUDO_MODE_REQUIRED'
+    );
+    if (!isSudoError) {
+      throw error;
+    }
+    await sessionManager.showSudoPromptAsync({ sso: false });
+    return await graphqlQuery();
+  }
+}
```
szdziedzic, line 21: Offline discussion feedback:

It seems like requiring sudo mode is problematic for now, it doesn't work for SSO auth and doesn't seem to work with `--non-interactive` mode, which is a big use case. For now, we decided to allow people to read sensitive env vars without sudo mode. This will require backend changes + dropping sudo logic from this PR.

szdziedzic (changes requested):

PR #2447, opened 2024-06-28
feat(eas-cli): Implement initial `worker:deploy` command

kitten:
## Summary

This implements the initial `worker:deploy` command, allowing us to deploy workers (`expo export -p web` in server-mode) to our hosting service.

For now, we'd like to get this merged and keep the command marked as `hidden`, since hosting is currently feature flagged. I'm not sure if we have a canary-release process, but worst case, we'd be shipping a hidden command here.

This is an **early version** of this command and we don't take customization/configuration into account yet, i.e. we don't have any configuration yet. We also don't yet allow the command to pass in a `deploymentIdentifier` (which will eventually have to be added) and we don't load up and bundle the worker with any environment variables.

## Set of changes

- Allow `devDomainName` to be initialized for a project
- Add deployment API call to create a deployment URL
- Tarball-up and upload worker deployments in the expected format by the `deployment-api`
- Receive list of asset URLs and upload assets to it
- Add interactive output


kitten:
/changelog-entry new-feature Add EAS Worker command (experimental/internal)

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+      nonInteractive: true,
+    });
+
+    const distPath = path.resolve(projectDir, 'dist');
+    const distClientPath = path.resolve(distPath, 'client');
+    const distServerPath = path.resolve(distPath, 'server');
+    if (!(await isDirectory(distPath))) {
+      throw new Error(
+        `No "dist/" folder found at ${distPath}. Prepare your project for deployment with "npx expo export"`
+      );
```
szdziedzic, line 54: Can we do it automatically for the user if `dist` is not present?

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+      const manifest = { env: {} };
+      yield ['manifest.json', JSON.stringify(manifest)];
+
+      const workerFiles = WorkerAssets.listWorkerFiles(distServerPath);
+      for await (const workerFile of workerFiles) {
+        yield [`server/${workerFile.normalizedPath}`, workerFile.data];
+      }
+    }
+
+    async function uploadTarballAsync(tarPath: string): Promise<any> {
```
szdziedzic, line 80: I believe we expect `json.result` to have a certain structure later on. Can we validate that structure here using Joi/zod and return a typed value instead of `any`.

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+            progress.percent = ++progress.transferred / progress.total;
+          } else {
+            progress.pending++;
+          }
+          updateProgress({ progress });
+        }
+      } catch (error: any) {
+        updateProgress({ isComplete: true, error });
+        Log.error(error);
+        return;
```
szdziedzic, line 158:
```suggestion
      } catch (error: any) {
        updateProgress({ isComplete: true, error });
        Log.error(error);
        throw error;
```

Shouldn't this error be thrown?

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+    } catch (error: any) {
+      progress.fail('Failed to create worker deployment');
+      Log.error(error);
+      return;
+    }
+    progress.succeed('Created worker deployment');
+
+    await uploadAssetsAsync(assetMap, deployResult.uploads);
+
+    const baseDomain = process.env.EXPO_STAGING ? 'staging.expo.app' : 'expo.app';
```
szdziedzic, line 189: does it make sense to also cover `expo.test` for `EXPO_LOCAL=1`?

packages/eas-cli/src/worker/assets.ts
```diff
@@ -0,0 +1,170 @@
+/* eslint-disable async-protect/async-suffix */
```
szdziedzic, line 1: What is the reason for this? Do you find `Async` suffixes annoying?

szdziedzic (changes requested):
awesome 👏 

I have a few comments/questions:

packages/eas-cli/package.json
```diff
@@ -78,6 +79,7 @@
     "semver": "7.5.4",
     "slash": "3.0.0",
     "tar": "6.2.1",
+    "tar-stream": "3.1.7",
```
byCedric, line 82: Is it possible to use the streaming functionality of `tar` itself? Avoid having to add another library. `tar` itself is actively maintained by isaacs, `tar-stream` seems to be updated less frequently.

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+        filePath: tarPath,
+        compress: false,
+        headers: {
+          accept: 'application/json',
+        },
+      });
+      if (response.status === 413) {
+        throw new Error(
+          'Upload failed! (Payload too large)\n' +
+            `The gzipped files in "dist/server/" (at: ${distServerPath}) exceed the maximum file size.`
```
byCedric, line 96: What's the maximum file size and/or what can the user do to get "unblocked"? Does the server return a special header with the maximum size?

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+        );
+      } else if (!response.ok) {
+        throw new Error(`Upload failed! (${response.statusText})`);
+      } else {
+        const json = await response.json();
+        if (!json.success || !json.result || typeof json.result !== 'object') {
+          throw new Error(json.message ? `Upload failed: ${json.message}` : 'Upload failed!');
+        }
+        return json.result;
+      }
```
byCedric, line 106: Also here, what can the user do to unblock themselves from these errors?

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,195 @@
...
+      const updateProgress = createProgressTracker({
+        total: progress.total,
+        message(ratio) {
+          const percent = `${Math.floor(ratio * 100)}`;
+          const details = chalk.dim(
+            `(${progress.pending} Pending, ${progress.transferred} Completed, ${progress.total} Total)`
+          );
+          return `Uploading client assets: ${percent.padStart(3)}% ${details}`;
+        },
+        completedMessage: 'Uploaded client assets for worker deployment',
```
byCedric, line 142:
```suggestion
        completedMessage: 'Uploaded assets for serverless deployment',
```

Not sure if it's nice to "leak" this implementation detail to the user. Since we name it "serverless" on our dashboard, might be good to use the same naming.

There are a couple more instances, e.g. through `ora` messages. Might be nice to clean these up too?

packages/eas-cli/src/worker/assets.ts
```diff
@@ -0,0 +1,170 @@
...
+  const basename = path.basename(__filename, path.extname(__filename));
+  const tmpdir = await fs.promises.realpath(os.tmpdir());
+  const random = randomBytes(4).toString('hex');
+  return path.resolve(tmpdir, `tmp-${basename}-${process.pid}-${random}`);
+}
+
+/** Normalizes given sourcemap sources to relative paths for a given root path */
+function formatSourcemap(rootPath: string, data: Buffer | string): Buffer | string {
+  try {
+    const cwd = process.cwd();
```
byCedric, line 38: `process.cwd` might be a bit hard to test, it works fine but might be nice to have a clear "rootPath" vs something like "projectRoot" or "workspaceRoot" that defaults to `process.cwd`, but can be overridden?

packages/eas-cli/src/worker/deployment.ts
```diff
@@ -0,0 +1,89 @@
...
+      deploymentVariables
+    );
+  } catch (error: any) {
+    const isMissingDevDomain = (error as GraphqlError)?.graphQLErrors?.some(e =>
+      ['APP_NO_DEV_DOMAIN_NAME'].includes(e?.extensions?.errorCode as string)
+    );
+
+    if (!isMissingDevDomain) {
+      throw error;
+    }
```
byCedric, line 29: Love this, actively guiding users through the errors ❤️

byCedric (approved):
Left a few nits, but its looking good overall. I think @szdziedzic has some other comments.

packages/eas-cli/src/commands/worker/deploy.ts
kitten: That's a great shout!
If you don't mind, I created a ticket for it, since there's several extra steps we'd have to do (make sure the export mode is correctly set up and check for Expo Router)
https://linear.app/expo/issue/ENG-13381/eas-cli-on-missing-dist-folder-offer-export

packages/eas-cli/src/commands/worker/deploy.ts
kitten: I always got double errors here when throwing, so it's either `Log.error` and not getting the stack trace, or `throw error` and getting the stack trace but not both 🤔 I'm a little confused on this tbh, and am not sure which is better, but happy to switch to just throwing in this particular process

packages/eas-cli/src/commands/worker/deploy.ts
kitten: We can't deploy to `expo.test`, since it's Cloudflare-based. For that, we'd have to simulate worker namespaces, which isn't easily possible with Miniflare/Cloudflare's test tooling. The idea here is to either:
- just deploy to the Cloudflare staging environment anyway, even if your API is local
- or; to have a development environment in Cloudflare which we'll just use when running locally

packages/eas-cli/src/worker/assets.ts
kitten: Oh sorry, I'll get that cleaned up.

On a side-note, I do actually find it redundant and annoying 😄 It was especially annoying in this case because of `packFilesIterable` which isn't even returning a promise, but an async generator. Anyway, I'll get this cleaned up 👀 Just forgot to do so. nice catch!

packages/eas-cli/package.json
kitten: I can try it

packages/eas-cli/src/commands/worker/deploy.ts
kitten: There's no path right now to unblock users, but it's 10MB gzipped. We'll have to explore what we can do in these cases to let users avoid this. One solution is to try to leave out sourcemaps and issue a warning to people, but if 10MB is still exceeded afterwards, we'll have to try some other solutions.

For server components, we may have to split up the server workers if we get close to reaching these limits.

packages/eas-cli/src/commands/worker/deploy.ts
kitten: So, ideally, we'll have to test which errors can occur. The problem is that Cloudflare's Upload API has several failure modes, some of which return no details at all.
Since we haven't run into any of these undocumented, non-error-coded errors, it's hard to predict when they'd occur 🤦 

Some of them can occur if JS is invalid, I think, but we'll need to test this by hand and take notes of what errors messages we get, since there's no documented list of error codes

kitten: Made a ticket for this: https://linear.app/expo/issue/ENG-13382/test-invalid-worker-uploads-and-take-notes-of-non-error-codes-cf-api

packages/eas-cli/src/commands/worker/deploy.ts
kitten: I'll have a look 👍 and will go through them again

packages/eas-cli/src/worker/assets.ts
kitten: I think we should be using the `projectPath` here. Nice catch!

packages/eas-cli/src/commands/worker/deploy.ts
kitten: We should eventually, but currently there is no result we use from the response body, and we do a simple `!json.success || !json.result || typeof json.result !== 'object'` sanity check. The response shape is currently only `success: boolean`, so there's no much data to convey just yet.

I'd like to keep this simple until we have a "real" payload to respond with, which is on the list of todos ✌️

packages/eas-cli/src/worker/assets.ts
kitten: Just checked, this is dead code and not needed anymore in the latest version 🎉 Forgot about that

packages/eas-cli/package.json
kitten: jk, `tar-stream` is a bit better suited since it allows us to edit and stream in files, which is something `tar` doesn't support at all

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -0,0 +1,193 @@
...
+    Log.warn('EAS Worker Deployments are in beta and subject to breaking changes.');
+
+    const {
+      getDynamicPrivateProjectConfigAsync,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(WorkerDeploy, {
+      nonInteractive: true,
+    });
+
+    const { projectId, projectDir, exp } = await getDynamicPrivateProjectConfigAsync();
```
szdziedzic, line 47: We can use new env vars pulled from EAS servers for config resolution here later on. I can add it in the future.

szdziedzic (approved):

packages/eas-cli/src/commands/worker/deploy.ts
kitten: That'd be awesome 🎉 We have yet to figure out env vars and secrets. I reckon, secrets _should_ be sourced locally, and depending on whether this is run on an EAS Custom Build step (TBD) or locally, they'd be combined with remote env vars.

The problem is that we haven't figured out yet how to source local env vars, i.e. secrets. We might have to try to crawl the Babel tree for them, or come up with a different solution that doesn't require the user to specify them explicitly in the project config

kadikraman (approved):
🚀

packages/eas-cli/src/commands/worker/deploy.ts
kitten: If you have any thoughts on the flow, we've got a ticket for this now, so if you want to throw in any comments, thoughts, suggestions, an implementation path (whatever really) I hope this makes that easier 🙌 https://linear.app/expo/issue/ENG-13428/environment-variables

PR #2460, opened 2024-07-16
add username and password flags to `eas login`

pvinis:
# Why

Context:
https://discord.com/channels/695411232856997968/1253063539598295141

Basically, as a freelancer, I need to login to multiple expo accounts in a day, and doing the whole logout/login dance is taking time.

My ideal scenario is that when i enter a directory/repo, eas will be on the right account, but to get there, we need a few steps. The first step is to allow `eas login` to take username and password flags, so that I can have these in a `.env` and with a simple npm script like `eas logout; eas login -u $EXPO_USERNAME -p $EXPO_PASSWORD`, I can at least run that when I enter the directory. That alone will make a huge difference.

# How

I added the flags. If they exist, the prompt will not ask for whatever was given.

# Test Plan

I used the `easd` alias that you suggest in the contributing file, and then I tried to login and out. I also tried like I described above, with direnv and `.envrc`, and it worked.

### DIFF
--- a/packages/eas-cli/src/commands/account/login.ts
+++ b/packages/eas-cli/src/commands/account/login.ts
@@ -14,6 +14,14 @@ export default class AccountLogin extends EasCommand {
       char: 's',
       default: false,
     }),
+    username: Flags.string({
+      description: 'Username',
+      char: 'u',
+    }),
+    password: Flags.string({
+      description: 'Password',
+      char: 'p',
+    }),
   };
 
   static override contextDefinition = {
@@ -22,11 +30,11 @@ export default class AccountLogin extends EasCommand {
 
   async runAsync(): Promise<void> {
     const {
-      flags: { sso },
+      flags: { sso, username, password },
     } = await this.parse(AccountLogin);
 
     const { sessionManager } = await this.getContextAsync(AccountLogin, { nonInteractive: false });
-    await sessionManager.showLoginPromptAsync({ sso });
+    await sessionManager.showLoginPromptAsync({ sso, username, password });
     Log.log('Logged in');
   }
 }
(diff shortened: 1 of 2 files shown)


szdziedzic (commented):
Hi @pvinis!

Thanks for your contribution. 

We already support programmatic access by using access tokens. I believe this is what you are looking for 😄.

https://docs.expo.dev/accounts/programmatic-access

pvinis:
@szdziedzic I am trying now, and I dont see this working.

- On my local machine, I have generated an expo token, and added it on the .env, and even .envrc for direnv.
- I tried running `pnpx eas-cli@latest login`, `EXPO_TOKEN="thetoken" pnpx eas-cli@latest login`, `pnpx eas-cli@latest login --non-interactive`, and all seem to ask me for an email and password.

What am I missing? Maybe the docs can be improved?

szdziedzic:
>I tried running pnpx eas-cli@latest login, EXPO_TOKEN="thetoken" pnpx eas-cli@latest login, pnpx eas-cli@latest login --non-interactive, and all seem to ask me for an email and password.

@pvinis If you use the `EXPO_TOKEN` env var you don't need to be running the `eas login` command. You can just run any other `eas` command like `EXPO_TOKEN=XYZ eas whoami` or `EXPO_TOKEN=XYZ eas build` and be authenticated as a user for who the PAT was generated.

However, I see that the current `eas login` behavior is confusing when you use `EXPO_TOKEN` as it doesn't indicate that you are already authed because you use `EXPO_TOKEN` 🤔. We will make it better 👍.

szdziedzic:
Please, let me know if it works for you!

pvinis:
Ah, I see. I need to be logged out, and then just run a build command or something that needs auth, and then it will try to find the expo_token by itself. Kinda funky, and def needs better docs.

pvinis:
Yea, it works like that. If I log out, then have the expo_token in .envrc with direnv, I just call `eas build` and it works, same with `eas whoami` etc.

Thanks for the info!

szdziedzic:
>def needs better docs

Thanks for the feedback! We can improve docs as well 👍.

PR #2461, opened 2024-07-16
[ENG-12793][eas-cli] make `eas login` and `eas whoami` behave more clearly when one is authenticated using `EXPO_TOKEN`

szdziedzic:
# Why

Using `EXPO_TOKEN` seems confusing to people.

https://github.com/expo/eas-cli/pull/2460

# How

When using the `eas whoami` command signal that a user is authenticated by using `EXPO_TOKEN`. It seems like people sometimes tend to forget that they have `EXPO_TOKEN` set, then they log in as a different account but `eas whoami` still returns the same value because they are authed using `EXPO_TOKEN` and it seems to be confusing.

Warn people when running `eas login` with the `EXPO_TOKEN` set that this is not necessary. 

# Test Plan

Test manually

`eas whoami` after and before
Screenshot 2024-07-16 at 16 39 28

`eas login` 
Screenshot 2024-07-16 at 16 52 52
Screenshot 2024-07-16 at 16 53 06
Screenshot 2024-07-16 at 16 53 39

### DIFF
--- a/packages/eas-cli/src/commands/account/login.ts
+++ b/packages/eas-cli/src/commands/account/login.ts
@@ -35,31 +35,13 @@ export default class AccountLogin extends EasCommand {
     } = await this.getContextAsync(AccountLogin, { nonInteractive: false });
 
     if (sessionManager.getAccessToken()) {
-      Log.warn(
-        'We detected that you have set EXPO_TOKEN in your environment. If EXPO_TOKEN is set, it takes precedence over the current session as an authentication method.'
+      throw new Error(
+        'EXPO_TOKEN is set in your environment, and is being used for all EAS authentication. To use username/password authentication, unset EXPO_TOKEN in your environment and re-run the command.'
       );
-      Log.warn(
-        'Even if you log in with a different account, EXPO_TOKEN will still be used if set.'
-      );
-      Log.warn(
-        `You don't need to use ${chalk.bold(
-          'eas login'
-        )} command if you are planning to use EXPO_TOKEN.`
-      );
-
-      const shouldContinue = await confirmAsync({
-        message: 'Do you want to continue?',
-      });
-      if (!shouldContinue) {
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/account/login.ts
```diff
@@ -25,7 +29,46 @@ export default class AccountLogin extends EasCommand {
...
     } = await this.parse(AccountLogin);
 
-    const { sessionManager } = await this.getContextAsync(AccountLogin, { nonInteractive: false });
+    const {
+      sessionManager,
+      maybeLoggedIn: { actor },
+    } = await this.getContextAsync(AccountLogin, { nonInteractive: false });
+
+    if (sessionManager.getAccessToken()) {
+      Log.warn(
```
wschurman, line 38: IMO we should just throw an error when `eas login` is run when the `EXPO_TOKEN` is set.

wschurman (changes requested):

packages/eas-cli/src/commands/account/view.ts
```diff
@@ -12,14 +12,19 @@ export default class AccountView extends EasCommand {
...
 
   async runAsync(): Promise<void> {
     const {
       maybeLoggedIn: { actor },
+      sessionManager,
     } = await this.getContextAsync(AccountView, { nonInteractive: true });
     if (actor) {
-      Log.log(chalk.green(getActorDisplayName(actor)));
+      const loggedInAs = sessionManager.getAccessToken()
+        ? `${getActorDisplayName(actor)} (by using EXPO_TOKEN)`
```
wschurman, line 25:
```suggestion
        ? `${getActorDisplayName(actor)} (authenticated using EXPO_TOKEN)`
```

wschurman (approved):
One wording change, otherwise lgtm.

PR #2466, opened 2024-07-19
[eas-cli] bump asset upload timeout from 90 to 180 seconds

quinlanj:
# Why

Extend the timeout from 90 seconds to 180 seconds, because some people have really slow internet connections. Using a p99 of the cloud function time of 29s, we essentially change the upload timeout from 60s to 150s

# background
Why we chose 90 seconds in the first place, from https://github.com/expo/eas-cli/pull/2085:

> For slow connections that take a while to upload, 60s is not enough of a timeout. For example, lets say there's only one asset to upload, and it takes even 1 second to upload and 60s to process (the max of the cloud function), it'll time out. The p99 of the processing cloud function is 29s, so changing this timeout to 90s gives a upload timeout of 60s.

> We also change the semantics of when to reset the timeout, and now reset it when the upload is retried. So, in the example above, let's say the asset takes 59 seconds to upload, but fails at the 58th second, it will be retried and the timeout will reset to give it an additional 60 seconds to upload.

# Test Plan

- [ ] sanity check that `eas update` still works on a test project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Bump asset upload timeout from 90 to 180 seconds. ([#2466](https://github.com/expo/eas-cli/pull/2466) by [@quinlanj](https://github.com/quinlanj))
+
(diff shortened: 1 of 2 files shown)


wschurman (approved):

quinlanj:
/changelog-entry chore bump asset upload timeout to 180 seconds

PR #2478, opened 2024-08-01
Add environment to Metadata

khamilowicz:
# Why

User should be able to specify `environment` in their build profile to select Environment Variables set to be used with their build

# How

* added `metadata` field to build `metadata` (`eas-build`)
* added `metadata` to build profile schema
* passed `metadata` from build profile to build metadata

# Test Plan

Added test

### DIFF
--- a/packages/eas-json/src/build/types.ts
+++ b/packages/eas-json/src/build/types.ts
@@ -69,6 +69,8 @@ export interface CommonBuildProfile {
 
   // credentials
   withoutCredentials?: boolean;
+
+  environment?: string;
 }
 
 interface PlatformBuildProfile extends Omit<CommonBuildProfile, 'autoIncrement'> {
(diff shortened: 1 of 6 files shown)


packages/eas-json/src/build/types.ts
```diff
@@ -69,6 +69,8 @@ export interface CommonBuildProfile {
 
   // credentials
   withoutCredentials?: boolean;
+
+  environment?: string;
```
szdziedzic, line 73:
```suggestion
  environment?: 'preview' | 'production' | 'development'
```

szdziedzic (changes requested):

szdziedzic (changes requested):
You probably need to create new release in https://github.com/expo/eas-build and bump the `@expo/eas-build-job` version before it lands so tests pass

szdziedzic (approved):
lgtm

We probably also need to add the `environment` filed description to `eas.schema.json`, but let's not do it now, because it's still "secret" and not user-facing. Can you create a linear task for it so we don't forget about it?

PR #2488, opened 2024-08-12
[eas-cli] Change behavior for ending EAS Update channel rollout

wschurman:
# Why

The previous behavior for ending a rollout on RTV 1 from branch `control` to branch `test` was to do one of the following outcomes:
- republish-and-revert: take the latest update from `test` and publish it to `control`, and set the channel pointer for all runtime versions to `control`.
- revert: just set the channel pointer for all runtime versions to `control`.

This was beneficial as it reverted the channel's branch mapping back to a constant branch pointer which can be better optimized on the backend.

But it also is a bit confusing from a UX perspective since when thinking about a rollout, it is most commonly that "rolling out a branch fully" puts all users on that branch.

# How

The fix to the UX problem is as follows:

The behavior for ending a rollout on RTV 1 from branch `control` to branch `test` is to do one of the following outcomes:
- roll-out-and-republish: take the latest update from `test` and re-publish it to `test`, and set the channel pointer for the rollout's runtime version to `test`.
- revert-and-republish: take the latest update from `control` and re-publish it to `control`, and set the channel pointer for the rollout's runtime version to `control`.

The tough part of this is that we end up with much more complex branch mappings for a channel once rollouts of multiple runtime versions have been run.

# Test Plan

Create rollout and end the rollout using interactive menu. See that in both scenarios the correct update is republished and the channel's branch mapping is correct.

### DIFF
--- a/packages/eas-cli/src/channel/branch-mapping.ts
+++ b/packages/eas-cli/src/channel/branch-mapping.ts
@@ -27,22 +27,25 @@ export type BranchMappingNode =
   | BranchMappingAlwaysTrue
   | BranchMappingObject
   | BranchMappingStatement;
-export type BranchMapping = {
-  version: number;
-  data: {
-    branchId: string;
-    branchMappingLogic: BranchMappingNode;
-  }[];
+
+export type BranchMappingDataItem = {
+  branchId: string;
+  branchMappingLogic: BranchMappingNode;
 };
 
-export type AlwaysTrueBranchMapping = {
+export type BranchMapping = {
   version: number;
-  data: [
-    {
-      branchId: string;
-      branchMappingLogic: BranchMappingAlwaysTrue;
-    },
-  ];
+  data: BranchMappingDataItem[];
+};
+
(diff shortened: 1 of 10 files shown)


wschurman:
* **#2488** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

wschurman (commented):
This is proving to be too complex. If we want to go with something like this, we'll need to add a new set of commands with a more general-purpose "branch mapping editor" rather than trying to infer a state transition from one branch mapping to the next for each type of command.

wschurman:
For posterity, here was the proposed change:

proposed change: (control branch, test branch)
- when rollout completed to 100%, point channel to test branch for the rollout's rtv, republish latest update for rtv from test branch to test branch
- when rollout reverted to 0%, point channel to control branch for the rollout's rtv, republish latest update for rtv from control branch to control branch

PR #2504, opened 2024-08-20
Implement env:init, env:load and env:unload commands

khamilowicz:
# Why

Commands streamlining usage of environment variables

* `eas env:init` - install direnv, set up .envrc file and .gitignore
* `eas env:load --environment production` - pulls envvars from given environment and loads them
* `eas env:unload` - unloads envvars

# Test Plan

Tested manually


packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+  };
+
+  async runAsync(): Promise<void> {
+    const {
+      flags: { 'non-interactive': nonInteractive },
+    } = await this.parse(EnvironmentVariableInit);
+
+    if (nonInteractive) {
+      throw new Error("Non-interactive mode is not supported for 'eas env:init'");
+    }
```
szdziedzic, line 34: I would just remove this flag from the command

packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+  'dotenv_if_exists .env;\ndotenv_if_exists .env.eas.local;\ndotenv_if_exists .env.local;\n';
+
+export default class EnvironmentVariableInit extends EasCommand {
+  static override description = 'setup environment variables';
+
+  static override hidden = true;
+
+  static override flags = {
+    ...EASNonInteractiveFlag,
+  };
```
szdziedzic, line 21:
```suggestion
```

packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+
+      const confirm = await confirmAsync({
+        message: `Do you want to add the direnv hook to ${shellConfigPath}?`,
+      });
+
+      if (!confirm) {
+        Log.log('Skipping adding the direnv hook to the shell config');
+        Log.log('You may need to add the direnv hook to your shell config manually.');
+        Log.log('Learn more: https://direnv.net/docs/hook.html');
+        return;
```
szdziedzic, line 62:
```suggestion
        throw new Error
```
to make it exit with non-zero code in such case

packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      const configContent = await readFile(shellConfigPath, 'utf8');
+
+      if (configContent.includes(direnvHookCmd)) {
+        Log.log('The direnv hook is already present in the shell config');
+        return;
```
szdziedzic, line 69: I would throw error here as well

packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      await appendFile(shellConfigPath, `\n${direnvHookCmd}\n`, 'utf8');
+      Log.log(`Added direnv hook to ${shellConfigPath}`);
+      await spawnAsync(...direnvInitCmd);
+    } else {
+      Log.log("Unable to determine the user's shell");
+      Log.log('You may need to add the direnv hook to your shell config manually.');
+      Log.log('Learn more: https://direnv.net/docs/hook.html');
+    }
```
szdziedzic, line 79: I believe we also need to throw an error here

packages/eas-cli/src/commands/env/init.ts
```diff
@@ -0,0 +1,231 @@
...
+        installArgs = ['install', '-y', 'direnv'];
+      } else if (stdout.includes('Fedora') || stdout.includes('CentOS')) {
+        Log.log('Detected a Red Hat-based Linux distribution.');
+        installCommand = 'sudo dnf';
+        installArgs = ['install', '-y', 'direnv'];
+      } else {
+        throw new Error('Your Linux distribution is not supported by this script.');
+      }
+    } else {
+      Log.log(`Your platform (${platform}) is not supported by this script.`);
```
szdziedzic, line 216:
```suggestion
     throw new Error(`Your platform (${platform}) is not supported by this script.`);
```

packages/eas-cli/src/commands/env/load.ts
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    const secretVariables = environmentVariables
+      .filter(({ value }) => value === null)
+      .map(({ name }) => name);
+
+    const envFileContent = environmentVariables
+      .filter((variable: EnvironmentVariableFragment) => variable.value !== null)
+      .map((variable: EnvironmentVariableFragment) => {
+        return `${variable.name}=${variable.value}`;
```
szdziedzic, line 68:
```suggestion
      .filter((variable => variable.value !== null)
      .map((variable) => {
        return `${variable.name}=${variable.value}`;
```

packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -24,7 +24,7 @@ export default class EnvironmentValuePull extends EasCommand {
     ...EASNonInteractiveFlag,
     path: Flags.string({
       description: 'Path to the result `.env` file',
-      default: '.env.local',
+      default: '.env.eas.local',
```
szdziedzic, line 27: TBH I would stick with `.env.local` as default, I believe that's what the average person expects, and `.env.local` works as well with Expo SDK by default but `.env.eas.local` doesn't.

szdziedzic (changes requested):
As discussed during the meeting this looks like a really cool idea and we definitely want to play with it more behind the `hidden` flag. Awesome! 🚀 

Few comments:

Issue #2519, opened 2024-08-27
eas-cli@11 breaks environment variable resolution


djMax:
### Summary

I have an eas.json like so:
```
{
  "cli": {
    "version": ">= 7.0.0"
  },
  "build": {
    "development": {
      "developmentClient": true,
      "channel": "development",
      "env": {
        "EXPO_PUBLIC_APP_TYPE": "flavora",
        "EXPO_PUBLIC_APP_ENV": "development",
      }
    },
```

and so on. In v10, these environment variables are correctly given to app.config.ts to do its job of building a dynamic environment aware config. In version 11, this has stopped working an no environment variables are defined, breaking generation of app.config.

### Managed or bare?

managed

### Environment

```
  expo-env-info 1.2.0 environment info:
    System:
      OS: macOS 14.6.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.12.2 - ~/.local/share/mise/installs/node/20/bin/node
      Yarn: 3.8.1 - ~/.local/share/mise/installs/node/20/bin/yarn
      npm: 10.5.0 - ~/.local/share/mise/installs/node/20/bin/npm
    Managers:
      CocoaPods: 1.15.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 23.5, iOS 17.5, macOS 14.5, tvOS 17.5, visionOS 1.2, watchOS 10.5
    IDEs:
      Android Studio: 2023.2 AI-232.10300.40.2321.11668458
      Xcode: 15.4/15F31d - /usr/bin/xcodebuild
    npmGlobalPackages:
      eas-cli: 11.0.0
    Expo Workflow: managed
```

### Error output

Project config: Slug for project identified by "extra.eas.projectId" (myapp-flavora) does not match the "slug" field (myapp-undefined). Learn more

### Reproducible demo or steps to reproduce from a blank project

I do not yet have one. This is a pretty major undocumented breaking change, so would love to either get a rationale or place to look first.

itajenglish:
+1 Experiencing the same issue

IjzerenHein:
Same here, we are trying to access the env-vars that are set by `eas-cli` in `app.config.ts` but it's no longer possible as of version 11

szdziedzic:
Thanks for the report!

[...]

PR #2521, opened 2024-08-27
[eas-cli] remove unnecessary static project config context definition from eas build command

szdziedzic:
# Why

https://github.com/expo/eas-cli/issues/2519

I was able to reproduce the issue and it is indeed true. It seems like we mistakenly added a static project config definition to the `eas build` command context in https://github.com/expo/eas-cli/commit/56510f092527e458aa9dca19feac0b5d425dac56#diff-cd2e00f4d286a4714f8736e14c635f4c3b98c402367cd71a83e4cd743034bd79. The output from it wasn't really used anywhere, but adding it caused the app configs to be evaluated without `buildProfile.env` in the first pass when trying the resolve project ID and slug at the begging of command execution and our slug/project ID assertions to fail. It only affected people with dynamic project IDs and slug fields.

# How

Remove unnecessary static context from context definitions, therefore avoid resolving project ID and slug from app config without env vars as a first thing when the `eas build` command is executed.

# Test Plan

Tested manually

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Remove unncessary static project config context definition from `eas build` command context, to fix resolving dynamic projest ID and slug for dynamic app configs. ([#2521](https://github.com/expo/eas-cli/pull/2521) by [@szdziedzic](https://github.com/szdziedzic))
+
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


szdziedzic:
* **#2521** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

khamilowicz (approved):

PR #2528, opened 2024-09-02
[eas-cli] remove @expo/config-types dependency

szdziedzic:
# Why

Rely on ExpoConfig type from the `@expo/config` package in all places for consistency

# How

Rely on ExpoConfig type from the `@expo/config` package in all places for consistency

# Test Plan

Tests

PR #2529, opened 2024-09-02
[eas-cli] use `npx expo config` command to resolve expo config if available

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C02123T524U/p1724970663050089?thread_ts=1724924614.773089&cid=C02123T524U
https://exponent-internal.slack.com/archives/C5ERY0TAR/p1725002816077089

Generally `@expo/config`, `@expo/prebuild-config`, and `@expo/config-plugins` are per SDK concept. Currently, in EAS CLI, we rely on one version of the `@expo/...` package to work well globally for all SDKs. It seems to sometimes bite us like in the case of https://exponent-internal.slack.com/archives/C02123T524U/p1724909908574979, because SDK 51 config wasn't compatible with SDK 50.

It's better to switch to using config-related functions shipped with `@expo/config`, `@expo/prebuild-config`, and `@expo/config-plugins`.  I believe we can assume that `getConfig`, `getPrebuildConfig`, and `compileModsAsync` shipped with these packages have stable API https://exponent-internal.slack.com/archives/C5ERY0TAR/p1725293002128849?thread_ts=1725002816.077089&cid=C5ERY0TAR.

# How

Use `getConfig`, `getPrebuildConfig`, and `compileModsAsync` from `@expo/config`, `@expo/prebuild-config`, and `@expo/config-plugins` shipped with Expo SDK installed in the user's project. If not available, fallback to versions installed per SDK. If we can assume that these guarantee stable API we can use types of these functions from packages installed for EAS CLI as well.

# Test Plan

Run builds for repro case of https://exponent-internal.slack.com/archives/C02123T524U/p1724909908574979 with SDK 51 config packages installed for EAS CLI (that previously broke it), see that it works and capabilities are correctly synced.

Initiate a new bare RN project using community CLI, and check that it works as well using fallback config loading.

### DIFF
--- a/packages/eas-cli/src/commands/update/__tests__/index.test.ts
+++ b/packages/eas-cli/src/commands/update/__tests__/index.test.ts
@@ -141,16 +141,7 @@ describe(UpdatePublish.name, () => {
 
     // Add configuration to the project that should not be included in the update
     const { appJson } = mockTestProject({
-      expoConfig: {
-        hooks: {
-          postPublish: [
-            {
-              file: 'custom-hook.js',
-              config: { some: 'config' },
-            },
-          ],
-        },
-      },
+      expoConfig: {},
     });
 
     const { platforms, runtimeVersion } = mockTestExport({ platforms: ['ios'] });
--- a/packages/eas-cli/src/commands/update/__tests__/index.test.ts
+++ b/packages/eas-cli/src/commands/update/__tests__/index.test.ts
@@ -140,9 +140,7 @@ describe(UpdatePublish.name, () => {
     const flags = ['--non-interactive', '--branch=branch123', '--message=abc'];
 
     // Add configuration to the project that should not be included in the update
-    const { appJson } = mockTestProject({
-      expoConfig: {},
-    });
+    const { appJson } = mockTestProject();
 
     const { platforms, runtimeVersion } = mockTestExport({ platforms: ['ios'] });
 
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -44,6 +44,8 @@ export async function createOrModifyExpoConfigAsync(
   }
 }
 
+let wasExpoConfigWarnPrinted = false;
+
 function getExpoConfigInternal(
   projectDir: string,
   opts: ExpoConfigOptionsInternal = {}
@@ -59,9 +61,12 @@ function getExpoConfigInternal(
     try {
       const expoConfig = require(projectExpoConfigPath);
       getConfig = expoConfig.getConfig;
-    } catch (error: any) {
-      Log.warn(`Failed to load getConfig function from ${projectExpoConfigPath}: ${error.message}`);
-      Log.warn('Falling back to the version of @expo/config shipped with the EAS CLI.');
+    } catch {
+      if (!wasExpoConfigWarnPrinted) {
+        Log.warn(`Failed to load getConfig function from ${projectExpoConfigPath}`);
+        Log.warn('Falling back to the version of @expo/config shipped with the EAS CLI.');
+        wasExpoConfigWarnPrinted = true;
+      }
       getConfig = _getConfig;
     }
     const { exp } = getConfig(projectDir, {
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -54,14 +54,13 @@ function getExpoConfigInternal(
       ...process.env,
       ...opts.env,
     };
+    const projectExpoConfigPath = path.join(projectDir, 'node_modules', '@expo', 'config');
     let getConfig: typeof _getConfig;
     try {
-      const expoConfig = require(`${projectDir}/node_modules/@expo/config`);
+      const expoConfig = require(projectExpoConfigPath);
       getConfig = expoConfig.getConfig;
     } catch (error: any) {
-      Log.warn(
-        `Failed to load getConfig function from ${projectDir}/node_modules/@expo/config: ${error.message}`
-      );
+      Log.warn(`Failed to load getConfig function from ${projectExpoConfigPath}: ${error.message}`);
       Log.warn('Falling back to the version of @expo/config shipped with the EAS CLI.');
       getConfig = _getConfig;
     }
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -62,6 +62,7 @@ function getExpoConfigInternal(
       Log.warn(
         `Failed to load getConfig function from ${projectDir}/node_modules/@expo/config: ${error.message}`
       );
+      Log.warn('Falling back to the version of @expo/config shipped with the EAS CLI.');
       getConfig = _getConfig;
     }
     const { exp } = getConfig(projectDir, {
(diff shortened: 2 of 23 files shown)


szdziedzic:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2529?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2530** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2529** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2528** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2528?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

wschurman (changes requested):
There are two approaches we could take, I don't have a strong preference:
1. Call the CLI (`npx expo config`).
2. `require` these modules in the project (what this PR does).

For `expo-updates` calls from EAS CLI we chose to use a CLI that returns JSON since it's a bit clearer to reviewers when a change is a breaking change, and as developers we're more attuned to knowing when a CLI change is breaking.

For option 2 to be stable in perpetuity, we need to add some constraint/test/something in the `@expo/config` package that calls it in the same way that we'd be introducing here in this PR to ensure that the function signature never changes. My recommendation would be an integration test in that package that is essentially the same as the updated `expoConfig.ts` and `entitlements.ts` files in this CLI.

I think if we can do option 2 (what this PR does) and add a signature stability API then this PR is good to go, but let's add that stability guarantee to the destination package before landing this to be sure it's doable.

wschurman:
Just a new piece of info: `npx expo config` has a `--json` flag to disable console during execution to prevent broken JSON!

szdziedzic:
Awesome! I will refactor it to use approach number 1 as in `expo-updates` then. Thanks @wschurman.

szdziedzic:
Oh, I see that it can be pretty straightforward for `npx expo config` however there still is `compileModsAsync` 🤔 I believe it still needs to be dynamically required, right? 🤔

szdziedzic:
Ok I believe `npx config --type introspect` should do the job here 🎉

szdziedzic:
@wschurman Thanks once again for the tips! I'm sorry that it took so long I was off the grid for the last 2 weeks and I've just got back to working normally. Can you please take a look at it again 🙏?

packages/eas-cli/src/project/expoConfig.ts
```diff
@@ -37,21 +41,52 @@ export async function createOrModifyExpoConfigAsync(
...
+
+    let exp: ExpoConfig;
+    if (isExpoInstalled(projectDir)) {
+      try {
+        const { stdout } = await spawnAsync(
+          'npx',
+          ['expo', 'config', '--json', ...(opts.isPublicConfig ? ['--type', 'public'] : [])],
+
+          {
+            cwd: projectDir,
```
wschurman, line 65: do we need to pass `env` into spawnAsync? (we do for the expo-updates CLI). Or does the process.env override on L52 work?

wschurman (approved):
One inline comment, otherwise LGTM. As far as releasing this goes, we'll definitely want to do some testing of a broad set of commands to ensure it works since almost every command calls it.

wschurman:
Would also be good to get one more approval on this before landing. Could help to remove some reviewers to better indicate who is responsible for review.

wschurman:
Narrowing reviewers so hopefully this gets reviewed.

szdziedzic:
@wschurman I plan to finally merge it today; I run both builds and updates with projects with different SDK versions and raw RN projects. 

It all seems to work. The issue that occurred after the previous `@expo/config` bump no longer occurs.

What do you think about marking a new release as major (even though it probably should be minor because this change should not be breaking) just for extra caution and extra warning for the users? I also plan to let Dev Success know about it so they can help people better if some weird things start to happen.

packages/eas-cli/src/project/expoConfig.ts
szdziedzic: Technically the override should work, but I passed envs explicitly either way so we are certain it won't break.

szdziedzic:
Ok I figured out that we also need to use `EXPO_NO_DOTENV=1` to keep CLI behavior consistent with what was it doing before. It's fixed.

packages/eas-cli/src/project/ios/entitlements.ts
```diff
@@ -22,14 +27,39 @@ export async function getManagedApplicationTargetEntitlementsAsync(
...
+            ...env,
+            EXPO_NO_DOTENV: '1',
+          },
+        }
+      );
+      expWithMods = JSON.parse(stdout);
+    } catch (err: any) {
+      if (!wasExpoConfigPluginsWarnPrinted) {
+        Log.warn(
+          `Failed to read the app config from the project using npx expo config command: ${err.message}.`
```
sjchmiela, line 50:
```suggestion
          `Failed to read the app config from the project using "npx expo config" command: ${err.message}.`
```
?

packages/eas-cli/src/project/ios/entitlements.ts
```diff
@@ -22,14 +27,39 @@ export async function getManagedApplicationTargetEntitlementsAsync(
...
+        );
+        Log.warn('Falling back to the version of @expo/config shipped with the EAS CLI.');
```
sjchmiela, line 52:
```suggestion
        Log.warn('Falling back to the version of "@expo/config" shipped with the EAS CLI.');
```

packages/eas-cli/src/commands/update/__tests__/index.test.ts
```diff
@@ -141,16 +141,7 @@ describe(UpdatePublish.name, () => {
...
-        hooks: {
-          postPublish: [
-            {
-              file: 'custom-hook.js',
-              config: { some: 'config' },
-            },
-          ],
-        },
-      },
+      expoConfig: {},
```
sjchmiela, line 144: Should this get removed?

sjchmiela (approved):
As far as I understand looks good.

khamilowicz (approved):

szdziedzic:
### Merge activity

* **Nov 5, 1:33 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/eas-cli/2529).

packages/eas-cli/src/project/ios/entitlements.ts
```diff
@@ -22,14 +27,39 @@ export async function getManagedApplicationTargetEntitlementsAsync(
...
-      platforms: ['ios'],
-      introspect: true,
-      ignoreExistingNativeFiles: await hasIgnoredIosProjectAsync(projectDir, vcsClient),
-    });
+
+    let expWithMods: ExportedConfig;
+    try {
+      const { stdout } = await spawnAsync(
+        'npx',
+        ['expo', 'config', '--json', '--type', 'introspect'],
```
wschurman, line 35: I think the issue might be that this command now does the config for both platforms (https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/config/configAsync.ts#L66-L75) while previously it only did the `getPrebuildConfigAsync` and `compileModsAsync` for ios.

PR #2536, opened 2024-09-05
feat(eas-cli): Add support for static worker deployments

kitten:
Related ENG-12989

# Why

This implements a quick check to ensure that we can differentiate between `output: "server"` and `output: "static"` and explicitly aborts the deployment if we detect `output: "single"` (which is also the current default)

This PR also implements support for the `output: "static"` export mode.

# How

When we go into `static` deployment mode, we now only upload client assets and skip server files from being added to the upload tarball.
In this mode, the entire `dist/` folder is uploaded as client assets.

See: https://github.com/expo/expo-easy-poc/pull/35

# Test Plan

- Manually tested


kadikraman (approved):
Tested locally, works like a charm 👌

PR #2543, opened 2024-09-10
[eas-cli] [ENG-10146] Allow submission of builds in progress

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-10146/fr-add-option-to-trigger-auto-submit-while-waiting-for-build-to

# How

When using `--latest` or selecting a build from a list when prompted, now the builds in-progress, in-queue and new are returned as well. All results are then sorted in decreasing order by `createdAt` and up to `limit` (4) are returned to the user to pick from

# Test Plan

Added tests for the function listing the builds

### Creating a build
Screenshot 2024-09-10 at 15 47 45

### In-progress build is returned when listing available builds. The description now includes the `status` as well, to differentiate between the builds in-progress and finished
Screenshot 2024-09-10 at 15 48 01

### The finished builds are also returned
Screenshot 2024-09-10 at 15 48 09

### Selecting in-progress build creates the submission
Screenshot 2024-09-10 at 15 48 23

### Created submission on the website
Screenshot 2024-09-10 at 15 48 54


radoslawkrzemien:
/changelog-entry new-feature Allow submitting builds in progress

szdziedzic (commented):
One general question: 

When a submission is created during a build command with `--auto-submit` it enters special waiting mode on our backend and is only put in the queue when we receive an update from Turtle in WWW that the build is finished or canceled when it fails. Will it behave here the same way? Or is it put in the queue straight away?

radoslawkrzemien:
@szdziedzic As far as I can tell it behaves exactly the same way - waiting for the build to finish first
Screenshot 2024-09-10 at 16 35 27

packages/eas-cli/src/submit/utils/builds.ts
```diff
@@ -8,14 +8,29 @@ export async function getRecentBuildsForSubmissionAsync(
...
-  return await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
+  const builds = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
     appId,
     limit,
     offset: 0,
     filter: {
       platform,
       distribution: DistributionType.Store,
-      status: BuildStatus.Finished,
+      status: BuildStatus.InProgress,
```
szdziedzic, line 18: What about `NEW` and `IN_QUEUE` builds? I believe we should handle submitting these as well.

packages/eas-cli/src/submit/utils/builds.ts
```diff
@@ -8,14 +8,29 @@ export async function getRecentBuildsForSubmissionAsync(
...
+      limit: remainingLimit,
+      offset: 0,
+      filter: {
+        platform,
+        distribution: DistributionType.Store,
+        status: BuildStatus.Finished,
+      },
+    });
+    builds.push(...finishedBuilds);
+  }
```
szdziedzic, line 34:
```suggestion
  }
  builds.sort((buildA, buildB) => (buildA.createdAt > buildB.createdAt ? -1 : 1));
```
I believe these should be sorted

szdziedzic (changes requested):
good job, it generally looks good 🚀, although I think we should also allow to submit new and in-queue builds as well

packages/eas-cli/src/submit/utils/builds.ts
radoslawkrzemien: I don't see why not, I'll add them

packages/eas-cli/src/submit/utils/builds.ts
radoslawkrzemien: I believe the results are already sorted for each of the calls, and in-progress builds should be newer than finished ones in general. But I have no problem with sorting them explicitly

packages/eas-cli/src/submit/ArchiveSource.ts
```diff
@@ -368,6 +369,7 @@ function formatBuildChoice(build: BuildFragment): prompts.Choice {
         ? chalk.bold(message.length > 200 ? `${message.slice(0, 200)}...` : message)
         : null,
     },
+    { name: 'Status', value: status.replace('_', ' ').toLowerCase() },
```
sjchmiela, line 372: nit: i'd rather have a mapping `Record<status, description>`

sjchmiela (approved):

radoslawkrzemien:
@szdziedzic any further feedback?

szdziedzic (approved):

PR #2544, opened 2024-09-10
chore(cli): Add `--non-interactive` flag to fix `worker:deploy` initialisation

kitten:
# Why

We were missing `--non-interactive` which meant that the command wasn't offering to initialize and tie the project to an EAS project when running `worker:deploy`

# How

Add missing flag parsing to `worker:deploy` command.

# Test Plan

Tested manually


szdziedzic (approved):

kadikraman (approved):
Looks good, verified locally 👍

PR #2545, opened 2024-09-10
feat(cli): Source `@expo/env` dotenv vars for worker deployments

kitten:
# Why

Deployments should allow environment variables to be added to them.

# How

Source environment variables using `@expo/env`, add them to the manifest, and send them to the deployment API to upload environment variables for deployments.

# Test Plan

- Manually tested


szdziedzic (commented):
Hmm, shouldn't we actually use server-side defined env vars for deployments? Why or why not? What is our approach here?

I believe that using @expo/env with EAS is one of the most commonly flagged pain points of EAS Update users. What I mean here is that you usually keep your local dev env vars in `.env.local`, but when you trigger a prod update they will unknowingly be used with the highest priority for you in your prod update bundle, so you need to be very careful when using EAS Update. 

By using @expo/env for worker deployments we might introduce the same confusion here 🤔

szdziedzic (approved):

PR #2548, opened 2024-09-11
feature(eas-cli): add `worker:alias` command

byCedric:
PRs #2550, #2551 are stacked on top of this PR, since this PR adds deployment alias mutations.

# Why

This adds `worker:alias` or `deploy:alias` command to assign different aliases from the CLI. These aliases affects the subdomain url.

# How

- Added `WorkerAlias` command, with (optional) flags `--id <deployment id>` and `--alias `<alias name>`

# Test Plan

<details><summary>When not providing the deployment ID, it will prompt the user to select the deployment:</summary>

worker-selection

</details>

<details><summary><code>$ eas worker:alias</code></summary>

worker-alias

</details>

<details><summary><code>$ eas worker:alias --id <deployment identifier></code></summary>

worker-alias-id

</details>

<details><summary><code>$ eas worker:alias --alias <alias name></code></summary>

worker-alias-name

</details>

<details><summary><code>$ eas worker:alias --id <deployment identifier> --alias <alias name></code></summary>

worker-alias-id-name

</details>


kitten (approved):

packages/eas-cli/src/commands/worker/alias.ts
```diff
@@ -0,0 +1,119 @@
...
+import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
+import Log from '../../log';
+import { ora } from '../../ora';
+import { promptAsync } from '../../prompts';
+import {
+  assignWorkerDeploymentAliasAsync,
+  selectWorkerDeploymentOnAppAsync,
+} from '../../worker/deployment';
+
+export default class WorkerAlias extends EasCommand {
```
szdziedzic, line 14: Should we support non-interactive mode in this command as well?

packages/eas-cli/src/commands/worker/alias.ts
```diff
@@ -0,0 +1,119 @@
...
+      deploymentId,
+      aliasName,
+    });
+
+    progress.succeed(
+      chalk`Alias {bold ${workerAlias.aliasName}} assigned to deployment {bold ${deploymentId}}`
+    );
+
+    const baseDomain = process.env.EXPO_STAGING ? 'staging.expo' : 'expo';
+    const aliasUrl = `https://${baseDomain}.dev/projects/${projectId}/serverless/deployments`;
```
szdziedzic, line 72: I believe we use this domain creation logic in multiple places. Would it make sense to move it to a function?

szdziedzic (approved):

packages/eas-cli/src/commands/worker/alias.ts
byCedric: Yes, that would be ideal. It's one of the follow-up tasks we have, basically adding `--json` to the `worker:alias` command.

The non-interactive mode is a bit implicit here, if you run with `--id` and `--alias`, it would not ask the user and just set the alias. But, I agree we should add the full `--non-interactive` and `--json` flags here too.

packages/eas-cli/src/commands/worker/alias.ts
byCedric: Yeah, that would be a nice cleanup. In this case I just copied it to get this working, its mostly used in 2 places now.

byCedric:
Merging this, and will open a follow-up PR to add `--json` and `--non-interactive`. The command itself is still hidden and should not be used by normal users yet.

PR #2550, opened 2024-09-11
feature(eas-cli): add `worker --prod` flag and hint for production deployments

byCedric:
This PR is stacked on top of #2548, since we use the alias mutation here too.

# Why

This adds the `--prod` flag to `eas worker`, assigning the deployment as production.

> Note, this also changes the URL of the project to use `/projects/<id>/serverless/deployments`. This is to avoid having to use or load `exp.owner` in these URLs.

# How

- Added hint when running `eas worker` without `--prod`
  without-prod
- Added optional `--prod` flag to `eas worker`
  with-prod

# Test Plan

- `$ eas worker`
- `$ eas worker --prod`

### DIFF
--- a/packages/eas-cli/src/commands/worker/deploy.ts
+++ b/packages/eas-cli/src/commands/worker/deploy.ts
@@ -1,3 +1,4 @@
+import { Flags } from '@oclif/core';
 import chalk from 'chalk';
 import fs from 'node:fs';
 import * as path from 'node:path';
@@ -8,7 +9,10 @@ import Log from '../../log';
 import { ora } from '../../ora';
 import { createProgressTracker } from '../../utils/progress';
 import * as WorkerAssets from '../../worker/assets';
-import { getSignedDeploymentUrlAsync } from '../../worker/deployment';
+import {
+  assignWorkerDeploymentProductionAsync,
+  getSignedDeploymentUrlAsync,
+} from '../../worker/deployment';
 import { UploadParams, batchUploadAsync, uploadAsync } from '../../worker/upload';
 
 const isDirectory = (directoryPath: string): Promise<boolean> =>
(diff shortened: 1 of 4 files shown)


kitten (approved):

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -36,6 +42,9 @@ export default class WorkerDeploy extends EasCommand {
   static override state = 'beta';
 
   static override flags = {
+    prod: Flags.boolean({
+      description: 'Deploy to production',
+    }),
```
szdziedzic, line 47: nit: if you want to you can even do
```suggestion
    prod: Flags.boolean({
      description: 'Deploy to production',
      default: false,
    }),
```
to make flag required, but with default value

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -20,11 +24,13 @@ const isDirectory = (directoryPath: string): Promise<boolean> =>
 interface DeployFlags {
   nonInteractive: boolean;
   json: boolean;
+  prod: boolean;
 }
 
 interface RawDeployFlags {
   'non-interactive': boolean;
   json: boolean;
+  prod?: boolean;
```
szdziedzic, line 33: and change it here
```suggestion
  prod: boolean;
```

packages/eas-cli/src/commands/worker/deploy.ts
```diff
@@ -217,20 +226,65 @@ export default class WorkerDeploy extends EasCommand {
...
+    if (!flags.prod) {
+      logDeployment({
+        dashboardUrl,
+        deploymentUrl: `https://${deployResult.fullName}.${expoBaseDomain}.app`,
+        isProduction: false,
+      });
+      return;
+    }
+
+    progress = ora('Assigning worker deployment to production');
```
szdziedzic, line 241:
```suggestion
    progress = ora('Assigning worker deployment to production').start();
```

szdziedzic (approved):
Did you run `yarn generate-graphql-code` after changing the mutation?

packages/eas-cli/src/commands/worker/deploy.ts
byCedric: Oh I see, to remove the sanitization part, right?

PR #2576, opened 2024-09-18
[eas-cli] Let iOS bundle identifier override always

sjchmiela:
# Why

In a submit job we're going to have an IPA/APK to submit. We don't want to resolve any bundle identifiers from the project, we can use one from that IPA/APK.

# How

Let the override env variable override bundle identifier resolution if the workflow is managed too.

# Test Plan

I used it to submit an IPA with workflows.


sjchmiela:
/changelog-entry new-feature Let application ID overrides work in any workflow, not only in bare workflow.

szdziedzic (approved):

PR #2597, opened 2024-09-24
refactor(eas-cli): allow Expo Router server deployments without client directory

byCedric:
# Why

Previously, we checked on the existence of both `/dist/client` and `/dist/server` to enable a deployment for API routes. With upcoming changes in Expo Router, we'd need to support a "server only" deployment. This means, we can't check `/dist/client` as it might not exist.

# How

- Check on `/dist/server/_expo/routes.json` to enable server deployments
- Make `/dist/client` optional during upload

# Test Plan

- `$ bun create expo ./test-server-only`
- `$ cd ./test-server-only`
- Create an API route, e.g. `app/test+api.ts` with:
  ```ts
  export function GET() {
    return Response.json({ ok: true });
  }
  ```
- Change the output type to `server` in `app.json`'s `web.output = server`.
- `$ rm -rf ./dist/client`
- `$ easd deploy`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Simplify the output of `eas deploy --json`. ([#2596](https://github.com/expo/eas-cli/pull/2596) by [@byCedric](https://github.com/byCedric))
+- Support deploying Expo Router server exports without client directory. ([#2597](https://github.com/expo/eas-cli/pull/2597) by [@byCedric](https://github.com/byCedric))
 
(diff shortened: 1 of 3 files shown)


byCedric:
@kitten the "wait until the deployment is ready before responding that the deployment is ready"-thing broke with this change 🙈

image

kitten (approved):
Approving since implementation itself is fine.

Re. the timing issue, this ticket would resolve these problems for sure: https://linear.app/expo/issue/ENG-13544/implement-closing-request-for-upload-logic

But weirdly enough, I can't spot any deferred logic in here that would otherwise cause this. So, not sure why this timing issue is happening

byCedric:
The mystery has been solved by @kitten, it's pending merge & deploy. Going to merge this PR as is; the erroneous case should be fixed by the pending PR.

PR #2602, opened 2024-09-26
[eas-cli] Disallow republishing an update that is being rolled-out

wschurman:
# Why

What it means to republish an update that is being rolled-out is undefined behavior. Rather than try to guess at the user's intent, just throw an error in this case and give guidance about other options to help them clarify intent.

Note that this doesn't hide the groups during interactive chooser, since that'd be even more confusing IMO.

Closes ENG-13613.

# How

Throw error when any update being republished is a rollout update.

# Test Plan

```
$ neas update:republish --branch main
✔ Load more update groups? › [Aug 19 16:23 by wschurman, runtimeVersion: 1.0.0] wat2
✔ The republished update group will appear only on: all
✔ Provide an update message. … dsadsasd
Cannot republish an update that is being rolled-out. Either complete the update rollout and then republish or publish a new rollout update.
    Error: update:republish command failed.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Disallow republishing an update that is being rolled-out. ([#2602](https://github.com/expo/eas-cli/pull/2602) by [@wschurman](https://github.com/wschurman))
+
 ### 🧹 Chores
 
 - Implement new `worker` deploy API flow. ([#2601](https://github.com/expo/eas-cli/pull/2601) by [@kitten](https://github.com/kitten)))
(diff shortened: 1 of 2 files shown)


wschurman:
* **#2602** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

quinlanj (approved):

quinlanj:
would be nice to add server side guards for this in case we havent already

wschurman:
> would be nice to add server side guards for this in case we havent already

Unfortunately a republish is just a normal publish, so there isn't a way to detect a republish on the server. (I think this is good and by design though, one less concept to keep track of)

PR #2611, opened 2024-10-07
[eas-cli] Add message about concurrency add-on

brentvatne:
# Why

Let people know that they can add more concurrencies easily

# How

Make the message more clear

# Test Plan

-

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -486,8 +486,8 @@ async function handleSingleBuildProgressAsync(
       spinner.succeed('Build finished');
       return { refetch: false };
     case BuildStatus.New:
-      spinner.text = `Build is waiting to enter the queue. Check your concurrency limit at ${link(
-        formatAccountSubscriptionsUrl(accountName)
+      spinner.text = `Build concurrency limit reached for your account, waiting for a concurrency to become available. Add additional concurrencies at ${link(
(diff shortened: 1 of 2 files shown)


brentvatne:
/changelog-entry chore Let folks know about the new concurrency add-on

brentvatne:
/changelog-entry chore Let folks know about the new concurrency add-on

packages/eas-cli/src/build/build.ts
```diff
@@ -486,8 +486,8 @@ async function handleSingleBuildProgressAsync(
       spinner.succeed('Build finished');
       return { refetch: false };
     case BuildStatus.New:
-      spinner.text = `Build is waiting to enter the queue. Check your concurrency limit at ${link(
-        formatAccountSubscriptionsUrl(accountName)
+      spinner.text = `Build concurrency limit reached for your account, waiting for a concurrency to become available. Add additional concurrencies at ${link(
```
keith-kurak, line 489:
```suggestion
      spinner.text = `Build concurrency limit reached for your account. Build will enter queue once a concurrency becomes available. Add additional concurrencies at ${link(
```
Just an idea... I think the original message communicates slightly better what's happening to the _build_.

keith-kurak (approved):
Left an idea for a verbiage tweak, but looks good to me!

PR #2613, opened 2024-10-09
[eas-cli] Add `configInput` to `submit:internal` output

sjchmiela:
# Why

When creating a Submission Entity I want to also provide the submission config. https://linear.app/expo/issue/ENG-13612/create-submission-entity-upon-successful-workflow-submission

# How

Added `configInput` next to `config`. Backwards compatible.

# Test Plan

`EXPO_LOCAL=1 /Users/sjchmiela/Developer/eas-cli/bin/run submit:internal --platform ios --id eebf0c9b-8f7a-41c0-9633-cb1576e17e23`:

Zrzut ekranu 2024-10-9 o 11 39 40

### DIFF
--- a/packages/eas-cli/src/commands/submit/internal.ts
+++ b/packages/eas-cli/src/commands/submit/internal.ts
@@ -92,6 +92,7 @@ export default class SubmitInternal extends EasCommand {
     });
 
     let config;
+    let configInput;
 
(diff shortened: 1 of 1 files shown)


sjchmiela:
Meh, I'll create the object with the properties I get from "prepare credentials" manually.

Issue #2600, opened 2024-09-25
"$schema" is not allowed


saiichihashimoto:
### Summary

When running `eas build` with a [$schema definition](https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition) in the eas.json, it fails with the following error:

```
eas.json is not valid.
- "$schema" is not allowed
    Error: build command failed.
```

My eas.json has `"$schema": "https://raw.githubusercontent.com/expo/eas-cli/main/packages/eas-json/schema/eas.schema.json"`, which is the one in [json schema store](https://www.schemastore.org/json/).

### Managed or bare?

managed

### Environment

  expo-env-info 1.2.0 environment info:
    System:
      OS: macOS 14.6
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
      Yarn: 1.22.19 - /opt/homebrew/bin/yarn
      npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
      Watchman: 2024.08.26.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.15.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.0, iOS 18.0, macOS 15.0, tvOS 18.0, visionOS 2.0, watchOS 11.0
    IDEs:
      Android Studio: 2024.1 AI-241.19072.14.2412.12360217
      Xcode: 16.0/16A242d - /usr/bin/xcodebuild
    Expo Workflow: bare

Detailed check results:

Expected package @expo/config-plugins@~8.0.0
Found invalid:
  @expo/config-plugins@7.8.4
  (for more info, run: npm why @expo/config-plugins)
Expected package @expo/prebuild-config@~7.0.0
Found invalid:
  @expo/prebuild-config@6.7.3
  (for more info, run: npm why @expo/prebuild-config)
Advice: Upgrade dependencies that are using the invalid package versions.

The following packages should be updated for best compatibility with the installed expo version:
  react@18.3.1 - expected version: 18.2.0
  react-native@0.75.3 - expected version: 0.74.5
Your project may not work correctly until you install the expected versions of the packages.
Found outdated dependencies
Advice: Use 'npx expo install --check' to review and upgrade your dependencies.

### Error output

eas.json is not valid.
- "$schema" is not allowed
[...]

PR #2624, opened 2024-10-15
Allow $schema in eas.json

saiichihashimoto:
# Why

JSON Schemas are an open and well used method of validating json configs and works well with [existing tools](https://www.npmjs.com/package/eslint-plugin-json-schema-validator). While the VSCode Expo plugin exists, working with existing tooling is ideal (and the expo already uses a json schema internally, so the additional lift is minimal).

# How

Added to the Joi config

Closes https://github.com/expo/eas-cli/issues/2600


saiichihashimoto:
@szdziedzic do I need to reach out to a reviewer? What's the process for moving this forward?

brentvatne:
this seems reasonable to me, but i defer to @byCedric

szdziedzic (approved):

szdziedzic:
Thanks!

PR #2628, opened 2024-10-17
[eas-cli] add `--with-eas-environment-variables-set` flag to `eas update` command

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C013ZK4SA12/p1725644220583309

We are planning to introduce enhancements to our env vars system soon. Currently, we only allow people to keep secrets on EAS servers. Secrets are not readable outside of EAS servers so of course EAS Update doesn't work with them. In a new system, people can also store readable env vars on our servers, which opens new possibilities about how EAS Update can interact with env vars!

In this PR I'm adding support for the `--with-eas-environment-variables-set production|preview|development` flag. If used, we use the flag's value to fetch readable env vars from our servers and use it in the `expo export` command to produce an update bundle based on these env vars. If the flag is used I'm using the `EXPO_NO_DOTENV` flag to disable Expo CLI from loading env vars from `.env` files and potentially unwillingly overwriting env vars fetched from our servers for a given environment.

Users seem to face big issues with using env vars with EAS Build & Update at the moment, originating from the fact that these 2 handle different, non-compatible ways of using env vars. Env vars in the build profile and server-side secrets vs local env vars + `.env`. With this flag it will be easy to avoid envs mismatch by doing `eas update ----with-eas-environment-variables-set environment` and `eas build` with `{ ..., environment: environment, ...} in the build profile!

# How

Add a new hidden (for now) flag `--with-eas-environment-variables-set` that indicates which environment to use. If used fetch env vars from server. Merge and overwrite `process.env` with newly fetched env vars. Use them to resolve the expo config. Don't use `.env` env var loading from Expo CLI if server-side env vars are used.

# Test Plan

Test manually. Run `eas update` with and without a new flag. Check that the expo config is resolved correctly.


szdziedzic:
* **#2628** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -197,7 +217,13 @@ export default class UpdatePublish extends EasCommand {
...
-    } = await getDynamicPublicProjectConfigAsync();
+    } = withEasEnvironmentVariablesSet
+      ? await resolveExpoConfigWithEasEnvironmentVariablesLoadedAsync({
+          environment: withEasEnvironmentVariablesSet,
+          graphqlClient,
+          getDynamicPublicProjectConfigAsync,
+        })
+      : await getDynamicPublicProjectConfigAsync();
 
     await maybeWarnAboutEasOutagesAsync(graphqlClient, [StatuspageServiceName.EasUpdate]);
```
wschurman, line 229: <i class='graphite__hidden'>[Re: lines +239 to +240]<br/><br/></i>

do we need it for these calls as well?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2628?utm_source=unchanged-line-comment">Graphite</a>.</span>

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -651,6 +692,99 @@ export default class UpdatePublish extends EasCommand {
...
+    exp: ExpoConfig;
+    projectDir: string;
+  }>;
+}): Promise<{
+  projectId: string;
+  exp: ExpoConfig;
+  projectDir: string;
+}> {
+  const { projectId } = await getDynamicPublicProjectConfigAsync();
+  await maybeUpdateProcessEnvWithEasEnvironmentVariablesAsync({
```
wschurman, line 774: should this revert the process.env back after executing the inner function? (the way we do in `getExpoConfigInternal`)

wschurman (changes requested):
I wonder if maybe this should live up a level in the contexts to better indicate where this flag might be necessary to apply (which commands it needs to be added to for consistency)? (`DynamicPublicProjectConfigContextField`, `PrivateProjectConfigContextField`, `OptionalPrivateProjectConfigContextField`). As it stands, I think we would at least need to add it to a few other update commands (roll-back-to-embedded, etc) in case the config resolved in those commands also depends on these env variables, which makes me think this might be good to move into the contexts. I might be wrong though. What do you think?

szdziedzic:
@wschurman @quinlanj I moved server side env vars fetching to context functions. This way they will be used every time when one class `getExpoConfig` function if `withServerSideEnvironment` context resolution option is set. It will be easier to not forget about using them for resolving config. 

I also added a context option to get a function that will fetch server side env vars for given project and environment.

Please let me know what you think about it now 🙏 

If it looks OK, I will add the same flag and logic to the rest of `eas update:xyz` commands.

khamilowicz (approved):

packages/eas-cli/src/commandUtils/context/contextUtils/loadServerSideEnvironmentVariablesAsync.ts
```diff
@@ -0,0 +1,58 @@
...
+  environment,
+  projectId,
+  graphqlClient,
+}: {
+  environment: EnvironmentVariableEnvironment;
+  projectId: string;
+  graphqlClient: ExpoGraphqlClient;
+}): Promise<Record<string, string>> {
+  // don't load environment variables if they were already loaded while executing a command
+  if (cachedServerSideEnvironmentVariables) {
```
wschurman, line 18: this should be memoized by environment instead of globally just in case a second call is added elsewhere during command execution that calls it with a different arg.

```
let cachedServerSideEnvironmentVariables: Map<
  EnvironmentVariableEnvironment,
  Record<string, string>
> = new Map();
```

```suggestion
  if (cachedServerSideEnvironmentVariables.has(environment)) {
```

wschurman (approved):
One inline comment. Also I believe we also need to add this to `OptionalPrivateProjectConfigContextField`?

For the follow-up PR that adds the argument elsewhere, I think there's a good chance we'll need to add the  `withServerSideEnvironment ` argument to every command that uses any of (`DynamicProjectConfig`, `ProjectConfig`, `OptionalProjectConfig`, `ServerSideEnvironmentVariables`) contexts? Maybe there's a clever way we could use typescript to enforce that the argument is supplied when any of those fields is used.

szdziedzic:
>Also I believe we also need to add this to OptionalPrivateProjectConfigContextField?

Added 🫡 

>For the follow-up PR that adds the argument elsewhere, I think there's a good chance we'll need to add the withServerSideEnvironment argument to every command that uses any of (DynamicProjectConfig, ProjectConfig, OptionalProjectConfig, ServerSideEnvironmentVariables) contexts? Maybe there's a clever way we could use typescript to enforce that the argument is supplied when any of those fields is used.

@wschurman  believe I was able to do so. Can you take a look once again please?

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -45,6 +47,20 @@ export type ContextOutput<
```
wschurman, line 47: Add utility type helper for inline comment below:

```suggestion

type GetContextType<Type> = {
  [Property in keyof Type]: any;
};

```

packages/eas-cli/src/commandUtils/EasCommand.ts
```diff
@@ -141,7 +161,21 @@ export default abstract class EasCommand extends Command {
...
+      withServerSideEnvironment,
+    }: C extends { getDynamicPrivateProjectConfigAsync: any }
+      ? GetContextAsyncArgsWithRequiredServerSideEnvironmentArgument
+      : C extends { getDynamicPublicProjectConfigAsync: any }
+        ? GetContextAsyncArgsWithRequiredServerSideEnvironmentArgument
+        : C extends { optionalPrivateProjectConfig: any }
+          ? GetContextAsyncArgsWithRequiredServerSideEnvironmentArgument
+          : C extends { getServerSideEnvironmentVariablesAsync: any }
+            ? GetContextAsyncArgsWithRequiredServerSideEnvironmentArgument
+            : GetContextAsyncArgsWithoutServerSideEnvironmentArgument
```
wschurman, line 178:
```suggestion
    }: C extends
      | GetContextType<typeof EasCommand.ContextOptions.ProjectConfig>
      | GetContextType<typeof EasCommand.ContextOptions.DynamicProjectConfig>
      | GetContextType<typeof EasCommand.ContextOptions.OptionalProjectConfig>
      | GetContextType<typeof EasCommand.ContextOptions.ServerSideEnvironmentVariables>
      ? GetContextAsyncArgsWithRequiredServerSideEnvironmentArgument
      : GetContextAsyncArgsWithoutServerSideEnvironmentArgument
```

wschurman (approved):
Awesome use of types! Made a few small adjustments but overall it's really cool to be able to have the typechecker tell us where this is needed.

Side note: I think we need to add the variable fetching code to `PrivateProjectConfigContextField` as well?

quinlanj (approved):

wschurman:
I see that if we add it to `PrivateProjectConfigContextField` and make it's context require the `withServerSideEnvironment` argument as I proposed, it reveals quite a number of commands that need it.

Looking at the commands though, I see a number of them just get the `projectId` out of the `privateProjectConfig`, so this might signal that a new context is needed that is the same as `ProjectConfig` but only exposes the `projectId`?

wschurman:
(assuming projectId is not allowed to be altered via a server-side environment variable since that'd be wonky)

szdziedzic:
@wschurman I added a separate PR in which I add new context

szdziedzic:
### Merge activity

* **Oct 25, 8:18 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/expo/eas-cli/2628).

PR #2629, opened 2024-10-17
Use the same names for visibility options in eas cli as on

khamilowicz:
# Why

[ENG-13849: When no `--visibility` input is provided `eas env:create` should ask about it by default](https://linear.app/expo/issue/ENG-13849/when-no-visibility-input-is-provided-eas-envcreate-should-ask-about-it)

[ENG-13850: Use the same names for visibility options in EAS CLI as on website](https://linear.app/expo/issue/ENG-13850/use-the-same-names-for-visibility-options-in-eas-cli-as-on-website)

`eas env:*` commands were using obsolete names for visibility options: `public` and `secret`, instead of `plaintext` and `encrypted`. It should be updated.
Also, user is prompted for the variable visibility if not provided.

# Test Plan

Updated tests

### DIFF
--- a/packages/eas-cli/src/commandUtils/flags.ts
+++ b/packages/eas-cli/src/commandUtils/flags.ts
@@ -1,10 +1,6 @@
 import { Flags } from '@oclif/core';
 
-import {
-  EnvironmentVariableEnvironment,
-  EnvironmentVariableScope,
-  EnvironmentVariableVisibility,
-} from '../graphql/generated';
+import { EnvironmentVariableEnvironment, EnvironmentVariableScope } from '../graphql/generated';
 
 // NOTE: not exactly true, but, provided mapToLowercase and upperCaseAsync
 // are used in tandem, it saves on unnecessary typying in commands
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -56,14 +52,9 @@ export const EASVariableFormatFlag = {
 };
 
 export const EASVariableVisibilityFlag = {
-  visibility: Flags.enum<EnvironmentVariableVisibility>({
+  visibility: Flags.enum({
```
szdziedzic, line 55:
```suggestion
  visibility: Flags.enum<'plaintext' | 'sensitive' | 'encrypted'>({
```

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -32,6 +31,19 @@ export async function promptVariableTypeAsync(
   });
 }
 
+export function parseVisibility(stringVisibility: string): EnvironmentVariableVisibility {
```
szdziedzic, line 34:
```suggestion
export function parseVisibility(stringVisibility: 'plaintext' | 'sensitive' | 'encrypted'): EnvironmentVariableVisibility {
```

szdziedzic (approved):

PR #2632, opened 2024-10-17
Use lower case environment across all eas env commands

khamilowicz:
# Why

[ENG-13846: Use lower-case environment names across all `eas env:xyz` commands for consistency](https://linear.app/expo/issue/ENG-13846/use-lower-case-environment-names-across-all-eas-envxyz-commands-for)

### DIFF
--- a/packages/eas-cli/src/utils/prompts.ts
+++ b/packages/eas-cli/src/utils/prompts.ts
@@ -81,7 +81,7 @@ export async function promptVariableEnvironmentAsync({
     return await selectAsync(
       'Select environment:',
       (availableEnvironments ?? Object.values(EnvironmentVariableEnvironment)).map(environment => ({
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/utils/prompts.ts
```diff
@@ -91,7 +91,7 @@ export async function promptVariableEnvironmentAsync({
     name: 'environments',
     type: 'multiselect',
     choices: Object.values(EnvironmentVariableEnvironment).map(environment => ({
-      title: environment,
+      title: environment.toLocaleLowerCase(),
```
szdziedzic, line 94: why `toLocaleLowerCase()` vs `toLowerCase()`?

szdziedzic (approved):

PR #2636, opened 2024-10-18
Make eas env:pull pull file env vars

khamilowicz:
# Why

File variables were not handled by the `env:pull` command. They should be downloaded just like the other values.

# How

* `env:pull` saves file type variables to `.eas/.env` directory and sets its value in `.env.local` file to its file path
* `env:push` skips file type variables

# Test Plan

Tested manually

### DIFF
--- a/packages/eas-cli/src/commands/env/pull.ts
+++ b/packages/eas-cli/src/commands/env/pull.ts
@@ -1,10 +1,18 @@
 import { Flags } from '@oclif/core';
 import * as fs from 'fs-extra';
+import path from 'path';
 
 import EasCommand from '../../commandUtils/EasCommand';
 import { EASEnvironmentFlag, EASNonInteractiveFlag } from '../../commandUtils/flags';
-import { EnvironmentVariableFragment } from '../../graphql/generated';
-import { EnvironmentVariablesQuery } from '../../graphql/queries/EnvironmentVariablesQuery';
+import {
+  EnvironmentSecretType,
+  EnvironmentVariableFragment,
+  EnvironmentVariableVisibility,
+} from '../../graphql/generated';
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -65,15 +75,30 @@
...
+      await fs.mkdir(envDir, { recursive: true });
+    }
+
+    const envFileContentLines = await Promise.all(
+      environmentVariables.map(async (variable: EnvironmentVariableWithFileContent) => {
+        if (variable.visibility === EnvironmentVariableVisibility.Secret) {
           return `# ${variable.name}=***** (secret variables are not available for reading)`;
         }
+        if (variable.type === EnvironmentSecretType.FileBase64 && variable.valueWithFileContent) {
+          const filePath = `${envDir}/${variable.name}`;
```
szdziedzic, line 93: can we use `path.join` to possibly make it work on Windows as well?

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -196,6 +196,10 @@ export default class EnvironmentVariablePush extends EasCommand {
     const variables: Record<string, string> = dotenv.parse(await fs.readFile(envPath, 'utf8'));
 
     for (const [name, value] of Object.entries(variables)) {
+      // Skip file variables
+      if (value.endsWith(`.eas/.env/${name}`)) {
+        continue;
+      }
```
szdziedzic, line 202: Maybe we can be verbose about it?

szdziedzic (changes requested):

szdziedzic (approved):

PR #2503, opened 2024-08-20
[eas-cli] Add update:edit command for editing rollout

wschurman:
# Why

This adds the ability to increase a rollout for an update.

# How

Add new command, `update:edit` that takes the rollout-percentage flag.

# Test Plan

```
$ neas update:edit 63809081-ee5f-4216-add2-b3412aae7a16
✔ New rollout percentage (min: 30, max: 100) … 31
Update group:
Platforms                 android, ios
Runtime Version           1.0.0
Message                   "wat2" (1 hour ago by wschurman)
Code Signing Key          N/A
Is Roll Back to Embedded  No
Rollout Percentage        31%
Group ID                  63809081-ee5f-4216-add2-b3412aae7a16

$ neas update:view 63809081-ee5f-4216-add2-b3412aae7a16
Update group:
Platforms                 android, ios
Runtime Version           1.0.0
Message                   "wat2" (1 hour ago by wschurman)
Code Signing Key          N/A
Is Roll Back to Embedded  No
Rollout Percentage        31%
Group ID                  63809081-ee5f-4216-add2-b3412aae7a16
```

PR #2638, opened 2024-10-18
[eas-cli] Add interactivity to eas update:edit command

wschurman:
# Why

When writing the docs for update-based rollouts, I noticed that a) this command was still hidden, and b) it didn't have the best UX for interactive mode (https://github.com/expo/eas-cli/pull/2503 didn't add it)

# How

a) unhide the command
b) Make group arg optional, add optional branch flag, and lead user through interactive process to choose an update group to edit if not supplied as an arg

Note that this doesn't filter update groups to those that are editable since it isn't currently possible to query. We can update this in the future if we want, but usability of the feature probably won't suffer much without it since it's likely the latest update group will always be the one chosen to be edited.

# Test Plan

`neas update:edit`
`neas update:edit --branch main`


wschurman:
* **#2638** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2638?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

quinlanj (approved):

PR #2517, opened 2024-08-26
[eas-cli] make `eas config` command not require authentication when running in `--eas-json-only`  mode

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/CLQ58M23X/p1724709887450179

# How

Make sure that we require authentication only if not run in the `--eas-json-only` mode

# Test Plan

test manually

PR #2648, opened 2024-10-22
[eas-cli] Create dynamic logged in context field and clean up erroneous SessionManager context field uses

wschurman:
# Why

Noticed while reviewing another PR that the `config` command was using the session management context, when that context should only be necessary to use in `login` and `logout` commands.

Digging in a bit, it was needed because the command only required being logged in in a certain conditional codepath, but there was no context construct available to support this.

Ref: https://github.com/expo/eas-cli/pull/2517
Ref: https://github.com/expo/eas-cli/pull/2461

Closes ENG-13892.

# How

1. Add a new context field, `DynamicLoggedIn`, which requires log in but only when the context function is called.
2. Change `config` command to use this.
3. Change logged in contexts to also expose the authentication info so that session isn't needed in `account:view` command.

# Test Plan

While both logged in and logged out:

```
neas account:view
neas config
neas config --eas-json-only
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Create dynamic logged in context field and clean up erroneous SessionManager context field uses. ([#2648](https://github.com/expo/eas-cli/pull/2648) by [@wschurman](https://github.com/wschurman))
+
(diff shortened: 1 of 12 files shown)


wschurman:
* **#2648** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

khamilowicz (approved):

szdziedzic (approved):
looks reasonable

PR #2650, opened 2024-10-24
Adds workflow:validate command

jonsamp:
# Why

This adds a command to validate YAML in workflow configuration files.

# Screenshots

## Valid YAML output
Screenshot 2024-10-24 at 12 04 22 AM

## Invalid YAML output
Screenshot 2024-10-24 at 12 04 28 AM

# Test Plan

Create a .yml file and test it with `eas workflow:validate path/to/file.yml`. You should see output like the screenshots above.
- Make sure non existent files result in an error
- Make sure non .yml/.yaml files result in an error
- Make sure bad YAML results in an error
- Make sure good YAML results in no error

### DIFF
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -58,7 +58,6 @@
     "indent-string": "4.0.0",
     "jks-js": "1.1.0",
     "joi": "17.11.0",
-    "js-yaml": "^4.1.0",
     "jsonwebtoken": "9.0.0",
     "keychain": "1.5.0",
     "log-symbols": "4.1.0",
@@ -87,6 +86,7 @@
     "untildify": "4.0.0",
     "uuid": "9.0.1",
     "wrap-ansi": "7.0.0",
+    "yaml": "2.6.0",
     "zod": "^3.23.8"
   },
   "devDependencies": {
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -58,6 +58,7 @@
     "indent-string": "4.0.0",
     "jks-js": "1.1.0",
     "joi": "17.11.0",
+    "js-yaml": "^4.1.0",
     "jsonwebtoken": "9.0.0",
     "keychain": "1.5.0",
     "log-symbols": "4.1.0",
--- a/packages/eas-cli/src/commands/workflow/validate.ts
+++ b/packages/eas-cli/src/commands/workflow/validate.ts
@@ -8,6 +8,7 @@ import { ora } from '../../ora';
 
 export class WorkflowValidate extends EasCommand {
   static override description = 'validate a workflow configuration yaml file';
+  static override hidden = true;
 
   static override args = [
     {
--- a/packages/eas-cli/src/commands/workflow/validate.ts
+++ b/packages/eas-cli/src/commands/workflow/validate.ts
@@ -1,0 +1,83 @@
+import fs from 'fs/promises';
+import yaml from 'js-yaml';
+import path from 'path';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import Log from '../../log';
+import { ora } from '../../ora';
+
+export class WorkflowValidate extends EasCommand {
+  static override description = 'validate a workflow configuration yaml file';
+
+  static override args = [
+    {
+      name: 'path',
+      description: 'Path to the workflow configuration YAML file (must end with .yml or .yaml)',
+      required: true,
+    },
+  ];
+
+  async runAsync(): Promise<void> {
+    const {
+      args: { path: filePath },
+    } = await this.parse(WorkflowValidate);
+
+    const spinner = ora().start('Validating the workflow YAML file…');
+
+    try {
+      await validateYAMLAsync(filePath);
+      spinner.succeed('Workflow configuration YAML is valid.');
+    } catch (error) {
+      spinner.fail('Workflow configuration YAML is not valid.');
+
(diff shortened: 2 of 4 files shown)


packages/eas-cli/package.json
```diff
@@ -58,6 +58,7 @@
     "indent-string": "4.0.0",
     "jks-js": "1.1.0",
     "joi": "17.11.0",
+    "js-yaml": "^4.1.0",
```
sjchmiela, line 61: FWIW in other projects we use [`yaml`](https://www.npmjs.com/package/yaml), not sure why. I guess it has the benefit of being more actively maintained? Would you suggest we move to `js-yaml`, should this move to `yaml` or do you expect it's ok to keep two separate YAML parsers?

packages/eas-cli/src/commands/workflow/validate.ts
```diff
@@ -0,0 +1,83 @@
...
+      required: true,
+    },
+  ];
+
+  async runAsync(): Promise<void> {
+    const {
+      args: { path: filePath },
+    } = await this.parse(WorkflowValidate);
+
+    const spinner = ora().start('Validating the workflow YAML file…');
```
sjchmiela, line 25: nit: does validating take so long that it requires a spinner?

packages/eas-cli/src/commands/workflow/validate.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+async function validateYAMLAsync(filePath: string): Promise<void> {
+  await checkIfFileExistsAsync(filePath);
+  checkIfValidYAMLFileExtension(filePath);
+
+  try {
+    const fileContents = await fs.readFile(filePath, 'utf8');
+    const parsedYaml = yaml.load(fileContents);
```
sjchmiela, line 68: It would be much better if the command could also validate the schema, currently only in `www`. I think this is a sensible reason to move the workflow schema to `eas-build-job` later…

Another viable approach would be to use `createWorkflowRun` with some `dry = true` flag, or something. It would _not_ create a workflow run or a workflow revision, only parse and validate with `www`.

yarn.lock
```diff
@@ -4,40 +4,32 @@
 
 "@0no-co/graphql.web@^1.0.1":
   version "1.0.4"
-  resolved "https://registry.yarnpkg.com/@0no-co/graphql.web/-/graphql.web-1.0.4.tgz#9606eb651955499525d068ce0ad8bea596286ce2"
```
sjchmiela, line 7: Do you use NPM to install dependencies? I think we use Yarn in `eas-cli`.

sjchmiela (approved):
- package manager discrepancy to fix
- YAML parser discrepancy to comment on or fix
- spinner to reconsider?

packages/eas-cli/src/commands/workflow/validate.ts
szdziedzic: I'm not sure if it's worth complicating things, but in some other places we use a spinner if the action takes > 1s:
https://github.com/expo/eas-cli/blob/d5a9e76fd470b2a941a4ba11656ec091e7066641/packages/eas-cli/src/build/utils/repository.ts#L130-L139

packages/eas-cli/src/commands/workflow/validate.ts
```diff
@@ -0,0 +1,83 @@
+import fs from 'fs/promises';
+import yaml from 'js-yaml';
+import path from 'path';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import Log from '../../log';
+import { ora } from '../../ora';
+
+export class WorkflowValidate extends EasCommand {
+  static override description = 'validate a workflow configuration yaml file';
```
szdziedzic, line 10:
```suggestion
export class WorkflowValidate extends EasCommand {
  static override description = 'validate a workflow configuration yaml file';
  static override hidden = true;
```
Let's hide it in our autogenerated docs + help command for now, I guess it's not top secret, but if it's not ready it's better to not put it there.

packages/eas-cli/src/commands/workflow/validate.ts
szdziedzic: Yeah, I guess that schema validation would have the biggest benefit for a user here, it would catch all of the typos and so on. Let's move the schema definition to `eas-build-job` and validate against it here.

jonsamp: I'm not sure how to move this to `eas-build-job` (not familiar with where that code is or what change to make there). Could we either pair or could one of you help me?

szdziedzic: >I'm not sure how to move this to eas-build-job (not familiar with where that code is or what change to make there). Could we either pair or could one of you help me?

Don't worry about it @jonsamp, this doesn't need to be a part of this PR and we can do it because we are familiar why how things work there 🙌 

>Also, ideally we'd write a JSON schema that we can validate against, and not only check that the YAML is valid. Adding that where the workflow runs would be great, then exporting it here would be nice, to validate on the server and on the client, and have the same set of rules.

I believe this is what @sjchmiela meant, we have this JSON schema validation implemented using Zod ready and working on both worker-level and WWW-level. We need to move it to `eas-build-job` which is a common lib for such definitions, that is shared between here, WWW, and other places 👍

szdziedzic (approved):
Awesome, thanks for applying suggested changes 🚀

PR #2653, opened 2024-10-24
[eas-cli] add simplified projectId-focused context

szdziedzic:
# Why

As suggested in the base PR we need a simple context focused on only getting the project ID, without specifying the server-side environment and resolving Expo config with it.

# How

Add such context. Make the `withServerSideEnvironment` argument required for standard project config context.

# Test Plan

Tests, test manually


wschurman (approved):

wschurman (approved):
Awesome!

PR #2656, opened 2024-10-24
env:pull uses local values of encrypted variables

khamilowicz:
# Why

[ENG-13929: eas env:pull overwrites all variables](https://linear.app/expo/issue/ENG-13929/eas-envpull-overwrites-all-variables)

`env:pull` should not overwrite existing values for encrypted variables, but leave them in `.env.local`.

# Test Plan

Tested manually

### DIFF
--- a/packages/eas-cli/src/commands/env/pull.ts
+++ b/packages/eas-cli/src/commands/env/pull.ts
@@ -1,14 +1,11 @@
 import { Flags } from '@oclif/core';
+import dotenv from 'dotenv';
 import * as fs from 'fs-extra';
 import path from 'path';
 
 import EasCommand from '../../commandUtils/EasCommand';
 import { EASEnvironmentFlag, EASNonInteractiveFlag } from '../../commandUtils/flags';
-import {
-  EnvironmentSecretType,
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/commands/env/pull.ts
```diff
@@ -101,14 +112,20 @@ export default class EnvironmentVariablePull extends EasCommand {
...
     await fs.writeFile(targetPath, filePrefix + envFileContentLines.join('\n'));
 
-    const secretEnvVariables = environmentVariables.filter(
-      (variable: EnvironmentVariableFragment) => variable.value === null
-    );
-    if (secretEnvVariables.length > 0) {
+    if (overridenSecretVariables.length > 0) {
+      Log.addNewLineIfNone();
+      Log.log(
+        `Following encrypted variables were overriden by local values: ${overridenSecretVariables.join(
```
szdziedzic, line 118:
```suggestion
        `Reused local values for following secrets: ${overridenSecretVariables.join(
```

szdziedzic (approved):

PR #2669, opened 2024-10-31
[eas-cli] Add `workflow:run`

sjchmiela:
# Why

# How

# Test Plan


packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+import { WorkflowProjectSourceType } from '../../graphql/generated';
+import { WorkflowRunMutation } from '../../graphql/mutations/WorkflowRunMutation';
+import Log, { link } from '../../log';
+import { getOwnerAccountForProjectIdAsync } from '../../project/projectUtils';
+import { uploadAccountScopedEasJsonAsync } from '../../project/uploadAccountScopedEasJsonAsync';
+import { uploadAccountScopedProjectSourceAsync } from '../../project/uploadAccountScopedProjectSourceAsync';
+
+export default class WorkflowRun extends EasCommand {
+  static override description = 'Run an EAS workflow';
+  static override aliases = ['workflow:run'];
```
szdziedzic, line 16: why is it needed? I believe if a command is in the `commands/workflow` directory and is put in `run.ts` file it will automatically be `eas workflow:run` 🤔

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+  static override usage = [chalk`workflow:run {dim [options]}`];
```
szdziedzic, line 17: do we need to override the default here?

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    Log.warn('Workflows are in beta and subject to breaking changes.');
+
+    const {
+      getDynamicPrivateProjectConfigAsync,
+      loggedIn: { graphqlClient },
+      vcsClient,
+      projectDir,
+    } = await this.getContextAsync(WorkflowRun, {
+      nonInteractive: flags['non-interactive'],
```
szdziedzic, line 43: Wait, how does it even work? 😮 

It doesn't seem that this command defines any flags, so `flags['nonInteractive']` will always be `undefined`? Why it isn't a TypeError 🤯?

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+  static override description = 'Run an EAS workflow';
+  static override aliases = ['workflow:run'];
+  static override usage = [chalk`workflow:run {dim [options]}`];
+
+  // TODO(@sjchmiela): Keep command hidden until workflows are live
+  static override hidden = true;
+  static override state = 'beta';
+
+  static override args = [{ name: 'file', description: 'Path to the workflow file to run' }];
+
```
szdziedzic, line 24:
```suggestion

static override flags = {
  ...EasNonInteractiveFlag,
}

```

packages/eas-cli/src/commands/workflow/run.ts
szdziedzic: It seems like we get type hints for existing flags defined in the `flags` field, but it doesn't warn/throw an error if you access a flag that doesn't exist. It's weird. It seems dangerous and error-prone

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+      getDynamicPrivateProjectConfigAsync,
+      loggedIn: { graphqlClient },
+      vcsClient,
+      projectDir,
+    } = await this.getContextAsync(WorkflowRun, {
+      nonInteractive: flags['non-interactive'],
+      withServerSideEnvironment: null,
+    });
+
+    const yamlConfig = await fs.promises.readFile(path.join(projectDir, args.file), 'utf8');
```
szdziedzic, line 47: Should we wrap it in a try-catch with some nice error message for a user?

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,99 @@
...
+        yamlConfig,
+      },
+      workflowRunInput: {
+        projectSource: {
+          type: WorkflowProjectSourceType.Gcs,
+          projectArchiveBucketKey,
+          easJsonBucketKey,
+        },
+      },
+    });
```
szdziedzic, line 90: maybe this should be wrapped with a try-catch with a nice error message too? what do you think?

packages/eas-cli/src/utils/progress.ts
```diff
@@ -29,7 +30,7 @@ export function createProgressTracker({
   let transferredSoFar = 0;
   let current = 0;
 
-  const timerLabel = String(Date.now());
+  const timerLabel = String(randomUUID());
```
szdziedzic, line 33: I'm just curious: why is it needed?

szdziedzic (changes requested):

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Do we need this flag? I removed it on purpose, because the command isn't interactive.

packages/eas-cli/src/utils/progress.ts
sjchmiela: Sure! I was getting errors that "a timer is being ended which hasn't been started" or something like that. This fixed it. I think the two promises (for uploading eas.json and project archive) were starting and finishing too quickly and, thus, having same `Date.now()`, conflicting with each other?

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Wrapped.

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Applied your suggestion adding the flag.

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Removed.

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Removed.

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    Log.warn('Workflows are in beta and subject to breaking changes.');
+
+    const {
+      getDynamicPrivateProjectConfigAsync,
+      loggedIn: { graphqlClient },
+      vcsClient,
+      projectDir,
+    } = await this.getContextAsync(WorkflowRun, {
+      nonInteractive: flags['non-interactive'],
```
szdziedzic, line 45: After giving it some more thought, there isn't an interactive way of using this command, so it's `--non-interactive` by default, right? So I think we can skip the flag altogether?
```suggestion
      nonInteractive: true,
```

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // TODO(@sjchmiela): Keep command hidden until workflows are live
+  static override hidden = true;
+  static override state = 'beta';
+
+  static override args = [{ name: 'file', description: 'Path to the workflow file to run' }];
+
+  static override flags = {
+    ...EASNonInteractiveFlag,
+  };
```
szdziedzic, line 25:
```suggestion
```
ditto, sorry

szdziedzic (approved):

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: That's what I thought. However, I figured maybe https://github.com/expo/eas-cli/blob/e45e90a7bad8c612aea5c8186b559b13a19a6a20/packages/eas-cli/src/commands/workflow/run.ts#L44-L47 can be internally interactive? Or https://github.com/expo/eas-cli/blob/e45e90a7bad8c612aea5c8186b559b13a19a6a20/packages/eas-cli/src/commands/workflow/run.ts#L58-L62?

PR #2671, opened 2024-11-05
fix(eas-cli): update EAS Metadata with new props for App Store 3.6.0

byCedric:
# Why

This changes a few properties from `apple.advisory.*` object in the store config. It's just an update to keep up with Apple's changes on the App Store.

- `seventeenPlus` - Dropped in favor of `ageRatingOverride`
- `ageRatingOverride` - New property to override the standard age rating
- `koreaAgeRatingOverride` - Same as `ageRatingOverride`, but for South Korea

If you see this, you can upgrade to use this new format:

- `eas metadata:pull`

> Yep, that's it.

Make sure you don't have any local changes, as this command will override your local store config.

# How

- Updated to `@expo/apple-utils@2.1.0`

# Test Plan

Both command should work fine for any listed app.

- `eas metadata:pull`
- `eas metadata:push`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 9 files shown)


EvanBacon (approved):
great work here, very slick

PR #2675, opened 2024-11-06
[eas-cli] improve fingerprint UX

quinlanj:
# Why

This PR does the following:
- Add logs to notify user fingerprint is getting computed
- Lets user know they can skip fingerprint computation if it's taking longer than 5 seconds
- Skips fingerprint computation if `EAS_SKIP_AUTO_FINGERPRINT` is set, as per https://www.notion.so/expo/Soft-fingerprinting-116e5b5735248071951cddcab32ae43d?pvs=4#116e5b57352480e89189eb7feb88dacd

# How
```
Compressing project files and uploading to EAS Build. Learn more: https://expo.fyi/eas-build-archive
✔ Uploaded to EAS 
⌛️ Computing project fingerprint taking longer than expected...
⏩ To skip this step, set the environment variable: EAS_SKIP_AUTO_FINGERPRINT=1
✔ Computed project fingerprint
```

# Test Plan

- [x] Tested a fingerprint computation that 'takes too long' (await 5 second promise)
- [x] Tested a fast fingerprint computation
- [x] Tested `eas build` with and without `EAS_SKIP_AUTO_FINGERPRINT=1`

### DIFF
--- a/packages/eas-cli/src/utils/fingerprintCli.ts
+++ b/packages/eas-cli/src/utils/fingerprintCli.ts
@@ -18,6 +18,12 @@ export async function createFingerprintAsync(
   sources: object[];
   isDebugSource: boolean;
 } | null> {
+  // @expo/fingerprint is exported in the expo package for SDK 52+
+  const fingerprintPath = silentResolveFrom(projectDir, 'expo/fingerprint');
+  if (!fingerprintPath) {
+    return null;
+  }
+
   if (process.env.EAS_SKIP_AUTO_FINGERPRINT) {
     Log.log('Skipping project fingerprint');
     return null;
@@ -30,11 +36,6 @@ export async function createFingerprintAsync(
 
   const spinner = ora(`Computing project fingerprint`).start();
   try {
-    // @expo/fingerprint is exported in the expo package for SDK 52+
-    const fingerprintPath = silentResolveFrom(projectDir, 'expo/fingerprint');
-    if (!fingerprintPath) {
-      return null;
-    }
     const Fingerprint = require(fingerprintPath);
     const fingerprintOptions: Record<string, any> = {};
     if (options.platform) {
@@ -46,7 +47,6 @@ export async function createFingerprintAsync(
     if (options.debug) {
       fingerprintOptions.debug = true;
     }
-    // eslint-disable-next-line @typescript-eslint/return-await
     const fingerprint = await Fingerprint.createFingerprintAsync(projectDir, fingerprintOptions);
     spinner.succeed(`Computed project fingerprint`);
     return fingerprint;
@@ -57,5 +57,6 @@ export async function createFingerprintAsync(
   } finally {
     // Clear the timeout if the operation finishes before the time limit
     clearTimeout(timeoutId);
+    spinner.stop();
   }
 }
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/utils/fingerprintCli.ts
```diff
@@ -15,22 +18,44 @@ export async function createFingerprintAsync(
...
+    if (options.platform) {
+      fingerprintOptions.platforms = [options.platform];
+    }
+    if (options.workflow === Workflow.MANAGED) {
+      fingerprintOptions.ignorePaths = ['android/**/*', 'ios/**/*'];
+    }
+    if (options.debug) {
+      fingerprintOptions.debug = true;
+    }
+    // eslint-disable-next-line @typescript-eslint/return-await
```
szdziedzic, line 49: I think it's no longer needed?
```suggestion
```

szdziedzic (changes requested):

packages/eas-cli/src/utils/fingerprintCli.ts
```diff
@@ -15,22 +18,44 @@ export async function createFingerprintAsync(
...
+    Log.log('⏩ To skip this step, set the environment variable: EAS_SKIP_AUTO_FINGERPRINT=1');
+  }, 5000);
+
+  const spinner = ora(`Computing project fingerprint`).start();
+  try {
+    // @expo/fingerprint is exported in the expo package for SDK 52+
+    const fingerprintPath = silentResolveFrom(projectDir, 'expo/fingerprint');
+    if (!fingerprintPath) {
+      return null;
+    }
```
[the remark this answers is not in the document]
quinlanj: I put this bit of code before the spinner. I've also cleaned up the spinner in `finally` for the future in case we decide to add `return` before the spinner succeeds

packages/eas-cli/src/utils/fingerprintCli.ts
```diff
@@ -20,17 +23,40 @@ export async function createFingerprintAsync(
...
+
+  if (process.env.EAS_SKIP_AUTO_FINGERPRINT) {
+    Log.log('Skipping project fingerprint');
+    return null;
   }
-  if (options.debug) {
-    fingerprintOptions.debug = true;
+
+  const timeoutId = setTimeout(() => {
+    Log.log('⌛️ Computing project fingerprint taking longer than expected...');
```
wschurman, line 33:
```suggestion
    Log.log('⌛️ Computing the project fingerprint is taking longer than expected...');
```

packages/eas-cli/src/utils/fingerprintCli.ts
```diff
@@ -20,17 +23,40 @@ export async function createFingerprintAsync(
...
+    if (options.platform) {
+      fingerprintOptions.platforms = [options.platform];
+    }
+    if (options.workflow === Workflow.MANAGED) {
+      fingerprintOptions.ignorePaths = ['android/**/*', 'ios/**/*'];
+    }
+    if (options.debug) {
+      fingerprintOptions.debug = true;
+    }
+    const fingerprint = await Fingerprint.createFingerprintAsync(projectDir, fingerprintOptions);
```
wschurman, line 50: I forgot this when I was reviewing the other PR re: doing a function call directly vs CLI call. One additional consideration/benefit of CLI is the process/subprocess model of using the CLI via spawn instead of a direct function call. By having this be in a subprocess rather than the CLI process, we could monitor that and even kill it with a `Promise.race` if it's taking too long (I think?).

wschurman (approved):

szdziedzic (approved):

PR #2678, opened 2024-11-08
Adds workflow:create command

jonsamp:
# Why

Adds `eas workflow:create` to make it faster to create workflow files in a project. This command creates a file like **.eas/workflows/workflow.yml**. The user can choose the file name or can provide it as an arugment.

# Screenshots

Screenshot 2024-11-08 at 8 50 58 AM
Screenshot 2024-11-08 at 8 51 27 AM
Screenshot 2024-11-08 at 8 52 04 AM

# Test Plan

Make sure that you can run the command as pictured above and that it correctly creates the workflow file. Also make sure that the unit tests pass.

# Future

Once we have docs for workflows that are public, I'd like to put a link to the docs in the default file this command creates. For now, I am just putting Hello World as a reasonable placeholder.


packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+    branches: ['*']
+
+jobs:
+  Hello World:
+    steps:
+      - run: echo "Hello, World"
+`;
+
+export class WorkflowCreate extends EasCommand {
+  static override description = 'create a new workflow configuration YAML file';
```
szdziedzic, line 24: Let's make it hidden until officially released

```suggestion
export class WorkflowCreate extends EasCommand {
  static override hidden = true;
  static override description = 'create a new workflow configuration YAML file';
```

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+      throw error;
+    }
+  }
+
+  private validateYamlExtension(fileName: string): void {
+    const fileExtension = path.extname(fileName).toLowerCase();
+    if (fileExtension !== '.yml' && fileExtension !== '.yaml') {
+      throw new Error('File must have a .yml or .yaml extension');
+    }
+  }
```
szdziedzic, line 80: I believe we are doing the same thing in `workflow:validate`:
https://github.com/expo/eas-cli/blob/fdb167df344516d376b395242e5300b64fff8d5c/packages/eas-cli/src/commands/workflow/validate.ts#L55-L61

Let's extract it to the reusable helper function vs class method.

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
+import chalk from 'chalk';
+import fs from 'fs/promises';
+import path from 'path';
+import prompts from 'prompts';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import Log from '../../log';
+
+const WORKFLOWS_DIR = '.eas/workflows';
```
szdziedzic, line 9: Do we want to make it work on Windows as well? Our general policy is that Windows is not officially supported and we recommend running things through WSL on Windows, but when possible we try to make it work (although we don't test against it)

I'm not sure if it will work on Windows, I believe it's safer to do:

```suggestion
const WORKFLOWS_DIR = path.join('.eas', 'workflows');
```

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private async createWorkflowFileAsync(fileName: string): Promise<void> {
+    this.validateYamlExtension(fileName);
+
+    const filePath = path.join(WORKFLOWS_DIR, fileName);
+
+    try {
+      await fs.access(filePath);
+      throw new Error(`Workflow file already exists: ${filePath}`);
```
szdziedzic, line 98: Do we want to prompt a user if they want to replace it with a new file?

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+    } catch (error) {
+      if ((error as any)?.code === 'ENOENT') {
+        await fs.writeFile(filePath, HELLO_WORLD_TEMPLATE);
+      } else {
+        throw error;
+      }
+    }
```
szdziedzic, line 105: nit: I believe that if we use `pathExists` from `fs-extra` here we don't need to worry about error handling, it's just `true` or `false`.

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+        process.exit(0);
+      }
+
+      fileName = response.fileName;
+    }
+
+    try {
+      await this.ensureWorkflowsDirectoryExistsAsync();
+      await this.createWorkflowFileAsync(fileName);
+      Log.withTick(`Created ${chalk.bold(path.join(WORKFLOWS_DIR, fileName))}`);
```
szdziedzic, line 68: nit: To be more consistent we can move this log to `createWorkflowFileAsync` (we log workflows directory creation in `ensureWorkflowsDirectoryExistsAsync`).

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -0,0 +1,107 @@
...
+  static override description = 'create a new workflow configuration YAML file';
+
+  static override args = [
+    {
+      name: 'name',
+      description: 'Name of the workflow file (must end with .yml or .yaml)',
+      required: false,
+    },
+  ];
+
```
szdziedzic, line 33:
```suggestion
  static override contextDefinition = {
    ...this.ContextOptions.ProjectDir,
  }
```
Let's add this context here. It will ensure that this command can be only run inside of a project directory vs some other random place. I believe we also need to use this context in the `workflow:validate` command.

szdziedzic (changes requested):

sjchmiela (approved):

PR #2687, opened 2024-11-13
[eas-cli] create fingerprint on each update

quinlanj:
# Why

This PR computes a fingerprint for each update. 
- If the update does not use the fingerprint policy, it calculates the fingerprint using the exported cli in the `expo` package for projects SDK 52+. 
- If the update uses the fingerprint policy, it just uses the fingerprint that was previously calculated using `expo-updates`.  

# How

- Modified fingerprint computation to handle platform-specific cases
- Updated the fingerprint data structure to use `fingerprintHash` instead of the full fingerprint object
- Added support for computing fingerprints for projects without expo-updates
- Introduced new types and interfaces for handling fingerprint information groups
- Created utility functions for async map operations and fingerprint calculations
- Updated GraphQL schema to support new fingerprint-related fields

# Test Plan

- [x] on build: verify fingerprints till work 
- [x] on update: verify fingerprint is calculated for new project with SDK 52
- [x] on update: verify no fingerprint is calculated for old project (< SDK 52)
- [x] on update: verify fingerprint is reused for project with fingerprint runtime policy

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -50,7 +50,7 @@ import { truthy } from '../utils/expodash/filter';
 import groupBy from '../utils/expodash/groupBy';
 import mapMapAsync from '../utils/expodash/mapMapAsync';
 import uniqBy from '../utils/expodash/uniqBy';
-import { createFingerprintsByKeyAsync } from '../utils/fingerprintCli';
+import { FingerprintOptions, createFingerprintsByKeyAsync } from '../utils/fingerprintCli';
 import { Client } from '../vcs/vcs';
 
 // update publish does not currently support web
@@ -908,12 +908,12 @@ export async function maybeCalculateFingerprintForRuntimeVersionInfoObjectsWitho
     runtimeToPlatformsAndFingerprintInfoAndFingerprintSourceMapping.filter(
       infoGroup => !infoGroup.fingerprintHash
     );
-  const fingerprintOptionsByRuntimeAndPlatform = new Map();
+  const fingerprintOptionsByRuntimeAndPlatform = new Map<string, FingerprintOptions>();
   for (const infoGroup of runtimesToComputeFingerprintsFor) {
     for (const platform of infoGroup.platforms) {
       const runtimeAndPlatform = `${infoGroup.runtimeVersion}-${platform}`;
       const options = {
-        platform,
+        platforms: [platform],
         workflow: workflowsByPlatform[platform],
         projectDir,
         env,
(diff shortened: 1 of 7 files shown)


quinlanj:
* **#2687** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @quinlanj and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/eas-cli/src/project/publish.ts
```diff
@@ -858,9 +873,119 @@ export function getRuntimeToPlatformsAndFingerprintInfoMappingFromRuntimeVersion
...
+        fingerprintHash:
+          runtimeVersionInfoObjects.map(
+            runtimeVersionInfoObject => runtimeVersionInfoObject.runtimeVersionInfo.fingerprintHash
+          )[0] ?? null,
+      };
+    }
+  );
+}
+
+export async function maybeCalculateFingerprintForRuntimeVersionInfoObjectsWithoutExpoUpdatesAsync({
```
Kudo, line 885: nit: the logic of this function is kind of complicated. would be good to split as smaller functions and add some unit test.
but fine if that requires too much effort.

Kudo (approved):

wschurman (changes requested):
Think we also need to add this to republish. Probably just pull in the fingerprint from the previously-published update assuming we are sure it won't change between publish and republish (don't think we have another option since we can't re-calculate the fingerprint on republish since republish can't use any info from the current project state, code signing suffers from this fwiw)

wschurman:
(if you want to do republish in a follow-up, request re-review on this one. this looks good otherwise)

wschurman:
Also, should roll-back-to-embedded updates have this as well? (every command that calls `publishUpdateGroupAsync` should have it I think?)

quinlanj:
@wschurman here is the PR for republishing (made separate because of extra dependencies on www changes): https://github.com/expo/eas-cli/pull/2708

I dont think we need to pass on fingerprint info for rollback to embedded. A 'rollback to embedded' directive would always be compatible with whatever build that is targeted. Also, it would technically have the fingerprint of the receiving build (ideally all builds of the same runtime have the same fingerprint, but that's not necessarily the case)

wschurman (approved):
> I dont think we need to pass on fingerprint info for rollback to embedded.

My guess is that it might be simpler product-wise to include them since they are just normal UpdateEntities (as are rollouts and republishes). Otherwise we'd need to special case the UI to explain why they don't have them for those updates?

> A 'rollback to embedded' directive would always be compatible with whatever build that is targeted. Also, it would technically have the fingerprint of the receiving build (ideally all builds of the same runtime have the same fingerprint, but that's not necessarily the case)

I believe they're only compatible with builds with the same code signing key (and therefore theoretically runtime version as we advise people to change when they do a key change). I think since the code signing configuration is part of the fingerprint we should probably attach the fingerprint to directives.

PR #2697, opened 2024-11-15
[ENG-14204][eas-cli] UX improvements for env:list command

szdziedzic:
# Why

Currently using the `env:list` command is confusing and annoying with this multi-select for environments and too noisy when an env var is secret or sensitive

# How

Make it less noisy.

Just use select for environment vs multi-select.

# Test Plan

Tested it manually


radoslawkrzemien (approved):
LGTM 🚀

PR #2713, opened 2024-11-20
All account-wide envvars should be global (for now)

khamilowicz:
# Why

[ENG-14274: Remove linking from eas-cli](https://linear.app/expo/issue/ENG-14274/remove-linking-from-eas-cli)

Initial release will not include linking account-wide variables to specific projects - they all should be global.
Without explicitly setting `isGlobal` flag, account-wide envvars would be created as non-global, without possibility to link it to anything.

# How

* Added `isGlobal` = true to requests
* Also, removed unnecessary check in non-interactive mode - user should be able to create an account-wide variable with environments without linking it to the current project.

# Test Plan

Updated tests

### DIFF
--- a/packages/eas-cli/src/graphql/mutations/EnvironmentVariableMutation.ts
+++ b/packages/eas-cli/src/graphql/mutations/EnvironmentVariableMutation.ts
@@ -24,6 +24,7 @@ type CreateVariableArgs = {
   visibility: EnvironmentVariableVisibility;
   environments: EnvironmentVariableEnvironment[];
   type: EnvironmentSecretType;
+  isGlobal?: boolean;
   fileName?: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/graphql/mutations/EnvironmentVariableMutation.ts
```diff
@@ -24,6 +24,7 @@ type CreateVariableArgs = {
   visibility: EnvironmentVariableVisibility;
   environments: EnvironmentVariableEnvironment[];
   type: EnvironmentSecretType;
+  isGlobal?: boolean;
```
szdziedzic, line 27: shouldn't you regenrate gql schema?

szdziedzic (changes requested):

packages/eas-cli/src/graphql/mutations/EnvironmentVariableMutation.ts
khamilowicz: I don't think so, `isGlobal` was added (but not used here) long time ago. It is defined in `CreateSharedEnvironmentVariableInput` and in `graphql.schema.json`

szdziedzic (approved):

PR #2748, opened 2024-12-09
[eas-cli] add support for broadcast push notifications option for push notifications capability

szdziedzic:
# Why

Companion to https://github.com/expo/expo/pull/32623 and https://github.com/expo/third-party/pull/113

https://exponent-internal.slack.com/archives/C02123T524U/p1728981742266059

We want to handle new broadcast push notifications option for push notifications capability

It's a bit atypical because there's no way to enable it by using entitlement (this is how it works for the rest of the capabilities using options). I talked about it with @brentvatne and we decided to add a new field to the app config to manage this option `ios.usesBroadcastPushNotifications`.

# How

If `usesBroadcastPushNotifications` is `true` and push notification entitlement is set use the correct option for the push notification capability in Apple's API.

Make sure that we update the capability at the correct times:
- when the entitlement is present but the capability is disabled - enable it (with to without a new option, depending on whether it is set)
- when the entitlement is missing but the capability is enabled (with or without broadcast) - disable it
- when the entitlement is present, the capability is enabled, no options are set on the remote side, and `usesBroadcastPushNotifications` is true locally - add a new option for capability
- when the entitlement is present, the capability is enabled, the broadcast option is set on the remote side, and `usesBroadcastPushNotifications` is false - remove the option
-  when the entitlement is present, the capability is enabled, and broadcast is defined both locally and remotely - do nothing
- when the entitlement is present, the capability is enabled, broadcast is not used both locally and remotely - do nothing

https://github.com/expo/third-party/pull/113 is a part of this change and needs to be merged first + new version of a package needs to be published - for local testing I used linked version

# Test Plan

Test each of these cases manually in EAS CLI on the test project and confirm that the correct capabilities are turned on using Apple's developer portal

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
@@ -6,6 +6,7 @@ import {
   CapabilityType,
   CapabilityTypeDataProtectionOption,
   CapabilityTypeOption,
+  CapabilityTypePushNotificationsOption,
   CloudContainer,
   MerchantId,
 } from '@expo/apple-utils';
@@ -19,7 +20,10 @@ export const EXPO_NO_CAPABILITY_SYNC = getenv.boolish('EXPO_NO_CAPABILITY_SYNC',
 
 type GetOptionsMethod<T extends CapabilityType = any> = (
   entitlement: JSONValue,
-  entitlementsJson: JSONObject
+  entitlementsJson: JSONObject,
+  additionalOptions: {
+    usesBroadcastPushNotifications?: boolean;
+  }
 ) => CapabilityOptionMap[T];
 
 const validateBooleanOptions = (options: any): boolean => {
@@ -78,7 +82,10 @@ const getDefinedOptions: GetOptionsMethod = entitlement => {
  */
 export async function syncCapabilitiesForEntitlementsAsync(
   bundleId: BundleId,
-  entitlements: JSONObject = {}
+  entitlements: JSONObject = {},
+  additionalOptions: {
+    usesBroadcastPushNotifications: boolean;
+  }
 ): Promise<{ enabled: string[]; disabled: string[] }> {
   if (EXPO_NO_CAPABILITY_SYNC) {
     return { enabled: [], disabled: [] };
@@ -91,7 +98,8 @@ export async function syncCapabilitiesForEntitlementsAsync(
 
   const { enabledCapabilityNames, request, remainingCapabilities } = getCapabilitiesToEnable(
     currentCapabilities,
-    entitlements
+    entitlements,
+    additionalOptions
   );
 
   const { disabledCapabilityNames, request: modifiedRequest } = getCapabilitiesToDisable(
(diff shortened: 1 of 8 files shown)


packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -488,7 +515,15 @@ export const CapabilityMapping: CapabilityClassifier[] = [
     entitlement: 'aps-environment',
     capability: CapabilityType.PUSH_NOTIFICATIONS,
     validateOptions: validateDevProdString,
-    getOptions: getDefinedOptions,
+    getOptions(entitlement, _entitlementsJson, { usesBroadcastPushNotifications }) {
+      Log.log('usesBroadcastPushNotifications', usesBroadcastPushNotifications);
```
szdziedzic, line 519:
```suggestion
```

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -488,7 +515,15 @@ export const CapabilityMapping: CapabilityClassifier[] = [
...
+      const option = entitlement ? CapabilityTypeOption.ON : CapabilityTypeOption.OFF;
+      if (option === CapabilityTypeOption.ON && usesBroadcastPushNotifications) {
+        Log.log('Using broadcast push notifications');
```
szdziedzic, line 522:
```suggestion
```

packages/eas-cli/src/credentials/ios/appstore/bundleIdCapabilities.ts
```diff
@@ -156,7 +167,23 @@ function getCapabilitiesToEnable(
...
+    // For push notifications, we should always update the capabaility if
+    // - settings are not defined in the existing capability, but usesBroadcastPushNotifications is enabled (we want to add settings for this capability)
+    // - settings are defined in the existing capability, but usesBroadcastPushNotifications is disabled (we want to remove settings for this capability)
+    const isPushNotificationsCapability =
+      staticCapabilityInfo.capability === CapabilityType.PUSH_NOTIFICATIONS;
+    const noSettingsAttributes = existing?.attributes.settings == null;
+    const hasBroadcastPushNotificationsDisabledBothLocalAndRemote =
+      !additionalOptions.usesBroadcastPushNotifications && noSettingsAttributes;
+    const hasBroadcastPushNotificationsEnabledBothLocalAndRemote =
+      additionalOptions.usesBroadcastPushNotifications && !noSettingsAttributes;
```
khamilowicz, line 179: IMO this logic should be in a separate function, it only handles one specific case and makes this code less readable.

khamilowicz, line 179: Can we replace it with `const isPushNotificationSameLocalAndRemove = additionalOptions.usesBroadcastPushNotifications === noSettings`? Do you think it would be clearer?

khamilowicz (approved):
Looks good, just some nits.

PR #2761, opened 2024-12-11
feat(eas-cli): Add `--dry-run` flag to `deploy` command

kitten:
# Why

This outputs a tarball to `deploy.tar.gz` if the `--dry-run` flag is passed, skipping all other work.

# How

- Added flag
- Add early abort with `fs.promises.copyFile` run

# Test Plan


kadikraman (approved):
Nice! This will be handy. Tested locally, works great 👍

Issue #2753, opened 2024-12-10
Equivalent functionality of secret:push causing not able to load env variable that uses custom env files


djaffer:
### Summary

The recent changes has been a big disappointment. Broke the code and having to spend hours to get environment variables fix. Why not just let user push variable to the eas including secrets. If the need is that critical have a flag that allows to keep using old. Expo/EAS upgrade is the most dreading part of developer experience. Please improve it.

### Managed or bare?

Managed

### Environment

expo-env-info 1.2.1 environment info:
    System:
      OS: Windows 11
    Binaries:
      Node: 20.17.0 - C:\Program Files\nodejs\node.EXE
      Yarn: 1.22.22 - ~\AppData\Roaming\npm\yarn.CMD
      npm: 10.8.1 - C:\Program Files\nodejs\npm.CMD
    IDEs:
      Android Studio: AI-231.9392.1.2311.11076708
    npmPackages:
      @expo/webpack-config: ^19.0.1 => 19.0.1
      expo: ^51.0.36 => 51.0.39
      react: 18.3.1 => 18.3.1
      react-dom: 18.3.1 => 18.3.1
      react-native: 0.76.5 => 0.76.5
      react-native-web: ~0.19.13 => 0.19.13

### Error output

You cannot change a secret variable to a non-secret variable.
Request ID: 
    Error: GraphQL request failed.

### Reproducible demo or steps to reproduce from a blank project

You cannot change a secret variable to a non-secret variable.
Request ID
    Error: GraphQL request failed.

felipesilva-plank:
any updates on this? #2765 is not merged yet

PR #2765, opened 2024-12-12
Support pushing existing secret and sensitive variables

khamilowicz:
# Why

[ENG-14486: Support updating secret variables with `eas env:push`](https://linear.app/expo/issue/ENG-14486/support-updating-secret-vars-with-eas-envpush)

https://github.com/expo/eas-cli/issues/2753

Currently, we don't support adding secrets using the `eas env:push` command. However, this behavior conflicts with the legacy `eas secrets:push` command, forcing some users to either convert their secrets to public variables or prefix them with `EXPO_SENSITIVE`. This pull request addresses this issue by enabling users to update existing secret or sensitive variables without any additional work.

Additionally, the `--force` option has been added to skip prompts.

Also, an unintended behavior has been fixed: pushing variables would overwrite the existing `environments` field of eas variables instead of appending to it. If a user has defined a variable in multiple environments and then pushes it to just one, the variable will only be linked to that specific environment.

# How

* A boolean `force` flag has been added.
* If a variable from the `.env` file is already present in `eas`, its `visibility` and `environments` are preserved.

# Test Plan

Tested manually


khamilowicz:
/changelog-entry bug-fix Support pushing existing secret and sensitive environment variables

giorgiofellipe:
It would be great to have it!

PR #2771, opened 2024-12-13
[eas-cli] show deployment errors

kadikraman:
# Why

The error messages from `403` upload error were not being shown in the terminal.

# How

Extract the error message from `403` errors and show to the user when applicable.

# Test Plan

Tested locally by uploading a build with 100+ assets.

<details>
<summary>
Before / After
</summary>

### Before
Screenshot 2024-12-13 at 11 47 08

### After
Screenshot 2024-12-13 at 15 37 55
</details>

### DIFF
--- a/packages/eas-cli/src/worker/upload.ts
+++ b/packages/eas-cli/src/worker/upload.ts
@@ -109,20 +109,25 @@ export async function uploadAsync(params: UploadParams): Promise<UploadResult> {
         return retry(error);
       }
 
+      const defaultErrorMessage = `Upload of "${filePath}" failed: ${response.statusText}`;
+
       if (
         response.status === 408 ||
         response.status === 409 ||
         response.status === 429 ||
         (response.status >= 500 && response.status <= 599)
       ) {
-        const message = `Upload of "${filePath}" failed: ${response.statusText}`;
         const text = await response.text().catch(() => null);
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/worker/upload.ts
```diff
@@ -109,20 +109,25 @@ export async function uploadAsync(params: UploadParams): Promise<UploadResult> {
...
         (response.status >= 500 && response.status <= 599)
       ) {
-        const message = `Upload of "${filePath}" failed: ${response.statusText}`;
         const text = await response.text().catch(() => null);
-        return retry(new Error(text ? `${message}\n${text}` : message));
+        return retry(new Error(text ? `${defaultErrorMessage}\n${text}` : defaultErrorMessage));
+      } else if (response.status === 403) {
+        const body = await response.json();
+        const message = body?.error ?? defaultErrorMessage;
+        throw new Error(message);
```
kitten, line 125: Should we do this on every error maybe for non-403s? I think it'd also be neat to make sure we handle failure causes of `response.json()`

kadikraman: So, check if `response.json().message` exists and show it instead of the default if it does?

kitten (approved):

PR #2786, opened 2024-12-20
[eas-cli] Upload `package.json` alongside `eas.json` when creating a workflow run

sjchmiela:
# Why

Thanks to @szdziedzic's changes to `www`, if we send a `package.json` along with `eas.json` when setting up a workflow run, EAS will pick the right images based on SDK and RN versions as `eas build` would.

# How

Added uploading `package.json` after uploading `eas.json`, added the new argument to the mutation.

# Test Plan

Tested manually.

Zrzut ekranu 2024-12-20 o 22 27 20
Zrzut ekranu 2024-12-20 o 22 27 58


szdziedzic (approved):
awesome, thanks! 🙏

PR #2800, opened 2025-01-06
[eas-cli] Silence all non-command output in non-interactive mode of eas env:exec

wschurman:
# Why

When using this programmatically to run a command, `env:exec` shouldn't modify the stdout of the sub-command. This makes it possible to run a JSON command, for example.

Concretely, this came up in https://github.com/expo/expo-github-action/pull/316/files#diff-9a4c8863ad4d04f246a265d420c404b948ffcd8902e7de5312c9205cd034fa2eR226 where we are trying to run `npx expo-updates fingerprint:generate` which outputs JSON.

# How

One solution (this solution) is to slightly overload the `non-interactive` flag to also silence all non-command output. Alternatively we could add a separate flag.

I chose to overload the existing flag since I couldn't think of a reason that a non-interactive scenario would want the `Log.log` calls. That being said, I'm open to adding a separate flag like `--raw` or `--passthrough` or something else if overloading the flag is not desired.

# Test Plan

Before:
```
$ neas env:exec preview "npx expo config --json --type public" --non-interactive
No environment variables with visibility "Plain text" and "Sensitive" found for the "preview" environment on EAS servers.

Running command: npx expo config --json --type public
[stdout] {"name":"pkfsa...
```

After:
```
$ neas env:exec preview "npx expo config --json --type public" --non-interactive
{"name":"pkfsa...
```

Other tests:

Does not print hello since subcommand had nonzero exit status:
```
$ neas env:exec preview "npx expo config --json --type publicd" --non-interactive && echo "hello"
```
Ensure output (no output) matches:
```
npx expo config --json --type publicd && echo "hello"
```

Prints hello after config (correct success exit status):
```
$ neas env:exec preview "npx expo config --json --type public" --non-interactive && echo "hello"
{"name":"pkfsa...
hello
```
Ensure output (no output) matches:
```
npx expo config --json --type public && echo "hello"
```

### DIFF
--- a/packages/eas-cli/src/commands/env/exec.ts
+++ b/packages/eas-cli/src/commands/env/exec.ts
@@ -53,9 +53,10 @@ export default class EnvExec extends EasCommand {
     },
   ];
 
+  private isNonInteractive: boolean = false;
+
   async runAsync(): Promise<void> {
     const { flags, args } = await this.parse(EnvExec);
-
     const parsedFlags = this.sanitizeFlagsAndArgs(flags, args);
 
     const {
@@ -65,6 +66,8 @@ export default class EnvExec extends EasCommand {
       nonInteractive: parsedFlags.nonInteractive,
     });
 
+    this.isNonInteractive = parsedFlags.nonInteractive;
+
     const environment =
       parsedFlags.environment ??
       (await promptVariableEnvironmentAsync({ nonInteractive: parsedFlags.nonInteractive }));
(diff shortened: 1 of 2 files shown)


wschurman:
* **#2800** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/env/exec.ts
```diff
@@ -103,6 +111,29 @@ export default class EnvExec extends EasCommand {
...
+    environmentVariables: Record<string, string>;
+  }): Promise<void> {
+    try {
+      const { stdout } = await spawnAsync('bash', ['-c', command], {
+        env: {
+          ...process.env,
+          ...environmentVariables,
+        },
+      });
+      Log.log(stdout);
```
szdziedzic, line 128: I believe that this will wait until command execution is completed and then log the whole command output all at once. I believe it's better to print logs as they appear (as we do in `--non-interactive` mode) to indicate to the user that the command execution is not stuck.

szdziedzic (changes requested):
Looks like a good idea overall, thanks 🙏 

I have one suggestion:

szdziedzic (approved):
Looks reasonable to me, thanks! 🙏

PR #2820, opened 2025-01-16
[ENG-14625][eas-cli] add `eas build:dev` command

szdziedzic:
# Why

https://exponent-internal.slack.com/archives/C1QLM27T9/p1734729481930599?thread_ts=1734708689.473259&cid=C1QLM27T9

Add `eas dev` command that:
1. calculates fingerprint
2. checks if we have simulator/emulator dev client builds with such a fingerprint hash on our servers
a. if we do -  it downloads and launches them on the emulator/simulator using the same logic we use for eas build:run
b. if not -  such a build is created and launched afterward 

# How

Do exactly what I mentioned above. Use as much code and functions from build logic and eas build:run that already exist to make it easier to maintain.

# Test Plan

Test manually

https://exponent-internal.slack.com/archives/C1QLM27T9/p1736987685783579?thread_ts=1734708689.473259&cid=C1QLM27T9

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -99,18 +99,29 @@ export interface BuildFlags {
   repack: boolean;
 }
 
-export async function runBuildAndSubmitAsync(
-  graphqlClient: ExpoGraphqlClient,
-  analytics: Analytics,
-  vcsClient: Client,
-  projectDir: string,
-  flags: BuildFlags,
-  actor: Actor,
-  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn,
-  buildProfilesOverride?: ProfileData<'build'>[],
-  downloadSimBuildAutoConfirm?: boolean,
-  envOverride?: Env
-): Promise<{
+export async function runBuildAndSubmitAsync({
+  graphqlClient,
+  analytics,
+  vcsClient,
+  projectDir,
+  flags,
+  actor,
+  getDynamicPrivateProjectConfigAsync,
+  buildProfilesOverride,
+  downloadSimBuildAutoConfirm,
+  envOverride,
+}: {
+  graphqlClient: ExpoGraphqlClient;
+  analytics: Analytics;
+  vcsClient: Client;
+  projectDir: string;
+  flags: BuildFlags;
+  actor: Actor;
+  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn;
+  buildProfilesOverride?: ProfileData<'build'>[];
+  downloadSimBuildAutoConfirm?: boolean;
+  envOverride?: Env;
+}): Promise<{
   buildIds: string[];
 }> {
   await vcsClient.ensureRepoExistsAsync();
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -106,7 +106,10 @@ export async function runBuildAndSubmitAsync(
   projectDir: string,
   flags: BuildFlags,
   actor: Actor,
-  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn
+  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn,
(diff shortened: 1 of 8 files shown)


szdziedzic:
* **#2820** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

brentvatne:
my main concern here is implementing this as a top level command, which feels like something we'd be committing to - and we may want to use the `eas dev` command for something else, that maybe does more than what we're doing here (perhaps in addition to this functionality). what if this was `eas build:run --dev` for now?

szdziedzic:
>what if this was eas build:run --dev for now? if that was how to access this functionality i'd be good with rolling it out right away

What do you think about `eas build:dev`?

szdziedzic:
>what if this was eas build:run --dev for now? if that was how to access this functionality i'd be good with rolling it out right away

@brentvatne What do you think about `eas build:dev`?

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -123,13 +126,15 @@ export async function runBuildAndSubmitAsync(
...
   const platforms = toPlatforms(flags.requestedPlatform);
-  const buildProfiles = await getProfilesAsync({
-    type: 'build',
-    easJsonAccessor,
-    platforms,
-    profileName: flags.profile ?? undefined,
-    projectDir,
-  });
+  const buildProfiles =
+    buildProfilesOverride ??
```
wschurman, line 130: Just curious (I'm not super familiar with build profiles), from a UX perspective should we require them to specify/choose an existing dev-client-enabled profile instead of creating a virtual profile for them? If not, are there any side effects of creating this "virtual" profile for them?

packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -106,7 +106,10 @@ export async function runBuildAndSubmitAsync(
   projectDir: string,
   flags: BuildFlags,
   actor: Actor,
-  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn
+  getDynamicPrivateProjectConfigAsync: DynamicConfigContextFn,
+  buildProfilesOverride?: ProfileData<'build'>[],
+  downloadSimBuildAutoConfirm?: boolean,
```
wschurman, line 111: nit: prefer to refactor the params here to be destructured object (can use vs code refactor thing) now that we're adding a boolean param. This way we get a named boolean param at the callsite.

wschurman (commented):
Few minor comments and one question about the build profile. Otherwise looks good to me. Will delegate to a reviewer more familiar with builds for approval though.

packages/eas-cli/src/commands/build/dev.ts
```diff
@@ -0,0 +1,158 @@
...
+    Log.newLine();
+
+    const builds = await BuildQuery.viewBuildsOnAppAsync(graphqlClient, {
+      appId: projectId,
+      filter: {
+        platform: toAppPlatform(platform),
+        fingerprintHash: fingerprint.hash,
+        status: BuildStatus.Finished,
+        simulator: platform === Platform.IOS ? true : undefined,
+        distribution: DistributionType.Internal,
```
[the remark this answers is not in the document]
szdziedzic: I don't think so, I believe `DistributionType.Simulator` is a legacy way of marking build as `simulator`.

cc: @radoslawkrzemien

packages/eas-cli/src/build/runBuildAndSubmit.ts
szdziedzic: Hmm, I think what you suggest here might be a better option 🤔 The Current implementation should "magically" work well out of the box for users, but I think it limits the customization of things (like selecting the `bun` version). I believe that using explicit profiles can also have the benefit of people having a good idea about what settings we apply under the hood for these builds 🤔 What do you think?

PR #2821, opened 2025-01-16
[eas-cli] release fingerprint:compare

quinlanj:
# Why

This PR releases `eas fingerprint:compare` from the hidden state. It also adds support for the following scenarios:

- `eas fingerprint:compare [HASH1]` : If provided alone, HASH1 is compared against the current project's fingerprint.
- `eas fingerprint:compare [HASH1] [HASH2]`: If two hashes are provided, HASH1 is compared against HASH2.

More details: https://www.notion.so/expo/Soft-fingerprinting-116e5b5735248071951cddcab32ae43d

# How

- Added support for comparing fingerprints using hashes via new command arguments `hash1` and `hash2`
- Improved platform detection from fingerprint sources
- Enhanced workflow handling for multi-platform projects
- Updated the GraphQL schema to use a dedicated Fingerprint type
- Refactored fingerprint comparison logic into separate functions for better maintainability
- Added proper error handling for missing or invalid fingerprints

# Test Plan

Test the following scenarios:
- Compare current project with a build, `EXPO_STAGING=1 ~/Documents/eas-cli/packages/eas-cli/bin/run fingerprint:compare --build-id  1aa82a8d-6fed-4c4c-ab93-ef07d0c7ca97`
-  Compare two specific fingerprint hashes `EXPO_STAGING=1 ~/Documents/eas-cli/packages/eas-cli/bin/run fingerprint:compare f0d6a916e73f401d428e6e006e07b12453317ba2 c71a7d475aa6f75291bc93cd74aef395c3c94eee`
- Compare a specific hash with current project `EXPO_STAGING=1 ~/Documents/eas-cli/packages/eas-cli/bin/run fingerprint:compare c71a7d475aa6f75291bc93cd74aef395c3c94eee`


quinlanj:
* **#2821** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2821?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2821?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/fingerprint/compare.ts
```diff
@@ -170,6 +174,186 @@ export default class FingerprintCompare extends EasCommand {
...
+    if (nonInteractive) {
+      throw new Error('Build ID must be provided in non-interactive mode');
+    }
+
+    const displayName = await getDisplayNameForProjectIdAsync(graphqlClient, projectId);
+    buildId = await selectBuildToCompareAsync(graphqlClient, projectId, displayName, {
+      filters: { hasFingerprint: true },
+    });
+    if (!buildId) {
+      throw new Error(); // exit, explanation already printed in prompt
```
wschurman, line 211:
```suggestion
      throw new Error('Must select build with fingerprint for comparison.');
```
so the output will have both:
```
No fingerprints have been computed for builds of project ${projectDisplayName}.
Must select build with fingerprint for comparison.
```

(this matches the exit pattern for similar cases in other commands I believe)

packages/eas-cli/src/commands/fingerprint/compare.ts
```diff
@@ -170,6 +174,186 @@ export default class FingerprintCompare extends EasCommand {
...
+
+  // Generate ignorePaths for mixed workflows
+  const ignorePaths = platforms
+    .filter(platform => workflowsByPlatform[appPlatformToPlatform(platform)] === Workflow.MANAGED)
+    .map(platform => `${appPlatformToString(platform)}/**/*`);
+
+  return { ignorePaths };
+}
+
+function inferPlatformsFromSource(fingerprint: Fingerprint): AppPlatform[] {
```
wschurman, line 339: talked about this offline, but I think if we can avoid this by limiting args or splitting out commands that'd be preferable. The main issue is that (I believe) it's possible to create a fingerprint for either platform or both that doesn't have either string being checked for (by using a specially-crafted .fingerprintIgnore or sourceSkips).

packages/eas-cli/src/commands/fingerprint/compare.ts
```diff
@@ -21,10 +22,33 @@ import { createFingerprintAsync, diffFingerprint } from '../../utils/fingerprint
...
+  nonInteractive: boolean;
+  json: boolean;
+}
 
 export default class FingerprintCompare extends EasCommand {
   static override description = 'compare fingerprints of the current project, builds and updates';
-  static override hidden = true;
+  static override strict = false;
+
+  static override args = [
```
wschurman, line 39: while the command does the correct thing to error when an invalid combination of args and flags is passed in, I think it might be a bit tough to explain the different combinations that are allowed and what each combination does, and thus potentially cause a UX issue. See my note about maybe splitting them up into different commands based on use case so we can better require and explain different combinations of args.

wschurman (changes requested):
Talked offline, but the summary of changes:

To avoid the platform inference, either:
a) if need to calculate fingerprint locally and need a platform, look for a linked build to get the platform from either from build-id flag or from builds linked to that fingerprint. If none is linked, throw error.
b) if need to calculate fingerprint locally, require build-id flag. get platform from build.
c) split this into two commands. I prefer this since the human-readable description of the allowed arg/flag combinations is simpler (see inline comment), but am happy to be overridden if others prefer something else:
    1. `fingerprint:diff hash1 hash2` - diffs two fingerprints by id/hash. no build/update arg for this one. instead it just simply diffs.
    2. `fingerprint:diff-local-directory-with-remote --build/--update` - diffs local with fingerprint used by a build or an update. this satisfies the requirement of the spec in notion. doesn't take a hash. command name should be something else, but I called it `diff-local-directory-with-remote` to demonstrate intent.
d) maybe get rid of local-remote compatibility in its own command altogether and instead just include it in the build info (build:info, build:list)? maybe it's too slow for this. could be behind a flag on those commands, i.e. `eas build:info <id> --compare-to-local-fingerprint` returns `{id: ...., isLocalFingerprintCompatible: bool}`. I guess maybe not this option since the diff would be helpful.

Couple additional questions:
1. What is the behavior when a debug fingerprint is compared to a non-debug fingerprint?

packages/eas-cli/src/commands/fingerprint/compare.ts
quinlanj: I chose to keep everything in one command because imo it's easier to use that way. I added examples in the CLI help section to clear things up

```
compare fingerprints of the current project, builds and updates

USAGE
  $ eas fingerprint:compare [HASH1] [HASH2] [--build-id <value>] [--json --non-interactive]

ARGUMENTS
  HASH1  If provided alone, HASH1 is compared against the current project's fingerprint.
  HASH2  If two hashes are provided, HASH1 is compared against HASH2.

FLAGS
  --build-id=<value>  Compare the fingerprint with the build with the specified ID
  --json              Enable JSON output, non-JSON messages will be printed to stderr.
  --non-interactive   Run the command in non-interactive mode.

DESCRIPTION
  compare fingerprints of the current project, builds and updates

EXAMPLES
  $ eas fingerprint:compare 	 # Compare fingerprints in interactive mode

  $ eas fingerprint:compare c71a7d475aa6f75291bc93cd74aef395c3c94eee 	 # Compare fingerprint against local directory

  $ eas fingerprint:compare c71a7d475aa6f75291bc93cd74aef395c3c94eee f0d6a916e73f401d428e6e006e07b12453317ba2 	 # Compare provided fingerprints

  $ eas fingerprint:compare --build-id 82bc6456-611a-48cb-8db4-5f9eb2ca1003 	 # Compare fingerprint from build against local directory
  ```

File: packages/eas-cli/src/commands/fingerprint/compare.ts [hunk 4]
```diff
@@ -54,70 +103,52 @@ export default class FingerprintCompare extends EasCommand {
...
+          `🔄 ${capitalizeFirstLetter(
+            prettyPrintFingerprint(firstFingerprint, firstFingerprintOrigin)
+          )} differs from ${prettyPrintFingerprint(secondFingerprint, secondFingerprintOrigin)}`
+        );
+      }
     } else {
-      Log.log(`🔄 Project fingerprint differs from build`);
+      printJsonOnlyOutput({ fingerprint1: firstFingerprint, fingerprint2: secondFingerprint });
+      return;
     }
```
wschurman, line 149: prefer guard clauses to reduce nested conditionals:

```suggestion
    if (json) {
      printJsonOnlyOutput({ fingerprint1: firstFingerprint, fingerprint2: secondFingerprint });
      return;
    }

    if (firstFingerprint.hash === secondFingerprint.hash) {
      Log.log(
        `✅ ${capitalizeFirstLetter(
          prettyPrintFingerprint(firstFingerprint, firstFingerprintOrigin)
        )} matches fingerprint from ${prettyPrintFingerprint(
          secondFingerprint,
          secondFingerprintOrigin
        )}`
      );
      return;
    }

    Log.log(
      `🔄 ${capitalizeFirstLetter(
        prettyPrintFingerprint(firstFingerprint, firstFingerprintOrigin)
      )} differs from ${prettyPrintFingerprint(secondFingerprint, secondFingerprintOrigin)}`
    );
```

wschurman (approved):
One small change, otherwise lgtm. TIL about the `examples` thing, very cool.

PR #2824, opened 2025-01-16
[eas-cli] Use `deploy` as top level hosting command

kadikraman:
# Why

Currently we have `worker` as the top level command for hosting, with `deploy` as an alias.

Based on actual usage, we should make `deploy` as the top level command. This has the added benefit of the subcommands being listed correctly in the `--help` command.

# How

- Renamed `worker` -> `worker:deploy` and `worker:alias` as aliases for backwards compatibility
- Small wording fixes

# Test Plan

| Before      | After |
| ----------- | ----------- |
| Screenshot 2025-01-16 at 14 26 47      | Screenshot 2025-01-16 at 14 26 34       |
| Screenshot 2025-01-16 at 14 27 53        | Screenshot 2025-01-16 at 14 27 41  |


packages/eas-cli/src/commands/deploy/alias.ts
```diff
@@ -34,8 +34,8 @@ interface RawDeployAliasFlags {
 
 export default class WorkerAlias extends EasCommand {
   static override description = 'Assign deployment aliases';
-  static override aliases = ['deploy:alias', 'deploy:promote'];
-  static override state = 'beta';
+  static override aliases = ['worker:alias', 'deploy:promote'];
```
kadikraman, line 37: Do we need to keep `worker:alias` around? I'd prefer to remove it, but we should technically keep it for backwards compatibility.

kitten: I'd love to keep this. It doesn't really disrupt anything/anyone as opposed to potentially causing confusion if it goes missing. We can always remove it in a future major version

kitten (approved):

packages/eas-cli/src/commands/deploy/index.ts
```diff
@@ -65,10 +65,9 @@ interface DeployInProgressParams {
 }
 
 export default class WorkerDeploy extends EasCommand {
-  static override description = 'Deploy your Expo web build';
-  static override aliases = ['deploy'];
```
kitten, line 69: We're missing a `worker:deploy` alias here, right? That one's maybe less important, but for completeness I'd add it.

sjchmiela (approved):

PR #2828, opened 2025-01-20
[eas-cli] [ENG-14633] Start autoIncremented builds at 1

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-14633/start-autoincrement-counter-at-0-so-first-build-has-buildnumber-==-1

# How

When versionCode/buildNumber is initialised ignore autoIncrement and use the initial version

# Test Plan

Added tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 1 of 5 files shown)


sjchmiela (approved):
Judging by surrounding code it looks to me like it makes sense, but I'm not the most familiar with this part of code so I'm defering final review to others.

szdziedzic (approved):
looks correct

PR #2829, opened 2025-01-20
[eas-cli] Detect valid JSON files during hosting upload

kadikraman:
# Why

The apple app site association file (needed for Universal linking on iOS) does not have a file extension, as it's simply served from `/.well-known/apple-app-site-association`.

However in order for it to be valid, the file need to return `Content-Type: application/json` header.

# How

When content type cannot be detected form the mime type, try parsing the file to check if it's a valid JSON and set the header accordingly.

# Test Plan

Deploy a website with an `apple-app-site-association` file and verify that the `Content-Type` header is set to `application/json`.

### DIFF
--- a/packages/eas-cli/src/worker/upload.ts
+++ b/packages/eas-cli/src/worker/upload.ts
@@ -29,6 +29,24 @@ const isCompressible = (contentType: string | null, size: number): boolean => {
   }
 };
 
+const getContentTypeAsync = async (filePath: string): Promise<string | null> => {
+  let contentType = mime.getType(path.basename(filePath));
+
+  if (!contentType) {
+    const fileContent = await readFile(filePath, 'utf-8');
+    try {
+      // check if file is valid JSON without an extension, e.g. for the apple app site association file
+      const parsedData = JSON.parse(fileContent);
+
+      if (parsedData) {
+        contentType = 'application/json';
+      }
+    } catch {}
+  }
+
+  return contentType;
+};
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/worker/upload.ts
```diff
@@ -77,7 +78,19 @@ export async function uploadAsync(params: UploadParams): Promise<UploadResult> {
...
+  if (!contentType) {
+    const fileContent = await readFile(filePath, 'utf-8');
+    try {
+      // check if file is valid JSON without an extension, e.g. for the apple app site association file
+      const parsedData = JSON.parse(fileContent);
+      if (parsedData) {
+        contentType = 'application/json';
+      }
+    } catch {}
+  }
```
kitten, line 92: nit: let's extract this function, so we can prepare for it to be moved into `assets.ts` into an earlier phase.

We could also do a quick check such as `/^\s*[{[]/` to determine whether it's worth calling `JSON.parse`, in case the JSON asset is quite large.

kitten (approved):

packages/eas-cli/src/worker/upload.ts
kadikraman: As discussed, the pre-check is probably not necessary as `JSON.parse` should fail quite quickly for non-JSON files.

Moved the util to a separate function 👍

kitten (approved):

PR #2830, opened 2025-01-21
[eas-cli] [ENG-14629] Improve AppleID validation

radoslawkrzemien:
# Why

https://linear.app/expo/issue/ENG-14629/improve-apple-id-input-validation

- user was asked to log into their Apple account
- prompt displayed for the username, pre-filled with value from cache
- value from cache contains invalid, unprintable characters (`\u0017`)
- user accepts "correct" username
- fails to log in
- user removes the cache file
- logs in successfully when manually typing "the same" username (but without invalid characters)

# How

The username value from prompt is cleared of any invalid unprintable characters before updating cache and using the value to log into Apple account

# Test Plan

Added tests for `resolveUserCredentialsAsync` function
Tested the behaviour for a username with reported invalid characters

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/resolveCredentials.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/resolveCredentials.ts
@@ -208,13 +208,14 @@ async function promptUsernameAsync(): Promise<string> {
   // the default value for quicker authentication.
   const lastAppleId = await getCachedUsernameAsync();
 
-  const { username } = await promptAsync({
+  let { username } = await promptAsync({
     type: 'text',
     name: 'username',
     message: `Apple ID:`,
     validate: (val: string) => val !== '',
     initial: lastAppleId ?? undefined,
   });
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/credentials/ios/appstore/resolveCredentials.ts
```diff
@@ -208,13 +208,14 @@ async function promptUsernameAsync(): Promise<string> {
...
 
-  const { username } = await promptAsync({
+  let { username } = await promptAsync({
     type: 'text',
     name: 'username',
     message: `Apple ID:`,
     validate: (val: string) => val !== '',
     initial: lastAppleId ?? undefined,
   });
+  username = username.replace(/[\x00-\x1F]/gi, '');
```
szdziedzic, line 218: can you add a comment here about historical context and what intention stands behind this regex?

szdziedzic (approved):

PR #2839, opened 2025-01-27
feat: add internal TestFlight group to submit

EvanBacon:
# Why

- During `eas submit -p ios`, automatically create an internal TestFlight group `Internal (Expo)` and add all App Managers to it.
- Assuming that non-exempt encryption is set before the build—`expo.ios.infoPlist.ITSAppUsesNonExemptEncryption` in the app.json—then all app admins will have instant access to new builds from TestFlight.
- This greatly streamlines the time-to-testing for developers.

---

- Should we make this feature opt-in? I feel like it's good enough to just have on by default.
- Should there be comprehensive eas.json config settings for configuring the group and adding other groups? Maybe that's out of scope for this PR?
- External groups require manual assignment which would need to take place after a build and not before it. For this, we'd likely want to run the API calls from the server.

# Test Plan

- In a project with `expo.ios.infoPlist.ITSAppUsesNonExemptEncryption` set:
- `eas build --auto-submit -p ios` -> Group exists in ASC -> App can be installed from TestFlight on iOS for app owners.
- I observed a server issue where you had to wait ~25s after creating the app to create a TestFlight group, this can only be observed by changing the bundle identifier and submitting for the first time (my account has a lot of apps).

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/ensureTestFlightGroup.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/ensureTestFlightGroup.ts
@@ -1,0 +1,255 @@
+import { App, BetaGroup, Session, User, UserRole } from '@expo/apple-utils';
+
+import Log from '../../../log';
+import { ora } from '../../../ora';
+import { confirmAsync } from '../../../prompts';
+
+const AUTO_GROUP_NAME = 'Team (Expo)';
+
+export async function ensureTestFlightGroupExistsAsync(app: App): Promise<void> {
+  // Ensure the app has an internal TestFlight group with access to all builds and app managers added.
+  // TODO: Should we expose this feature somehow?
+  const max = false;
+
+  const group = await ensureInternalGroupAsync(app);
+
+  const admins = await User.getAsync(
+    app.context,
+    !max
+      ? undefined
+      : // Querying all visible apps for all users can be expensive so we only do it when there's a chance we may need to reconcile.
+        {
+          query: {
+            includes: ['visibleApps'],
+          },
+        }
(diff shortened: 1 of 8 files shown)


packages/eas-cli/src/credentials/ios/appstore/ensureTestFlightGroup.ts
```diff
@@ -0,0 +1,201 @@
...
+      /** 'ENTITY_ERROR.ATTRIBUTE.INVALID.INVALID_CHARACTERS' */
+      code: string;
+      /** 'An attribute value has invalid characters.' */
+      title: string;
+      /** 'App Name contains certain Unicode symbols, emoticons, diacritics, special characters, or private use characters that are not permitted.' */
+      detail: string;
+    }[];
+  };
+} {
+  return 'data' in error && 'errors' in error.data && Array.isArray(error.data.errors);
```
szdziedzic, line 175: we can use `zod` or `joi` to validate this schema if we want to

szdziedzic (approved):
>Should we make this feature opt-in?

I think we can let it be on by default, but maybe we want to add an env var to disable it, so make it opt-out?

>Should there be comprehensive eas.json config settings for configuring the group and adding other groups? Maybe that's out of the scope for this PR?

It sounds like a good option, although what you built here can be just a default behavior when nothing else is specified and we can add a config later.

>External groups require manual assignment which would need to take place after a build and not before it. For this, we'd likely want to run the API calls from the server.

I talked about sth like this last week with @jonsamp and we believe the workflows can be handy here. We can just add one more job to the workflow that will do it 😄

PR #2841, opened 2025-01-27
[eas-cli] Use Git to pack Git projects

sjchmiela:
# Why

Currently, when `eas-cli` create a project tarball for EAS Build or Workflows, unless the user has `requireCommit: true` in `eas.json`, the project is being packed with a plain `makeShallowCopyAsync` which copies all the files that are deemed ignored by `ignore` package and `.easignore` or `.gitignore` files.

This is suboptimal for `eas workflow:run` experience — a developer with a project with Git expects the repository to exist on EAS.

# How

This was a big pain.

Previously, we had `GitClient` which required commit and its subclass, `GitNoCommitClient` which overrode some functions to allow uncommitted changes. On top of that, `GitClient` copied project with `git clone`, while `GitNoCommitClient` copied it with `makeShallowCopyAsync`. This meant `GitClient`'s copy would include `.git`, whereas `GitNoCommitClient` would not. Moreover, thus, `GitClient` would adhere to `.gitignore` and `GitNoCommitClient` would _not_ — only to `.easignore`.

This pull request:
- merges the two Git clients into one with `requireCommit` option
- this causes `.git` directory to exist in copies with `requireCommit = false`
- changes how we log changes caused by Git case sensitivity change (we need to handle the case where there is a preexisting dirty tree)
- after we clone the repository, we delete files that should have been ignored according to `.easignore`
- afterwards, we make a shallow copy of all the files over to the clone. This copies live changes to the working tree over to the clean clone. (They may only exist in `requireCommit = false` situation.)
- adds better support for `.easignore` to `isFileIgnoredAsync`.

As far as I can see the behavior looks like this:

Bez tytułu-2025-01-11-0010

The breaking change in here is that we're going to start adhering to `.easignore` even if someone has `requireCommit = true`.

# Test Plan

I used `easd build:inspect` a bunch to verify it works like I expect. I encourage fellow reviewers to do the same.


packages/eas-cli/src/vcs/clients/git.ts
```diff
@@ -386,3 +433,51 @@ async function setGitCaseSensitivityAsync(
...
+  await setGitCaseSensitivityAsync(true, cwd);
+  try {
+    const uncommitedChangesAfterCaseSensitivityChange = await gitStatusAsync({
+      showUntracked: true,
+      cwd,
+    });
+
+    if (
+      uncommitedChangesAfterCaseSensitivityChange !== uncommittedChangesBeforeCaseSensitivityChange
+    ) {
```
szdziedzic, line 458: what is the format of these 2? is there a chance that git will return files in different order causing us to enter this if statement even if nothing changed? maybe we should operate on sets here?

packages/eas-cli/src/vcs/clients/git.ts
```diff
@@ -386,3 +433,51 @@ async function setGitCaseSensitivityAsync(
...
+
+async function assertEnablingGitCaseSensitivityDoesNotCauseNewUncommittedChangesAsync(
+  cwd: string
+): Promise<void> {
+  // Remember uncommited changes before case sensitivity change
+  // for later comparison so we log to the user only the files
+  // that were marked as changed after the case sensitivity change.
+  const uncommittedChangesBeforeCaseSensitivityChange = await gitStatusAsync({
+    showUntracked: true,
+    cwd,
```
szdziedzic, line 445: we are using `rootPath` as cwd for it here, but previously we used `this.maybeCwdOverride`

does this change matter?

radoslawkrzemien (approved):
I *think* it looks good, but I'm not too familiar with this part of the codebase

packages/eas-cli/src/vcs/clients/git.ts
sjchmiela: `rootPath` is supposed to be root of the Git repository and should take `maybeCwdOverride` into consideration:

Zrzut ekranu 2025-01-30 o 20 01 44

szdziedzic (approved):
looks reasonable

PR #2842, opened 2025-01-28
feat(eas-cli): Use a fallback generated name during EAS Submit

EvanBacon:
# Why

If the name you selected was taken, it's a bit of a hassle to pick a new one. We had a fallback error message that said to use `--app-name` but this isn't a real flag afaict, and the fallback was never called!

Screenshot 2025-01-27 at 4 56 31 PM

# How

This change catches the duplicate name error and warns that a generated name will be used instead. This name can be instantly changed from the App Store Connect website later so it seems like a better default to ensure complex commands like `eas build --auto-submit` are less prone to failure.

# Test Plan

- Added some unit tests to emulate the responses of invalid fixtures.

- Create a project with the name `expo`
- Run `eas submit`
- Observe a warning log like `App name "expo" is already taken. Using generated name "expo (6eb7d6)" which can be changed later from https://appstoreconnect.apple.com.`
- Submit works as expected.
- Name can be manually changed from the website.

Same process but with `expo🚀` will warn:
- `App name "expo🚀" contains invalid characters. Using sanitized name "expo-" which can be changed later from https://appstoreconnect.apple.com.`

### DIFF
--- a/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
+++ b/packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
@@ -1,6 +1,7 @@
-import { App, BundleId } from '@expo/apple-utils';
+import { App, BundleId, RequestContext } from '@expo/apple-utils';
 import { JSONObject } from '@expo/json-file';
 import chalk from 'chalk';
+import { randomBytes } from 'node:crypto';
 
 import { getRequestContext, isUserAuthCtx } from './authenticate';
 import { AuthCtx, UserAuthCtx } from './authenticateTypes';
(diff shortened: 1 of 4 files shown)


packages/eas-cli/src/credentials/ios/appstore/ensureAppExists.ts
```diff
@@ -215,3 +216,120 @@ export async function ensureAppExistsAsync(
...
+      /** 'ENTITY_ERROR.ATTRIBUTE.INVALID.INVALID_CHARACTERS' */
+      code: string;
+      /** 'An attribute value has invalid characters.' */
+      title: string;
+      /** 'App Name contains certain Unicode symbols, emoticons, diacritics, special characters, or private use characters that are not permitted.' */
+      detail: string;
+    }[];
+  };
+} {
+  return 'data' in error && 'errors' in error.data && Array.isArray(error.data.errors);
```
szdziedzic, line 334: We can try to validate this data against a scheme by using `zod` or `joi` instead of writing if statements manually

szdziedzic (approved):
awesome

PR #2843, opened 2025-01-28
feat(eas-cli): Ensure non-exempt encryption status

EvanBacon:
# Why

Apple requires the non-exempt encryption status be defined on every build before it can be submitted to TestFlight or app review. We can use the Info.plist flag to only ask users one time and reuse that answer for every build. This is a better version of https://github.com/expo/expo/pull/34487

Prompting in the CLI saves users six clicks in App Store Connect for every iOS build.

# How

I copied the bundle identifier prompting logic to a new function for setting non-exempt encryption, made it more optional, and added it to `eas build -p ios`.

If the user indicates that they use non-exempt encryption then we won't write anything this will enable them to see the in-depth warnings in ASC about this option. They can still skip the prompt by manually adding the value to their app.json.

# Test Plan

- Run `eas build -p ios` -> prompted "app uses standard encryption?" -> enter / Y -> app builds and doesn't ask again.
- Unit tests for all conditions.

### DIFF
--- a/packages/eas-cli/src/project/ios/exemptEncryption.ts
+++ b/packages/eas-cli/src/project/ios/exemptEncryption.ts
@@ -1,10 +1,4 @@
-import {
-  ExpoConfig,
-  getConfigFilePaths,
-  getProjectConfigDescription,
-  modifyConfigAsync,
-} from '@expo/config';
-import assert from 'assert';
+import { ExpoConfig, getProjectConfigDescription, modifyConfigAsync } from '@expo/config';
 import chalk from 'chalk';
 
 import Log, { learnMore } from '../../log';
@@ -43,24 +37,12 @@ async function configureNonExemptEncryptionAsync({
   exp: ExpoConfig;
   nonInteractive: boolean;
 }): Promise<void> {
-  const paths = getConfigFilePaths(projectDir);
   const description = getProjectConfigDescription(projectDir);
   if (nonInteractive) {
     Log.warn(
       chalk`${description} is missing {bold ios.infoPlist.ITSAppUsesNonExemptEncryption} boolean. Manual configuration is required in App Store Connect before the app can be tested.`
     );
   }
-
-  if (paths.dynamicConfigPath) {
-    Log.warn(
-      chalk`${description} is missing {bold ios.infoPlist.ITSAppUsesNonExemptEncryption} boolean and it cannot be updated automatically. You will need to configure non-exempt encryption status in App Store Connect for this build. {dim Learn more: ${learnMore(
-        'https://developer.apple.com/documentation/Security/complying-with-encryption-export-regulations'
-      )}}`
-    );
-    return;
-  }
-
-  assert(paths.staticConfigPath, 'app.json must exist');
 
   let onlyExemptEncryption = await confirmAsync({
     message: `iOS app only uses standard/exempt encryption? ${chalk.dim(
@@ -126,7 +108,7 @@ async function configureNonExemptEncryptionAsync({
       },
     };
 
-    Log.log(chalk.cyan(`Add the following to your Expo config`));
+    Log.log(chalk.cyan(`Add the following to ${description}:`));
     Log.log();
     Log.log(JSON.stringify(edits, null, 2));
     Log.log();
--- a/packages/eas-cli/src/project/ios/exemptEncryption.ts
+++ b/packages/eas-cli/src/project/ios/exemptEncryption.ts
@@ -1,0 +1,176 @@
+import { ExpoConfig, getConfigFilePaths } from '@expo/config';
+import { IOSConfig } from '@expo/config-plugins';
+import { getInfoPlistPathFromPbxproj } from '@expo/config-plugins/build/ios/utils/getInfoPlistPath';
+import { Platform, Workflow } from '@expo/eas-build-job';
+import plist from '@expo/plist';
+import assert from 'assert';
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import path from 'node:path';
+
+import { readAppJson } from '../../build/utils/appJson';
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/project/ios/exemptEncryption.ts
```diff
@@ -0,0 +1,196 @@
...
+  vcsClient,
+  nonInteractive,
+}: {
+  projectDir: string;
+  exp: ExpoConfig;
+  vcsClient: Client;
+  nonInteractive: boolean;
+}): Promise<void> {
+  const workflow = await resolveWorkflowAsync(projectDir, Platform.IOS, vcsClient);
+  assert(workflow === Workflow.MANAGED, 'This function should be called only for managed projects');
```
szdziedzic, line 31: we only call it for managed workflow
```ts
if (ctx.workflow === Workflow.MANAGED) {
    ...
    await ensureNonExemptEncryptionIsDefinedForManagedProjectAsync(ctx);
  }
```
so I guess we can skip it to save some execution milliseconds 🤔?

packages/eas-cli/src/project/ios/exemptEncryption.ts
```diff
@@ -0,0 +1,196 @@
...
+
+export async function getNonExemptEncryptionAsync(
+  projectDir: string,
+  exp: ExpoConfig,
+  vcsClient: Client,
+  xcodeContext?: { targetName?: string; buildConfiguration?: string }
+): Promise<void> {
+  const workflow = await resolveWorkflowAsync(projectDir, Platform.IOS, vcsClient);
+
+  if (workflow === Workflow.GENERIC) {
```
szdziedzic, line 58: wait, don't we run this whole exempt encryption logic only for managed workflow?

packages/eas-cli/src/project/ios/exemptEncryption.ts
```diff
@@ -0,0 +1,196 @@
...
+    const nonExemptEncryption = infoPlist['ITSAppUsesNonExemptEncryption'];
+    const buildConfigurationDesc =
+      xcodeContext?.targetName && xcodeContext?.buildConfiguration
+        ? ` (target = ${xcodeContext.targetName}, build configuration = ${xcodeContext.buildConfiguration})`
+        : '';
+    assert(
+      nonExemptEncryption !== undefined,
+      `Could not read non-exempt encryption setting from Xcode project${buildConfigurationDesc}.`
+    );
+    return nonExemptEncryption;
```
szdziedzic, line 88: if we run it only for managed does it mean we should delete it? or maybe we should do it for generic as well?

I guess that we should do it for generic as well, right?

packages/eas-cli/src/project/ios/exemptEncryption.ts
```diff
@@ -0,0 +1,196 @@
...
+        'https://developer.apple.com/documentation/Security/complying-with-encryption-export-regulations'
+      )
+    )}`
+  );
+
+  let { onlyExemptEncryption } = await promptAsync({
+    name: 'onlyExemptEncryption',
+    type: 'confirm',
+    message: `iOS app only uses standard/exempt encryption? `,
+    // message: `Does your app use non-exempt encryption?`,
```
szdziedzic, line 143: we probably want to delete it?
```suggestion
```

szdziedzic (commented):
awesome! thanks for working on it 🙏 

few comments:

brentvatne:
i approve of the approach, will leave the actual code review up to @szdziedzic

RRaideRR:
Hey @EvanBacon ,

just to let you know: I'm having a dynamic app.config.ts. 

When I try to run the build, I get the following error:

```
eas build --platform ios --profile production
Environment variables with visibility "Plain text" and "Sensitive" loaded from the "production" environment on EAS: APP_VARIANT, EXPO_PUBLIC_API_URL, GOOGLE_LOGIN_WEB_CLIENT_ID.

✔ iOS app only uses standard/exempt encryption? Learn more … yes
Cannot read properties of undefined (reading 'ITSAppUsesNonExemptEncryption')
    Error: build command failed.
```

The solution is of course to add to my app.config.ts:

```
      infoPlist: {
        ITSAppUsesNonExemptEncryption: false,
      },
```


However: I'm not really happy that the build process just errors and does not give clear instructions to the user what to do.

SzymonGonet:
Hi, I'm encountering the error "Cannot read properties of undefined (reading 'ITSAppUsesNonExemptEncryption')".
My app.json has the following configuration for iOS:

  "ios": {
    "config": {
      "usesNonExemptEncryption": false
    },
    "infoPlist": {
      "ITSAppUsesNonExemptEncryption": false
    }
  }
When I run the build with:

eas build --platform ios --profile preview
I'm prompted with:

iOS app only uses standard/exempt encryption? Learn more … (Y/n)

After confirming, I receive the error:

Cannot read properties of undefined (reading 'ITSAppUsesNonExemptEncryption')
Error: build command failed.
Do you have any suggestions on what might be causing this issue or how I can resolve it?

SzymonGonet:
Hi, I'm encountering the error "Cannot read properties of undefined (reading 'ITSAppUsesNonExemptEncryption')". 
My `app.json` has the following configuration for iOS:

```json
  "ios": {
    "config": {
      "usesNonExemptEncryption": false
    },
    "infoPlist": {
      "ITSAppUsesNonExemptEncryption": false
    }
  }
```

When I run the build with:

```bash
eas build --platform ios --profile preview
```

I'm prompted with:

> iOS app only uses standard/exempt encryption? Learn more … (Y/n)

After confirming, I receive the error:

```bash
Cannot read properties of undefined (reading 'ITSAppUsesNonExemptEncryption')
Error: build command failed.
```

Do you have any suggestions on what might be causing this issue or how I can resolve it?

PR #2877, opened 2025-02-05
[eas-cli] Prevent using non-initialized `Ignore`

sjchmiela:
# Why

`eas-cli@15.0.0-15.0.3` doesn't interpret `.easignore` right because I haven't initialized the `Ignore` class properly.

# How

Make it so it's not possible to use `Ignore` without initializing it.

# Test Plan

Tests should pass.


szdziedzic (approved):
awesome

PR #2878, opened 2025-02-05
[eas-cli] Raise threshold for "big tarball" warning

sjchmiela:
# Why

Adding `.git` directory to the tarball makes it bigger. Users notice this warning and want it to be actionable, but currently adding `.git` to `.easignore` does nothing.

# How

For now, raise threshold for warning from 100 MiB to 150 MiB. In the next pull request I'm going to fix `.git` exclusion via `.easignore`.

# Test Plan

CI should pass.

### DIFF
--- a/packages/eas-cli/src/project/uploadAccountScopedProjectSourceAsync.ts
+++ b/packages/eas-cli/src/project/uploadAccountScopedProjectSourceAsync.ts
@@ -1,7 +1,10 @@
 import chalk from 'chalk';
 import fs from 'node:fs';
 
-import { makeProjectTarballAsync } from '../build/utils/repository';
+import {
+  makeProjectTarballAsync,
+  maybeWarnAboutProjectTarballSize,
+} from '../build/utils/repository';
 import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
 import { AccountUploadSessionType } from '../graphql/generated';
 import Log, { learnMore } from '../log';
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/project/uploadAccountScopedProjectSourceAsync.ts
```diff
@@ -36,15 +39,7 @@ export async function uploadAccountScopedProjectSourceAsync({
...
-          projectTarball.size
-        )}. You can reduce its size and the time it takes to upload by excluding files that are unnecessary for the build process in ${chalk.bold(
-          '.easignore'
-        )} file. ${learnMore('https://expo.fyi/eas-build-archive')}`
-      );
-    }
+    maybeWarnAboutProjectTarballSize(projectTarball.size);
 
     if (projectTarball.size > 2 * 1024 * 1024 * 1024) {
       throw new Error('Project archive is too big. Maximum allowed size is 2GB.');
```
szdziedzic, line 45: do we want to move it as well to separate function?

szdziedzic (approved):

packages/eas-cli/src/project/uploadAccountScopedProjectSourceAsync.ts
sjchmiela: 👍 
https://github.com/expo/eas-cli/blob/242f949236145b40477bba9a51ad0b9a331ffd34/packages/eas-cli/src/build/utils/repository.ts#L186-L198

PR #2879, opened 2025-02-05
[eas-cli] Support excluding `.git` via `.easignore`

sjchmiela:
# Why

Currently it's not possible to exclude `.git` from being uploaded in project tarball.

# How

In `makeShallowCopyAsync()` we first clone the repository, then we delete files ignored by `.easignore` and then we copy working directory over.

The "delete files ignored by `.easignore`" step is governed by `git ls-files --exclude-from=.easignore`. `git ls-files` is never going to consider `.git` directory to be excluded, even if specified in `.easignore`, so we need to special-case the `.git` directory and remove it manually if specified.

# Test Plan

Added `.git` to a test `.easignore`, ran `eas build:inspect` and confirmed the copied directory did _not_ contain `.git` folder.


szdziedzic (approved):
awesome

Issue #2883, opened 2025-02-05
V15 breaks EAS build (Path is a directory: rm returned EISDIR (is a directory))


benjaminkomen:
### Summary

All our builds fail since version 15.0.0 of eas-cli was released.

### Managed or bare?

managed

### Environment

```
 expo-env-info 1.2.2 environment info:
    System:
      OS: macOS 15.3
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.18.2 - ~/.nvm/versions/node/v20.18.2/bin/node
      Yarn: 1.22.22 - /opt/homebrew/bin/yarn
      npm: 10.9.2 - /opt/homebrew/bin/npm
      Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 24.2, iOS 18.2, macOS 15.2, tvOS 18.2, visionOS 2.2, watchOS 11.2
      Android SDK:
        API Levels: 30, 31, 33, 34, 35
        Build Tools: 30.0.3, 33.0.1, 34.0.0, 35.0.0
        System Images: android-34 | Google APIs ARM 64 v8a, android-34 | Google Play ARM 64 v8a, android-35 | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2024.2 AI-242.23726.103.2422.12816248
      Xcode: 16.2/16C5032a - /usr/bin/xcodebuild
    npmPackages:
      expo: ~52.0.7 => 52.0.7
      react: 18.3.1 => 18.3.1
      react-dom: 18.3.1 => 18.3.1
      react-native: 0.76.2 => 0.76.2
      react-native-web: ~0.19.13 => 0.19.13
    npmGlobalPackages:
      eas-cli: 14.7.1
      expo-cli: 6.3.7
    Expo Workflow: managed
```

### Reproducible demo or steps to reproduce from a blank project

Our GitHub Actions workflow that creates EAS builds is failing since yesterday, after version 15.0.0 was published:

```
Run expo/expo-github-action/preview@v8
  with:
    command: eas build --profile preview --platform all --non-interactive --no-wait
    comment: true
    working-directory: /home/runner/work/[REDACTED]
    github-token: ***
  env:
    EXPO_TOKEN: ***
    output_type: merge
    COMMIT_NUMBER: 101[REDACTED]
[...]

PR #2884, opened 2025-02-05
[eas-cli] Fix deleting ignored submodules when archiving project

sjchmiela:
# Why

`git ls-files` returns submodules as a single filepath entry, so `rm` must remove it with `-rf`. Fixes https://github.com/expo/eas-cli/issues/2883.

# How

Added `{ recursive: true, force: true }` to `fs.rm`.

# Test Plan

I tried `easd build:inspect -p android -s archive -o ~/testtest --force` on a repo with a submodule in `.easignore`. It succeeded and the submodule was not included in the folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


benjaminkomen:
I'm not sure if it would be easy to add a test to prevent regression of this scenario? Maybe a future task.

szdziedzic (approved):

PR #2885, opened 2025-02-05
[eas-cli] Skip `makeShallowCopyAsync` if `requireCommit`

sjchmiela:
# Why

`requireCommit: true` setups expect worktree to be clean (e.g. encrypted files to be locked).

# How

Made it so if `requireCommit` is true we don't copy files over.

# Test Plan

I have enabled `git-crypt` in my test repository, added a secret file, ran `easd build:inspect -p android -s archive -o ~/test-production --force`. The clone did contain secret file in plaintext.

Then I enabled `requireCommit: true` and repeated the command. The clone did contain the file encrypted.

### DIFF
--- a/packages/eas-cli/src/vcs/clients/git.ts
+++ b/packages/eas-cli/src/vcs/clients/git.ts
@@ -223,10 +223,15 @@ export default class GitClient extends Client {
       await setGitCaseSensitivityAsync(isCaseSensitive, rootPath);
     }
 
-    // After we create the shallow Git copy, we copy the files
-    // again. This way we include the changed and untracked files
-    // (`git clone` only copies the committed changes).
-    await makeShallowCopyAsync(rootPath, destinationPath);
+    if (!this.requireCommit) {
+      // After we create the shallow Git copy, we copy the files
+      // again. This way we include the changed and untracked files
+      // (`git clone` only copies the committed changes).
+      //
+      // We only do this if `requireCommit` is false because `requireCommit: true`
+      // setups expect no changes in files (e.g. locked files should remain locked).
+      await makeShallowCopyAsync(rootPath, destinationPath);
+    }
   }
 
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/vcs/clients/git.ts
```diff
@@ -223,10 +223,15 @@ export default class GitClient extends Client {
...
-    // (`git clone` only copies the committed changes).
-    await makeShallowCopyAsync(rootPath, destinationPath);
+    if (!this.requireCommit) {
+      // After we create the shallow Git copy, we copy the files
+      // again. This way we include the changed and untracked files
+      // (`git clone` only copies the committed changes).
+      //
+      // We only do this if `requireCommit` is false because `requireCommit: true`
+      // setups expect no changes in files (e.g. locked files should remain locked).
+      await makeShallowCopyAsync(rootPath, destinationPath);
```
szdziedzic, line 233: I'm not sure if I understand. Why should we not do `git clone` when `requireCommit: true` is present?

sjchmiela: It comes down to secrets and locked / unlocked repository (if using `git-crypt`).

Looks like `requireCommit` flag wasn't being used just as a "it's a good practice to always commit your changes", but more like "make tarball look exactly like a cloned repository at this commit would". In the former understanding it's not a big deal if we copy unlocked files to the tarball because the user _did_ make a commit. In the latter understanding `requireCommit` is more like `makeTarballResembleClone` or something.

See discussion in [here](https://exponent-internal.slack.com/archives/C017N0N99RA/p1738752553121949) and how it broke Brent's setup with encrypted files (he was using `requireCommit: true` and was expecting encrypted files to be locked and worktree clean).

szdziedzic (approved):

PR #2892, opened 2025-02-07
[eas-cli] Allow normalizing root path to fix Windows not ignoring files

sjchmiela:
# Why

Since v15 files that should be ignored get added to the tarball on Windows.

# How

After adding some debugging I was able to find that `src`, `srcFilePath` don't match on path separators which breaks `ignore.ignores`.

Zrzut ekranu 2025-02-7 o 21 57 02

Added `path.normalize` if `__NORMALIZE` is `1`.

---

Also noticed we used wrong `cwd` for is-tracked check.

# Test Plan

Going to release this and test.


brentvatne:
is the `__NORMALIZE` env var something special on windows or is this something you're exposing to test the change?

sjchmiela:
Something I'm exposing to test the change

PR #2904, opened 2025-02-11
[eas-cli] Use Git to resolve root path in no-VCS mode

sjchmiela:
# Why

I was thinking we can offer `EAS_NO_VCS=1` as a simple escape hatch for people that are getting weird Git-related issues. One of them may be an old version of Git that does not support our `git clone` properly.

# How

In `getRootPath` of no-VCS mode, return first of those values (in order):
- `EAS_PROJECT_ROOT`
- Git root path
- `process.cwd()`

# Test Plan

Added a basic test.

### DIFF
--- a/packages/eas-cli/src/vcs/clients/noVcs.ts
+++ b/packages/eas-cli/src/vcs/clients/noVcs.ts
@@ -1,18 +1,53 @@
-import { Ignore, getRootPath, makeShallowCopyAsync } from '../local';
+import spawnAsync from '@expo/spawn-async';
+import path from 'path';
+
+import Log from '../../log';
+import { Ignore, makeShallowCopyAsync } from '../local';
 import { Client } from '../vcs';
 
 export default class NoVcsClient extends Client {
+  private readonly cwdOverride?: string;
+
+  constructor(options: { cwdOverride?: string } = {}) {
+    super();
+    this.cwdOverride = options.cwdOverride;
+  }
+
   public async getRootPathAsync(): Promise<string> {
-    return getRootPath();
+    // Honor `EAS_PROJECT_ROOT` if it is set.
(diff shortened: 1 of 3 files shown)


csulit:
Suddenly our git becomes this big. And all of our changes from new eas-build are not reflected in IOS and ANDROID production (IOS crashes in app opening) 

image

packages/eas-cli/src/vcs/clients/noVcs.ts
```diff
@@ -1,18 +1,54 @@
...
+      ).stdout.trim();
+    } catch (err) {
+      Log.warn(`Failed to get Git root path with \`git rev-parse --show-toplevel\`.`, err);
+      Log.warn('Falling back to using current working directory as project root.');
+      Log.warn(
+        'You can set `EAS_PROJECT_ROOT` environment variable to let eas-cli know where your project is located.'
+      );
+    }
+
+    return path.resolve(resolveRoot, process.env.EAS_PROJECT_ROOT ?? '.');
```
szdziedzic, line 41:
```suggestion
    // If `EAS_PROJECT_ROOT` is not set, try to get the root path from Git.
    try {
      return (
        await spawnAsync('git', ['rev-parse', '--show-toplevel'], {
          cwd: this.cwdOverride,
        })
      ).stdout.trim();
    } catch (err) {
      Log.warn(`Failed to get Git root path with \`git rev-parse --show-toplevel\`.`, err);
      Log.warn('Falling back to using current working directory as project root.');
      Log.warn(
        'You can set `EAS_PROJECT_ROOT` environment variable to let eas-cli know where your project is located.'
      );
    }

    return path.resolve(process.cwd(), process.env.EAS_PROJECT_ROOT ?? '.');
```

szdziedzic (approved):

PR #2924, opened 2025-02-27
[eas-cli] improve applicationIdSuffix error message

kadikraman:
# Why

When building bare RN projects with EAS CLI with `applicationIdSuffix`, the CLI will error due to being unable to determine the application ID from the `build.gradle`. The workaround for this is not obvious.

# How

Update the error message to suggest adding the full application ID in `productFlavors` which is the simplest workaround.

# Test Plan

Verified locally.


szdziedzic (approved):
awesome

there are some errors marked by GHA

PR #2925, opened 2025-02-27
[eas-cli] Fix `.git` x `.easignore` interaction

sjchmiela:
# Why

Thanks to [a conversation on Bluesky](https://bsky.app/profile/sjchmiela.bsky.social/post/3lj6k2xo7rk2g) it turned out [this check](https://github.com/expo/eas-cli/blob/3a434940eb44dfa8dddc42af055793134ca9016f/packages/eas-cli/src/vcs/clients/git.ts#L255-L257) is invalid (always true), because `Ignore` has `DEFAULT_IGNORE`, so even if a user had no `.git` in `.easignore`, especially if they had `requireCommit: true`, `.git` would be removed.

# How

- Added a check to disallow `.easignore` if one has `requireCommit: true`.
- Changed logic around `Ignore` so we have separate `createForCopying` (ignores `.git` and `node_modules` always) and `createForChecking` (does not ignore anything by default). I don't like it, but it should work and is the best of what I could come up with without further refactoring.

# Test Plan

Added more tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fixed `.git` being always unexpectedly removed if you had `requireCommit: true` and `.easignore` present. ([#2925](https://github.com/expo/eas-cli/pull/2925) by [@sjchmiela](https://github.com/sjchmiela))
+
 ### 🧹 Chores
 
 - Fix `eas fingerprint:compare` URL generation and pretty prints. ([#2909](https://github.com/expo/eas-cli/pull/2909) by [@quinlanj](https://github.com/quinlanj))
--- a/packages/eas-cli/src/vcs/__tests__/local-test.ts
+++ b/packages/eas-cli/src/vcs/__tests__/local-test.ts
@@ -18,7 +18,7 @@ describe(Ignore, () => {
       '/root'
     );
 
-    const ignore = await Ignore.createAsync('/root');
+    const ignore = await Ignore.createForCopyingAsync('/root');
     expect(ignore.ignores('aaa')).toBe(true);
     expect(ignore.ignores('bbb')).toBe(false);
     expect(ignore.ignores('dir/aaa')).toBe(true);
@@ -35,7 +35,7 @@ describe(Ignore, () => {
       '/root'
     );
 
-    const ignore = await Ignore.createAsync('/root');
+    const ignore = await Ignore.createForCopyingAsync('/root');
     expect(ignore.ignores('aaa')).toBe(false);
     expect(ignore.ignores('bbb')).toBe(false);
     expect(ignore.ignores('ccc')).toBe(true);
@@ -52,7 +52,7 @@ describe(Ignore, () => {
       '/root'
     );
 
-    const ignore = await Ignore.createAsync('/root');
+    const ignore = await Ignore.createForCopyingAsync('/root');
     expect(ignore.ignores('aaa')).toBe(true);
     expect(ignore.ignores('bbb')).toBe(false);
     expect(ignore.ignores('node_modules/aaa')).toBe(true);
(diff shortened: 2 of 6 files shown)


szdziedzic (approved):
looks ok

thespacemanatee:
@sjchmiela this is a breaking change for us and it is not apparent at first glance how to fix this issue. We have `requireCommit=true` because of EAS Workflows and `.easignore` to prevent some irrelevant files from being uploaded during `eas build`.

sjchmiela:
Based on http://expo.fyi/eas-build-archive .easignore should never have been used if requireCommit=true, so my perspective was v15 introduced a bug which this fixed.

Main purpose of requireCommit is to ensure `eas build` and running a build from GitHub works the same. We can't adhere to .easignore when running from GitHub.

sjchmiela:
Is it possible in v14 .easignore wasn't being used at all?

sjchmiela:
Ohh, I see what you're saying — "if a release requires me to change stuff in my code to proceed, it's a breaking change". I get that.

What I was going for was "hey guys this `.easignore` you have in your workspace is doing nothing, just remove it because you may be thinking it's doing something and it's not". Sorry.

Let me write out how I things played out:
- in `eas-cli@<15` `requireCommit: true` allowed existence of `.easignore`, but ignored it (is this a clear behavior? I don't think so.)
- in `eas-cli@15.0.0` `requireComit: true` not only allowed existence of `.easignore`, but started removing files based on it. This was an unwanted breaking change and a bug.
- in `eas-cli@15.0.13` `requireCommit: true` disallowed existence of `.easignore` hoping to make the behavior clearer.

What do you think is the best course of action here now?

thespacemanatee:
@sjchmiela I guess my question then is, why is 2. an unwanted change? Our project is a monorepo where some modules are not part of the build process, ignoring them saves `eas build` several 100mb of upload bandwidth, and was even recommended by the cli. Having `requireCommit: true` and having `.easignore` was the solution for us to get EAS Workflows working while having a small `eas build` footprint. How would you update our setup?

sjchmiela:
Ohh, so what you're saying is that for you adding support for `.easignore` to `requireCommit: true` was an improvement! I think it goes against what has been the case for `requireCommit: true` [for at least 3 years](https://github.com/expo/fyi/pull/76/files) and I worry switching it back again could break others' projects in unexpected ways.

Would `requireCommit: false` work for you? It _does_ remove files based on `.easignore`!

thespacemanatee:
@sjchmiela switching it to false would break eas workflows right? I recall that it needed to be switched on for the build step to work. All I'm asking is if there's a solution where I can run eas workflows with easignore in the project.

thespacemanatee:
@sjchmiela For more context on the confusing directives from expo, here is what we can when we remove `.easignore` from the project and run `eas build`:

CleanShot 2025-03-08 at 14 34 26

This suggestion from the CLI goes against the `requireCommit: true` requirement set by EAS Workflows. What would you suggest we do here?

sjchmiela:
> I recall that it needed to be switched on for the build step to work.

If this was the case in the past, I think with the changes in `eas-cli@15` this should no longer be necessary. 

> For more context on the confusing directives from expo (…)

You're absolutely right, this warning shouldn't be printed if we don't expect the user to be allowed to have `.easignore`.

> What would you suggest we do here?

I can see how `.easignore`, if you know what you're doing (and you clearly know 🙏), may be helpful even in `requireCommit: true` mode.

My current idea is to release v16 which will support `.easignore` in `requireCommit: true`, but print a warning (hideable with an environment variable). Let me talk to some folks and hopefully do this today.

thespacemanatee:
@sjchmiela I'm pretty sure it was a requirement somewhere in our pipeline, I think it was the GitHub build action. That was pretty recent and we've been on 15 for awhile now so I doubt that's changed. In any case, we are stuck on 15.0.12 until the v16 you proposed. It's just a bit unfortunate that Expo introduces breaking changes frequently on minor patches.

thespacemanatee:
@sjchmiela This seems like a pretty good suggestion: https://github.com/expo/eas-cli/issues/1757#issuecomment-1817421226

Just show a warning instead of breaking existing workflows? Whether not it is working "as expected" should be up to the developer. In our case, the status quo was fine and now a lot of our flows are messed up.

sjchmiela:
> It's just a bit unfortunate that Expo introduces breaking changes frequently on minor patches.

I'm really sorry for this. I have underestimated people's projects' complexities when introducing the v15.0.0 change and many of the v15.0.X patches were just cleaning up the aftermath.

This change we're commenting on should have been a warning from the start. https://github.com/expo/eas-cli/pull/2942

Apart from this `eas-cli@15...` debacle, should we improve in this regard in other places too?

 > This seems like a pretty good suggestion: https://github.com/expo/eas-cli/issues/1757#issuecomment-1817421226

Indeed, it's exactly what I have been thinking of. Implemented in https://github.com/expo/eas-cli/pull/2942.

> Also, documentation like these: https://docs.expo.dev/build-reference/easignore/ make me think that `.easignore` should be used. I think we should keep using it to save Expo 400mb of storage every build, wdyt?

Yea, `.easignore` documentation turns out to be scattered across `docs.expo.dev` and https://github.com/expo/fyi/blob/main/eas-build-archive.md. I'll make them better.

thespacemanatee:
> > It's just a bit unfortunate that Expo introduces breaking changes frequently on minor patches.
> 
> I'm really sorry for this. I have underestimated people's projects' complexities when introducing the v15.0.0 change and many of the v15.0.X patches were just cleaning up the aftermath.
> 
> This change we're commenting on should have been a warning from the start. https://github.com/expo/eas-cli/pull/2942
> 
> Apart from this `eas-cli@15...` debacle, should we improve in this regard in other places too?
> 
>  > This seems like a pretty good suggestion: https://github.com/expo/eas-cli/issues/1757#issuecomment-1817421226
> 
> Indeed, it's exactly what I have been thinking of. Implemented in https://github.com/expo/eas-cli/pull/2942.
> 
> > Also, documentation like these: https://docs.expo.dev/build-reference/easignore/ make me think that `.easignore` should be used. I think we should keep using it to save Expo 400mb of storage every build, wdyt?
> 
> Yea, `.easignore` documentation turns out to be scattered across `docs.expo.dev` and https://github.com/expo/fyi/blob/main/eas-build-archive.md. I'll make them better.

I really appreciate your efforts in addressing every single one of my points 🙏 I understand it's not easy to account for everyone's workflow. Hopefully in a few minor patches down we can arrive at a solution that fixes the regression while warning about any problems that may arise from having both .easignore and requireCommit=true.

If there are any further issues we will create separate issues for them. Thanks again!

PR #2929, opened 2025-03-05
[eas-cli] suggest using eas build:dev when using simulator/emulator dev client configuration

szdziedzic:
# Why

Suggest using `eas build:dev` if one runs `eas build` resulting in simulator/emulator dev client build.

# How

Add log to suggest using `eas build:dev`

# Test Plan

Tests

### DIFF
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -311,9 +311,9 @@ export default class Build extends EasCommand {
         return true;
       })
     ) {
-      Log.addNewLineIfNone();
+      Log.newLine();
       Log.log(
-        `🔎 You are using a build configuration that could benefit from using ${chalk.bold(
+        `🔎 TIP: You are using a build configuration that could benefit from using ${chalk.bold(
           'eas build:dev'
         )} command. Run it to install and run cached development build, or create a new one if a compatible build doesn't exist yet.`
       );
--- a/packages/eas-cli/src/commands/build/index.ts
+++ b/packages/eas-cli/src/commands/build/index.ts
@@ -1,5 +1,5 @@
 import { Platform } from '@expo/eas-build-job';
-import { EasJsonAccessor, EasJsonUtils, ResourceClass } from '@expo/eas-json';
+import { BuildProfile, EasJsonAccessor, EasJsonUtils, ResourceClass } from '@expo/eas-json';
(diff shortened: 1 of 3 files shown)


szdziedzic:
* **#2929** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2929?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -268,6 +271,43 @@ export default class Build extends EasCommand {
...
+          }
+        }
+        return true;
+      })
+    ) {
+      Log.addNewLineIfNone();
+      Log.log(
+        `🔎 You are using a build configuration that could benefit from using ${chalk.bold(
+          'eas build:dev'
+        )} command. Run it to install and run cached dev client build or create a new one if it doesn't exist.`
```
brentvatne, line 307:
```suggestion
        )} command. Run it to install and run cached development build, or create a new one if a compatible build doesn't exist yet.`
```

brentvatne (approved):
let's give it a try!

packages/eas-cli/src/commands/build/index.ts
```diff
@@ -268,6 +271,43 @@ export default class Build extends EasCommand {
       requestedPlatform,
     };
   }
+
+  private maybeSuggestUsingEasBuildDev(buildProfiles?: ProfileData<'build'>[]): void {
+    // suggest using eas build:dev if the build configuration results in simulator/emulator dev client build
+    if (
+      buildProfiles?.some(({ profile, platform }) => {
```
brentvatne, line 278: let's also not show this message on CI

brentvatne:
it might also be useful to check for a build with the same profile with a matching fingerprint and let people know about that

szdziedzic:
>it might also be useful to check for a build with the same profile with a matching fingerprint and let people know about that

I added this message at the end of the build process. I believe we would need to add the check at the beginning instead, right?

I believe I can merge this PR the way it is and we can figure out the best way to add the fingerprint check in a separate PR.

PR #2932, opened 2025-03-06
Add share command

gabrieldonadel:
# Why

Closes ENG-14993

# How

# Test Plan

### DIFF
--- a/packages/eas-cli/src/commands/share.ts
+++ b/packages/eas-cli/src/commands/share.ts
@@ -1,0 +1,157 @@
+import { Platform } from '@expo/eas-build-job';
+import { Flags } from '@oclif/core';
+import fg from 'fast-glob';
+import fs from 'fs-extra';
+import path from 'path';
+
+import EasCommand from '../commandUtils/EasCommand';
+import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
+import { EASNonInteractiveFlag } from '../commandUtils/flags';
+import { UploadSessionType } from '../graphql/generated';
+import { ShareBuildMutation } from '../graphql/mutations/ShareBuildMutation';
+import Log from '../log';
+import { promptAsync } from '../prompts';
+import { uploadFileAtPathToGCSAsync } from '../uploads';
+import { createProgressTracker } from '../utils/progress';
+
+export default class BuildUpload extends EasCommand {
+  static override description = 'upload a local build and generate a sharable link';
+
+  static override flags = {
+    platform: Flags.enum<Platform.IOS | Platform.ANDROID>({
+      char: 'p',
+      options: [Platform.IOS, Platform.ANDROID],
+    }),
+    'build-path': Flags.string({
+      description: 'Path for the local build',
+    }),
+    ...EASNonInteractiveFlag,
+  };
+
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectId,
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    const { flags } = await this.parse(BuildUpload);
+    const { 'build-path': buildPath } = flags;
+    const {
+      projectId,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(BuildUpload, {
+      nonInteractive: false,
+    });
+
+    const platform = await this.selectPlatformAsync(flags.platform);
+    const localBuildPath = await resolveLocalBuildPathAsync({
+      platform,
+      buildPath,
+    });
+
+    Log.log('Uploading your app archive to EAS Share');
+    const bucketKey = await uploadAppArchiveAsync(graphqlClient, localBuildPath);
+
+    // @TODO: use Share build mutation
+    const { debugInfoUrl } = await ShareBuildMutation.uploadLocalBuildAsync(
+      graphqlClient,
+      projectId,
+      { bucketKey },
+      {
+        // @TODO: read the fingerprint from the local build
+        hash: '',
+      }
+    );
+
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/share.ts
```diff
@@ -0,0 +1,305 @@
...
+        const selectedAnswer = await promptAsync({
+          name: 'fingerprint',
+          message: `The provided fingerprint hash ${manualFingerprintHash} does not match the fingerprint hash of the build ${buildFingerprintHash}. Which fingerprint do you want to use?`,
+          type: 'select',
+          choices: [
+            { title: manualFingerprintHash, value: manualFingerprintHash },
+            { title: buildFingerprintHash, value: buildFingerprintHash },
+          ],
+        });
+        fingerprint = String(selectedAnswer.fingerprint);
```
szdziedzic, line 83: why is it needed?

packages/eas-cli/src/commands/share.ts
```diff
@@ -0,0 +1,305 @@
...
+
+async function resolveLocalBuildPathAsync(
+  platform: Platform,
+  inputBuildPath?: string
+): Promise<string> {
+  const applicationArchivePatternOrPath =
+    inputBuildPath ??
+    (platform === Platform.ANDROID
+      ? 'android/app/build/outputs/**/*.{apk,aab}'
+      : 'ios/build/Build/Products/*simulator/*.app');
```
szdziedzic, line 130: I'm not sure if these are compatible with the default artifact path of `eas build --local`. It would create a build and move it to `projectDir/somename.extenstion` and not `projectDir/ios|android/...`. Are these paths the result path as of `npx expo:run` build?

packages/eas-cli/src/commands/share.ts
```diff
@@ -0,0 +1,305 @@
...
+  simulator: boolean;
+};
+
+async function extractAppMetadataAsync(
+  buildPath: string,
+  platform: Platform
+): Promise<AppMetadata> {
+  let developmentClient = false;
+  let fingerprintHash: string | undefined;
+  const simulator = platform === Platform.IOS;
```
szdziedzic, line 229: is it even a valid use case to share `.ipa` builds? I'm just curious

packages/eas-cli/src/commands/share.ts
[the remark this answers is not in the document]
gabrieldonadel: Brent actually suggested us to rename this to `upload`

packages/eas-cli/src/commands/share.ts
gabrieldonadel: promptAsync answers are typed as `any`, so we need to do this to prevent TS from complaining
https://github.com/expo/eas-cli/blob/f21aa7d99bf14325d7c3ce1a18eb9289d4b5dece/node_modules/@types/prompts/index.d.ts#L107-L108

packages/eas-cli/src/commands/share.ts
gabrieldonadel: That's right, let's not support .ipa's for now

packages/eas-cli/src/commands/share.ts
gabrieldonadel: yes, I've updated it and now it should match `npx expo:run `

packages/eas-cli/src/commands/upload.ts
```diff
@@ -0,0 +1,321 @@
...
+  buildPath: string,
+  platform: Platform
+): Promise<AppMetadata> {
+  let developmentClient = false;
+  let fingerprintHash: string | undefined;
+  const simulator = platform === Platform.IOS;
+
+  const basePath = platform === Platform.ANDROID ? 'assets/' : buildPath;
+  const fingerprintFilePath =
+    platform === Platform.ANDROID ? 'fingerprint' : 'EXUpdates.bundle/fingerprint';
```
quinlanj, line 257: Normally, when we run eas build and eas update, the fingerprint is automatically calculated (in SDK 52+) and the sources are uploaded. However, during the build process when a fingerprint update policy is declared — I recall that the fingerprint gets calculated after the prebuild step. That might be when these files are being generated.

Idealy, we should figure out how to upload or output the fingerprint sources at the moment when `EXUpdates.bundle/fingerprint` (and the Android equivalent) is created. Otherwise, we'll end up with just the fingerprint hash, which isn’t all that useful on its own.

packages/eas-cli/src/commands/upload.ts
```diff
@@ -0,0 +1,321 @@
...
+            { title: buildFingerprintHash, value: buildFingerprintHash },
+          ],
+        });
+        fingerprint = String(selectedAnswer.fingerprint);
+      }
+
+      await FingerprintMutation.createFingerprintAsync(graphqlClient, projectId, {
+        hash: fingerprint,
+      });
+    }
```
quinlanj, line 92: See my other comment about the fingerprint sources — they’re not very useful if we only have the hash without the corresponding source files. I can update the website in the meantime to improve the viewing experience (right now it just shows a generic error when the source is missing). But ideally, we should find a way to upload the source when the fingerprint is generated

quinlanj (approved):

packages/eas-cli/src/commands/upload.ts
gabrieldonadel: I agree, but there's still some value in storing just the hash if the source is not available, because if another user in the team calculates the local fingerprint hash and it's a match, then they could just download the build from EAS instead of building locally

packages/eas-cli/src/commands/upload.ts
gabrieldonadel: I agree, we will probably need to store this info somewhere, potentially inside a node_modeles folder or something like that

PR #2948, opened 2025-03-15
[ci] Pin to older tj-actions/changed-files commit.

randomhajile:
# Why

`tj-actions/changed-files` has been compromised. See discussion [here](https://github.com/tj-actions/changed-files/issues/2463).

# How

Pinned to a uncompromised commit. I've also disabled this workflow for the repo.


ide (approved):
Looks good, and good that our repositories have not run the bad version of the Action.

PR #2937, opened 2025-03-08
[ENG-15002][eas-cli] fingerprint:generate

quinlanj:
# Why

First pass of `eas fingerprint:generate` in hidden mode.

- Interactive fingerprint generation:
`eas fingerprint:generate`
Screenshot 2025-03-16 at 6 31 51 PM

- Non Interactive fingerprint generation:
`eas fingerprint:generate --json --non-interactive -p android`
Will print the fingerprint json to stdout

# Test Plan

- [ ] ran `eas fingerprint:generate`
- [ ] ran `eas fingerprint:generate --json --non-interactive -p android`

PR #2951, opened 2025-03-16
[eas-cli] add support for build profiles in fingerprint generation

quinlanj:
# How you can help

This is my current understanding of how a build profile can affect a project's fingerprint. @szdziedzic and @sjchmiela, please let me know if it is correct, or if there are any nuances I am missing!

- A fingerprint contains the [following components](https://staging.expo.dev/accounts/quintest110/projects/staging-sdk-52/fingerprints/653b4ea74e6e03eda64fa4815e8a1dd8430b57ba) which we use to determine compatibility at an app's native layer
- In `eas.json`, we have a bunch of build profiles which define a bunch of parameters like build machine types, update channels, etc. However, it is the set of environment variables produced (specifying build profile parameter as input) which ultimately affects the fingerprint. **Is there any other way a build profile can affect a fingerprint?** 

For example, I can define the TEST env var to differ based on the build profile that is chosen:
Screenshot 2025-03-18 at 3 33 58 PM

Next, I configure my `app.config.js` to determine the bundleIdentifier based on the `TEST` env var:
```
function getBundleIdentifier() {
  if (process.env.TEST === "production") {
    return "com.quintest110.stagingsdk52.production";
  } else if (process.env.TEST === "development") {
    return "com.quintest110.stagingsdk52.development";
  } else if (process.env.TEST === "preview") {
    return "com.quintest110.stagingsdk52.preview";
  }
  return "com.quintest110.stagingsdk52";
}
```

Since the fingerprint relies on the app config, I can see the fingerprint is changed when I run `eas fingerprint:generate -p ios --profile production` vs `eas fingerprint:generate -p ios --profile development` [here](https://staging.expo.dev/accounts/quintest110/projects/staging-sdk-52/fingerprints/compare/4424134ca5d1f07725826b3a5a504eba2e570379/081bce3372dff93836ff55c0344b896aebaf97ef) 

# Why
This PR adds support for build profiles in `eas fingeprint:generate` by allowing the user to pass in the `--profile` flag. 

# How
After a build profile is chosen, the cli takes the generated Env and temporarily sets those env vars while calculating the fingerprint

# Tests

- [ ] Different fingerprints with expected diffs are generated with `eas fingerprint:generate -p ios --profile production` vs `eas fingerprint:generate -p ios --profile development` [here](https://staging.expo.dev/accounts/quintest110/projects/staging-sdk-52/fingerprints/compare/4424134ca5d1f07725826b3a5a504eba2e570379/081bce3372dff93836ff55c0344b896aebaf97ef)

### DIFF
--- a/packages/eas-cli/src/commands/fingerprint/generate.ts
+++ b/packages/eas-cli/src/commands/fingerprint/generate.ts
@@ -1,6 +1,9 @@
+import { Env, Platform } from '@expo/eas-build-job';
+import { EasJsonAccessor } from '@expo/eas-json';
 import { Flags } from '@oclif/core';
 
 import { getExpoWebsiteBaseUrl } from '../../api';
+import { evaluateConfigWithEnvVarsAsync } from '../../build/evaluateConfigWithEnvVarsAsync';
 import EasCommand from '../../commandUtils/EasCommand';
 import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
 import {
@@ -12,6 +15,7 @@ import { AppQuery } from '../../graphql/queries/AppQuery';
 import Log, { link } from '../../log';
 import { promptAsync } from '../../prompts';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+import { getProfilesAsync } from '../../utils/profiles';
 
 export default class FingerprintGenerate extends EasCommand {
   static override description = 'generate fingerprints from the current project';
@@ -19,14 +23,19 @@ export default class FingerprintGenerate extends EasCommand {
   static override hidden = true;
 
   static override examples = [
-    '$ eas fingerprint:generate',
-    '$ eas fingerprint:generate --json --non-interactive -p android',
+    '$ eas fingerprint:generate  \t # Generate fingerprint in interactive mode',
+    '$ eas fingerprint:generate --profile preview  \t # Generate a fingerprint using the "preview" build profile',
+    '$ eas fingerprint:generate --json --non-interactive --platform android  \t # Output fingerprint json to stdout',
   ];
 
   static override flags = {
     platform: Flags.enum({
       char: 'p',
       options: ['android', 'ios'],
+    }),
+    profile: Flags.string({
+      char: 'e',
+      description: 'Name of the build profile from eas.json.',
     }),
     ...EasNonInteractiveAndJsonFlags,
   };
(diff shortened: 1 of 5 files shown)


quinlanj:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2951?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#2951** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2951?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2937** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/2937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

wschurman (commented):
How does this relate to EAS Environment Variables? (the `environment` flag in `eas update`). I was under the assumption that we didn't add the environment flag to `eas build` since this was specified in the eas json profile and they were automatically loaded. As a point of reference, for doing something similar in the github action we use `eas env:exec`: https://github.com/expo/expo-github-action/blob/main/src/fingerprintUtils.ts#L72

packages/eas-cli/src/commands/fingerprint/generate.ts
```diff
@@ -101,3 +144,13 @@ export async function selectRequestedPlatformAsync(): Promise<AppPlatform> {
   });
   return requestedPlatform;
 }
+
+function appPlatformtoPlatform(appPlatform: AppPlatform): Platform {
```
wschurman, line 148:
```suggestion
function appPlatformToPlatform(appPlatform: AppPlatform): Platform {
```

wschurman:
(requesting changes mainly to put back on your queue for discussion of environment variables, not for inline comments)

quinlanj:
@wschurman makes sense, i updated the command so that it takes the environment arg instead of `profile`.

packages/eas-cli/src/commands/fingerprint/generate.ts
```diff
@@ -64,12 +74,21 @@ export default class FingerprintGenerate extends EasCommand {
       }
       platform = await selectRequestedPlatformAsync();
     }
+
+    let env: Env | undefined;
+    if (environment) {
+      env = environment
+        ? { ...(await getServerSideEnvironmentVariablesAsync()), EXPO_NO_DOTENV: '1' }
+        : {};
+    }
```
wschurman, line 83:
```suggestion
    const env = environment
      ? { ...(await getServerSideEnvironmentVariablesAsync()), EXPO_NO_DOTENV: '1' }
      : undefined;
```

wschurman: Also regarding `EXPO_NO_DOTENV`, I see this is the behavior for `eas update` for calling into the expo cli (which I assume fingerprint does under the hood hence why we add it to this call), but is it also the behavior for `eas build`?

wschurman (approved):

packages/eas-cli/src/commands/fingerprint/generate.ts
quinlanj: i passed `EXPO_NO_DOTENV` because it is also passed to grab the config elsewhere in `eas-cli` (npx expo config is a shim over expo/config): https://github.com/expo/eas-cli/blob/22bc21ec9c5571cbbe428e89213e5e459f7de767/packages/eas-cli/src/project/expoConfig.ts#L59-L72

under the hood, the fingerprint library calls into the `expo/config` library  https://github.com/expo/eas-cli/blob/22bc21ec9c5571cbbe428e89213e5e459f7de767/packages/eas-cli/src/project/expoConfig.ts#L59-L72

PR #2957, opened 2025-03-20
[eas-cli] Always sync update configuration with eas update:configure

brentvatne:
# Why

We sync updates config from app.json/config.js when we run `eas build` but not when running `eas update:configure`.

# How

- Always sync native config from app.json/config.js when running `eas update:configure`, even when already set up. This is a safe change
- Do not sync native config when running `eas update`

# Test Plan

Create a project, configure updates with `eas update:configure`, run prebuild. Change updates configuration in app.json, run `eas update:configure` again - it should sync the changes to the native directories. Change updates configuration in app.json, run `eas update` - it should not sync changes.

### DIFF
--- a/packages/eas-cli/src/update/configure.ts
+++ b/packages/eas-cli/src/update/configure.ts
@@ -409,15 +409,13 @@ export async function ensureEASUpdateIsConfiguredAsync({
       workflows,
     });
 
-  if (projectChanged || !hasExpoUpdates) {
-    await ensureEASUpdateIsConfiguredNativelyAsync(vcsClient, {
-      exp: expWithUpdates,
-      projectDir,
-      platform,
-      workflows,
-      env,
-    });
-  }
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/update/configure.ts
```diff
@@ -409,7 +411,7 @@ export async function ensureEASUpdateIsConfiguredAsync({
       workflows,
     });
 
-  if (projectChanged || !hasExpoUpdates) {
+  if (forceNativeConfigSync || projectChanged || !hasExpoUpdates) {
```
brentvatne, line 414: did this instead of deleting the conditional entirely, because `ensureEASUpdateIsConfiguredAsync` is also called from `eas update`. this minimizes the impact of the change in this PR and ensures that it only applies to `eas update:configure`

wschurman (approved):

PR #2968, opened 2025-03-28
[eas-cli] Improve unauthorized error for `eas deploy --prod`

sjchmiela:
# Why

When we move Workflows to scoped token which has at most DEVELOPER permission, `type: deploy, prod: true` jobs will fail with an ominous

```
You don't have the required permissions to perform this operation.

This can sometimes happen if you are logged in as incorrect user.
Run eas whoami to check the username you are logged in as.
Run eas login to change the account.

Original error message: Entity not authorized: AppDevDomainNameEntity[00000000-0000-0000-0000-000000000000] (viewer = ScopedAccountActorViewerContext[16d62ef3-a219-47ca-936b-89958f6cb1f8], action = CREATE, ruleIndex = -1)
Request ID: 863a1317-882b-4cab-b8f2-c5f5d63b6f29

```

# How

Special-cased the `--prod` && `AppDevDomainEntity` && `CREATE` && `UNAUTHORIZED_ERROR` error.

Not sure if we should call out "as an app admin" in the error message. Open to suggestions.

# Test Plan

Ran `eas deploy --prod` as a developer-permission-only user. I've updated message in code since taking the screenshot.

Zrzut ekranu 2025-03-28 o 21 21 02

### DIFF
--- a/packages/eas-cli/src/commands/deploy/index.ts
+++ b/packages/eas-cli/src/commands/deploy/index.ts
@@ -5,11 +5,13 @@ import chalk from 'chalk';
 import fs from 'node:fs';
 import * as path from 'node:path';
 
+import { getHostingDeploymentsUrl } from '../../build/utils/url';
 import EasCommand from '../../commandUtils/EasCommand';
 import { EASEnvironmentFlag, EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
 import { EnvironmentVariableEnvironment } from '../../graphql/generated';
-import Log from '../../log';
+import Log, { link } from '../../log';
 import { ora } from '../../ora';
+import { getOwnerAccountForProjectIdAsync } from '../../project/projectUtils';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 import { createProgressTracker } from '../../utils/progress';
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/deploy/index.ts
```diff
@@ -299,6 +300,25 @@ export default class WorkerDeploy extends EasCommand {
...
+            err.extensions?.errorCode === 'UNAUTHORIZED_ERROR' &&
+            err.message.includes('AppDevDomainNameEntity') &&
+            err.message.includes('CREATE')
+          );
+        })
+      ) {
+        throw new Error(
+          `You have specified the new deployment should be a production deployment, but a production domain has not been set up yet for this app.\n\nRun ${chalk.bold(
+            'eas deploy --prod'
+          )} as an app admin on your machine to set up a production domain and then try again.`
```
kadikraman, line 318: Another way to do this would be to promote an existing deployment to production on the dashboard https://expo.dev/accounts/[account]/projects/[project]/hosting/deployments?view=deployments

```suggestion
          )} as an app admin on your machine or promote an existing deployment to production on the Hosting Dashbaord to set up a production domain and then try again.`
```

kadikraman (approved):

PR #2982, opened 2025-04-08
[eas-cli] Add build:download command

gabrieldonadel:
# Why

We want to introduce a feature in `expo-cli` to allow users to download remote builds instead of always building locally if there is an available build with a matching fingerprint. For this "cache" layer to work we need a way to download EAS builds through eas-cli 

# How

Introduce a new `build:download`command

# Test Plan

### Setup

 create a couple of builds on expo.dev

### Actions
  - Run `easd build:download --platform [ios|android] --fingerprint [hash]`
  - Check if the build was downloaded correctly
  
### Test Reproduction (Screenshots/Videos/Terminal Output)

image

### DIFF
--- a/packages/eas-cli/src/commands/build/download.ts
+++ b/packages/eas-cli/src/commands/build/download.ts
@@ -8,6 +8,7 @@ import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/creat
 import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
 import { AppPlatform, BuildFragment, BuildStatus, DistributionType } from '../../graphql/generated';
 import { BuildQuery } from '../../graphql/queries/BuildQuery';
+import { toAppPlatform } from '../../graphql/types/AppPlatform';
 import Log from '../../log';
 import { promptAsync } from '../../prompts';
 import { getEasBuildRunCachedAppPath } from '../../run/run';
@@ -50,17 +51,14 @@ export default class Download extends EasCommand {
       enableJsonOutput();
     }
 
-    const selectedPlatform = await resolvePlatformAsync(nonInteractive, platform);
+    const selectedPlatform = await resolvePlatformAsync({ nonInteractive, platform });
     const build = await this.getBuildAsync(graphqlClient, projectId, selectedPlatform, fingerprint);
-    const buildPath = await this.getPathToBuildAppAsync(build, selectedPlatform);
+    const buildArtifactPath = await this.getPathToBuildArtifactAsync(build, selectedPlatform);
     if (jsonFlag) {
-      const jsonResults: { path?: string } = {};
-      if (buildPath) {
-        jsonResults.path = buildPath;
-      }
+      const jsonResults = { path: buildArtifactPath };
       printJsonOnlyOutput(jsonResults);
     } else {
-      Log.log(`Build downloaded at ${chalk.bold(buildPath)}`);
+      Log.log(`Build downloaded to ${chalk.bold(buildArtifactPath)}`);
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/build/download.ts
```diff
@@ -0,0 +1,138 @@
...
+      build.artifacts.applicationArchiveUrl,
+      platform,
+      cachedAppPath
+    );
+  }
+}
+
+async function resolvePlatformAsync(
+  nonInteractive: boolean,
+  platform?: string
```
sjchmiela, line 118: Those are simple function, but as a rule of thumb I think it's good to default to `{ nonInteractive, platform }` rather than `nonInteractive, platform` for function arguments.

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    Log.succeed(`🎯 Found successful build with matching fingerprint on EAS servers.`);
+    return builds[0];
+  }
+
+  async getPathToBuildAppAsync(build: BuildFragment, platform: AppPlatform): Promise<string> {
+    const cachedAppPath = getEasBuildRunCachedAppPath(build.project.id, build.id, platform);
+    if (await pathExists(cachedAppPath)) {
+      Log.newLine();
+      Log.log(`Using cached app...`);
```
sjchmiela, line 100: Is it a build or an app? ("app" feels like a one-off word here.)

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -0,0 +1,138 @@
...
+  nonInteractive: boolean,
+  platform?: string
+): Promise<AppPlatform> {
+  if (nonInteractive && !platform) {
+    throw new Error('Platform must be provided in non-interactive mode');
+  }
+
+  if (platform && Object.values(AppPlatform).includes(platform.toUpperCase() as AppPlatform)) {
+    return platform.toUpperCase() as AppPlatform;
+  }
```
sjchmiela, line 126: nit: a

```ts
const PlatformToAppPlatform: Record<Platform, AppPlatform> = {
  [Platform.IOS]: AppPlatform.Ios,
  [Platform.ANDROID]: AppPlatform.Android,
};
```

feels cleaner / safer.

EDIT: Actually, there's s `toAppPlatform` already.

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    const selectedPlatform = await resolvePlatformAsync(nonInteractive, platform);
+    const build = await this.getBuildAsync(graphqlClient, projectId, selectedPlatform, fingerprint);
+    const buildPath = await this.getPathToBuildAppAsync(build, selectedPlatform);
+    if (jsonFlag) {
+      const jsonResults: { path?: string } = {};
+      if (buildPath) {
+        jsonResults.path = buildPath;
+      }
```
sjchmiela, line 60: I think as it stands `buildPath` will always return a non-empty string (or it will throw)?

sjchmiela, line 60:
```suggestion
      const jsonResults = { path: buildPath };
```

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -0,0 +1,138 @@
...
+      printJsonOnlyOutput(jsonResults);
+    } else {
+      Log.log(`Build downloaded at ${chalk.bold(buildPath)}`);
```
sjchmiela, line 63:
```suggestion
      Log.log(`Build downloaded to ${chalk.bold(buildPath)}`);
```

sjchmiela: Do we want to ask follow-up question to run the build as we do in `maybeDownloadAndRunSimulatorBuildsAsync`?

sjchmiela (changes requested):
Some small improvements suggestions

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: I don't think we, at least not for the initial release, my understanding is that if users want to run a build they should use `eas build:run`

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: Sorry for the confusion, updated it.

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: updated

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: Updated to use `toAppPlatform `

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: That's correct, fixed

sjchmiela (approved):

quinlanj (approved):

PR #2983, opened 2025-04-09
[eas-cli] Add non-interactive and json flags to upload command

gabrieldonadel:
# Why

As part of the integration with expo-cli we should support a non-interactive mode when running `eas upload`

# How

 Add non-interactive and json flags to `upload` the command

# Test Plan

### Setup

 create build your app locally using Xcode or Android Studio

### Actions
  - Run `easd upload --non-interactive --platform ios --json`
  - Navigate to http://staging.expo.dev/accounts/gabrieldonadel/projects/fabric-tester/builds/bc7fe643-a5db-40e2-9cac-5362a8601423
  - Observe your local build

### DIFF
--- a/packages/eas-cli/src/commands/upload.ts
+++ b/packages/eas-cli/src/commands/upload.ts
@@ -11,7 +11,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { getBuildLogsUrl } from '../build/utils/url';
 import EasCommand from '../commandUtils/EasCommand';
 import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
-import { EASNonInteractiveFlag } from '../commandUtils/flags';
+import { EasNonInteractiveAndJsonFlags } from '../commandUtils/flags';
 import {
   DistributionType,
   LocalBuildArchiveSourceType,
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/upload.ts
```diff
@@ -107,10 +122,24 @@ export default class BuildUpload extends EasCommand {
       { distribution: DistributionType.Internal, fingerprintHash: fingerprint, developmentClient }
     );
 
-    Log.withTick(`Here is a sharable link of your build: ${getBuildLogsUrl(build)}`);
+    if (jsonFlag) {
+      printJsonOnlyOutput({ url: getBuildLogsUrl(build) });
+    } else {
+      Log.withTick(`Shareable link to the build: ${getBuildLogsUrl(build)}`);
+    }
```
quinlanj, line 129: Preference to organize code like this, because if you add on extra logic later (e.g., printing other logs), having the early return keeps the structure clean and prevents deeply nested conditionals. It also separates the JSON-only output path from the interactive logging path.

```suggestion
    if (jsonFlag) {
      printJsonOnlyOutput({ url: getBuildLogsUrl(build) });
      return;
    } 
    
    Log.withTick(`Shareable link to the build: ${getBuildLogsUrl(build)}`);
    
```

quinlanj (approved):

PR #2985, opened 2025-04-11
[eas-cli] Allow no in dev-client flag for build:download

gabrieldonadel:
# Why

Users should be able to download non dev-client builds (e.g. builds with `--configuration Release`)

# How

Allow no in dev-client flag for build:download

# Test Plan

Run `easd build:download -p ios --fingerprint "5daec8fdf29fd8023ce92fae92f0e33d2eadac33" --no-dev-client`

### DIFF
--- a/packages/eas-cli/src/commands/build/download.ts
+++ b/packages/eas-cli/src/commands/build/download.ts
@@ -29,6 +29,7 @@ export default class Download extends EasCommand {
     }),
     'dev-client': Flags.boolean({
       description: 'Filter only dev-client builds.',
+      allowNo: true,
     }),
     ...EasNonInteractiveAndJsonFlags,
   };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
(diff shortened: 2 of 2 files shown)


packages/eas-cli/src/commands/build/download.ts
```diff
@@ -29,6 +29,7 @@ export default class Download extends EasCommand {
     }),
     'dev-client': Flags.boolean({
```
sjchmiela, line 30: According to types, `dev-client` flag is `boolean` (meaning no `undefined`). Is this true? Do we consciously default to `--dev-client`?

sjchmiela (approved):

packages/eas-cli/src/commands/build/download.ts
gabrieldonadel: Sorry but, according to which types? I changed this mostly because the value I get when not passing `--dev-client`
is `undefined`. This basically allows us to 3 different queries, all builds, dev-client builds only and non-dev-client build only  

image

sjchmiela: Cool, what you pasted is what I would expect. I guess the types are wrong.

Zrzut ekranu 2025-04-11 o 15 18 46

PR #2991, opened 2025-04-15
[eas-cli] change launch asset file extension

quinlanj:
# Why

This PR changes the file extension detection for launch assets to either `.hbc` or `.js`. If the file extension cannot be determined, we fall back to `.bundle`. It would be great to have this so we can display a UI with someone's update and show whether the bundle is hermes bytecode or js. 

## Documentation
- In [our v1 spec](https://docs.expo.dev/technical-specs/expo-updates-1/#manifest-body), we mention that we dont use the `fileExtension` field for the launch asset:

> launchAsset: A special asset that is the entry point of the application code. The fileExtension field will be ignored for this asset and SHOULD be omitted. 

- In the client code, it doesnt look like we care about the launch asset file extension for [Android updates](https://github.com/expo/expo/blob/177b7e763db899ecb6404441d706215d18508309/packages/expo-updates/android/src/main/java/expo/modules/updates/manifest/ExpoUpdatesUpdate.kt#L53) and [iOS updates](https://github.com/expo/expo/blob/65bd4188bcffad05394fd81ada883e125b512ee3/packages/expo-updates/ios/EXUpdates/Update/ExpoUpdatesUpdate.swift#L36)

# How

- Made `loadAssetMapAsync` function private by removing the `export` keyword since it's only used internally.
- Updated the `collectAssetsAsync` function to dynamically determine the file extension from the bundle path using `path.extname()`.
- Added a fallback to `.bundle` when `path.extname()` returns an empty string (which happens when there's no extension).

# Test Plan

- [x] tested with hermes and jsengine to confirm we correctly mark `.hbc` and `.js` in common cases
- [x] tested on android and ios, ensured updates are correctly applied

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Update upload command to display build info. ([#2990](https://github.com/expo/eas-cli/pull/2990) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+- change launch asset file extension. ([#2991](https://github.com/expo/eas-cli/pull/2991) by [@quinlanj](https://github.com/quinlanj))
 
 ## [16.3.1](https://github.com/expo/eas-cli/releases/tag/v16.3.1) - 2025-04-11
 
--- a/packages/eas-cli/src/project/__tests__/publish-test.ts
+++ b/packages/eas-cli/src/project/__tests__/publish-test.ts
@@ -424,7 +424,7 @@ describe(collectAssetsAsync, () => {
     expect(await collectAssetsAsync(inputDir)).toEqual({
       android: {
         launchAsset: {
-          fileExtension: '.bundle',
+          fileExtension: '.js',
           contentType: 'application/javascript',
           path: `${inputDir}/bundles/android.js`,
         },
@@ -432,7 +432,7 @@ describe(collectAssetsAsync, () => {
       },
       ios: {
         launchAsset: {
-          fileExtension: '.bundle',
+          fileExtension: '.js',
           contentType: 'application/javascript',
           path: `${inputDir}/bundles/ios.js`,
         },
@@ -440,7 +440,7 @@ describe(collectAssetsAsync, () => {
       },
       web: {
         launchAsset: {
-          fileExtension: '.bundle',
+          fileExtension: '.js',
           contentType: 'application/javascript',
           path: `${inputDir}/bundles/web.js`,
         },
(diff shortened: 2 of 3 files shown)


wschurman (approved):
I think this is theoretically safe. Though it'd be good to grep for `.bundle`/`bundle` in expo-updates and EAS server code to be sure. Also adding @Kudo to ensure this doesn't conflict with his recent work on assets on client.

Kudo (approved):
my embedded asset change doesn't change the way to handle remote updates. also looks good to me

wschurman:
One question: what is the embedded launch asset's filename? Has it ever been possible that it and its hash are equal to that of an OTA update and therefore when said OTA update occurs it doesn't need to download it? Rephrasing this question another way: do we need to make the same change for the embedded bundle?

Kudo:
> what is the embedded launch asset's filename? Has it ever been possible that it and its hash are equal to that of an OTA update and therefore when said OTA update occurs it doesn't need to download it? 

embedded launch asset would be `file:///android_asset/index.android.bundle` on android. it won't copy to local file system (the `.expo-internal` directory) and not conflict with ota update. it would possible for hash matched and not download launch asset from ota again.

> Rephrasing this question another way: do we need to make the same change for the embedded bundle?

you mean rename the `index.android.bundle` to something like `index.android.hbc`? it's not necessary from i can tell. or do we want to display the ui including embedded asset type?

quinlanj:
@wschurman confirming we don't send down the fileExtension of the launch asset from www: https://github.com/expo/universe/blob/f5958bd0190be307fd464db0d5c9ea63b84c3320/server/www/src/data/entities/updates/UpdatesUtils.ts#L1092-L1095

Double checked it with postman too:

```
-------ExpoManifestBoundary-uuMDwIy1TBahDdXwVrOirA
Content-Disposition: form-data; name="manifest"
Content-Type: application/json

{"id":"bae303c0-8cb5-4c16-a10d-d5f056b3a2ac","createdAt":"2025-04-10T00:55:30.965Z","runtimeVersion":"1.0.0","launchAsset":{"hash":"xRj6dzchwC5WBYiMhiJSnU9jCZDlOdpUL4QUH7gFxxw","key":"2659c006eeb013fd046de61606865e6f","contentType":"application/javascript","url":"https://staging-assets.eascdn.net/dmRPmYGZjlcDFWHrQfJvy2yyj0hhGiTFL2Lc83EHGhc?project=4a0663e3-6834-4a9b-8ac1-f82c0d29b7f9"},"assets":[{"hash":"i2Gkx-9w3JJ1PwSUl2SC9m_UFQ7CPfx3KrZeEDc6-lU","key":"778ffc9fe8773a878e9c30a6304784de","contentType":"image/png","fileExtension":".png","url":"https://staging-assets.eascdn.net/idnX8z03q4vGLzhYMPaQGBADjhqvWe67_afPqJ1vGzM?project=4a0663e3-6834-4a9b-8ac1-f82c0d29b7f9"},
...
```

PR #2996, opened 2025-04-22
[eas-cli] Detect when expo metro config is used

kadikraman:
# Why

After upgrading to SDK 53 I get this warning when running `eas build` even though I'm definitely using the expo metro config.

Screenshot 2025-04-22 at 20 08 27

# How

Update the check to match that of [Expo Doctor](https://github.com/expo/expo/blob/main/packages/expo-doctor/src/checks/MetroConfigCheck.ts#L20) which does detect this.

# Test Plan

Run `easd build` locally and confirm the warning is no longer shown.

### DIFF
--- a/packages/eas-cli/src/project/metroConfig.ts
+++ b/packages/eas-cli/src/project/metroConfig.ts
@@ -16,10 +16,11 @@ export async function validateMetroConfigForManagedWorkflowAsync(
   }
 
   const metroConfig = await loadConfigAsync(ctx.projectDir);
-  const hasHashAssetFilesPlugin = metroConfig.transformer?.assetPlugins?.find((plugin: string) =>
-    plugin.match(/expo-asset[/|\\]tools[/|\\]hashAssetFiles/)
-  );
-  if (!hasHashAssetFilesPlugin) {
+  
+  // This is a custom property that we inject to ensure cache invalidation between projects.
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/project/metroConfig.ts
```diff
@@ -16,10 +16,11 @@ export async function validateMetroConfigForManagedWorkflowAsync(
   }
 
   const metroConfig = await loadConfigAsync(ctx.projectDir);
-  const hasHashAssetFilesPlugin = metroConfig.transformer?.assetPlugins?.find((plugin: string) =>
-    plugin.match(/expo-asset[/|\\]tools[/|\\]hashAssetFiles/)
-  );
-  if (!hasHashAssetFilesPlugin) {
+  
+  // This is a custom property that we inject to ensure cache invalidation between projects.
+  const isExtendingExpoMetroConfig = metroConfig.transformer.hasOwnProperty('_expoRelativeProjectRoot');
```
brentvatne, line 21: do we need to keep the old logic for sdk 52 and below?

kadikraman: Good point, looks like this property is actually [relatively new](https://github.com/expo/expo/commit/5ba784171ea0a7b58989414161a8fd8a1d75ad7e) - I'll add the old check back as a fallback.

brentvatne (approved):

brentvatne:
thank you!

kitten (approved):

PR #2998, opened 2025-04-24
[eas-cli] Add specific error for WAF block events (EAS Hosting)

kitten:
# Why

The Cloudflare CDN WAF might block the requests from specific sources when a request contains suspicious contents, a rule has caught the content, or the IP has been rate limited by their bot net protection.

All we can do in this case is:
- Tell the user this has been a CDN-level error
- Quote the Request ID (aka Ray ID) so we can look up the error

Here's an example of what this might look like:
image

# How

- Add check for `Content-Type: text/html` and output non-JSON error message

# Test Plan

- Tested manually locally

### DIFF
--- a/packages/eas-cli/src/worker/upload.ts
+++ b/packages/eas-cli/src/worker/upload.ts
@@ -130,11 +130,25 @@ export async function uploadAsync(params: UploadParams): Promise<UploadResult> {
       }
 
       const getErrorMessageAsync = async (): Promise<string> => {
-        const body = await response.json().catch(() => null);
-        return body?.error ?? `Upload of "${filePath}" failed: ${response.statusText}`;
+        const rayId = response.headers.get('cf-ray');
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/worker/upload.ts
```diff
@@ -130,11 +130,25 @@ export async function uploadAsync(params: UploadParams): Promise<UploadResult> {
...
+          return `Upload of "${filePath}" failed: ${message}`;
+        } else {
+          const json = await response.json().catch(() => null);
+          return json?.error ?? `Upload of "${filePath}" failed: ${response.statusText}`;
+        }
       };
 
       if (
+        response.status === 408 ||
         response.status === 408 ||
```
kadikraman, line 152: This line is duplicated.

kadikraman (approved):

PR #3003, opened 2025-04-28
[eas-cli] Allow any suffix for gradle command build type

khamilowicz:
# Why

[ENG-15618: Support different suffixes of custom gradle commands](https://linear.app/expo/issue/ENG-15618/support-different-suffixes-of-custom-gradle-commands)

Currently when user specifies a custom gradle command that does not end with `*Release` or `*Debug` the command is not parsed correctly. Build type and flavour are not parsed, which causes problems with setting code version and app version down the line.

# How

- Changed the regexp that parses gradle command to be more flexible, allowing any suffix that starts with a capital letter.

# Test Plan

- add a custom gradle command to `eas.json`, for example 
```
 "development": {
      "android": {
        "gradleCommand": ":app:assembleAbcStaging"
      }
    },
```
- add a `buildType` and packageFlavor to `build.gradle` file:
```
productFlavors {  
        abc { // New flavor
            versionCode 3
            versionName "abc"
        }
    }
buildTypes {
        staging {}
        debug {... }
        release {...}
}
```

- start an android build from cli.
- started build should not warn about invalid configuration
- on website, build version should be set to the value assigned in flavour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+-- Allow non-standard build types in custom gradle command. ([#3000](https://github.com/expo/eas-cli/pull/3003) by [@khamilowicz](https://github.com/khamilowicz))
+
 ## [16.4.0](https://github.com/expo/eas-cli/releases/tag/v16.4.0) - 2025-05-02
 
 ### 🎉 New features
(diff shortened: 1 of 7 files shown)


sjchmiela (changes requested):
While we're here, can we:
- make sure we only lowercase the first letter of the build type?
- instead of plain allowing build types other than "release" and "debug", can we print a warning if we see a non-standard build type (hidable with an env var)? This way if someone accidentally sets up `buildCommand: "assembleRelease "`, they'll see the problem more easily (trailing space)
- add a link to Regex to https://developer.android.com/build/build-variants#build-types

sebryu:
Hey @sjchmiela . Is there any chance to land it soon? 
I can volunteer to resolve the conflicts after granting write permissions to repo or I can push a fork with resolved conflicts to speed this up.

sjchmiela (approved):

sebryu (approved):
Looks great!

PR #3004, opened 2025-04-30
[eas-cli] Add missing sourcemap normalization to worker deployments

kitten:
# Why

The deployment API used to take care of this before https://github.com/expo/eas-hosting/pull/184
The previous assumption here was that Cloudflare would now handle sourcemaps with missing `file` and `sourcesRoot` correctly. However, we still need to supply the `file` entry. This seems to be a bug on the Cloudflare site, and I believe this is non-spec complliant 

As per the spec, https://sourcemaps.info/spec.html
> Line 3: An optional name of the generated code that this source map is associated with. 

As per ECMA 426, https://tc39.es/ecma426/#sec-source-map-format
> The file field is an optional name of the [generated code](https://tc39.es/ecma426/#sec-terms-and-definitions-generated-code) that this source map is associated with. It's not specified if this can be an [URL](https://tc39.es/ecma426/#external-whatwg-url), relative path name, or just a base name. Source map generators may choose the appropriate interpretation for their contexts of use.

Typically, this means, omitting `file` means that the targeted file for the sourcemap is the **basename** i.e. the sourcemap's filename without the `.map` extension. However, this seems to break in Cloudflare.

# How

- Readd normalization which was previously on the API layer

# Test Plan

- Tested upload output to be normalized manually

### DIFF
--- a/packages/eas-cli/src/worker/assets.ts
+++ b/packages/eas-cli/src/worker/assets.ts
@@ -142,18 +142,20 @@ export async function createManifestAsync(
 }
 
 interface RawSourceMap {
-    version: number;
-    sources: string[];
-    names: string[];
-    sourceRoot?: string;
-    sourcesContent?: string[];
-    mappings: string;
-    file: string;
+  version: number;
+  sources: string[];
+  names: string[];
+  sourceRoot?: string;
+  sourcesContent?: string[];
+  mappings: string;
+  file: string;
 }
 
 const sourceMapExtensionRe = /\.map$/;
 
-async function normalizeSourceMap(file: RecursiveFileEntry): Promise<WorkerFileEntry | undefined> {
+async function normalizeSourceMapAsync(
+  file: RecursiveFileEntry
+): Promise<WorkerFileEntry | undefined> {
   // This logic normalizes source maps like Wrangler does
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/worker/assets.ts
```diff
@@ -141,6 +141,53 @@ export async function createManifestAsync(
...
+  names: string[];
+  sourceRoot?: string;
+  sourcesContent?: string[];
+  mappings: string;
+  file: string;
+}
+
+const sourceMapExtensionRe = /\.map$/;
+
+async function normalizeSourceMapAsync(
```
douglowder, line 156: Is there any way to add a unit test for this? E.g. https://github.com/cloudflare/workers-sdk/blob/34c71ce9208ffceefe718fc9ae7282ef95e2f2be/packages/wrangler/src/__tests__/deploy.test.ts#L12220

kitten: There is, but I'll close this PR for now :sweat_smile: this was opened as a test, since I had reasonable confidence this would resolve a source map desymbolication issue. However, it turns out that confidence was unfounded entirely and this didn't fix this issue, which isn't even on our end. That means this PR should be unnecessary for now

PR #3014, opened 2025-05-13
[eas-cli] Add support for uploading for ipa files

gabrieldonadel:
# Why

A user requests the ability to upload internal distribution builds in our Canny (https://expo.canny.io/feature-requests/p/upload-local-build-to-eas). Although we can't collect the same information as when a build is generated through EAS, we should allow users to upload IPA files.

# How

Add support for uploading IPA files and update the simulator detection logic. In order to determine if an iOS app was built for a simulator, we should read the `DTPlatformName` value from the main `Info.plist`.

# Test Plan

Tested manually with multiple different file formats 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Add support for uploading iOS internal distribution local builds. ([#3014](https://github.com/expo/eas-cli/pull/3014) by [@gabrieldonadel](https://github.com/gabrieldonadel))
+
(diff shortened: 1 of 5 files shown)


quinlanj (approved):

Lucienzera:
[Q] Even if we upload the ipa files here, the devices would still need to be provisioned right? 

Thanks

gabrieldonadel:
> [Q] Even if we upload the ipa files here, the devices would still need to be provisioned right?
> 
> Thanks

Yes, for a build to be installable, it needs to be signed using the certificates specifics for the device

PR #3022, opened 2025-05-20
Move to yarn 4.9.1

douglowder:
# Why

yarn 1.22.21 is extremely old and has many issues. This change brings the yarn version up to date. Our server packages in universe are already at yarn 4.7.0.

# How

- Modified the top level package.json
- Added required `.yarnrc.yml` and `.yarn/releases`
  - Disabled pnp and set node linker to use node_modules
- Adjustments to scripts
- Adjustments to workflows
- Removed Node 18 workflow

# Test Plan

- CI should pass
  - All workflows seem to be passing now except for the "size" workflow, which is hitting https://github.com/oclif/oclif/issues/759
- These changes did not seem to cause issues locally, the CLI works correctly


quinlanj (approved):

wschurman (approved):
Double check failing `Size/compare` workflow, otherwise lgtm.

douglowder:
@wschurman @quinlanj : this is going to be more complex than it first appeared.

There seems to be no way to make the oclif pack scripts work correctly with any yarn other than v1, unless we upgrade the oclif version we are using. So we have several options:

- Stick with yarn v1
- Move to yarn v4
  - Upgrade oclif dependencies
  - Go through all of the source code and types, to upgrade them to the newer oclif API (lots of changes)
- Move to a different package manager
  - Bun is not supported
  - See if pnpm or npm will work instead

I'll put this PR back in draft mode for now and we can discuss this more after appjs conference.

PR #3030, opened 2025-05-24
[eas-cli] workflow:list and workflow-run:list commands

douglowder:
# Why

Provide the ability to list workflows and workflow runs in the CLI.

```sh
$ eas workflow:list --help
List workflows for the current project

USAGE
  $ eas workflow:list [--json]

FLAGS
  --json  Enable JSON output, non-JSON messages will be printed to stderr.

DESCRIPTION
  List workflows for the current project
```

```sh
$ eas workflow-run:list --help
List workflow runs for the current project

USAGE
  $ eas workflow-run:list [--workflowId <value>] [--status
    ACTION_REQUIRED|CANCELED|FAILURE|IN_PROGRESS|NEW|PENDING_CANCEL|SUCCESS] [--json] [--limit <value>]

FLAGS
  --json                Enable JSON output, non-JSON messages will be printed to stderr.
  --limit=<value>       The number of items to fetch each query. Defaults to 10 and is capped at 100.
  --status=<option>     If present, filter the returned runs to select those with the specified status
                        <options: ACTION_REQUIRED|CANCELED|FAILURE|IN_PROGRESS|NEW|PENDING_CANCEL|SUCCESS>
  --workflowId=<value>  If present, filter the returned runs to select those for the specified workflow ID

DESCRIPTION
  List workflow runs for the current project
```

# How

- New fragments in `graphql/types/App.ts`
- New queries in `graphql/queries/AppQuery.ts`
- New commands in `commands/workflow` and `commands/workflow-run`

# Test Plan

- Tested locally in an EAS project with workflows
- CI should pass


wschurman (commented):
Code looks good. Wonder if we should consider making workflow ID required for the workflow run list (fan-out queries are pretty expensive, not sure how useful it'd be to see runs from multiple workflows interleaved?), and then using the toplevel `WorkflowQuery` instead of going through the app?

Also, what's the oclif topic pattern we use for sub-objects? Should it be:
`eas workflow:run:list` vs `eas workflow-run:list` vs `eas workflow:list-runs` vs `eas workflow --id <id> --runs`

douglowder:
> Code looks good. Wonder if we should consider making workflow ID required for the workflow run list (fan-out queries are pretty expensive, not sure how useful it'd be to see runs from multiple workflows interleaved?), and then using the toplevel `WorkflowQuery` instead of going through the app?
> 

Thanks @wschurman ! I was not aware that `workflowRunsPaginated` was a fan-out, I had thought the runs were indexed on the app ID. I'll take a look in universe to see the details.

The main use case I had in mind was to see all the recent runs, and especially to see if there were too many runs still in process. Next up is a CLI command to cancel workflows (singly or in bulk). It made sense to have the query on the app, as for these use cases, you only care about the one app.

> Also, what's the oclif topic pattern we use for sub-objects? Should it be: `eas workflow:run:list` vs `eas workflow-run:list` vs `eas workflow:list-runs` vs `eas workflow --id <id> --runs`

Good question -- I was attempting to avoid adding a third level, as we only use that pattern for the build number commands.

wschurman:
A fan-out (with limit) should be fine as long as that's the behavior we want (see all the recent runs). I assume that's how the `App.runs` resolver is implemented so your use of it is fine. My comment was more about whether it made sense to interleave runs from all workflows belonging to an app or to just have them be per-workflow. I guess looking at other products it might make sense to see all recent runs, so this fan-out is correct.

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,135 @@
...
+    'list recent workflow runs for this project, with their IDs, statuses, and timestamps';
+
+  static override flags = {
+    /*
+    workflowId: Flags.string({
+      description:
+        'If present, filter the returned runs to select those for the specified workflow ID',
+      required: false,
+    }),
+     */
```
wschurman, line 73: remove until needed

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export type WorkflowResult = {
+  id: string;
+  name?: string | null | undefined;
+  fileName: string | null;
+  createdAt: string | null;
+  updatedAt: string | null;
+};
+
+export default class ProjectWorkflowList extends EasCommand {
```
wschurman, line 15:
```suggestion
export default class WorkflowList extends EasCommand {
```

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,135 @@
...
+        startedAt: edge.node.createdAt,
+        finishedAt,
+        workflowId: edge.node.workflow.id,
+        workflowName: edge.node.workflow.name ?? null,
+        workflowFileName: edge.node.workflow.fileName ?? null,
+      };
+    });
+}
+
+export default class ProjectWorkflowRunList extends EasCommand {
```
wschurman, line 62:
```suggestion
export default class WorkflowRunList extends EasCommand {
```

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  async runAsync(): Promise<void> {
+    const { flags } = await this.parse(ProjectWorkflowRunList);
+    const {
+      projectId,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(ProjectWorkflowRunList, {
+      nonInteractive: true,
+    });
+
```
wschurman, line 101: add call similar to other commands:
```
if (flags.json) {
      enableJsonOutput();
    }
```
then use `printJsonOnlyOutput` for JSON output.

(ditto for list command)

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
     assert(data.app, 'GraphQL: `app` not defined in server response');
     return data.app.byFullName;
   },
+  async byIdWorkflowsAsync(
+    graphqlClient: ExpoGraphqlClient,
+    appId: string
+  ): Promise<AppWorkflowsFragment> {
```
wschurman, line 79: nit: update to return array since method name indicates that it does
```suggestion
  ): Promise<AppWorkflowsFragment['workflows']> {
```

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
...
+            }
+            ${print(AppWorkflowsFragmentNode)}
+          `,
+          { appId },
+          { additionalTypenames: ['App'] }
+        )
+        .toPromise()
+    );
+    assert(data.app, 'GraphQL: `app` not defined in server response');
+    return data.app.byId;
```
wschurman, line 100:
```suggestion
    return data.app.byId.workflows;
```

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
...
+  },
+  async byIdWorkflowRunsAsync(
+    graphqlClient: ExpoGraphqlClient,
+    appId: string,
+    limit?: number
+  ): Promise<AppWorkflowRunsFragment> {
```
wschurman, line 106: ditto

```suggestion
  ): Promise<AppWorkflowRunsFragment['runs']> {
```

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
...
+            }
+            ${print(AppWorkflowRunsFragmentNode)}
+          `,
+          { appId, limit },
+          { additionalTypenames: ['App'] }
+        )
+        .toPromise()
+    );
+    assert(data.app, 'GraphQL: `app` not defined in server response');
+    return data.app.byId;
```
wschurman, line 128:
```suggestion
    return data.app.byId.runs;
```

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,68 @@
...
+      projectId,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(ProjectWorkflowList, {
+      nonInteractive: true,
+    });
+
+    const byId = await AppQuery.byIdWorkflowsAsync(graphqlClient, projectId);
+    if (!byId) {
+      throw new Error(`Could not find project with id: ${projectId}`);
+    }
```
wschurman, line 39: (after other suggested change)
```suggestion
    const workflows = await AppQuery.byIdWorkflowsAsync(graphqlClient, projectId);
```

(`byId`/`workflows` is never falsy, can remove error)

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,135 @@
...
+    .filter(edge => {
+      if (workflowId && edge.node.workflow.id !== workflowId) {
+        return false;
+      }
+      if (workflowFileName && edge.node.workflow.fileName !== workflowFileName) {
+        return false;
+      }
+      if (status && edge.node.status !== status) {
+        return false;
+      }
```
wschurman, line 42: I think we probably want these to take place server-side so that the limit is accurate? (add the filters to `workflowRunsPaginated` query).

I'm thinking it might make sense to also add a query to get a workflow by file name, in which case we could fetch the workflow and then make `workflowRunsPaginated` only take a `workflowId` and `status` (so there's no need for `fileName` filter).

So, concretely:
1. Add new query to `WorkflowQuery`, `byAppIdAndFileName(appId: ID!, fileName: String!): Workflow!`
2. Update `workflowRunsPaginated` `WorkflowRunFilterInput` to take `workflowId` and `status`.

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,135 @@
...
+        'If present, filter the returned runs to select those for the specified workflow ID',
+      required: false,
+    }),
+     */
+    workflow: Flags.string({
+      description:
+        'If present, filter the returned runs to select those for the specified workflow file name',
+      required: false,
+    }),
+    status: Flags.string({
```
wschurman, line 79:
```suggestion
    status: Flags.enum({
```

wschurman (changes requested):

packages/eas-cli/src/commands/workflow/runs.ts
douglowder: @wschurman to really do what you suggest here would require universe changes, I believe (I didn't see support for these filters in the resolvers there).

We can do (1) above with the existing schema, by fetching all workflows for the app and finding the right ID for the filename. However, I don't see any support for (2) taking a status value, unless we actually modify universe.

For now, I'll make the changes to at least limit the query correctly if the filename is passed in.

wschurman: Yep, I think the universe changes are a prerequisite for this PR. I believe all our lists filtering is done server side.

douglowder: Well, I think we can still merge this PR, provided we remove the status filtering. We can add that later once universe supports it.

wschurman: I think we'd need to remove the `workflowName` filtering as well then. For example:
- I have two workflows, A and B
- The last 10 runs are A, followed by 5 B runs.
- If I ran `eas workflow:runs --workflowName B --limit 10` it'd return empty

(I think, if I'm reading this correctly)

douglowder: @wschurman I am now using different queries when getting runs by workflow filename, to avoid client side filtering on filename. I am still filtering client side on status for now.

This should be safe to merge, and once we add the correct server side filters in universe, we can rework these commands to use them.

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
     assert(data.app, 'GraphQL: `app` not defined in server response');
     return data.app.byFullName;
   },
+  async byIdWorkflowsAsync(
+    graphqlClient: ExpoGraphqlClient,
+    appId: string
+  ): Promise<Partial<Workflow>[]> {
```
wschurman, line 79: I think what you had before was more correct as far as nullability goes. My comment was more to just return the array of workflows to match the function name, but returning just `AppWorkflowsFragment` works, too.

```suggestion
  ): Promise<AppWorkflowsFragment['workflows']> {
```
or
```suggestion
  ): Promise<AppWorkflowsFragment> {
```

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
...
+            }
+            ${print(AppWorkflowsFragmentNode)}
+          `,
+          { appId },
+          { additionalTypenames: ['App'] }
+        )
+        .toPromise()
+    );
+    assert(data.app, 'GraphQL: `app` not defined in server response');
+    return (data.app.byId.workflows as Partial<Workflow>[]) ?? [];
```
wschurman, line 100: then no need to cast and no type information is lost via Partial

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,64 @@ export const AppQuery = {
...
+  },
+  async byIdWorkflowRunsAsync(
+    graphqlClient: ExpoGraphqlClient,
+    appId: string,
+    limit?: number
+  ): Promise<Partial<WorkflowRun>[]> {
```
wschurman, line 106: ditto

```suggestion
  ): Promise<AppWorkflowRunsFragment['runs']> {
```
or
```suggestion
  ): Promise<AppWorkflowRunsFragment> {
```
or, if you want to keep the array map and return the array of nodes:
```suggestion
  ): Promise<AppWorkflowRunsFragment['runs']['edges'][number]['node'][]> {
```

this way no type information is lost and partial doesn't purposefully reduce typesafety

packages/eas-cli/src/graphql/queries/WorkflowQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+import { withErrorHandlingAsync } from '../client';
+import { WorkflowRun, WorkflowRunsForWorkflowIdQuery } from '../generated';
+import { WorkflowRunsFragmentNode } from '../types/Workflow';
+
+export const WorkflowQuery = {
+  async byIdRunsAsync(
+    graphqlClient: ExpoGraphqlClient,
+    workflowId: string,
+    limit?: number
+  ): Promise<Partial<WorkflowRun>[]> {
```
wschurman, line 16:
```suggestion
  ): Promise<WorkflowRunsFragment['runs']['edges'][number]['node'][]> {
```

packages/eas-cli/src/graphql/queries/WorkflowQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+            }
+            ${print(WorkflowRunsFragmentNode)}
+          `,
+          { workflowId, limit },
+          { additionalTypenames: ['Workflow'] }
+        )
+        .toPromise()
+    );
+    assert(data.workflows, 'GraphQL: `workflows` not defined in server response');
+    return data.workflows.byId.runs.edges.map(edge => edge.node as Partial<WorkflowRun>) ?? [];
```
wschurman, line 38:
```suggestion
    return data.workflows.byId.runs.edges.map(edge => edge.node);
```

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
+import EasCommand from '../../commandUtils/EasCommand';
+import { EasJsonOnlyFlag } from '../../commandUtils/flags';
+import { AppQuery } from '../../graphql/queries/AppQuery';
+import Log from '../../log';
+import formatFields from '../../utils/formatFields';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+export type WorkflowResult = {
```
wschurman, line 8:
```suggestion
type WorkflowResult = {
```

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
...
+    const result: WorkflowResult[] = workflows.map(workflow => ({
+      id: workflow.id ?? null,
+      name: workflow.name ?? null,
+      fileName: workflow.fileName ?? null,
+      createdAt: workflow.createdAt ?? null,
+      updatedAt: workflow.updatedAt ?? null,
+    }));
+
+    if (flags.json) {
+      enableJsonOutput();
```
wschurman, line 48: Looking at the other commands, most tend to call this right after `getContext` in order to preempt any accidental console logs in code before the command calls `printJsonOnlyOutput`.

Prefer to do a call to it on L36 separately from the `printJsonOnlyOutput` call:
```
if (flags.json) {
  enableJsonOutput();
}
```

I think we do this as a defensive measure in case something like `AppQuery.byIdWorkflowsAsync` were to accidentally output in a future revision.

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,139 @@
...
+      nonInteractive: true,
+    });
+
+    const workflowFileName = flags.workflow;
+    const status = flags.status;
+    const limit = flags.limit ?? 10;
+
+    let runs: Partial<WorkflowRun>[] = [];
+    if (workflowFileName) {
+      const workflows = await AppQuery.byIdWorkflowsAsync(graphqlClient, projectId);
```
wschurman, line 98: While definitely an improvement over the fan-out, doing a full list load in order to find a single object is still decently slow. I think it should be pretty quick and easy to add a `byAppIdAndFileName(appId: ID!, fileName: String!): Workflow!` resolver to `WorkflowQuery`? Or is there a reason we don't want to add it (I'm not as familiar with the data model)? (cc @sjchmiela).

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,139 @@
...
+        workflow => workflow.fileName === workflowFileName
+      );
+      if (workflowsFiltered.length > 1) {
+        throw new Error(`Found multiple workflows with the same file name: ${workflowFileName}`);
+      }
+      if (!workflowsFiltered.length || !workflowsFiltered[0].id) {
+        Log.warn(`No workflows found with file name: ${workflowFileName}`);
+      } else {
+        const workflowId = workflowsFiltered[0].id;
+        runs = await WorkflowQuery.byIdRunsAsync(graphqlClient, workflowId ?? '', limit);
```
wschurman, line 109: can avoid needing the nullish coalescing operator with a slight refactor

```suggestion
      const workflowId = workflowsFiltered[0]?.id;
      if (!workflowId) {
        Log.warn(`No workflows found with file name: ${workflowFileName}`);
      } else {
        runs = await WorkflowQuery.byIdRunsAsync(graphqlClient, workflowId, limit);
      }
```

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+    } else {
+      runs = await AppQuery.byIdWorkflowRunsAsync(graphqlClient, projectId, limit);
+    }
+
+    const result = processWorkflowRuns(runs, { workflowFileName, status });
+
+    if (flags.json) {
+      enableJsonOutput();
```
wschurman, line 118: ditto as other command

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,139 @@
...
+      printJsonOnlyOutput(result);
+      return;
+    }
+
+    Log.addNewLineIfNone();
+    result.forEach(run => {
+      Log.log(
+        formatFields([
+          { label: 'Run ID', value: run.id ?? '-' },
```
wschurman, line 127: I think runs should always have an ID? (ditto with a lot of the other fields below)

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
+import EasCommand from '../../commandUtils/EasCommand';
+import { EasJsonOnlyFlag } from '../../commandUtils/flags';
+import { AppQuery } from '../../graphql/queries/AppQuery';
+import Log from '../../log';
+import formatFields from '../../utils/formatFields';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+export type WorkflowResult = {
+  id: string | null;
```
wschurman, line 9: I think every workflow should have an ID and fileName and createdAt?

wschurman (changes requested):
I agree 100% that when a workflow file is provided it should do a different query, thanks for bringing that up! But also I think the part of my initial comment about adding a `byAppIdAndFileName` is still the path forward for the new pattern so that we're just loading one row instead of N rows in order to find one row. If the other reviewers disagree or there is a bit of context I'm missing about why we can't add it, I'm definitely open to having my mind changed.

Other than that, some comments about adhering to existing patterns around JSON output (preemptively calling `enableJsonOutput` for safety), as well as some typescript type comments to retain non-nullability of a lot of fields (avoiding `Partial`).

packages/eas-cli/src/commands/workflow/runs.ts
douglowder: Ok I've added it, see https://github.com/expo/universe/pull/20312 . Once that is merged, I'll rework this PR to use it, thanks @wschurman

douglowder:
> Other than that, some comments about adhering to existing patterns around JSON output (preemptively calling `enableJsonOutput` for safety), as well as some typescript type comments to retain non-nullability of a lot of fields (avoiding `Partial`).

@wschurman I've reworked the code to use the new GraphQL queries and filters added to universe, and resolved most of the issues you called out.

packages/eas-cli/src/graphql/queries/WorkflowRunQuery.ts
```diff
@@ -84,4 +92,92 @@ export const WorkflowRunQuery = {
...
+  ): Promise<WorkflowRunFragment[]> {
+    validateLimit(limit);
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .query<WorkflowRunsForAppIdFileNameAndStatusQuery>(
+          gql`
+            query WorkflowRunsForAppIdFileNameAndStatusQuery(
+              $appId: ID!
+              $fileName: String!
+              $status: WorkflowRunStatus!
```
wschurman: I think we can just make status optional and then we can combine these two queries (`byAppIdAndFileNameAsync` and `byAppIdFileNameAndStatusAsync`)?

```suggestion
    status?: WorkflowRunStatus,
    limit?: number
  ): Promise<WorkflowRunFragment[]> {
    validateLimit(limit);
    const data = await withErrorHandlingAsync(
      graphqlClient
        .query<WorkflowRunsForAppIdFileNameAndStatusQuery>(
          gql`
            query WorkflowRunsForAppIdFileNameAndStatusQuery(
              $appId: ID!
              $fileName: String!
              $status: WorkflowRunStatus
```

packages/eas-cli/src/graphql/queries/AppQuery.ts
```diff
@@ -60,4 +73,114 @@ export const AppQuery = {
...
+    limit?: number
+  ): Promise<WorkflowRunFragment[]> {
+    validateLimit(limit);
+    const data = await withErrorHandlingAsync(
+      graphqlClient
+        .query<AppByIdWorkflowRunsFilteredByStatusQuery>(
+          gql`
+            query AppByIdWorkflowRunsFilteredByStatusQuery(
+              $appId: String!
+              $status: WorkflowRunStatus!
```
wschurman: ditto, should be able to merge these queries (I think?):


```suggestion
    status?: WorkflowRunStatus,
    limit?: number
  ): Promise<WorkflowRunFragment[]> {
    validateLimit(limit);
    const data = await withErrorHandlingAsync(
      graphqlClient
        .query<AppByIdWorkflowRunsFilteredByStatusQuery>(
          gql`
            query AppByIdWorkflowRunsFilteredByStatusQuery(
              $appId: String!
              $status: WorkflowRunStatus
```

packages/eas-cli/src/graphql/queries/WorkflowRunQuery.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable graphql/template-strings */
```
wschurman: remove? (lint passes without it)
```suggestion
```

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
...
+  name?: string | null | undefined;
+  fileName: string | null;
+  createdAt: string | null;
+  updatedAt: string | null;
+};
```
wschurman:
```suggestion
type WorkflowResult = {
  id: string;
  name?: string | null | undefined;
  fileName: string;
  createdAt: string;
  updatedAt: string;
};
```

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const workflows = await AppQuery.byIdWorkflowsAsync(graphqlClient, projectId);
+
+    const result: WorkflowResult[] = workflows.map(workflow => ({
+      id: workflow.id ?? null,
+      name: workflow.name ?? null,
+      fileName: workflow.fileName ?? null,
+      createdAt: workflow.createdAt ?? null,
+      updatedAt: workflow.updatedAt ?? null,
+    }));
```
wschurman:
```suggestion
    const result: WorkflowResult[] = workflows.map(workflow => ({
      id: workflow.id,
      name: workflow.name,
      fileName: workflow.fileName,
      createdAt: workflow.createdAt,
      updatedAt: workflow.updatedAt,
    }));
```

packages/eas-cli/src/commands/workflow/list.ts
wschurman: Same comment applies for this revision.

packages/eas-cli/src/commands/workflow/list.ts
```diff
@@ -0,0 +1,67 @@
...
+    Log.addNewLineIfNone();
+    result.forEach(workflow => {
+      Log.log(
+        formatFields([
+          { label: 'ID', value: workflow.id ?? '-' },
+          { label: 'Name', value: workflow.name ?? '-' },
+          { label: 'File name', value: workflow.fileName ?? '-' },
+          { label: 'Created At', value: workflow.createdAt ?? '-' },
+          { label: 'Updated At', value: workflow.updatedAt ?? '-' },
+        ])
```
wschurman:
```suggestion
        formatFields([
          { label: 'ID', value: workflow.id },
          { label: 'Name', value: workflow.name ?? '-' },
          { label: 'File name', value: workflow.fileName },
          { label: 'Created At', value: workflow.createdAt },
          { label: 'Updated At', value: workflow.updatedAt },
        ])
```

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -225,7 +225,7 @@ async function waitForWorkflowRunToEndAsync(
           spinner.warn('Workflow run is waiting for action.');
           break;
 
-        case WorkflowRunStatus.PendingCancel:
+        // case WorkflowRunStatus.PendingCancel:
```
wschurman: realize this is out of the scope of this PR, but let's just remove this case. @sjchmiela removed it from the API

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,151 @@
...
+        gitCommitMessage: run.gitCommitMessage ?? null,
+        gitCommitHash: run.gitCommitHash ?? null,
+        startedAt: run.createdAt ?? null,
+        finishedAt,
+        workflowId: run.workflow?.id ?? null,
+        workflowName: run.workflow?.name ?? null,
+        workflowFileName: run.workflow?.fileName ?? null,
+      };
+    });
+}
```
wschurman: no need for client side filtering any more, also can improve nullability of types that are never null now that the common fragment input is used

```suggestion
type WorkflowRunResult = {
  id: string;
  status: string;
  gitCommitMessage: string | null;
  gitCommitHash: string | null;
  startedAt: string;
  finishedAt: string;
  workflowId: string;
  workflowName: string | null;
  workflowFileName: string;
};

function processWorkflowRuns(runs: WorkflowRunFragment[]): WorkflowRunResult[] {
  return runs.map(run => {
    const finishedAt = run.status === WorkflowRunStatus.InProgress ? null : run.updatedAt;
    return {
      id: run.id,
      status: run.status,
      gitCommitMessage: run.gitCommitMessage ?? null,
      gitCommitHash: run.gitCommitHash ?? null,
      startedAt: run.createdAt,
      finishedAt,
      workflowId: run.workflow.id,
      workflowName: run.workflow.name ?? null,
      workflowFileName: run.workflow.fileName,
    };
  });
}
```

packages/eas-cli/src/commands/workflow/runs.ts
wschurman: Same comment applies

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,151 @@
...
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(WorkflowRunList, {
+      nonInteractive: true,
+    });
+
+    const workflowFileName = flags.workflow;
+    const status = flags.status;
+    const limit = flags.limit ?? 10;
+
+    let runs: WorkflowRunFragment[] = [];
```
wschurman:
```suggestion
    let runs: WorkflowRunFragment[];
```
(so typescript catches fallthrough cases, which we don't want to allow)

packages/eas-cli/src/commands/workflow/runs.ts
```diff
@@ -0,0 +1,151 @@
...
+      Log.log(
+        formatFields([
+          { label: 'Run ID', value: run.id ?? '-' },
+          { label: 'Workflow', value: run.workflowFileName ?? '-' },
+          { label: 'Status', value: run.status ?? '-' },
+          { label: 'Started At', value: run.startedAt ?? '-' },
+          { label: 'Finished At', value: run.finishedAt ?? '-' },
+          { label: 'Git Commit Message', value: run.gitCommitMessage ?? 'null' },
+          { label: 'Git Commit Hash', value: run.gitCommitHash ?? 'null' },
+        ])
```
wschurman:
```suggestion
        formatFields([
          { label: 'Run ID', value: run.id },
          { label: 'Workflow', value: run.workflowFileName },
          { label: 'Status', value: run.status },
          { label: 'Started At', value: run.startedAt },
          { label: 'Finished At', value: run.finishedAt },
          { label: 'Git Commit Message', value: run.gitCommitMessage ?? 'null' },
          { label: 'Git Commit Hash', value: run.gitCommitHash ?? 'null' },
        ])
```

wschurman (changes requested):
Looks solid with the new queries (functionally correct status filtering!).

Main comments that remain:
- Can combine the queries by making status optional (graphql understands undefined I think)
- Can improve nullability of fields of the intermediate types (`WorkflowResult`, `WorkflowRunResult`) now that all operate on common fragment types that have non-optional fields (instead of previously when `Partial` made types optional).
- Same comment about needing to move the call to `enableJsonOutput()` to higher up in the command method (separate from `printJsonOnlyOutput` call). Should follow the pattern used in most/all other commands due to the side-effect nature of this function call.

wschurman (approved):
LGTM. Double check the lint failures before landing.

PR #3031, opened 2025-05-25
[eas-json] Allow a string of 1-50 characters as a track name

algoORgoal:
# Why

Previously, the Android submit track in eas.json only allowed a fixed set of values ("beta", "alpha", "internal", "production"). However, Google Play supports custom track names, and users may need to specify arbitrary track names for advanced workflows. This PR enables more flexible usage by allowing any string of 1-50 characters as a track name.
Track name follows the rule defined by Google textfield. See the image below.

1. When textfield is empty
    Empty Trackname textfield
2. When textfield has an input within 50 characters
    Track name textfield within 50 characters
3. When textfied has an input over 50 characters
    Track name textfield over 50 characters

# How

- Updated the Joi runtime schema in `packages/eas-json/src/submit/schema.ts` to allow any string of 1-50 characters for the Android submit track property.
- Updated the JSON schema in `packages/eas-json/schema/eas.schema.json` to add `minLength` and `maxLength` for the track property, ensuring IDE and documentation support matches runtime validation.
- Added and updated tests in `packages/eas-json/src/__tests__/submitProfiles-test.ts` to cover:
  - Valid custom track names within 50 characters
  - Invalid track names (empty string, over 50 characters)
  - Default value behavior when track is omitted

# Test Plan

- Ran the test suite to ensure all new and existing tests pass.
- Verified that:
  - A custom track name (e.g., `my-custom-track`) is accepted.
  - An empty string or a string longer than 50 characters is rejected.
  - Omitting the track property defaults to `internal`.

If possible, please also provide terminal output and/or screenshots demonstrating your test/reproduction.
Test results on terminal

### DIFF
--- a/packages/eas-json/src/__tests__/submitProfiles-test.ts
+++ b/packages/eas-json/src/__tests__/submitProfiles-test.ts
@@ -100,6 +100,99 @@ test('android config with serviceAccountKeyPath set to env var', async () => {
   }
 });
 
+test('android config without track name', async () => {
+  await fs.writeJson('/project/eas.json', {
+    submit: {
+      production: {
+        android: {
+          serviceAccountKeyPath: './path.json',
+          releaseStatus: 'completed',
+        },
+      },
+    },
+  });
+
+  const accessor = EasJsonAccessor.fromProjectPath('/project');
+  const androidProfile = await EasJsonUtils.getSubmitProfileAsync(
+    accessor,
+    Platform.ANDROID,
+    'production'
+  );
+
+  expect(androidProfile).toEqual({
+    serviceAccountKeyPath: './path.json',
+    track: 'internal',
+    releaseStatus: 'completed',
+    changesNotSentForReview: false,
+  });
+});
+
+test('android config with non-empty string track name within 50 characters', async () => {
+  await fs.writeJson('/project/eas.json', {
+    submit: {
+      production: {
+        android: {
+          serviceAccountKeyPath: './path.json',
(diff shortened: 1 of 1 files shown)


algoORgoal:
/changelog-entry [bug-fix] [Allow a string of 1-50 characters as a track name]

algoORgoal:
@quinlanj  I tried to follow the pr description to update the changelog, but it seems like it's not working. Can you get an assistance? Thank you.

algoORgoal:
/changelog-entry bug-fix Allow a string of 1-50 characters as a track name

kadikraman:
Thanks for the PR. This also requires some API schema changes on our part so we'll need to tackle those first.

In the meanwhile since the actual change is quite small, you can unblock yourself by adding the change on your local project with [patch-package](https://github.com/ds300/patch-package) (or [yarn patch](https://yarnpkg.com/cli/patch) if you're using yarn, [bun patch](https://bun.sh/docs/install/patch) if you're using bun etc)

kibs:
@kadikraman it seems this was fixed in the eas-build-job, some time ago. Any chances for this to be merged to main?

sjchmiela:
I think respective changes landed in https://github.com/expo/eas-cli/pull/3163. Backend should also support free-form track names. Do you still see issues when trying to use them with latest `eas-cli`?

sjchmiela:
Closed in https://github.com/expo/eas-cli/pull/3163

kibs:
Perfect, will test!

PR #3048, opened 2025-06-09
[eas-cli] Implement workflow:cancel

douglowder:
# Why

Allow user to cancel one or more workflow runs, as requested in @jonsamp 's [workflow CLI design document](https://www.notion.so/expo/EAS-CLI-workflow-commands-8bc5ccc4f9f949c8adfec1347817113c).

The CLI will print a separate message (success or error message) for each workflow run ID that it attempts to cancel.

```bash
$ eas workflow:cancel --help
Cancel one or more workflow runs. If no workflow run IDs are provided, you will be prompted to select IN_PROGRESS runs to cancel.

USAGE
  $ eas workflow:cancel [--non-interactive]

FLAGS
  --non-interactive  Run the command in non-interactive mode.

DESCRIPTION
  Cancel one or more workflow runs. If no workflow run IDs are provided, you will be prompted to select IN_PROGRESS runs
  to cancel.

```

Example:

```bash
$ eas workflow:cancel
? Select IN_PROGRESS workflow runs to cancel ›  
Instructions:
    ↑/↓: Highlight option
    ←/→/[space]: Toggle selection
    a: Toggle all
    enter/return: Complete answer
◯   01975c6b-3ec2-73ca-b117-987b434056e7 - updates-e2e-enabled.yml, Publish packages, 2025-06-11T00:37:01.762Z
◯   01975c67-7950-74a8-8771-6e2abdd28a82 - updates-e2e-enabled.yml, sync changelog (#37353), 2025-06-11T00:32:54.608Z
◯   01975c66-1c3c-784b-8c4a-f7f95414dbf4 - updates-e2e-enabled.yml, bump expo-image versions, 2025-06-11T00:31:25.244Z
```
# How

- Added a new mutation to cancel a single workflow run by ID
- Added the new command

# Test Plan

- CI should pass

### DIFF
--- a/packages/eas-cli/src/commands/workflow/cancel.ts
+++ b/packages/eas-cli/src/commands/workflow/cancel.ts
@@ -1,13 +1,14 @@
-import fs from 'fs';
-
-import { WorkflowRunResult } from './runs';
 import EasCommand from '../../commandUtils/EasCommand';
+import { processWorkflowRuns } from '../../commandUtils/workflows';
+import { WorkflowRunStatus } from '../../graphql/generated';
 import { WorkflowRunMutation } from '../../graphql/mutations/WorkflowRunMutation';
+import { AppQuery } from '../../graphql/queries/AppQuery';
 import Log from '../../log';
+import { promptAsync } from '../../prompts';
 
 export default class WorkflowRunCancel extends EasCommand {
   static override description =
-    'Cancel one or more workflow runs. Pass in the --fromJson flag to cancel all runs from a JSON file created with `eas workflow:runs --json`.';
+    'Cancel one or more workflow runs. If no workflow run IDs are provided, you will be prompted to select IN_PROGRESS runs to cancel.';
 
   static override strict = false;
 
@@ -19,6 +20,7 @@ export default class WorkflowRunCancel extends EasCommand {
   async runAsync(): Promise<void> {
     const { argv } = await this.parse(WorkflowRunCancel);
     const {
+      projectId,
       loggedIn: { graphqlClient },
     } = await this.getContextAsync(WorkflowRunCancel, {
       nonInteractive: true,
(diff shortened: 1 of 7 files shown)


douglowder:
> Since this cancels one at a time anyways (graphql in the for loop), I want to say JSON parsing should be left to tools like `jq` and then piped to `eas` via xargs or something (which IMO is the standard for tasks like this).
> 
> If we did want to do parsing like this in the command, we probably would want to do so very deliberately independently of this feature and do it for all of our single-object commands. i.e. if we add it to this, users would most likely expect it to work for things like `build:cancel`, `build:download`, etc..., taking the input from `eas build:list`

Thanks @wschurman ! After some more thought, it seems more useful to actually query for IN_PROGRESS runs and present a prompt when the user does not pass in any run IDs. I've reworked the command to do that (see example in the PR description).

packages/eas-cli/src/commands/workflow/cancel.ts
```diff
@@ -0,0 +1,83 @@
...
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    const { argv } = await this.parse(WorkflowRunCancel);
+    const {
+      projectId,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(WorkflowRunCancel, {
+      nonInteractive: true,
```
wschurman: this command is no longer non-interactive, so we should probably add the non-interactive flag and respect it for the prompt below
```suggestion
      nonInteractive,
```

packages/eas-cli/src/commands/workflow/cancel.ts
```diff
@@ -0,0 +1,83 @@
...
+  static override description =
+    'Cancel one or more workflow runs. If no workflow run IDs are provided, you will be prompted to select IN_PROGRESS runs to cancel.';
+
+  static override strict = false;
+
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectId,
+    ...this.ContextOptions.LoggedIn,
+  };
+
```
wschurman, line 23: static override flags = {
    ...EASNonInteractiveFlag,
  };

packages/eas-cli/src/commands/workflow/cancel.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Custom parsing of argv
+    const tokens = [...argv];
+    const workflowRunIds: Set<string> = new Set();
+    if (tokens.length > 0) {
+      tokens.forEach(token => {
+        workflowRunIds.add(token);
+      });
+    } else {
+      // Run the workflow run list query and select runs to cancel
```
wschurman, line 52: and then require nonInteractive === false here, otherwise throw:


```suggestion
      if (nonInteractive) {
        throw new Error('Must supply workflow run IDs as arguments when in non-interactive mode`);
      }

      // Run the workflow run list query and select runs to cancel
```

wschurman (approved):

PR #3084, opened 2025-07-01
[eas-cli] no warning in update:configure when no build profiles

brentvatne:
# Why

I was writing some instructions for testing an updates feature and when I ran `eas update:configure` in the project it threw a TypeError:

```
eas update:configure
💡 The following process will configure your project to use EAS Update. These changes only apply to your local project files and you can safely revert them at any time.
✔ Overwrote updates.url "https://u.expo.dev/69109f5f-9770-496f-b74f-c85277aa217d" with "https://updates-test-proxy.expo.app/69109f5f-9770-496f-b74f-c85277aa217d"

All builds of your app going forward will be eligible to receive updates published with EAS Update.

We were not able to configure eas.json. Error: TypeError: Cannot convert undefined or null to object.

🎉 Your app is configured to use EAS Update!

- Learn more about other capabilities of EAS Update: https://docs.expo.dev/eas-update/introduction/
```

It seems we assume that if you have an **eas.json** then you must be using EAS Build, but that is no longer a valid assumption (you may be using it for just submit, or for configuring the CLI for update).

# How

Bail out when no `build` key rather than erroring. Remove warning about running `eas build:configure` when no **eas.json** is present.

# Test Plan

Ran it locally, no type error printed.

### DIFF
--- a/packages/eas-cli/src/update/configure.ts
+++ b/packages/eas-cli/src/update/configure.ts
@@ -305,20 +305,22 @@ async function ensureEASUpdateIsConfiguredNativelyAsync(
 export async function ensureEASUpdateIsConfiguredInEasJsonAsync(projectDir: string): Promise<void> {
   const easJsonPath = EasJsonAccessor.formatEasJsonPath(projectDir);
 
+  // Skip if eas.json doesn't exist.
   if (!(await fs.pathExists(easJsonPath))) {
-    Log.warn(
-      `EAS Build is not configured. If you'd like to use EAS Build with EAS Update, run ${chalk.bold(
-        'eas build:configure'
-      )}.`
-    );
     return;
   }
 
+  // Add channel to all build profiles.
   try {
     const easJsonAccessor = EasJsonAccessor.fromProjectPath(projectDir);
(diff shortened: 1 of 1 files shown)


quinlanj (approved):

packages/eas-cli/src/update/configure.ts
```diff
@@ -305,20 +305,22 @@ async function ensureEASUpdateIsConfiguredNativelyAsync(
 export async function ensureEASUpdateIsConfiguredInEasJsonAsync(projectDir: string): Promise<void> {
   const easJsonPath = EasJsonAccessor.formatEasJsonPath(projectDir);
 
+  // Skip if eas.json doesn't exist.
   if (!(await fs.pathExists(easJsonPath))) {
-    Log.warn(
-      `EAS Build is not configured. If you'd like to use EAS Build with EAS Update, run ${chalk.bold(
-        'eas build:configure'
-      )}.`
-    );
```
wschurman, line 313: I think we still want something like this warning, maybe just a more precise detection of whether EAS Build is configured. i.e. change the detection to be more than just `eas.json` being present.

PR #3085, opened 2025-07-01
feat(eas-cli): Implement new batched EAS Hosting asset upload protocol

kitten:
# Why

This implements the new EAS Hosting asset upload protocol. This should improve parallel asset upload performance by grouping asset uploads into fewer requests. Because some assets can be small, the cost associated with making requests (even in parallel, and even with `keepAlive`) from scratch is likely high and outstrips the performance gains from the parallelization we have.

# How

The new EAS Hosting assets protocol, when active, changes the `/deploy` endpoint to return an `upload` result paramter. This contains batches of SHA-512 arrays. When this parameter is received, `eas-cli` _may_ use this parameter to send the assets as batches instead.

Each batch is encoded as a `multipart/form-data` request payload. The payload contains a non-standard `Content-Length` header for each part, which switches modes in the `deployment-api`, allowing it to complete the upload a little faster.

# Test Plan

- Manually test this against the new protocol in staging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🧹 Chores
 
+- Update the EAS Hosting worker deployment progress indicator to use a progress bar for asset uploads instead ([#3099](https://github.com/expo/eas-cli/pull/3099) by [@kitten](https://github.com/kitten))
+
 ## [16.14.1](https://github.com/expo/eas-cli/releases/tag/v16.14.1) - 2025-07-08
 
 ### 🧹 Chores
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Implement new EAS Hosting asset upload protocol, which improves performance by grouping files into batched upload API calls ([#3085](https://github.com/expo/eas-cli/pull/3085) by [@kitten](https://github.com/kitten))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/commands/deploy/index.ts
```diff
@@ -32,6 +32,8 @@ import {
   getDeploymentUrlFromFullName,
 } from '../../worker/utils/logs';
 
+const MAX_UPLOAD_SIZE = 5e8; // 5MB
```
kadikraman: `5e8` is 500MB (the comment is incorrect). Also could we name this sth like `MAX_TOTAL_UPLOAD_SIZE` or add a comment? At first glance it seemed like a per-asset limit.

kadikraman (commented):
Due to the nature of multipart uploads, the progress indicator becomes less informative since we have a lot of parallelisation now (I cropped the first 20 seconds of the video but it basically stays at 0% for ages)

As discussed, we should change this to be either a progress bar or a `x / y mb uploaded` text similar to EAS Build.

https://github.com/user-attachments/assets/5b030953-f553-4edf-bdc4-2bcfe5814f4b

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Implement new EAS Hosting asset upload protocol, which improves performance by grouping files into batched upload API calls ([#3085](https://github.com/expo/eas-cli/pull/3085) by [@kitten](https://github.com/kitten))
```
kadikraman, line 11: Do we need an entry for the new progress bar as well? Looks like all hosting deployments will have this now.

kadikraman (approved):

CHANGELOG.md
kitten: Hm, I assumed since the progress bar PR was merged into this PR, it should be fine to ship under one changelog entry? Not sure what our policy for this is. To me, it's basically a tweak that'll make the asset upload protocol change more obvious

Issue #3091, opened 2025-07-06
GraphQL generation fails due to service API change


douglowder:
### Summary

The generation of GraphQL code (`yarn --cwd packages/eas-cli generate-graphql-code`) fails due to a change in the Expo services GraphQL API (https://github.com/expo/universe/pull/20817).

- 

### Managed or bare?

Managed

### Environment

You can see the problem in this test run: https://github.com/expo/eas-cli/actions/runs/16092986663/job/45412059171?pr=3090

### Error output

https://github.com/expo/eas-cli/actions/runs/16092986663/job/45412059171?pr=3090

### Reproducible demo or steps to reproduce from a blank project

https://github.com/expo/eas-cli/actions/runs/16092986663/job/45412059171?pr=3090

PR #3092, opened 2025-07-06
[eas-cli] Remove link and unlink commands

kadikraman:
# Why

Resolves https://github.com/expo/eas-cli/issues/3091

We removed the unreleased link environments feature from the website and api, but omitted removing the code from here.

# How

Remove the `link` and `unlink` commands, and the corresponding API code.

# Test Plan

- tests still pass
- lint and typescheck passes
- `yarn generate-graphql-code` works


packages/eas-cli/src/utils/statuspageService.ts
```diff
@@ -26,6 +26,7 @@ const humanReadableServiceName: Record<StatuspageServiceName, string> = {
   [StatuspageServiceName.EasUpdate]: 'EAS Update',
   [StatuspageServiceName.GithubApiRequests]: 'GitHub API Requests',
   [StatuspageServiceName.GithubWebhooks]: 'Github Webhooks',
+  [StatuspageServiceName.EasWorkflows]: 'EAS Workflows',
```
kadikraman, line 29: Added this because after updating the gql code with `yarn generate-graphql-code` the typecheck for this failed because a new enum has been added.

wschurman (approved):

PR #3093, opened 2025-07-06
Workflow to test and verify GraphQL

douglowder:
# Why

# How

# Test Plan


douglowder:
* **#3093** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3093?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

douglowder:
Verified that workflow fails as expected on a GraphQL change without checking in generated files. https://github.com/expo/eas-cli/runs/45432973916

packages/eas-cli/scripts/verify-graphql.sh
```diff
@@ -0,0 +1,30 @@
...
+trap cleanup EXIT
+
+cp graphql.schema.json graphql.schema.json.bak
+cp src/graphql/generated.ts src/graphql/generated.ts.bak
+
+yarn generate-graphql-code
+if cmp -s graphql.schema.json graphql.schema.json.bak; then
+    echo "GraphQL schema is the same"
+    if cmp -s src/graphql/generated.ts src/graphql/generated.ts.bak; then
+        echo "GraphQL generated code is the same"
```
wschurman:
```suggestion
        echo "GraphQL generated code is up-to-date"
```

packages/eas-cli/scripts/verify-graphql.sh
```diff
@@ -0,0 +1,30 @@
...
+set -eox pipefail
+# If this script exits, trap it first and clean up
+trap cleanup EXIT
+
+cp graphql.schema.json graphql.schema.json.bak
+cp src/graphql/generated.ts src/graphql/generated.ts.bak
+
+yarn generate-graphql-code
+if cmp -s graphql.schema.json graphql.schema.json.bak; then
+    echo "GraphQL schema is the same"
```
wschurman:
```suggestion
    echo "GraphQL schema is up-to-date"
```

packages/eas-cli/scripts/verify-graphql.sh
```diff
@@ -0,0 +1,30 @@
...
+    if cmp -s src/graphql/generated.ts src/graphql/generated.ts.bak; then
+        echo "GraphQL generated code is the same"
+        exit 0
+    else
+        echo "GraphQL generated code has changed, please commit the changes"
+        exit 1
+    fi
+else
+    echo "GraphQL schema has changed, please commit the changes"
```
wschurman:
```suggestion
    echo "GraphQL schema has changed on the server. Run `yarn generate-graphql-code` and commit the changes."
```

packages/eas-cli/scripts/verify-graphql.sh
```diff
@@ -0,0 +1,30 @@
...
+cp src/graphql/generated.ts src/graphql/generated.ts.bak
+
+yarn generate-graphql-code
+if cmp -s graphql.schema.json graphql.schema.json.bak; then
+    echo "GraphQL schema is the same"
+    if cmp -s src/graphql/generated.ts src/graphql/generated.ts.bak; then
+        echo "GraphQL generated code is the same"
+        exit 0
+    else
+        echo "GraphQL generated code has changed, please commit the changes"
```
wschurman:
```suggestion
        echo "GraphQL code has changed but has not been regenerated. Run `yarn generate-graphql-code` and commit the changes.""
```

.github/workflows/test-graphql.yml
```diff
@@ -0,0 +1,29 @@
+name: Test GraphQL generation
```
wschurman: my suggestion might be a bit long, idk

```suggestion
name: Ensure GraphQL schema and generated code is up-to-date
```

wschurman (changes requested):
I like the broad idea here: catch cases where we may have unintentionally made an erroneous breaking change to our graphql.

For most GraphQL APIs it is fine for a schema not to be up-to-date since by definition GraphQL is designed to be backwards compatible, and normally non-backwards-compatible changes are disallowed in most graphql APIs. We treat our first-party (non-public) graphql API slightly differently since we can know if something is going to be used, still being used, and we can remove things that aren't in use. For this particular case with environment variable stuff being removed, I think it's unfortunately just kinda one of those things we've become used to as people writing code in website and CLI (stale graphql schema and code needing to be regenerated). In this case, it did require removing an unused/hidden query which is slightly abnormal (we got lucky that it wasn't used but could tell on the server that it wasn't since there were no rows in the DB).

So, what does this all mean: if done correctly (this case was not quite as smooth but still within the bounds of what we consider normal occurrences), we can add or remove stuff from the server and it doesn't necessarily mean this is broken, just stale, which is valid for a GraphQL client. But staleness (as this test is written currently) will also happen every time the GraphQL changes on the server, which is just about every day, so this test will fail every day (if cron is kept) and will fail on a large portion of PRs.

My proposal is to figure out a way to narrow the scope of the condition this is trying to test to only catch cases that cause `tsc`/generation errors. So:
1. Run generate. If it fails, fail the test. (would have caught this case appropriately).
2. Run `tsc`, if it fails, fail the test (catches cases that may unintentionally change types requiring manual attention or a server revert).
3. Otherwise, pass the test. (so that we don't get false-positives for normal schema changes like adding new fields).

packages/eas-cli/scripts/verify-graphql.sh
```diff
@@ -0,0 +1,30 @@
...
+    if cmp -s src/graphql/generated.ts src/graphql/generated.ts.bak; then
+        echo "GraphQL generated code is up-to-date"
+        exit 0
+    else
+        echo "GraphQL code has changed but has not been regenerated. Run `yarn generate-graphql-code` and commit the changes.""
+        exit 1
+    fi
+else
+    echo "GraphQL schema has changed on the server. Run `yarn generate-graphql-code` and commit the changes."
+    exit 1
```
douglowder:
```suggestion
    echo "GraphQL schema has changed on the server. Running `yarn tsc` to check for breaking changes..."
    yarn tsc
    # If there are no errors, then the schema is backwards compatible
    exit 0
```

douglowder: @wschurman Does this look correct for handling server side GraphQL changes?

wschurman: I think it'd need to be something like:
```
yarn generate-graphql-code
if cmp -s graphql.schema.json graphql.schema.json.bak; then
    echo "GraphQL schema is up-to-date"
    exit 0
else
    yarn tsc
    # If there are no errors, then the schema is backwards compatible
    exit 0
fi
```

douglowder: Excellent, that's the same as my suggestion, is it not?

wschurman: I think so. The difference is that I don't think we need to compare the `generated.ts` since there's no way to just get a new schema but not regenerate the ts.

douglowder: Hmm I thought I was comparing `generated.ts` only if the schema JSON is unchanged.

wschurman (approved):
LGTM. Sorry for the confusion.

douglowder:
### Merge activity

* **Jul 7, 7:27 PM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/eas-cli/3093).

PR #3095, opened 2025-07-07
[eas-cli] Add support for `workflow:run` inputs

sjchmiela:
# Why

We want to allow users to provide inputs to workflow runs.

# How

Added support for providing inputs with `-F` flag. It follows [`gh-cli`'s example](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#running-a-workflow).

Zrzut ekranu 2025-07-7 o 14 44 09

I am explicitly not supporting `-f` because the backend does not allow "raw field with no type coercion". It always tries to resolve from whatever is the input to the right type.

# Test Plan

Ran `easd workflow:run upload-download-test.yml -F name=test -F test=test2` and the workflow run defined two inputs: `name` and `test`.

- [workflow run](https://expo.dev/accounts/sjchmiela/projects/test-production/workflows/0197f5ea-a397-734f-81e0-8a5e78e3aa90) from `easd workflow:run inputs-test.yml`
- [workflow run](https://expo.dev/accounts/sjchmiela/projects/test-production/workflows/0197f5f0-7206-73e2-b741-409decc6e9ca) from `easd workflow:run inputs-test.yml -F name=Doug`

### DIFF
--- a/packages/eas-cli/src/commands/workflow/run.ts
+++ b/packages/eas-cli/src/commands/workflow/run.ts
@@ -162,6 +162,41 @@ export default class WorkflowRun extends EasCommand {
       }
     }
 
+    let inputs: Record<string, unknown> | undefined;
+
+    // Check for stdin input
+    const stdinData = await maybeReadStdinAsync();
+
+    const inputsFromFlags = [...(flags.input ?? [])];
+
+    // Validate that both stdin and -F flags are not provided simultaneously
+    if (stdinData && inputsFromFlags.length > 0) {
+      throw new Error(
+        'Cannot use both stdin JSON input and -F flags simultaneously. Please use only one input method.'
+      );
+    }
+
+    if (stdinData) {
+      inputs = parseJsonInputs(stdinData);
+    } else if (inputsFromFlags.length > 0) {
+      inputs = parseInputs(inputsFromFlags);
+    }
+
+    // Parse workflow inputs from YAML and prompt for missing required inputs
+    const inputSpecs = parseWorkflowInputsFromYaml(yamlConfig);
+    if (!flags['non-interactive']) {
+      inputs = await maybePromptForMissingInputsAsync({ inputSpecs, inputs: inputs ?? {} });
+    }
+
+    if (inputs && Object.keys(inputs).length > 0) {
+      Log.addNewLineIfNone();
+      Log.newLine();
+      Log.log('Running with inputs:');
+      for (const [key, value] of Object.entries(inputs)) {
+        Log.log(`- ${chalk.bold(key)}: ${JSON.stringify(value)}`);
+      }
+    }
+
     let projectArchiveBucketKey: string;
     let easJsonBucketKey: string | null = null;
     let packageJsonBucketKey: string | null = null;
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -168,6 +176,34 @@ export default class WorkflowRun extends EasCommand {
 
     let workflowRunId: string;
 
+    let inputs: Record<string, unknown> | undefined;
+
+    // Check for stdin input
+    const stdinData = await maybeReadStdinAsync();
```
douglowder: I think you need to check for the `non-interactive` flag before doing this?

sjchmiela: This is for supporting
```
echo '{"name": "Doug"}' | npx eas-cli workflow:run ...
```

This does not interactively ask the user for inputs. Also, later it checks `process.stdin.isTTY`. Is this sufficient? (I trusted Claude it is.)

douglowder: Yes that sounds ok!

douglowder: Just tried it on the demo workflow provided in this PR. I notice that if you have the input be required, but without any default, the `workflow:run` command ignores the missing input and runs anyway, and then fails ("fails to start with the API").

sjchmiela: Oh really… let me see

sjchmiela: Did you test with `easd` (the "local" version of `eas-cli`) or last published `eas-cli`? It fails for me:

```
Failed to start the workflow with the API.
Unexpected server error (Expo error ID: 40cfc8f3-70b7-4158-a0fa-403049aa355a).
Request ID: 40cfc8f3-70b7-4158-a0fa-403049aa355a
    Error: GraphQL request failed.
```

https://expoio.sentry.io/issues/6739689722/?environment=production&project=130569&query=is%3Aunresolved&referrer=issue-stream&stream_index=2

Looks like I forgot Zod errors are not automatically parsed to validation errors. Thanks for testing this case!

sjchmiela: I see your edited comment now, sorry. I'll make it so that it fails with a more readable error message. I don't think we need to add interactive prompt yet? Or, I can add it in another pull request.

douglowder: Yes I'm testing with a local build of your changes.
If we don't support interactive prompt yet, that's fine. I would suggest that we can at least throw a validation error if the input has `required: true` but no default value set.

sjchmiela: Zrzut ekranu 2025-07-11 o 12 48 36

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -168,6 +221,41 @@ export default class WorkflowRun extends EasCommand {
...
+    if (stdinData) {
+      inputs = parseJsonInputs(stdinData);
+    } else if (inputsFromFlags.length > 0) {
+      inputs = parseInputs(inputsFromFlags);
+    }
+
+    // Parse workflow inputs from YAML and prompt for missing required inputs
+    const inputSpecs = parseWorkflowInputsFromYaml(yamlConfig);
+    if (!flags['non-interactive']) {
+      inputs = await maybePromptForMissingInputsAsync({ inputSpecs, inputs: inputs ?? {} });
```
douglowder: I think it is very important that we prompt for inputs BEFORE uploading anything to EAS. Right now, this is the flow I'm seeing for a repo where the uploaded archive is 33 MB:

```
$ easd workflow:run .eas/workflows/input-test.yml 
Using workflow file from /Users/dlowder/iosProjects/react-native-tvos/tools/rntv-workflows/.eas/workflows/input-test.yml

Compressing project files and uploading to EAS. Learn more: https://expo.fyi/eas-build-archive
✔ Compressed project files 5s (34.6 MB)
✔ Uploaded project archive to EAS 2s
✔ Uploaded eas.json to EAS 
✔ Uploaded package.json to EAS 

Some required inputs are missing. Please provide them:
✔ name (The name to say hello to) … Doug


Running with inputs:
- name: "Doug"
```

douglowder (changes requested):
See inline

packages/eas-cli/src/commands/workflow/run.ts
sjchmiela: Fixed!

```
~/Developer/test-local/mobile (test-5 [$!?]) easd workflow:run .eas/workflows/cache.yml -F key=test
(node:56568) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Using workflow file from /Users/sjchmiela/Developer/test-local/mobile/.eas/workflows/cache.yml

Running with inputs:
- key: "test"

Compressing project files and uploading to EAS. Learn more: https://expo.fyi/eas-build-archive
✔ Uploaded project archive to EAS 1s
✔ Uploaded eas.json to EAS 
✔ Uploaded package.json to EAS 

See logs: http://expo.test/accounts/stanley/projects/test-local/workflows/0197fb1e-8c0d-73c4-9c38-348720c3bf7e
```

---

```
~/Developer/test-local/mobile (test-5 [$!?]) easd workflow:run .eas/workflows/cache.yml            
(node:57492) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Using workflow file from /Users/sjchmiela/Developer/test-local/mobile/.eas/workflows/cache.yml

Some required inputs are missing. Please provide them:
✔ key (Cache key) … test

Running with inputs:
- key: "test"

Compressing project files and uploading to EAS. Learn more: https://expo.fyi/eas-build-archive
✔ Uploaded project archive to EAS 1s
✔ Uploaded eas.json to EAS 
✔ Uploaded package.json to EAS 

See logs: http://expo.test/accounts/stanley/projects/test-local/workflows/0197fb1f-aa49-7050-b30f-5f0d75ed00da
```

douglowder (approved):

PR #3099, opened 2025-07-09
 feat(eas-cli): Implement new asset upload progress bar for `deploy` command

kitten:
This attempts to make the upload progress for assets more representative of the progress we know of in the background. We're replacing the 'activity-style' progress tracker with `cli-progress`, which is also used in approximately this style in other parts of the CLI.

Screenshot 2025-07-09 at 18 04 18
Screenshot 2025-07-09 at 18 02 29

### DIFF
--- a/packages/eas-cli/src/worker/upload.ts
+++ b/packages/eas-cli/src/worker/upload.ts
@@ -250,7 +250,7 @@ interface UploadProgressBar {
 
 export function createProgressBar(label = 'Uploading assets'): UploadProgressBar {
   const queueProgressBar = new cliProgress.SingleBar(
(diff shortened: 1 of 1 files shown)


kadikraman (commented):
This still has some teething issues with very large uploads I think. I tried it on my mega-project and the upload finished on 72% & 93s remaining though it all got uploaded correctly.

Could we ensure it ends at 100%?

Screenshot 2025-07-09 at 19 51 23


<details>
  <summary>Also in my case the ETA eas ticking up, not down. Here's the video.</summary>

https://github.com/user-attachments/assets/b886fb61-a0b7-4c79-a95c-b8e3adfd23f9

</details>

PR #3090, opened 2025-07-04
New commands workflow:view and workflow:logs

douglowder:
# Why

Adding the new `workflow:logs` command (see the [workflow CLI design document](https://www.notion.so/expo/EAS-CLI-workflow-commands-8bc5ccc4f9f949c8adfec1347817113c)) and the related command `workflow:view`, to provide more details on workflow runs, jobs, and their associated logs.

```bash
$ eas workflow:view --help
view details for a workflow run, including jobs. If no run ID is provided, you will be prompted to select
from recent workflow runs for the current project.

USAGE
  $ eas workflow:view [ID] [--json] [--non-interactive]

ARGUMENTS
  ID  ID of the workflow run to view

FLAGS
  --json             Enable JSON output, non-JSON messages will be printed to stderr.
  --non-interactive  Run the command in non-interactive mode.

DESCRIPTION
  view details for a workflow run, including jobs. If no run ID is provided, you will be prompted to select
from recent
  workflow runs for the current project.
```
```bash
$ eas workflow:logs --help
view logs for a workflow run, selecting a job and step to view. You can pass in either a workflow run ID or a job ID. 
If no ID is passed in, you will be prompted to select from recent workflow runs for the current project.

USAGE
  $ eas workflow:logs [ID] [--json] [--non-interactive] [--allSteps]

ARGUMENTS
  ID  ID of the workflow run or workflow job to view logs for

FLAGS
  --allSteps         Print all logs, rather than prompting for a specific step. This will be automatically set when in
                     non-interactive mode.
  --json             Enable JSON output, non-JSON messages will be printed to stderr.
  --non-interactive  Run the command in non-interactive mode.

DESCRIPTION
  view logs for a workflow run, selecting a job and step to view. You can pass in either a workflow run ID or a job ID.
If no ID is passed in, you will be prompted to select from recent workflow runs for the current project.
```
# How

Implemented the new commands. New queries and and a new fragment for workflow jobs have been added.

A state machine has been implemented to ensure correct flow for the `workflow:logs` command, allowing the user to drill down from workflow runs, to jobs, to viewing logs for an individual step. Where it makes sense, the selection prompts allow backtracking to search for a different run, job, etc.

# Test Plan

- CI should pass
- Tested locally with different inputs, and with different flag values
- New unit tests added for `workflow:logs`

PR #3103, opened 2025-07-15
[eas-cli] Fix batch uploads being subject to early-termination issues corrupting multipart upload parts

kitten:
# Why

We're basically using `Readable.from(asyncGenerator)` and `for await (const chunk of fs.createReadStream(filePath))` to read files. However, if the `createReadStream()` ends with an `'end'` event, in some Node.js versions they're subject to an early termination timing bug. When this happens, the file output ends earlier than expected, which causes the upload to become corrupted.

# How

- Switch to `fs.promises.open` and `filehandle.read` to work around async-iterable early termination bug
- Double-check the file's length (from the known size) while uploading
- Double-check the file's checksum (from the computed SHA-512) while uploading
- Use `Readable.from(asyncIterator, { objectMode: false })` to enter binary mode

This ensures that:
- we're now working around the early termination issue
- we're preventing file modifications during uploads from corrupting the upload
- we're properly using the `Readable` buffering and backpressure should work as expected

This issue is likely triggered in only some Node.js (22?) versions when:
- the `createReadStream` starts buffering
- the `Readable` hasn't filled the internal buffer yet
- the `Readable` then receives the `'end'` event discarding the internal buffer and ending early

# Test Plan

- `createReadStreamAsync` is now used for the legacy uploads as well, which should prevent this issue from being inconsistent and tied only to the new protocol (which also enforces the new length and checksum checks)
- While I've been unable to reproduce this locally so far, the new implementation isn't subject to the same code paths, which caused this issue. Because `AsyncGenerator`s don't buffer, the only buffer we're now subject to is `Readable.from`'s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - Refactor capabilities syncing to avoid provisioning profiles becoming invalid. ([#3088](https://github.com/expo/eas-cli/pull/3088) by [@vonovak](https://github.com/vonovak))
+- Avoid corrupted EAS Hosting batch upload bodies and double-check length and checksums while uploading. ([#3103](https://github.com/expo/eas-cli/pull/3103) by [@kitten](https://github.com/kitten))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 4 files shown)


douglowder (approved):
@kitten #3090 has the fix for the failing GraphQL test -- once that goes in, you should be able to rebase and merge! Thanks for this.

PR #3110, opened 2025-07-23
[eas-cli] Upgrade form-data to 4.0.4

kadikraman:
# Why

https://security.snyk.io/vuln/SNYK-JS-FORMDATA-10841150

# How

Upgraded the `form-data` dependency as per the security recommendation.

# Test Plan

Tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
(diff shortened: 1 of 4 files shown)


kadikraman:
> Do we also want to upgrade `axios` to bump `form-data`? Or should we make a Yarn override or something?

Oof, I should have checked that. I don't think we can safely update `axios` since we're on `0.28.1` and their form-data version bump is in [v1.11.0](https://github.com/axios/axios/releases/tag/v1.11.0). Seems safe enough to pin it though, so I'll do that.

Edit - updated! Also updated `@types/node-fetch` to latest as the current one had a peer dep to axios@3

Edit 2 - nope, can't bump the types package as typecheck will fail, so reverted that change

Screenshot 2025-07-23 at 17 58 35

douglowder:
Superseded by #3111

PR #3113, opened 2025-07-27
[TINY] SSO Login: print browser url

BackSlasher:
# Why
I'm working in a weird setup where the browser won't launch from cli. This required me to visit the url manually from my browser. To get the url, I had to add a `console.log` to the cli's code directly. Figured this change will make it easier for similar users to login, without exposing sensitive data

# How

Added small `log` call

# Test Plan
Tested manually on the compiled code, not the best

### DIFF
--- a/packages/eas-cli/src/user/expoSsoLauncher.ts
+++ b/packages/eas-cli/src/user/expoSsoLauncher.ts
@@ -96,6 +96,7 @@ export async function getSessionUsingBrowserAuthFlowAsync({
         );
         const port = address.port;
         const authorizeUrl = buildExpoSsoLoginUrl(port);
+        Log.log(`If the browser won't open, visit this link manually: ${authorizeUrl}`);
         void openBrowserAsync(authorizeUrl);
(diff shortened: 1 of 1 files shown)


packages/eas-cli/src/user/expoSsoLauncher.ts
```diff
@@ -96,6 +96,7 @@ export async function getSessionUsingBrowserAuthFlowAsync({
         );
         const port = address.port;
         const authorizeUrl = buildExpoSsoLoginUrl(port);
+        Log.log(`If the browser won't open, visit this link manually: ${authorizeUrl}`);
```
wschurman, line 99:
```suggestion
        Log.log(`If your browser doesn't automatically open, visit this link to log in: ${authorizeUrl}`);
```

wschurman (approved):
May also want to add the same log to https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/api/user/expoSsoLauncher.ts

PR #3121, opened 2025-07-30
[eas-cli] Only make EXPO_PUBLIC_ env vars plain text

kadikraman:
# Why

Currently, when you `eas env:push`, all env vars will be plain text unless prefixed with `EXPO_SENSITIVE`. This is unexpected, as we don't have any documentation or references to any kind for `EXPO_SENSITIVE` being used for sensitive env vars. Instead the expected and widely documented behaviour is of `EXPO_PUBLIC_` env vars being inlined in frontend code.

Therefore it makes sense to flip this logic so only `EXPO_PUBLIC_` env vars are made plain text, and the rest sensitive.

# How

When uploading bulk env vars, make `EXPO_PUBLIC_` values plain text, and the rest sensitive.

# Test Plan

Uploaded the following file via `easd env:push`

```
# .env.local
EXPO_PUBLIC_SOMETHING=should-be-plain-text
API_KEY=should-be-sensitive
```

The visibility is not as expected:

Screenshot 2025-07-30 at 10 29 55

### DIFF
--- a/packages/eas-cli/src/commands/env/push.ts
+++ b/packages/eas-cli/src/commands/env/push.ts
@@ -246,9 +246,9 @@ export default class EnvPush extends EasCommand {
         name,
         value,
         environments,
-        visibility: name.startsWith('EXPO_SENSITIVE')
-          ? EnvironmentVariableVisibility.Sensitive
-          : EnvironmentVariableVisibility.Public,
+        visibility: name.startsWith('EXPO_PUBLIC_')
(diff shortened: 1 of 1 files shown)


kadikraman:
Checking in the changed schema would also fix the gql check, but I don't believe that's the right thing to do as the changes are unrelated to this PR.

Suggested a fix for the gql check in a separate PR: https://github.com/expo/eas-cli/pull/3122

sjchmiela (approved):

PR #3138, opened 2025-08-10
[eas-cli] Add `eas run` command

sjchmiela:
# Why

# How

# Test Plan

### DIFF
--- a/packages/eas-cli/src/commands/workflow/run.ts
+++ b/packages/eas-cli/src/commands/workflow/run.ts
@@ -79,8 +79,15 @@ const EXIT_CODES = {
 
 export default class WorkflowRun extends EasCommand {
   static override description = 'run an EAS workflow';
+  static override aliases = ['run'];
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/utils/__tests__/workflowFile-test.ts
```diff
@@ -0,0 +1,114 @@
+import path from 'path';
+import { vol } from 'memfs';
```
douglowder, line 2:
```suggestion
import { vol } from 'memfs';
import path from 'path';
```

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -79,8 +79,15 @@ const EXIT_CODES = {
 
 export default class WorkflowRun extends EasCommand {
   static override description = 'run an EAS workflow';
+  static override aliases = ['run'];
```
douglowder, line 82: We also have `build:run`, so I am a little concerned about customer confusion with adding the alias.

PR #3154, opened 2025-08-27
Add better description to update:rollback

douglowder:
# Why

Make it easier for customers to use `eas update:rollback` in CI.

# How

Add description text on the underlying commands that do support the non-interactive flag.

### DIFF
--- a/packages/eas-cli/src/commands/update/rollback.ts
+++ b/packages/eas-cli/src/commands/update/rollback.ts
@@ -3,37 +3,61 @@ import { Flags } from '@oclif/core';
 import UpdateRepublish from './republish';
 import UpdateRollBackToEmbedded from './roll-back-to-embedded';
 import EasCommand from '../../commandUtils/EasCommand';
+import { EASNonInteractiveFlag } from '../../commandUtils/flags';
+import Log from '../../log';
 import { promptAsync } from '../../prompts';
 
 export default class UpdateRollback extends EasCommand {
   static override description = 'roll back to an embedded update or an existing update';
 
   static override flags = {
+    ...EASNonInteractiveFlag,
     'private-key-path': Flags.string({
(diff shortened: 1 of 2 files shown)


douglowder:
* **#3154** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/update/rollback.ts
```diff
@@ -3,37 +3,61 @@ import { Flags } from '@oclif/core';
...
 
     const privateKeyPathArg = flags['private-key-path']
       ? ['--private-key-path', flags['private-key-path']]
       : [];
     if (choice === 'published') {
+      Log.debug(
+        'Rolling back to a published update with private key path: ' +
+          JSON.stringify(privateKeyPathArg)
+      );
       await UpdateRepublish.run(privateKeyPathArg);
```
wschurman: I think we'd need to supply `--non-interactive` arg as well?

douglowder: We did, see line 14

wschurman: Ah, sorry I was unclear in my review, I meant we need to also pass `--non-interactive` to the two commands this calls into.

douglowder: After digging into the code, it does appear to be too complex to have this command combine arguments from both republish and roll-back-to-embedded. So instead, I've just added more text to the command description for users who need non-interactive.

wschurman (approved):

PR #3165, opened 2025-09-08
Improvements to workflow:validate

douglowder:
# Why

Improve the validation of workflow files:

- Add client-side checks for valid syntax, and compatibility with the user's EAS build profiles
- Make validation usable with future improvements to `workflow:create`

# How

- New `validation.ts` in `commandUtils` with the various validation checks separated cleanly into different functions
- `workflow:validate` command refactored to use the new utility and wrap errors
- The new code reads the workflow schema from `<ExpoAPIBaseURL>/v2/workflows/schema`
- For testing purposes only, a local copy of the schema can be used by setting EXPO_TESTING_WORKFLOW_SCHEMA_PATH
- An existing check in `workflowFile.ts` has been modified to remove the requirement for a Github repository, as that is no longer required to run workflows

# Test Plan

- Tested against different workflow templates, and against workflows in a real project.
- Tested against the www server from the main branch running locally, to verify recent server side changes (e.g. https://github.com/expo/universe/pull/21950)

### DIFF
--- a/packages/eas-cli/src/commandUtils/workflow/validation.ts
+++ b/packages/eas-cli/src/commandUtils/workflow/validation.ts
@@ -1,0 +1,196 @@
+import { EasJsonAccessor, EasJsonUtils } from '@expo/eas-json';
+import { InvalidEasJsonError, MissingEasJsonError } from '@expo/eas-json/build/errors';
+import { CombinedError } from '@urql/core';
+import * as YAML from 'yaml';
+
+import { AccountFragment } from '../../graphql/generated';
+import { WorkflowRevisionMutation } from '../../graphql/mutations/WorkflowRevisionMutation';
+import Log from '../../log';
+import { createValidator, getReadableErrors } from '../../metadata/utils/ajv';
+import { WorkflowFile } from '../../utils/workflowFile';
+import { ExpoGraphqlClient } from '../context/contextUtils/createGraphqlClient';
+
+export async function validateWorkflowFileAsync(
+  workflowFileContents: { yamlConfig: string; filePath: string },
+  projectDir: string,
+  graphqlClient: ExpoGraphqlClient,
+  projectId: string
+): Promise<void> {
+  const parsedYaml = parsedYamlFromWorkflowContents(workflowFileContents);
+  Log.debug(`Parsed workflow: ${JSON.stringify(parsedYaml, null, 2)}`);
+
+  // Check if the parsed result is empty or null
+  Log.debug(`Validating workflow is not empty...`);
+  validateWorkflowIsNotEmpty(parsedYaml);
+
+  const workflowSchema = await fetchWorkflowSchemaAsync();
+
+  // Check that all job types are valid
+  Log.debug(`Validating workflow job types...`);
+  validateWorkflowJobTypes(parsedYaml, workflowSchema);
+
+  // Check for build jobs that do not match any EAS build profiles
+  Log.debug(`Validating workflow build jobs...`);
+  await validateWorkflowBuildJobsAsync(parsedYaml, projectDir);
+
+  // Check that result passes validation against workflow schema
+  Log.debug(`Validating workflow structure...`);
+  validateWorkflowStructure(parsedYaml, workflowSchema);
+
+  // Check for other errors using the server-side validation
+  Log.debug(`Validating workflow on server...`);
+  await validateWorkflowOnServerAsync(graphqlClient, projectId, workflowFileContents);
+}
+
+export function logWorkflowValidationErrors(
+  error: unknown,
+  account: AccountFragment | undefined,
+  projectName: string | undefined
+): void {
+  if (error instanceof MissingEasJsonError) {
+    throw new Error(
+      'Workflows require a valid eas.json. Please run "eas build:configure" to create it.'
+    );
+  } else if (error instanceof InvalidEasJsonError) {
+    throw new Error(
+      'Workflows require a valid eas.json. Please fix the errors in your eas.json and try again.\n\n' +
+        error.message
+    );
+  } else if (error instanceof YAML.YAMLParseError) {
+    Log.error(`YAML syntax error: ${error.message}`);
+  } else if (error instanceof CombinedError) {
+    WorkflowFile.maybePrintWorkflowFileValidationErrors({
+      error,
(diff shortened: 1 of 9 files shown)


douglowder:
* **#3165** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3165?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3165?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commandUtils/workflow/validation.ts
```diff
@@ -0,0 +1,196 @@
...
+): Promise<void> {
+  await WorkflowRevisionMutation.validateWorkflowYamlConfigAsync(graphqlClient, {
+    appId: projectId,
+    yamlConfig: workflowFileContents.yamlConfig,
+  });
+}
+
+async function validateWorkflowBuildJobsAsync(parsedYaml: any, projectDir: string): Promise<void> {
+  const jobs = jobsFromWorkflow(parsedYaml);
+  const buildJobs = jobs.filter(job => job.value.type === 'build');
```
sjchmiela: we might want to also include `repack` here, it uses build profiles too

packages/eas-cli/src/commandUtils/workflow/validation.ts
```diff
@@ -0,0 +1,196 @@
...
+  if (buildJobs.length === 0) {
+    return;
+  }
+  const easJsonAccessor = EasJsonAccessor.fromProjectPath(projectDir);
+
+  const buildProfileNames = new Set(
+    easJsonAccessor && (await EasJsonUtils.getBuildProfileNamesAsync(easJsonAccessor))
+  );
+  const invalidBuildJobs = buildJobs.filter(
+    job => !buildProfileNames.has(job.value.params.profile)
```
sjchmiela: `profile` can be an interpolated value, like `${{ needs.setup.outputs.build_profile }}-android` (e.g. `"simulator-android"`)

packages/eas-cli/src/commandUtils/workflow/validation.ts
```diff
@@ -0,0 +1,196 @@
...
+  yamlConfig: string;
+}): Promise<any> {
+  const parsedYaml = YAML.parse(workflowFileContents.yamlConfig);
+  return parsedYaml;
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+async function fetchWorkflowSchemaAsync(): Promise<any> {
+  const schemaUrl =
+    process.env.EXPO_WORKFLOW_SCHEMA_URL ?? 'https://api.expo.dev/v2/workflows/schema';
```
sjchmiela: EXPO_? EAS_? you never know! haha

packages/eas-cli/src/commandUtils/workflow/validation.ts
```diff
@@ -0,0 +1,196 @@
...
+function jobsFromWorkflow(parsedYaml: any): any[] {
+  return Object.entries(parsedYaml?.jobs).flatMap(([key, value]: [string, any]) => {
+    return {
+      key,
+      value,
+    };
+  });
+}
+
+function jobTypesFromWorkflowSchema(workflowJsonSchema: any): string[] {
```
sjchmiela, line 205: oh wow. we really need to jump through hoops because of not having `WorkflowConfigZ` in public repo…

packages/eas-cli/src/commandUtils/workflow/validation.ts
```diff
@@ -0,0 +1,196 @@
...
+    throw new Error(
+      'Workflows require a valid eas.json. Please fix the errors in your eas.json and try again.\n\n' +
+        error.message
+    );
+  } else if (error instanceof YAML.YAMLParseError) {
+    Log.error(`YAML syntax error: ${error.message}`);
+  } else if (error instanceof CombinedError) {
+    WorkflowFile.maybePrintWorkflowFileValidationErrors({
+      error,
+      accountName: account?.name ?? '',
```
sjchmiela: I think we can remove https://github.com/expo/eas-cli/blob/4571910c4d9c518d42e46a9c90e30429fd45d417/packages/eas-cli/src/utils/workflowFile.ts#L66-L76, we no longer expect `www` to throw with `GITHUB_NOT_FOUND_ERROR` (it was the case back in the beginning of 2025 when we required GitHub repository to run workflows).

This should let you remove `accountName` and `projectName` etc

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: hmm do we have any CLI code that can construct the real string from the interpolated value?

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: Maybe EAS_TESTING_WORKFLOW_SCHEMA_URL, to ensure that customers know that this is for Expo internal testing purposes?

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: Ok I've removed that.

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: Done.

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: In looking into this more, it seems likely that interpolated values might only be resolved at runtime, so we can't include them in a static check like this. I'll add a check for a regex that matches "${{...}}" and exclude those profiles from the validation check.

packages/eas-cli/src/commandUtils/workflow/validation.ts
douglowder: I've reworked this code to make testing easier, and to ensure that the schema is fetched correctly when EXPO_STAGING or EXPO_LOCAL is enabled.

PR #3175, opened 2025-09-12
Add artifact URLs to workflow:view

douglowder:
# Why

To allow CLI users to get artifacts of builds or workflow jobs directly using the CLI, and make the URLs available in JSON output also (for easier use by automation scripts).

# How

- Add artifact info to WorkflowJobFragment
- Add this info to the results of `workflow:view`
- For workflow jobs that are of type=BUILD, use the existing build query to get the build artifact URLs for the build and show those as well
- Make some improvements to the formatting of workflow job outputs

# Test Plan

- Tested against different EAS projects
- CI should pass


douglowder:
* **#3175** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3175?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3175?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/build/utils/formatBuild.ts
```diff
@@ -10,6 +10,14 @@ import { link } from '../../log';
 import { appPlatformDisplayNames } from '../../platform';
 import formatFields from '../../utils/formatFields';
 
+export function buildArtifactFromBuild(build: BuildFragment): string | null {
+  return (
+    build.artifacts?.buildUrl ??
+    build.artifacts?.applicationArchiveUrl ??
```
sjchmiela: `buildUrl` and `applicationArchiveUrl` are the same. Not sure why one of them wasn't deprecated when the other was added… Maybe we should keep only one?

https://github.com/expo/universe/blob/98da058a2028b0f616c815b23fb3c5849f364836/server/www/src/graphql/types/Build.ts#L401-L402

packages/eas-cli/src/build/utils/formatBuild.ts
```diff
@@ -10,6 +10,14 @@ import { link } from '../../log';
...
+    build.artifacts?.buildArtifactsUrl ??
```
sjchmiela: Build artifacts are a separate thing which we might want to mention separately. A build may produce an application archive or it may not, it may produce build artifacts, but it may not. It may produce both or none.

packages/eas-cli/src/build/utils/formatBuild.ts
```diff
@@ -97,7 +105,7 @@ export function formatGraphQLBuild(build: BuildFragment): string {
           case GraphQLBuildStatus.Errored:
             return null;
           case GraphQLBuildStatus.Finished: {
-            const url = build.artifacts?.buildUrl;
+            const url = buildArtifactFromBuild(build);
```
sjchmiela: As far as I remember a build may produce application archive _and_ then fail. Like if there's a custom build which first does the build, uploads the application archive and then tries to send a Slack message or maybe the post-build hook fails. We may want to show build artifact if it exists, not only when build has succeeded.

packages/eas-cli/src/commands/workflow/view.ts
```diff
@@ -1,18 +1,145 @@
...
-import { WorkflowRunByIdWithJobsQuery } from '../../graphql/generated';
+import { WorkflowJobType, WorkflowRunByIdWithJobsQuery } from '../../graphql/generated';
+import { BuildQuery } from '../../graphql/queries/BuildQuery';
 import { WorkflowRunQuery } from '../../graphql/queries/WorkflowRunQuery';
 import Log, { link } from '../../log';
-import formatFields from '../../utils/formatFields';
+import formatFields, { FormatFieldsItem } from '../../utils/formatFields';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
+type WorkflowArtifact = {
```
sjchmiela: Why does it make sense to define the type here instead of using some GraphQL generated type?

packages/eas-cli/src/commandUtils/workflow/types.ts
```diff
@@ -44,3 +44,46 @@ export type WorkflowRunWithJobsResult = WorkflowRunResult & {
...
+export type WorkflowUpdateJobOutput = {
+  first_update_group_id: string; // ID of the first update group. You can use it to e.g. construct the update URL for a development client deep link.
+  updates_json: string; // Stringified JSON array of update groups. Output of `eas update --json`.
+};
+
+export type WorkflowCustomJobOutput = {
+  [key: string]: unknown;
+};
+
+export type WorkflowJobOutput =
```
sjchmiela: Hmm… in my opinion we should not be typing these outputs or using them in this nice user-facing manner (as nice and user-friendly it is). I think `outputs` are meant to be consumed by workflow run only, while it's running, to coordinate jobs execution.

If we want to mention updates created by a given job, we should query `workflowJob { turtleJobRun { updateGroups { ... } } }`.

If we want to mention builds found by a given job (`get-build` job), we should query `workflowJob { turtleBuild { ... } }`. (In other words — if the user is interested in `outputs`, they're interested in it because they're actively developing a workflow and want to know what's available so they can write further logic or interpolate in a Slack message. If they're interested in the Build that a job has produced, they should learn about its details from `Build`-processing code, not `outputs`-processing code).

For submissions and TestFlight I don't have a good solution right now, but ideally in the future we would be able to do `workflowJob { turtleJobRun { submissions { ... } } }`.

packages/eas-cli/src/commands/workflow/view.ts
```diff
@@ -118,10 +260,43 @@ export default class WorkflowView extends EasCommand {
...
+        }
+      }
+      if (job.artifacts?.length) {
+        Log.gray(chalk.dim('  Artifacts:'));
+        job.artifacts.forEach(artifact => {
+          Log.log(
+            formatFields([
+              { label: '    ID', value: artifact.id },
+              { label: '    Name', value: artifact.name },
+              { label: '    Content Type', value: artifact?.contentType ?? 'null' },
```
sjchmiela: FYI if the artifact is missing `contentType` and `fileSizeBytes` it's a way of `www` telling us "job said it would upload the artifact, but it didn't". After the job completes we look at the artifacts it said it will produce and we update their content type and file size bytes.

We've seen that _sometimes_ they wouldn't get filled _even though_ the objects were successfully created, but I think since https://github.com/expo/universe/pull/21681 this problem is gone.

packages/eas-cli/src/commands/workflow/view.ts
```diff
@@ -118,10 +260,43 @@ export default class WorkflowView extends EasCommand {
...
         ])
       );
+      if (job.errors.length > 0) {
+        Log.gray(chalk.dim('  Errors:'));
+        job.errors.forEach(error => {
+          Log.log(formatFields([{ label: `    ${error.title}`, value: `${error.message}` }]));
+        });
+      }
+      if (job.outputs) {
+        const outputs = processedOutputs(job);
```
sjchmiela, line 177: My suggestion would be to present `outputs` as a (formatted) JSON object.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: Good point

packages/eas-cli/src/commandUtils/workflow/types.ts
douglowder: I thought it would be ok to type them because we specifically document these types publicly. See for example https://docs.expo.dev/eas/workflows/syntax/#build

sjchmiela: My comment is not about types / no types, but something earlier, that outputs are just that – outputs, result of a "function" (in the mathematical sense) – and I think they're better identified by their keys, not labels of properties they represent.

Maybe I'm ok with adding types where all properties are optional, but I don't think we should put human-readable labels on them (in which case typing outputs makes less sense / is less beneficial)?

This also creates this tension (maybe only in my mind) between "showing all outputs" and "showing too many properties about the product (build, updates, deployment)". For instance, maybe it doesn't make sense to list all outputted build properties one after another like this? Maybe it'd be better to say `Build {shortId}, {version} ({buildNumber}), {profile}`?

I don't think it's necessarily a blocker, I just want to avoid confusion that "`job.outputs` is the best representation of a job's product". I think it may contain uninteresting properties (`git_commit_hash`? `first_update_group_id`?) and it may lack interesting properties (like `build_download_url`).

packages/eas-cli/src/build/utils/formatBuild.ts
douglowder: I now show buildUrl and buildArchiveUrl, and ignore applicationArchiveUrl. We should probably deprecate applicationArchiveUrl on the server side?

packages/eas-cli/src/build/utils/formatBuild.ts
douglowder: Let's discuss this more, as this is not how build:view is currently implemented.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: This type is now defined in terms of GraphQL generated types.

packages/eas-cli/src/commandUtils/workflow/types.ts
douglowder: I've removed these types, and am showing output fields in an unopinionated way using the actual key strings.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: I've redone the handling of outputs.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: Good to know!

packages/eas-cli/src/build/utils/formatBuild.ts
douglowder: I'm now showing these URLs separately (omitting the applicationArchiveUrl which is a duplicate).

packages/eas-cli/src/build/utils/formatBuild.ts
```diff
@@ -8,7 +8,46 @@ import {
 } from '../../graphql/generated';
 import { link } from '../../log';
 import { appPlatformDisplayNames } from '../../platform';
-import formatFields from '../../utils/formatFields';
+import formatFields, { FormatFieldsItem } from '../../utils/formatFields';
+
+export function formatGraphQLBuildArtifacts(build: BuildFragment): FormatFieldsItem[] {
+  const fields = [
+    {
+      label: 'Build Archive URL',
```
sjchmiela:
```suggestion
      label: 'Application Archive URL',
```
I think this is The Name

Image

Image

packages/eas-cli/src/commands/workflow/view.ts
```diff
@@ -7,12 +10,54 @@ import {
...
+
+const processedOutputs: (job: WorkflowJobResult) => FormatFieldsItem[] = job => {
+  const result: FormatFieldsItem[] = [];
+  const keys = job.outputs ? Object.keys(job.outputs) : [];
+  keys.forEach(key => {
+    result.push({
+      label: `    ${key}`,
+      get value(): string {
+        const value = job.outputs[key];
+        return typeof value === 'string' ? value : JSON.stringify(value, null, 2);
```
sjchmiela, line 52: fwiw I think `JSON.stringify(value, null, 2)` would work for strings too

packages/eas-cli/src/commands/workflow/view.ts
```diff
@@ -118,10 +174,62 @@ export default class WorkflowView extends EasCommand {
...
+      }
+      if (job.outputs) {
+        const outputs = processedOutputs(job);
+        if (outputs.length > 0) {
+          Log.gray(chalk.dim('  Outputs:'));
+          Log.log(formatFields(outputs));
+        }
+      }
+      if (job.type === WorkflowJobType.Build) {
+        if (job.outputs?.build_id) {
```
sjchmiela: Maybe `fragment WorklfowJob` should include `turtleBuild { ...BuildFragment }`? This also removes the n+1 problem (as low as n will be). I think it'll be be better to rely on `turtleBuild` field than `outputs.build_id`.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: You are right! I think it would add quotes around the string though...

packages/eas-cli/src/commands/workflow/view.ts
douglowder: Ah I didn't realize that turtleBuild was just a BuildFragment! 😃

packages/eas-cli/src/build/utils/formatBuild.ts
douglowder: Ok I fixed this -- we show "in progress" if the build is in that state, otherwise we show artifacts that exist regardless of build state.

packages/eas-cli/src/build/utils/formatBuild.ts
douglowder: Fixed.

packages/eas-cli/src/commands/workflow/view.ts
douglowder: Made this change and simplified the code, thanks for the suggestion!

douglowder:
### Merge activity

* **Sep 15, 6:42 PM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/eas-cli/3175).

PR #3194, opened 2025-09-24
Improvements to workflow:create

douglowder:
# Why

The existing `workflow:create` command simply creates a very bare-bones workflow, that is not very useful to developers who need to know how to fully use workflows for managing builds, updates, and other EAS operations.

# How

- Define four different workflow templates, initialized as JSON objects.
- After the user selects which type of workflow (build, update, deploy, custom), sanity checks are made, EAS build and EAS update are configured if needed, and any missing build profiles are created in `eas.json`.
- Once the user has made their selections, the final workflow YAML is dynamically generated and written to disk with an explanatory header.
- After writing the workflow, next steps are provided for any further configuration that may be needed, especially if build credentials or app store credentials are needed.

In future, we can easily add more templates that use other workflow job types (fingerprint, repack, etc.).

# Test Plan

- Tested against different projects
- CI should pass

### DIFF
--- a/packages/eas-cli/src/commandUtils/workflow/creation.ts
+++ b/packages/eas-cli/src/commandUtils/workflow/creation.ts
@@ -6,6 +6,8 @@ import {
   addIosDevelopmentBuildProfileToEasJsonAsync,
   addProductionBuildProfileToEasJsonIfNeededAsync,
   buildProfilesFromProjectAsync,
+  hasBuildConfigureBeenRunAsync,
+  hasUpdateConfigureBeenRunAsync,
   isBuildProfileForDevelopment,
   isIosBuildProfileForSimulator,
 } from './buildProfileUtils';
@@ -261,6 +263,34 @@ function nextStepForAppSubmission(): string {
     .trimStart();
 }
 
+export function howToRunWorkflow(
+  workflowFileName: string,
+  workflowStarter: WorkflowStarter
+): string {
+  const lines = [
+    `To run this workflow manually, use the command "eas workflow:run .eas/workflows/${workflowFileName}".`,
+  ];
+  if (workflowStarter.template?.on?.push?.branches?.length > 0) {
+    const branches = workflowStarter.template.on.push.branches;
+    if (branches.length === 1 && branches[0] === '*') {
+      lines.push(
+        'This workflow is also configured to run automatically when code is pushed to any branch.'
+      );
+    } else if (branches.length === 1) {
+      lines.push(
+        `This workflow is also configured to run automatically when code is pushed to the "${branches[0]}" branch.`
+      );
+    } else {
+      lines.push(
+        `This workflow is also configured to run automatically when code is pushed to the following branches: ${branches.join(
+          ', '
+        )}.`
+      );
+    }
+  }
+  return lines.join('\n');
+}
+
 export const workflowStarters: WorkflowStarter[] = [
   {
     displayName: 'Custom',
@@ -464,19 +494,40 @@ export async function ensureProductionBuildProfileExistsAsync(
 export async function customizeTemplateIfNeededAsync(
   workflowStarter: WorkflowStarter,
   projectDir: string,
-  exp: ExpoConfig
+  expoConfig: ExpoConfig
 ): Promise<any> {
+  // Ensure EAS Build is configured
   switch (workflowStarter.name) {
     case WorkflowStarterName.BUILD:
-      Log.debug('Adding build jobs to template...');
-      return await addBuildJobsToDevelopmentBuildTemplateAsync(projectDir, workflowStarter);
+    case WorkflowStarterName.DEPLOY:
     case WorkflowStarterName.UPDATE:
-      if (!exp?.updates?.url) {
+      if (!(await hasBuildConfigureBeenRunAsync({ projectDir, expoConfig }))) {
+        throw new Error(
+          'EAS Build is not configured for this project. Please run "eas build:configure" to configure it.'
+        );
+      }
+      break;
+    default:
+      break;
+  }
+  // Ensure EAS Update is configured
+  switch (workflowStarter.name) {
+    case WorkflowStarterName.DEPLOY:
+    case WorkflowStarterName.UPDATE:
+      if (!(await hasUpdateConfigureBeenRunAsync({ projectDir, expoConfig }))) {
         throw new Error(
           'EAS Update is not configured for this project. Please run "eas update:configure" to configure it.'
         );
       }
-      return workflowStarter;
+      break;
+    default:
+      break;
+  }
+  // Customize template
+  switch (workflowStarter.name) {
+    case WorkflowStarterName.BUILD:
+      Log.debug('Adding build jobs to template...');
+      return await addBuildJobsToDevelopmentBuildTemplateAsync(projectDir, workflowStarter);
     case WorkflowStarterName.DEPLOY:
       return await ensureProductionBuildProfileExistsAsync(projectDir, workflowStarter);
     default:
--- a/packages/eas-cli/src/commands/workflow/create.ts
+++ b/packages/eas-cli/src/commands/workflow/create.ts
@@ -13,6 +13,7 @@ import {
   WorkflowStarter,
   WorkflowStarterName,
   customizeTemplateIfNeededAsync,
+  howToRunWorkflow,
   workflowStarters,
 } from '../../commandUtils/workflow/creation';
 import {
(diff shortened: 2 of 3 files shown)


douglowder:
* **#3194** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3194?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3194?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -48,42 +55,90 @@ export class WorkflowCreate extends EasCommand {
...
-        process.exit(0);
+      const { exp: expPossiblyWithoutEasUpdateConfigured, projectId } =
+        await getDynamicPrivateProjectConfigAsync();
+
+      let fileName = argFileName;
+
+      let workflowStarter: WorkflowStarter;
+      if (flags.template) {
+        workflowStarter =
+          workflowStarters.find(template => template.name === flags.template) ??
```
sjchmiela: nit: would it be better to fail if not found instead of fall back to the first one?

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -48,42 +55,90 @@ export class WorkflowCreate extends EasCommand {
...
+          type: 'text',
+          name: 'fileName',
+          message: 'What would you like to name your workflow file?',
+          initial: workflowStarter.defaultFileName,
+          validate: value => {
+            try {
+              WorkflowFile.validateYamlExtension(value);
+              return true;
+            } catch (error) {
+              return error instanceof Error ? error.message : 'Invalid file name';
```
sjchmiela: Would be nice to mention it's about YAML extension, I think

packages/eas-cli/src/commands/workflow/create.ts
```diff
@@ -48,42 +55,90 @@ export class WorkflowCreate extends EasCommand {
...
+            }
+          },
+        });
+        if (!response.fileName) {
```
sjchmiela: Can we instead require some input? Would it be better?

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
...
+import Log from '../../log';
+import { promptAsync } from '../../prompts';
+import { easCliVersion } from '../../utils/easCli';
+
+export enum WorkflowStarterName {
+  BUILD = 'build',
+  UPDATE = 'update',
+  MAESTRO = 'maestro',
+  CUSTOM = 'custom',
+}
```
sjchmiela, line 23: nit: I think we could make the names less code-like? Like, we could have `publish-preview-updates-on-main`, `create-development-builds` and stuff?

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
...
+    },
+  },
+  jobs: {},
+};
+
+const PUBLISH_UPDATE_TEMPLATE = {
+  name: 'Publish updates',
+  on: {
+    push: {
+      branches: ['main'],
```
sjchmiela: Do we want to publish on pull requests too?

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
...
+      type: 'custom',
+      runs_on: 'linux-medium',
+      image: 'latest',
+      steps: [
+        {
+          uses: 'eas/checkout',
+        },
+        {
+          name: 'Hello World',
+          id: 'hello-world',
```
sjchmiela: We need to prefer users to do `_` instead of `-` because our `${{ }}` parser is based on JavaScript syntax and, when presented `steps.hello-world.outputs.key` it thinks it's `(steps.hello) - (world.outputs.key)` which doesn't work.

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
...
+    const buildProfile = await getBuildProfileAsync(projectDir, Platform.IOS, profileName);
+    if (buildProfile.channel) {
+      channels.add(buildProfile.channel);
+    }
+  }
+  const channelsToUpdate = (
+    await promptAsync({
+      type: 'multiselect',
+      name: 'selectedChannels',
+      message: 'Select update channels to add to workflow',
```
sjchmiela: I _think_ we tend to recommend people to use Git branches as updates branches names (and branches to equal channels)? Or sth like that, Jon may know more. (In which case the best `type: update` job is one without any params)

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
...
+  name: 'Publish updates',
+  on: {
+    push: {
+      branches: ['main'],
+    },
+  },
+  jobs: {},
+};
+
+const MAESTRO_TEST_TEMPLATE = {
```
sjchmiela: Maybe we want to do both iOS and Android tests?

packages/eas-cli/src/commandUtils/workflow/creation.ts
```diff
@@ -0,0 +1,318 @@
+import { ExpoConfig } from '@expo/config';
+import { Platform } from '@expo/eas-build-job';
+import { BuildProfile, EasJsonAccessor, EasJsonUtils } from '@expo/eas-json';
+
+import { buildProfileNamesFromProjectAsync } from './validation';
+import Log from '../../log';
+import { promptAsync } from '../../prompts';
+import { easCliVersion } from '../../utils/easCli';
+
+export enum WorkflowStarterName {
```
sjchmiela, line 18: What might also be nice (maybe in another PR) is the "golden workflow" https://docs.expo.dev/eas/workflows/examples/deploy-to-production/, maybe with the possiblity of adding Maestro tests before submitting?

Or maybe we want more purpose-based workflows, like
- run this on `main`: ensure development and production builds exist, submit production builds after approval
- run this on pull requests: create development builds, do simulator builds (or repack existing), run maestro tests?

packages/eas-cli/src/commands/workflow/create.ts
douglowder: I reworked this to let the user try a different name if their input fails YAML extension check.

packages/eas-cli/src/commandUtils/workflow/creation.ts
douglowder: If we don't allow users to type in a name, then these names will no longer be user-facing.

packages/eas-cli/src/commandUtils/workflow/creation.ts
douglowder: We could offer this as a choice menu too...

packages/eas-cli/src/commandUtils/workflow/creation.ts
douglowder: I'll sync with @jonsamp on that. I think I just copied that from our current docs.

packages/eas-cli/src/commandUtils/workflow/creation.ts
douglowder: Yes I'm hoping to add more complex workflows in future PRs!

packages/eas-cli/src/commands/workflow/create.ts
douglowder: Agreed -- I removed the template flag so now the user just selects from the prompt.

packages/eas-cli/src/commands/workflow/create.ts
douglowder: Fixed.

packages/eas-cli/src/commandUtils/workflow/creation.ts
douglowder: Fixed.

jonsamp:
I dogfooded this and I have some feedback about what templates we should expose and what outputs we should show. I wrote them up in a Notion doc (easier to collaborate on): https://www.notion.so/expo/workflow-create-feedback-27ee5b57352480f7a826d2bd0904aee4

Please let me know if you want to sync on this on call, I'm around!

douglowder:
Thanks I’ll take a look!

PR #3203, opened 2025-10-01
Support workflow:run from a Git ref

douglowder:
# Why

Add a `--ref` flag so that `eas workflow:run` can initiate a workflow run from an existing ref in the project's GitHub repository.

- Avoids the need to upload the entire project from a local machine
- Ensures that the workflow starts from a known reference point in the project's source control, with no unexpected local changes

# How

- If the ref flag is set, the run command will ensure that the ref is valid, and that the workflow can be used for that ref (using the new mutation for this from https://github.com/expo/universe/pull/22132)
- If the ref flag is set, the run command starts the run from that ref (using https://github.com/expo/universe/pull/20161)
- If the user chooses ref = HEAD, `git rev-parse` is called to convert that to a commit
- The flow is designed to work the same for running from GitHub ref as running from local, so inputs and waiting for results are supported 

# Test Plan

- Tested in an existing project with workflows with and without inputs
- CI should pass

### DIFF
--- a/packages/eas-cli/src/commands/workflow/run.ts
+++ b/packages/eas-cli/src/commands/workflow/run.ts
@@ -40,6 +40,7 @@
  * ```
  */
 
+import spawnAsync from '@expo/spawn-async';
 import { Flags } from '@oclif/core';
 import { CombinedError } from '@urql/core';
 import chalk from 'chalk';
@@ -55,6 +56,7 @@ import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/creat
 import { EASNonInteractiveFlag, EasJsonOnlyFlag } from '../../commandUtils/flags';
 import {
   WorkflowProjectSourceType,
+  WorkflowRevision,
   WorkflowRunByIdQuery,
   WorkflowRunStatus,
 } from '../../graphql/generated';
@@ -78,7 +80,8 @@ const EXIT_CODES = {
 };
 
 export default class WorkflowRun extends EasCommand {
-  static override description = 'run an EAS workflow';
+  static override description =
+    'run an EAS workflow. The entire local project directory will be packaged and uploaded to EAS servers for the workflow run, unless the --ref flag is used.';
 
   static override args = [{ name: 'file', description: 'Path to the workflow file to run' }];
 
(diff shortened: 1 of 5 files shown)


douglowder:
* **#3203** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -125,26 +133,70 @@ export default class WorkflowRun extends EasCommand {
...
+
+    if (flags.ref) {
+      // Run from git ref
+      const fileName = path.basename(args.file);
+      // Find the real commit, make sure the ref is valid
+      gitRef = (
+        await spawnAsync('git', ['rev-parse', flags.ref], {
+          cwd: projectDir,
+        })
+      ).output[0].trim();
```
sjchmiela, line 155: fwiw we might not do that too, one of the first things we're doing in the mutation is resolve the Git ref -- https://github.com/expo/universe/blob/95273847260d7801a517ab03169add2484b524a4/server/www/src/graphql/mutations/WorkflowRunMutation.ts#L210-L213 (and theoretically it might be that a user's local repo knows of more refs than the remote knows about)

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -125,26 +133,70 @@ export default class WorkflowRun extends EasCommand {
...
+        );
+      } catch (err) {
+        throw new Error(
+          `Failed to find or create workflow revision for ${fileName} at ${flags.ref}: ${err}`
+        );
+      }
+      Log.debug(`Workflow revision: ${JSON.stringify(revisionResult, null, 2)}`);
+      if (!revisionResult) {
+        throw new Error(
+          `Failed to find or create workflow revision for ${fileName} at ${flags.ref}`
```
sjchmiela, line 178: not sure if we want to be so explicit about what we're doing under the hood (what does the user care if we're finding, creating or fetching)

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -206,81 +258,95 @@ export default class WorkflowRun extends EasCommand {
...
-          )} jobs will not work. Run ${chalk.bold(
-            'eas build:configure'
-          )} to configure your project for builds.`
-        );
+    if (gitRef) {
+      // Run from git ref
+      let runResult: { id: string };
+      try {
+        runResult = await WorkflowRunMutation.createWorkflowRunFromGitRefAsync(graphqlClient, {
+          workflowRevisionId: workflowRevisionId ?? '',
```
sjchmiela, line 266: we might as well throw an error if we don't have a `workflowRevisionId` present, it is required currently

PR #3211, opened 2025-10-07
[eas-cli] Pass `_expo/.routes.json` for SSG exports as `routes.json` to Hosting deployments

kitten:
# Why

See: https://github.com/expo/expo/pull/40243

This adds support for passing the `headers` configuration to static deployments.

# How

We'll discover the `routes.json` in a few recognised paths and pass it as a special `routes.json` payload to the `deployment-api`. This is then (for static/single exports) recognised and stored, serving the `headers` configuration for page results.

# Test Plan

- Can only be E2E tested when the corresponding API change is deployed to staging
- `--dry-run` to validate that `assets.json` doesn't contain the `_expo/.routes.json` and has instead passed it into `/routes.json` in the tarball

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Support `_expo/.routes.json` metadata file exported by `@expo/cli` for headers configuration on static deployments ([#3211](https://github.com/expo/eas-cli/pull/3211) by [@kitten](https://github.com/kitten))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores


packages/eas-cli/src/worker/assets.ts
```diff
@@ -12,6 +12,8 @@ import { EnvironmentVariableEnvironment } from '../build/utils/environment';
 import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
 import { EnvironmentVariablesQuery } from '../graphql/queries/EnvironmentVariablesQuery';
 
+const EXPO_ROUTES_PATHS = new Set(['_expo/routes.json', '_expo/.routes.json', '.expo-routes.json']);
```
byCedric, line 15: Why do we have 3 different styles of writing the routes?

kitten: Because I couldn't decide which one looks best for when we'll maybe open this up to customisation 😄 In all seriousness though, the problem here is mostly that I am trying to give ourselves a few options, since I noticed the `export` emission logic is still in a weird state. I'm expecting us to maybe move things around.

- `_expo/routes.json` is added in case we decide to just merge the server/client output for this file
- `_expo/.routes.json` is what we'll actually go for
- `.expo-routes.json` is what might be a little more user-friendly if we allow for manual customisation

byCedric: Alright, well, might be good to eventually focus on a single format. But we can do that later on.

Issue #3212, opened 2025-10-07
Latest break workflows and local cli wiht pull:env


wcastand:
### Build/Submit details page URL

https://expo.dev/accounts/green-got/projects/green-got/workflows/0199bf64-5ee0-78b8-9f19-5e3d3e875e4a

### Summary

the latest release from less than an hour ago seems to break the cli
`npx -y eas-cli@latest env:pull --environment development`

breaks with the error 

`Invalid environment. Use one of 'production', 'preview', or 'development'.`

i tried  `npx -y eas-cli@latest env:pull development --environment development` but same error locally and on eas.

### Managed or bare?

managed

### Environment

 expo-env-info 2.0.7 environment info:
    System:
      OS: macOS 26.0.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.16.0 - ~/.nvm/versions/node/v22.16.0/bin/node
      Yarn: 1.22.22 - ~/.nvm/versions/node/v22.16.0/bin/yarn
      npm: 10.9.2 - ~/.nvm/versions/node/v22.16.0/bin/npm
      Watchman: 2025.09.15.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.16.2 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2025.1 AI-251.26094.121.2513.14007798
      Xcode: 26.0.1/17A400 - /usr/bin/xcodebuild
    npmPackages:
      expo: 54.0.12 => 54.0.12
      expo-router: ~6.0.10 => 6.0.10
      expo-updates: ~29.0.12 => 29.0.12
      react: 19.1.0 => 19.1.0
      react-dom: 19.1.0 => 19.1.0
      react-native: 0.81.4 => 0.81.4
    npmGlobalPackages:
      eas-cli: 16.15.0
    Expo Workflow: bare

16/16 checks passed. No issues detected!

### Error output

Invalid environment. Use one of 'production', 'preview', or 'development'.

### Reproducible demo or steps to reproduce from a blank project

try to use the cli npx -y eas-cli@latest env:pull --environment development

valentinjadot:
Indeed! Same problem here:

[...]

PR #3216, opened 2025-10-08
Switch to lowercase in exec and pull

kadikraman:
# Why

Resolves https://github.com/expo/eas-cli/issues/3212

I switched to lowercase in https://github.com/expo/eas-cli/pull/3213 but missed exec and pull.

# How

Switched from `toUpperCase` to `toLowerCase` in `exec` and `pull`.
Updated `isEnvironment` validator to support both (both are supported in the api).

> [!NOTE]
> Not adding test here in favour of get this fixed quick, but longer term we need to add tests for these commands. This should have been caught by CI.

# Test Plan

```sh
easd env:pull --environment development
```

### DIFF
--- a/packages/eas-cli/src/build/utils/environment.ts
+++ b/packages/eas-cli/src/build/utils/environment.ts
@@ -17,7 +17,7 @@ const BuildProfileEnvironmentToEnvironment: Record<Environment, EnvironmentVaria
 
 export function isEnvironment(env: string): env is EnvironmentVariableEnvironment {
   return Object.values(EnvironmentVariableEnvironment).includes(
-    env as EnvironmentVariableEnvironment
+    env.toLowerCase() as EnvironmentVariableEnvironment
   );
 }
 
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/build/utils/environment.ts
```diff
@@ -17,7 +17,7 @@ const BuildProfileEnvironmentToEnvironment: Record<Environment, EnvironmentVaria
 
 export function isEnvironment(env: string): env is EnvironmentVariableEnvironment {
```
sjchmiela, line 18: Do we want to delete this function?

sjchmiela (approved):

packages/eas-cli/src/build/utils/environment.ts
kadikraman: Yes, but I'll do this in a follow-up PR. With tests. Want to make the smallest change we can here to get the commands working as before.

PR #3220, opened 2025-10-09
[eas-cli] [eas-json] support custom environment variable environments

kadikraman:
# Why

Resolves ENG-16068

We currently only support three enum environments: `development`, `preview` and `production`. We're extending this to allow for custom environment names.

# How

- Removed the `isEnvironment` check.
- Update the `eas.json` schema (also in www, website, turtle and here https://github.com/expo/eas-build/pull/616)
- Fetch current environments from the api when prompted

# Test Plan

- Updated unit tests
- Verified commands

### Create

```sh
# should have prompt
eas env:create

# create env var on development env
eas env:create --environment development --name "EXPO_PUBLIC_TEST" --value "value" --visibility plaintext --force --non-interactive --scope project

# create env var on development env (environment as implicit first arg)
eas env:create development --name "EXPO_PUBLIC_TEST" --value "value" --visibility plaintext --force --non-interactive --scope project

# create env var on custom env
eas env:create --environment custom --name "EXPO_PUBLIC_TEST" --value "value" --visibility plaintext --force --non-interactive --scope project

# create env var on custom env (environment as implicit first arg)
eas env:create custom --name "EXPO_PUBLIC_TEST" --value "value" --visibility plaintext --force --non-interactive --scope project
```

### List

```sh
eas env:list custom
eas env:list --environment custom
```

### Delete

```sh
eas env:delete --variable-environment custom --variable-name EXPO_PUBLIC_TEST
```

### Pull
```sh
eas env:pull custom
eas env:pull --environment custom
```

### Push
```sh
eas env:push another-custom
eas env:push --environment another-custom
```

### Update
```sh
eas env:update
```

### Get
```sh
eas env:get custom --variable-name EXPO_PUBLIC_TEST
```

### Exec
```sh
eas env:exec custom 'echo $EXPO_PUBLIC_TEST'
```

| Create      | List |
| ----------- | ----------- |
| Screenshot 2025-10-10 at 20 05 11      | Screenshot 2025-10-10 at 20 06 05       |

### Example permission error

Reading and deleting custom variables with custom envs will work, create and edit will show this error.

Screenshot 2025-10-10 at 20 10 27


packages/eas-cli/src/utils/prompts.ts
```diff
@@ -1,10 +1,33 @@
...
+): Promise<string[]> {
+  try {
+    const environments = await EnvironmentVariablesQuery.environmentVariableEnvironmentsAsync(
+      graphqlClient,
+      projectId
+    );
+    return environments;
+  } catch {
+    Log.warn('Failed to fetch custom environments from API, falling back to default environments.');
+    return [];
```
sjchmiela: This feels somewhat unexpected / wrong to me. Shouldn't we throw an error?

packages/eas-cli/src/utils/prompts.ts
```diff
@@ -74,49 +115,93 @@ export async function promptVariableVisibilityAsync(
...
 
 export async function promptVariableEnvironmentAsync({
   nonInteractive,
   selectedEnvironments,
   multiple = false,
-  availableEnvironments,
-}: EnvironmentPromptArgs & { multiple?: boolean }): Promise<
-  EnvironmentVariableEnvironment[] | EnvironmentVariableEnvironment
-> {
+  canEnterCustomEnvironment = false,
```
sjchmiela, line 133: When is it false and when is it true and why?

kadikraman: It's true for `create` and `edit`. For other commands it doesn't make sense e.g if you're `view`ing your env vars you're already getting shown a list on all envs on your account (plus the default envs) to pick from, so it makes no sense to free text a custom env to pull from - it would always be empty.

sjchmiela: Ah ok so `env:list` is interactive and may ask the user for an environment they want to list the variables from in which case it doesn't make sense to allow freeform input. Sorry for misunderstanding!

kadikraman: No, `env:list` isn't a command, I shouldn't have used code braces there. We have:
```
create - can set environment
delete
exec
get
list
pull
push - can set environment
update - can set environment
```

PR #3226, opened 2025-10-15
fix: workflow:logs should show build and repack logs

douglowder:
# Why

Build and repack jobs do not have logs at the same URL as other workflow jobs, so no logs were showing for these job types.

# How

- Update BuildFragment to include the log file URL
- Check for build jobs, and add the correct queries and fetch calls for those
- Update and fix tests
- Bump `@expo/eas-build-job` to 1.0.239 to fix TS error in environment type

# Test Plan

- Tested with various workflow runs, including ones with custom, build, repack, and update jobs.
- CI should pass.

PR #3227, opened 2025-10-15
feat: workflow:run --wait should show status updates

douglowder:
# Why

If a user executes `eas workflow:run --wait`, it would be desirable to show status updates for the individual jobs and steps in the workflow while the workflow run is in progress.

# How

Added code that leverages some of the same utilities used by `workflow:logs`. 

During the run, the logs for the various workflow jobs are retrieved and parsed, and the progress of each job is shown in the text of a spinner, showing the current step and the last few lines of the step's logs.

If the workflow run fails, the output will show jobs that failed, along with the failed step for each, and any logs for the failed steps.

Other changes:

- Bumped the "ora" package to the latest version, to pick up bug fixes
- Removed deprecated Node 18 from test runs
- Refactored the code in `workflow:run` that parses inputs, moving that code to a file in `commandUtils`
- Modified the Zod type for inputs to align with server side changes in https://github.com/expo/universe/pull/21950

# Test Plan

- Testing with different workflows
- CI should pass

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['18', '20', '22']
+        node: ['20', '22']
         include:
           - node: '20'
             coverage: true


douglowder:
* **#3227** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3227?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3226** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3226?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -120,6 +129,9 @@ export default class WorkflowRun extends EasCommand {
     const { flags, args } = await this.parse(WorkflowRun);
 
     if (flags.json) {
+      if (flags.wait) {
```
sjchmiela, line 132: It might make sense to allow waiting for workflow completion and _then_ print the final JSON?

sjchmiela (approved):

packages/eas-cli/src/commands/workflow/run.ts
douglowder: I was concerned that this would show mixed JSON (the final results) and non-JSON (spinners, status updates) in the output. Is this ok?

douglowder: For now, let's merge this as is, and we can create a separate task for correctly having both --json and --wait. I plan to do a PR like this one for `eas build` when the customer waits for output, so we can implement that correctly then. Maybe the status updates can go to stderr, and JSON can go to stdout...

PR #3234, opened 2025-10-21
[eas-cli] Create portable project archives

sjchmiela:
# Why

Project archives created on one platform can be extracted on another platform with incompatible metadata or mode bits. In particular, Windows read-only directories can be represented as read-only POSIX directories in the tarball, which can make worker extraction fail when child files need to be created.

Using `portable: true` consistently also strips owner/group metadata and avoids making Windows a special case.

Supersedes the approach in #3489 and avoids changing worker extraction as proposed in #3663.

# How

Create project tarballs with `portable: true` on all platforms.

# Test Plan

- `yarn workspace eas-cli test src/build/utils/__tests__/repository.test.ts`
- `yarn workspace eas-cli typecheck`
- `yarn run -T oxfmt CHANGELOG.md packages/eas-cli/src/build/utils/repository.ts packages/eas-cli/src/build/utils/__tests__/repository.test.ts`

### DIFF
--- a/packages/eas-cli/src/build/utils/repository.ts
+++ b/packages/eas-cli/src/build/utils/repository.ts
@@ -143,9 +143,16 @@ export async function makeProjectTarballAsync(vcsClient: Client): Promise<LocalF
 
   try {
     await vcsClient.makeShallowCopyAsync(shallowClonePath);
-    await tar.create({ cwd: shallowClonePath, file: tarPath, prefix: 'project', gzip: true }, [
-      '.',
-    ]);
+    await tar.create(
+      {
+        cwd: shallowClonePath,
+        file: tarPath,
+        prefix: 'project',
+        gzip: true,
+        portable: true,
+      },
+      ['.']
+    );
   } catch (err) {
     clearTimeout(timer);
     if (spinner.isSpinning) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- [eas-cli] Create portable project archives on all platforms to normalize cross-platform tar metadata and permissions. ([#3234](https://github.com/expo/eas-cli/pull/3234) by [@sjchmiela](https://github.com/sjchmiela))
 - [eas-cli] Remove hardcoded `builderEnvironment.image` override in `eas build:resign`. ([#3661](https://github.com/expo/eas-cli/pull/3661) by [@hSATAC](https://github.com/hSATAC))
 - [eas-cli] Fix `eas update --json` intermittently failing with JSON parse errors during "Computing project fingerprints" by passing `silent: true` to `@expo/fingerprint` to suppress subprocess stdout pollution. ([#3659](https://github.com/expo/eas-cli/pull/3659) by [@Mookiies](https://github.com/Mookiies))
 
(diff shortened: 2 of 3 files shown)


douglowder (approved):

packages/eas-cli/src/build/utils/repository.ts
```diff
@@ -143,9 +143,17 @@ export async function makeProjectTarballAsync(vcsClient: Client): Promise<LocalF
...
-      '.',
-    ]);
+    await tar.create(
+      {
+        cwd: shallowClonePath,
+        file: tarPath,
+        prefix: 'project',
+        gzip: true,
+        // Use portable mode on Windows to ensure proper permissions when extracting on Unix/Linux
+        ...(process.platform === 'win32' && { portable: true }),
```
kitten: I don't think we'd need to apply this only to windows?

sjchmiela: Correct me if I'm wrong — if we apply this to Linux too, it may so happen that:
- right now, a user has a `./bin/postcheckout.sh` file marked as executable
- they have an EAS Build lifecycle hook that does `./bin/postcheckout.sh`
- they've been happily `eas build`-ing for years
- we make archive "portable" not only on Windows (which is the broken scenario this is fixing), but also on Linux
- `./bin/postcheckout.sh` is **no longer** marked as executable in the archive
- EAS Build lifecycle hook breaks because `./bin/postcheckout.sh` is no longer excutable.

I don't want to break this flow — I consider Windows to be the odd environment here we need to put a workaround for. What do you think?

kitten: `portable` doesn't discard the mode byte, but instead masks it, so executable bits and such shouldn't be lost. The typical mask is `0o7777`.

I wouldn't of course trust node-tar here, but this is easily verified and we should verify it. Their own docs say:
> Additionally, mode is set to a "reasonable default" for most unix systems, based on a umask value of 0o22.

But they don't clarify what that reasonable default is. That said, verifying this and keeping it consistent between both platforms makes more sense to me.

Coming at it from the other angle: if we have a `+x` file in a git repo, we shouldn't see _any_ different behaviour no matter if we're tar-ing on Windows or on macOS/Linux. If this was an issue, this wouldn't be a fix, but just a change that affirms the bug. However, conceptually speaking, `portable` should do the right thing here, so assuming `tar` is implemented correctly, it should keep `umode` around and only mask it to a valid value.

Also worth noting, I'm pretty sure we usually use the `tar` _binary_ on macOS/Linux. If we are, this change wouldn't affect Windows either, which would make this check also redundant, and again, make it explicitly inconsistent

sswrk:
I spent a couple of hours today on this issue, here's what I found:

The problem is when a directory is set as read-only in NTFS. This works different than read-only directory in POSIX – read-only directory in Windows doesn't stop you from creating files inside it. In POSIX, read-only directory means that you can't create files in that directory. Now, when we're creating a tar in Windows, the mode of read-only directories is set to `0o555` in tar. When extracting this tar on a VM, we're seeing an error because it's forbidden to create files in such directory. So it's the "translation" from NTFS to POSIX that causes us trouble here. 

I think it would make sense to apply a mask to force the write bits on, just for directories in Windows.


Side note: if you do `chmod 555` on a non-empty directory in your project mac/Linux and then try doing `eas build`, it fails to create a tarball with `permission denied` error, because it fails to copy the files.

kitten (approved):

sswrk (approved):
Since we're sticking with the `portable` flag approach, I'm also confirming here that this indeed does fix the case I described above (I've reproduced it).

PR #3236, opened 2025-10-21
[eas-cli] Display usage warnings on `eas:build`

mackenco:
# Why

[Linear issue](https://linear.app/expo/issue/ENG-16908/add-warning-to-cli). We show a banner (on Free and Starter plans) in the website when an account is at >=85% of their build limit for that month. Adding this to `eas build` to surface the same info in the CLI 

# How

* Created a function `maybeWarnAboutUsageOveragesAsync` which pulls the account's number of builds in the billing period and limit. If the number of builds are >= 85% of the limit, display a warning at the beginning of the command. Display slightly different text for free vs paid. 

Screenshot 2025-10-24 at 11 57 59 AM

# Test Plan

I have a free account that has 29/30 builds; I'd run the command against that. Happy to give out the login for this account if people would like to QA it locally.

### DIFF
--- a/packages/eas-cli/src/utils/usage/__tests__/checkForOverages-test.ts
+++ b/packages/eas-cli/src/utils/usage/__tests__/checkForOverages-test.ts
@@ -88,7 +88,7 @@ describe('maybeWarnAboutUsageOveragesAsync', () => {
   it('displays a warning for a Free plan with high build usage', async () => {
     mockGetUsageForOverageWarningAsync.mockResolvedValue(
       createMockAccountUsage({
-        buildPlanMetrics: [createMockPlanMetric({ value: 85, limit: 100 })],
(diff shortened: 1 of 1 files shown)


tchayen (approved):

macksal:
This has introduced a bug in 16.25.0+. It evaluates the project config without first taking `env` from the relevant build profile. In my case it took down my CI because my config will throw if certain values aren't set in the environment. (I'm pinned to `latest` eas-cli, oops).

You can see that later steps in this command will only call `getDynamicPrivateProjectConfigAsync` after getting `env` from the build profile. We should probably follow that pattern or move this logic to a later step where env is already available.

mackenco:
@macksal this should be resolved in [16.28.0](https://github.com/expo/eas-cli/blob/main/CHANGELOG.md#16280---2025-11-20)

PR #3237, opened 2025-10-21
Add missing env for local builds with latest plugin

douglowder:
# Why

Currently, the CLI is using an old version of `eas-cli-local-build-plugin` for builds with `eas build --local`. This is not ideal, as this means that local builds are not using the most recent `build-tools` and other packages.

# How

- Modify `local.ts` so that the version of the plugin matches the version of `@expo/eas-build-job` used by the CLI
- Add some missing environment variables that are required in the most recent releases of `@expo/build-tools`

# Test Plan

- Testing with local builds on my dev machine
- CI should pass

### DIFF
--- a/packages/eas-cli/src/build/local.ts
+++ b/packages/eas-cli/src/build/local.ts
@@ -1,13 +1,14 @@
-import { Job, Metadata } from '@expo/eas-build-job';
+import { Job, Metadata, version } from '@expo/eas-build-job';
 import spawnAsync from '@expo/spawn-async';
 import { ChildProcess } from 'child_process';
 import semver from 'semver';
 
+import { getExpoApiBaseUrl } from '../api';
 import Log from '../log';
 import { ora } from '../ora';
 
 const PLUGIN_PACKAGE_NAME = 'eas-cli-local-build-plugin';
-const PLUGIN_PACKAGE_VERSION = '1.0.171';
+const PLUGIN_PACKAGE_VERSION = version; // should match version of @expo/eas-build-job
 
 export enum LocalBuildMode {
   /**
@@ -60,6 +61,7 @@ export async function runLocalBuildAsync(
       ...env,
       ...process.env,
       EAS_LOCAL_BUILD_WORKINGDIR: options.workingdir ?? process.env.EAS_LOCAL_BUILD_WORKINGDIR,
+      __API_SERVER_URL: getExpoApiBaseUrl(),
       ...(options.skipCleanup || options.skipNativeBuild
         ? { EAS_LOCAL_BUILD_SKIP_CLEANUP: '1' }
         : {}),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - workflow:run can use --wait and --json together. ([#3239](https://github.com/expo/eas-cli/pull/3239) by [@douglowder](https://github.com/douglowder))
+- Add missing env for local builds with latest plugin. ([#3237](https://github.com/expo/eas-cli/pull/3237) by [@douglowder](https://github.com/douglowder))
 
 ### 🧹 Chores
 
--- a/packages/eas-cli/package.json
+++ b/packages/eas-cli/package.json
@@ -12,7 +12,7 @@
     "@expo/code-signing-certificates": "0.0.5",
     "@expo/config": "10.0.6",
     "@expo/config-plugins": "9.0.12",
-    "@expo/eas-build-job": "1.0.239",
+    "@expo/eas-build-job": "1.0.243",
     "@expo/eas-json": "16.23.0",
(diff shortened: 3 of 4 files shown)


douglowder:
* **#3237** <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3237?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/expo/eas-cli/3237?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/build/local.ts
```diff
@@ -60,6 +61,8 @@ export async function runLocalBuildAsync(
       ...env,
       ...process.env,
       EAS_LOCAL_BUILD_WORKINGDIR: options.workingdir ?? process.env.EAS_LOCAL_BUILD_WORKINGDIR,
+      __API_SERVER_URL: getExpoApiGraphqlUrl(),
+      EAS_BUILD_ID: 'local',
```
sjchmiela: Hmm, this doesn't seem right. I admit I don't think I, for one, have remembered about local builds when developing latest changes in `build-tools`. Maybe we need to be more careful about `nullthrows(env.EAS_BUILD_ID)` and handle it better in there, instead of providing a `'local'`? Like, instead of `nullthrows`, just skip some stuff (like uploading job outputs). Maintaining bith `--local` and non-local is so weird…

packages/eas-cli/src/build/local.ts
```diff
@@ -60,6 +61,8 @@ export async function runLocalBuildAsync(
       ...env,
       ...process.env,
       EAS_LOCAL_BUILD_WORKINGDIR: options.workingdir ?? process.env.EAS_LOCAL_BUILD_WORKINGDIR,
+      __API_SERVER_URL: getExpoApiGraphqlUrl(),
```
sjchmiela:
```suggestion
      __API_SERVER_URL: getExpoApiBaseUrl(),
```

https://github.com/expo/universe/blob/823cfc36d623cc5f8d62e222576c12c154e27f92/server/www/k8s/production/www.env#L37 -> https://github.com/expo/universe/blob/5aff3d6e18cc68a0c253f945823ab8d9b7449b31/server/www/src/Env.ts#L62 -> https://github.com/expo/universe/blob/a9d0e51eb7512befc6bc9fc8d78326d40f8ac467/server/www/src/data/entities/turtle-job-run/TurtleJobRunUtils.ts#L160

packages/eas-cli/src/build/local.ts
douglowder: I'll take a look at changing `build-tools` to look for `env.EAS_BUILD_LOCAL` or something similar, and skip the steps that don't make sense in local builds.

sjchmiela: fwiw I think we can use https://github.com/expo/turtle-v2/blob/d7f1f40c22a279e8dd87c32decfea0ddd071b7b6/src/services/worker/src/env.ts#L37-L38

sjchmiela: vs https://github.com/expo/eas-build/blob/e6912b908dffbfa2293d59ae8af97ec12a57ec35/packages/local-build-plugin/src/build.ts#L33-L34

lol at `EAS_BUILD=true` vs `1`

packages/eas-cli/src/build/local.ts
```diff
@@ -3,11 +3,12 @@ import spawnAsync from '@expo/spawn-async';
 import { ChildProcess } from 'child_process';
 import semver from 'semver';
 
+import { getExpoApiGraphqlUrl } from '../api';
```
douglowder:
```suggestion
import { getExpoApiBaseUrl } from '../api';
```

packages/eas-cli/src/build/local.ts
douglowder: Ok I fixed the issue with the build ID, see changes in https://github.com/expo/eas-build/pull/624

packages/eas-cli/src/api.ts
```diff
@@ -112,3 +112,7 @@ export function getEASUpdateURL(projectId: string, manifestHostOverride: string
 export function getExpoApiWorkflowSchemaURL(): string {
   return getExpoApiBaseUrl() + '/v2/workflows/schema';
 }
+
+export function getExpoApiGraphqlUrl(): string {
```
sjchmiela: I don't think this is used?

sjchmiela (approved):

packages/eas-cli/src/api.ts
douglowder: Yep we can revert that change now

douglowder:
### Merge activity

* **Oct 23, 9:08 PM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.dev/github/pr/expo/eas-cli/3237).

PR #3250, opened 2025-11-05
[eas-cli] kick off update patch generation if enabled

quinlanj:
# Why

If a user has enabled bundle diffing in their app config (`updates.isBundleDiffingEnabled`), we optimistically kick off patch generation against the second most recent update on the branch. We do this by doing a `HEAD` request for the launch asset with diff headers.

In the future we can be smarter about this by getting stats from clickhouse

# How

- Updated the GraphQL schema with new fields and types
- Updated package dependencies to canary versions:
  - `@expo/config`: 10.0.6 → 12.0.11-canary-20251031-b135dff
  - `@expo/config-plugins`: 9.0.12 → 54.0.3-canary-20251031-b135dff
  - `@expo/prebuild-config`: 8.0.17 → 54.0.7-canary-20251031-b135dff

- Added bundle diffing functionality:
  - Implemented `prewarmDiffingAsync` to optimize update delivery
  - Added `isBundleDiffingEnabled` utility function
  - Moved `UpdatePublishPlatform` type to utils file for better organization

- Updated build script to use `typecheck-for-build` instead of direct tsc call

# Test Plan

- [x] test against a branch with a prior update
- [x] test against a branch with no prior update

### DIFF
--- a/packages/eas-cli/src/update/utils.ts
+++ b/packages/eas-cli/src/update/utils.ts
@@ -293,20 +293,16 @@ export async function prewarmDiffingAsync(
   }[];
 
   for (const update of newUpdates) {
-    try {
-      const manifest = JSON.parse(update.manifestFragment);
-      const launchAssetKey: string | undefined = manifest.launchAsset?.storageKey;
-      const requestedUpdateId: string = update.id;
-      if (!launchAssetKey || !requestedUpdateId) {
-        continue;
-      }
-      toPrewarm.push({
-        update,
-        launchAssetKey,
-      });
-    } catch {
-      // ignore updates without launch assets
+    const manifest = JSON.parse(update.manifestFragment);
+    const launchAssetKey: string | undefined = manifest.launchAsset?.storageKey;
+    const requestedUpdateId: string = update.id;
+    if (!launchAssetKey || !requestedUpdateId) {
+      continue;
     }
+    toPrewarm.push({
+      update,
+      launchAssetKey,
+    });
   }
 
   await Promise.allSettled(
@@ -330,12 +326,12 @@ export async function prewarmDiffingAsync(
           launchAssetKey,
         ]);
         const first = signed?.[0];
-        if (!first?.url || !first?.headers) {
+        if (!first) {
           return;
         }
 
         const headers: Record<string, string> = {
-          ...(first.headers as Record<string, string>),
+          ...(first.headers as Record<string, string> | undefined),
           'expo-current-update-id': nextMostRecentUpdateId,
           'expo-requested-update-id': update.id,
           'expo-embedded-update-id': DUMMY_EMBEDDED_UPDATE_ID,


quinlanj:
* **#3250** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3250?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3250?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,108 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
     },
   };
 }
+
+export function isBundleDiffingEnabled(exp: ExpoConfig): boolean {
+  return exp.updates?.enableBsdiffPatchSupport === true;
```
wschurman: from in-person discussion in meeting, probably better to just use an `any` cast until the config is released, then upgrade and remove the any cast.

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,108 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
...
+}
+
+export function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
```
wschurman: can this be simplified by using `Promise.race` and `setTimeout` from `timers/promises`?

```typescript
export async function withTimeoutAsync<T>(promise: Promise<T>, ms: number): Promise<T> {
  return await Promise.race([
    promise,
    (async () => {
      await setTimeout(ms);
      throw new Error(`Timed out after ${ms} ms`);
    })(),
  ]);
}
```
or
```typescript
export async function withTimeoutAsync<T>(promise: Promise<T>, ms: number): Promise<T> {
  return await Promise.race([
    promise,
    setTimeout(ms).then(() => {
      throw new Error(`Timed out after ${ms} ms`);
    }),
  ]);
}
```

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,108 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
...
+      });
+    } catch {
+      // ignore updates without launch assets
+    }
+  }
+
+  await Promise.allSettled(
+    toPrewarm.map(async ({ update, launchAssetKey }) => {
+      try {
+        // Check to see if there's a second most recent update so we can pre-emptively generate a patch for it
```
wschurman, line 311: I'm having a hard time wrapping my head around whether this needs to take into account per-update rollouts? I _think_ the answer is no, since the control will by definition be the second-most-recent update, but I may be forgetting something.

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,108 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
...
+
+        const headers: Record<string, string> = {
+          ...(first.headers as Record<string, string>),
+          'expo-current-update-id': nextMostRecentUpdateId,
+          'expo-requested-update-id': update.id,
+          'expo-embedded-update-id': DUMMY_EMBEDDED_UPDATE_ID,
+          'a-im': 'bsdiff',
+        };
+
+        const controller = new AbortController();
```
wschurman: I think `AbortSignal.timeout` is equivalent

wschurman (changes requested):

packages/eas-cli/src/update/utils.ts
quinlanj: i scrapped `withTimeoutAsync` and replaced it with the Abort timeout signal

packages/eas-cli/src/update/utils.ts
quinlanj: should be fine in terms of correctness. in terms of usefulness, if the control update was published much earlier than the rolled out one, there's a good chance we've got people who are running the control update so the patch would be useful. but if they are published close together, the patch will not be a useful optimization in that case

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,90 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
...
+      if (!launchAssetKey || !requestedUpdateId) {
+        continue;
+      }
+      toPrewarm.push({
+        update,
+        launchAssetKey,
+      });
+    } catch {
+      // ignore updates without launch assets
+    }
```
wschurman: what portion of the code in the `try` can fail? from my read it gracefully handles updates without launch assets so that wouldn't throw. I think the only thing that could throw is the JSON.parse?

I think prefer to not wrap in try/catch here.

```suggestion
    const manifest = JSON.parse(update.manifestFragment);
    const launchAssetKey: string | undefined = manifest.launchAsset?.storageKey;
    const requestedUpdateId: string = update.id;
    if (!launchAssetKey || !requestedUpdateId) {
      continue;
    }
    toPrewarm.push({
      update,
      launchAssetKey,
    });
```

packages/eas-cli/src/update/utils.ts
```diff
@@ -269,3 +274,90 @@ export function getBranchDescription(branch: UpdateBranchFragment): FormattedBra
...
+        const signed = await AssetQuery.getSignedUrlsAsync(graphqlClient, update.id, [
+          launchAssetKey,
+        ]);
+        const first = signed?.[0];
+        if (!first?.url || !first?.headers) {
+          return;
+        }
+
+        const headers: Record<string, string> = {
+          ...(first.headers as Record<string, string>),
```
wschurman: I think URL is required? Or is this checking empty string?

Also since headers is optional I think we should continue with the fetch even if headers is required.


```suggestion
        if (!first) {
          return;
        }

        const headers: Record<string, string> = {
          ...(first.headers as Record<string, string> | undefined),
```

quinlanj:
@wschurman confirming `signedUrls` does return an overriden URL: https://github.com/expo/universe/blob/4eb8286b1e2e257ba13896e5d71836770df887d4/server/www/src/graphql/queries/AssetQuery.ts#L230

PR #3254, opened 2025-11-10
[eas-cli] remove plugins evaluation from workflow:run

Kudo:
# Why

customer requested feature to run `eas workflow:run` without node_modules installation

# How

if people having custom config-plugins, the `eas workflow:run` will need node_modules installation. this pr tries to remove the requirements

# Test Plan

```sh
$ bunx create-expo-app -t blank@sdk-54 sdk54
$ cd sdk54
$ npx expo install expo-build-properties
$ bunx eas-cli workflow:create
$ rm -rf node_modules
$ bunx eas-cli workflow:run .eas/workflows/test.yml
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🧹 Chores
 
 - Fix yarn watch. ([#3251](https://github.com/expo/eas-cli/pull/3251) by [@quinlanj](https://github.com/quinlanj))
+- Remove the **node_modules** installation requirement for `workflow:run` command. ([#3254](https://github.com/expo/eas-cli/pull/3254) by [@kudo](https://github.com/kudo))
 
 ## [16.26.0](https://github.com/expo/eas-cli/releases/tag/v16.26.0) - 2025-10-29
 


packages/eas-cli/src/commands/workflow/run.ts
```diff
@@ -135,7 +135,9 @@ export default class WorkflowRun extends EasCommand {
     const {
       projectId,
       exp: { slug: projectName },
-    } = await getDynamicPrivateProjectConfigAsync();
+    } = await getDynamicPrivateProjectConfigAsync({
+      skipPlugins: true,
```
sjchmiela, line 139: Oh interesting! Didn't know this was a thing! And I guess config plugins are not expected to influence the project ID anyhow, right?

Kudo: > I guess config plugins are not expected to influence the project ID anyhow, right?

a config plugin mod can theoretically do anything like changing the project id in app.json, or even removing itself from the plugins array. however, that would break much assumptions. i expect people would not do that. it's just a bad idea.

that said, it should be safe the skip plugins when getting project id and project slug.

PR #3257, opened 2025-11-14
[eas-cli] fix: When warning about usage overages, the build profile's env was not applied before evaluating config

macksal:
# Why

#3236 was recently merged, adding a check for the EAS account's build usage and issuing a warning before running a build. It introduced a bug: it evaluates the project's config to get the EAS projectId, but it doesn't apply the `env` from the build profile beforehand. All other code that reads the config will first

This created an issue in my CI builds where I am pinned to the latest CLI version. Because my `app.config.ts` depends on certain values being set in the environment (and throws if they aren't): this caused the entire build command to fail. It seems sensible to keep the previous behaviour where `eas build` will always apply `env` from the relevant build profile before evaluating `app.config.ts`.

# How

Move the code that issues the warning to deeper within the runBuildAndSubmitAsync flow, after the build profile has been read, and the `env` from that profile is available and used when evaluating the project config.

# Test Plan

The test added in #3236 is still passing, and this PR removes the problematic behaviour (evaluating the project config without applying env from the build profile).

### DIFF
--- a/packages/eas-cli/src/build/runBuildAndSubmit.ts
+++ b/packages/eas-cli/src/build/runBuildAndSubmit.ts
@@ -52,6 +52,7 @@ import {
 import { checkExpoSdkIsSupportedAsync } from '../project/expoSdk';
 import { validateMetroConfigForManagedWorkflowAsync } from '../project/metroConfig';
 import {
+  getOwnerAccountForProjectIdAsync,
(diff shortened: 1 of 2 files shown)


macksal:
/changelog-entry bug-fix When warning about usage overages, the build profile's env was not applied before evaluating config

PR #3260, opened 2025-11-17
[eas-cli] Fix usage warning bug in `eas:build`

mackenco:
# Why

Received a [bug report ](https://github.com/expo/eas-cli/pull/3236#issuecomment-3532051546) that the `eas:build` command was failing on latest. We were checking the project config too early. 

# How

Moved the `maybeWarnAboutUsageOveragesAsync` call later into the command; now it's called in `prepareAndStartBuildAsync`, after we know we have a config. I put it early in this function because 1) I wanted to warn _before_ we start the build 2) I wanted it to come before most of the other info logged by the command but 3) after `createBuildContextAsync` is run, since this grabs the `accountId`. 

# Test Plan

I overrode the logic in `maybeWarnAboutUsageOveragesAsync` to always warn locally. I think it's a little odd that it comes after the incremented log, but felt like a logical place in the code for this to live. 
Multiple builds: 
Multi build

Single build:
Screenshot 2025-11-17 at 1 39 46 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,5 +37,6 @@
   "volta": {
     "node": "20.11.0",
     "yarn": "1.22.21"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -409,6 +411,15 @@ async function prepareAndStartBuildAsync({
     env,
   });
 
+  if (!hasWarnedAboutUsageOverages && !flags.localBuildOptions.localBuildMode) {
```
mackenco, line 414: This this function could run multiple times (per build profile), we want to track if we've warned and only do it once.

douglowder (approved):

PR #3269, opened 2025-11-30
Allow exit from credentials command

douglowder:
# Why

The command `eas credentials` does not provide a way to exit other than control-C, if no platform flag is passed in.

# How

Add an "Exit" choice to the platform prompt.

# Test Plan

Tested locally.

### DIFF
--- a/packages/eas-cli/src/credentials/manager/SelectPlatform.ts
+++ b/packages/eas-cli/src/credentials/manager/SelectPlatform.ts
@@ -3,7 +3,7 @@ import { ManageIos } from './ManageIos';
 import { Analytics } from '../../analytics/AnalyticsManager';
 import { DynamicConfigContextFn } from '../../commandUtils/context/DynamicProjectConfigContextField';
 import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
-import { selectPlatformAsync } from '../../platform';
+import { selectPlatformWithExitOptionAsync } from '../../platform';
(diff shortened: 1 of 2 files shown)


douglowder:
* **#3269** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/credentials/manager/SelectPlatform.ts
```diff
@@ -20,7 +20,7 @@ export class SelectPlatform {
   ) {}
 
   async runAsync(): Promise<void> {
-    const platform = await selectPlatformAsync(this.flagPlatform);
+    const platform = await selectPlatformAsync(this.flagPlatform, true /* allowExit */);
```
jonsamp: Just curious: why are we passing in a parameter here if it's hard coded to `true`? Is there a future-proofing or code consistency reason?

douglowder: This is an optional parameter that adds the "Exit" option to the prompt shown by this method. This method is used in several places, including places where the exit option should not be shown. I can add a wrapper method instead (e.g. `selectPlatformAsyncWithExitOption`​) to make this a little clearer.

PR #3279, opened 2025-12-26
Feat: submit by local machine

linrongda:
# Why

The current EAS submit only supports cloud-based submission, and the `what-to-test` feature is exclusively available on the enterprise plan.

# How

I added the `packages/eas-cli/src/submit/utils/local.ts` to support local submission, external distribution and etc. by fastlane.

Now, you can use `--local` to submit your app by local machine, and use `--fastlane-args` to support other fastlane args, like `--distribute_external` and etc.

example:

```sh
eas submit --platform ios --profile production --non-interactive --local --path=./example.ipa --fastlane-args="--distribute_external"
```

# Test Plan

I am not good at ts, so I just test in the project in https://github.com/west2-online/fzuhelper-app/pull/318 , the corresponding action is https://github.com/west2-online/fzuhelper-app/actions/runs/20535766528

I also add some unit tests
```sh
npx jest packages/eas-cli/src/submit/utils/__tests__/local-test.ts --runInBand --colors
npx jest packages/eas-cli/src/submit/utils/__tests__/splitArgsString-test.ts --runInBand --colors
```

### DIFF
--- a/packages/eas-cli/src/commands/submit.ts
+++ b/packages/eas-cli/src/commands/submit.ts
@@ -145,7 +145,7 @@ export default class Submit extends EasCommand {
       withServerSideEnvironment: null,
     });
 
-    const flags = this.sanitizeFlags(rawFlags);
+    const flags = this.sanitizeFlags(rawFlags as RawCommandFlags);
 
     await maybeWarnAboutEasOutagesAsync(graphqlClient, [StatuspageServiceName.EasSubmit]);
 
--- a/packages/eas-cli/src/submit/utils/local.ts
+++ b/packages/eas-cli/src/submit/utils/local.ts
@@ -1,11 +1,13 @@
 import { Platform } from '@expo/eas-build-job';
 import { spawn, spawnSync } from 'child_process';
 import fs from 'fs-extra';
+import StreamZip from 'node-stream-zip';
 import os from 'os';
 import path from 'path';
 
 import { AppStoreConnectApiKeyQuery } from '../../graphql/queries/AppStoreConnectApiKeyQuery';
 import Log from '../../log';
+import { parseBinaryPlistBuffer } from '../../utils/plist';
 import { SubmissionContext } from '../context';
 import {
   AscApiKeySource,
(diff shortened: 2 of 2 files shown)


linrongda:
@douglowder
Hello! Would you please take your spare time to review my PR?
Thank you very much!

the-kenneth:
A `--local` flag on eas submit would be useful for us as well. Supporting Android would be great too! 🙏 

Also why is the `what-to-test` arg behind a enterprise plan? Its a very simple arg; its puzzling for the need to have to write a custom fastlane upload script to just get that arg.

linrongda:
> A `--local` flag on eas submit would be useful for us as well. Supporting Android would be great too! 🙏

The app developed by my team is only available in mainland China, and we have no plan to submit it to Google Play Store, so I can't test the Android submission. If you can help me improving this function, I would be grateful.

> Also why is the `what-to-test` arg behind a enterprise plan? Its a very simple arg;

#3023 said that:
This uses related features in other EAS services to provide changelog of a submission. At the moment, this consumes extra computing resources during the process so we're going to make this enabled only for enterprise plan.

> its puzzling for the need to have to write a custom fastlane upload script to just get that arg.

EAS build (iOS) is based on fastlane, so I guess that EAS submit is also based on fastlane on their server.

To add the changelog, you can use it like:

```sh
eas submit --platform ios --profile production --non-interactive --local --path=./example.ipa --what-to-test="changelog" --fastlane-args="--distribute_external"
```

sjchmiela:
Hey @linrongda! I appreciate your pull request very much! You've done a good job.

Unfortunately, merging this in would mean we'd take on the responsibility of maintaining this which is not something we can do. There's so much other stuff to build that is going to help EAS users have better experience!

This being said:
- while indeed `changelog` is limited to Enterprise customers currently in `eas submit`, it is _not_ limited to Enterprise customers if you use [`type: testflight` workflow job](https://docs.expo.dev/eas/workflows/pre-packaged-jobs/#testflight). I encourage you to try it! You can expect more and more improvements coming to workflows in the coming months. Let me know if you have hard time writing a workflow that uses this workflow job to submit an app. [Workflows can accept inputs](https://docs.expo.dev/eas/workflows/syntax/#onworkflow_dispatchinputs).
- you are correct that EAS Submit uses Fastlane under the hood, so writing… even a simple shell script that mimics what `eas submit --local` would do should not be too hard of a problem.
- as far as I know, the only missing part for the script are credentials — you'd just need to generate a new ASC API key and use it for the submission.

PR #3282, opened 2026-01-06
refactor(eas-cli): Remove `npx expo` invocations and config/entitlements fallbacks

kitten:
See: https://exponent-internal.slack.com/archives/C017N0N99RA/p1758469493137929
Resolves: https://github.com/expo/expo/issues/40227

# Why

> [!NOTE]
> There's still a lot of leftover reliance on `@expo/config` (outdated), `@expo/prebuild-config`, and some other internals, which are prone to failing and not being updated. A lot of them don't have replacement code paths that are obvious but stood out to @byCedric when implementing some Launch features and replicating/replacing some logic from `eas-cli`

All invocations of `npx` are prone to failure and unnecessary:
- `npx expo install` is redundant if we can invoke the Expo CLI directly (which forks if needed automatically)
- `npx expo config --json` is unreliable, if npx adds additional output to the standard output. It also can fail if there's a global provider of `expo`, another `expo` in the npm-resolved `PATH`, or for various other reasons
- the same applies to the entitlements call

Additionally, we have a few fallbacks that are invalid. The entitlements fallback doesn't work and replaces the failed `npx expo config` command with obscure output, failing to also print its `Log.warn`s in some cases, which are confusing.

The `@expo/config` fallback has to be left in place currently as there's no prerequisite of `expo` being installed (which there should be) for `getExpoConfigInternalAsync` to be invokable. The check for `isExpoInstalled` is also faulty as it doesn't resolve `expo` but instead checks the `package.json`, but to keep this PR leaner, this isn't addressed here.

The fallbacks are generally for older versions of the SDK which are now unsupported.

# How

- Replace all `npx expo` invocations with a direct `expo` CLI invocation (code already existed for this)
- Remove entitlements and config error fallbacks, which were confusing

# Test Plan

Most of these changes are speculative and untested right now.

### DIFF
--- a/packages/eas-cli/src/commandUtils/new/commands.ts
+++ b/packages/eas-cli/src/commandUtils/new/commands.ts
@@ -7,6 +7,7 @@ import { canAccessRepositoryUsingSshAsync, runGitCloneAsync } from '../../onboar
 import { PackageManager, installDependenciesAsync } from '../../onboarding/installDependencies';
 import { runCommandAsync } from '../../onboarding/runCommand';
 import { ora } from '../../ora';
+import { expoCommandAsync } from '../../utils/expoCli';
 
 export async function cloneTemplateAsync(targetProjectDir: string): Promise<string> {
   const githubUsername = 'expo';
@@ -48,16 +49,8 @@ export async function installProjectDependenciesAsync(
   });
 
   const dependencies = ['expo-updates', '@expo/metro-runtime'];
-  for (const dependency of dependencies) {
-    spinner.text = `Installing ${chalk.bold(dependency)}`;
-    await runCommandAsync({
-      cwd: projectDir,
-      command: 'npx',
-      args: ['expo', 'install', dependency],
-      showOutput: false,
-      showSpinner: false,
-    });
-  }
+  spinner.text = `Installing ${dependencies.map(dep => chalk.bold(dep)).join(', ')}`;
+  await expoCommandAsync(projectDir, ['install', ...dependencies], { silent: true });
   spinner.succeed(`Installed project dependencies`);
 }
 
--- a/packages/eas-cli/src/project/expoConfig.ts
+++ b/packages/eas-cli/src/project/expoConfig.ts
@@ -5,7 +5,6 @@ import Joi from 'joi';
 import path from 'path';
 
 import { isExpoInstalled } from './projectUtils';
-import Log from '../log';
 import { spawnExpoCommand } from '../utils/expoCli';
 
 export type PublicExpoConfig = Omit<
@@ -41,8 +40,6 @@ export async function createOrModifyExpoConfigAsync(
   }
 }
 
-let wasExpoConfigWarnPrinted = false;
-
 async function getExpoConfigInternalAsync(
   projectDir: string,
   opts: ExpoConfigOptionsInternal = {}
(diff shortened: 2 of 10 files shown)


douglowder (approved):
Thanks for doing this!! I’ll have a look to see if the tests can be improved in a separate PR, and also have a look at removing some of the fallbacks that are no longer needed with the current supported Expo SDK.

packages/eas-cli/src/commandUtils/new/commands.ts
```diff
@@ -48,16 +49,8 @@ export async function installProjectDependenciesAsync(
...
-    await runCommandAsync({
-      cwd: projectDir,
-      command: 'npx',
-      args: ['expo', 'install', dependency],
-      showOutput: false,
-      showSpinner: false,
-    });
-  }
+  spinner.text = `Installing ${dependencies.map(dep => chalk.bold(dep)).join(', ')}`;
+  await expoCommandAsync(projectDir, ['install', ...dependencies], { silent: true });
```
sjchmiela, line 53: 🙏

packages/eas-cli/src/project/expoConfig.ts
```diff
@@ -56,35 +53,16 @@ async function getExpoConfigInternalAsync(
...
-            `Failed to read the app config from the project using "npx expo config" command: ${err.message}.`
-          );
-          Log.warn('Falling back to the version of "@expo/config" shipped with the EAS CLI.');
-          wasExpoConfigWarnPrinted = true;
+      const { stdout } = await spawnExpoCommand(
+        projectDir,
+        ['config', '--json', ...(opts.isPublicConfig ? ['--type', 'public'] : [])],
+        {
+          env: {
+            EXPO_NO_DOTENV: '1',
```
sjchmiela, line 61: what about opts.env?

packages/eas-cli/src/project/expoConfig.ts
```diff
@@ -56,35 +53,16 @@ async function getExpoConfigInternalAsync(
...
+          },
         }
-        exp = getConfig(projectDir, {
```
sjchmiela, line 82: I presume we _want_ to throw an error now if expo is installed but it fails?

packages/eas-cli/src/project/ios/entitlements.ts
```diff
@@ -1,69 +1,27 @@
...
-        wasExpoConfigPluginsWarnPrinted = true;
-      }
-      const { exp } = await getPrebuildConfigAsync(projectDir, { platforms: ['ios'] });
-      expWithMods = await compileModsAsync(exp, {
-        projectRoot: projectDir,
-        platforms: ['ios'],
-        introspect: true,
-        ignoreExistingNativeFiles: await hasIgnoredIosProjectAsync(projectDir, vcsClient),
-      });
+  const { stdout } = await spawnExpoCommand(
```
sjchmiela, line 16: When should we use `spawnExpoCommand` and when should we use `expoCommandAsync`?

sjchmiela (approved):
godspeed

packages/eas-cli/src/project/expoConfig.ts
kitten: that was dead code: https://github.com/expo/eas-cli/blob/bb46b4fbb47a51b1a2ab3facf5f53069e248de7a/packages/eas-cli/src/project/expoConfig.ts#L49-L52

I don't think we want to do this, but just pointing it out, since I think this should be removed 😆

packages/eas-cli/src/project/expoConfig.ts
kitten: yep, https://github.com/expo/eas-cli/blob/bb46b4fbb47a51b1a2ab3facf5f53069e248de7a/packages/eas-cli/src/utils/expoCli.ts#L91-L98

also `getConfig` below needs to be replaced/removed, or the invariant needs to be enforced more strictly in the future

packages/eas-cli/src/project/ios/entitlements.ts
kitten: one exists purely to preserve the log pass-through. If we don't want that anymore, we should unify or delete one of them. `spawnExpoCommand` does what it says on the tin, it spawns the command, nothing more. Hence, `expoCommandAsync` is left-over with the "legacy" function name that wrapped stdout

PR #3297, opened 2026-01-09
[gha] Add conditional checks for GCP upload step

sjchmiela:
It takes a minute https://github.com/expo/eas-cli/actions/runs/20850818613/job/59905175877?pr=3286 and is unnecessary unless the other step runs too


sjchmiela:
Will include in https://github.com/expo/eas-cli/pull/3303

PR #3298, opened 2026-01-09
[infra] Cancel EAS runs when system-tests workflow is canceled

sjchmiela:
# Why

- Canceling the GitHub Actions workflow should cancel the underlying EAS build/workflow runs to avoid wasting resources.

# How

- Start the iOS build and Android workflow with `--no-wait` and capture their IDs from JSON.
- Poll the EAS build/workflow statuses to determine success/failure.
- Persist IDs to `GITHUB_ENV` and cancel them in a `if: cancelled()` step.

# Test Plan

https://github.com/expo/eas-cli/actions/runs/20859556010/job/59935355509

### DIFF
--- a/.github/workflows/worker-system-tests.yml
+++ b/.github/workflows/worker-system-tests.yml
@@ -85,8 +85,16 @@ jobs:
 
           BUILD_STATUS=""
           WORKFLOW_STATUS=""
+          TIMEOUT_SECONDS=1800
+          START_TIME=$(date +%s)
 
           while true; do
+            ELAPSED=$(( $(date +%s) - START_TIME ))
+            if [ "$ELAPSED" -ge "$TIMEOUT_SECONDS" ]; then
+              echo "Timeout: polling exceeded 30 minutes"
+              exit 1
+            fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/worker-system-tests.yml
```diff
@@ -47,30 +47,106 @@ jobs:
...
-          echo "iOS build exit code: $EXIT_IOS"
-          echo "Android workflow exit code: $EXIT_ANDROID"
+          BUILD_STATUS=""
+          WORKFLOW_STATUS=""
 
-          # Fail if either failed
-          if [ $EXIT_IOS -ne 0 ] || [ $EXIT_ANDROID -ne 0 ]; then
-            echo "One or more builds failed"
-            exit 1
+          while true; do
```
douglowder, line 91: Should there be a failsafe timeout for this loop?

sjchmiela: Shouldn't be a problem to add one! (Technically builds have max run time of 2 hours and if workflow run wouldn't get marked with a terminal status we should ideally fix the _underlying_ issue, but yea.)

PR #3303, opened 2026-01-12
[gha] Merge setup, build, and deploy into single job

sjchmiela:
## Why

The build-and-deploy-worker workflow had three separate jobs (setup, build-workers, deploy) with complex inter-job dependencies and redundant conditional logic. I think if we put it all in one job it should make stuff faster.

## How

Consolidated all three jobs into a single `build-and-deploy` job with conditional steps:

1. **Check**: Verify if tarballs for the current SHA exist in GCS (when deploying)
2. **Build**: Execute EAS workflow if needed (PR testing or cache miss)
3. **Upload**: Upload tarballs to GCS with SHA-based naming
4. **Deploy**: Copy tarballs to environment-specific names (when deploying)

Key simplifications:
- **Assumption**: Main branch always specifies an environment (staging/production), so `environment == ''` only occurs on PR builds
- **SHA-based caching**: Check cache by SHA regardless of branch - if the code is identical (same SHA), reuse existing tarballs
- **Clearer conditions**: Each step has a single clear purpose without overlapping branch-specific logic

I also added back building Linux tarball on Linux not to mix CPU architectures accidentally.

## Test Plan

[Build on PR](https://github.com/expo/eas-cli/actions/runs/20920297113/job/60104039095?pr=3303). [Deployment from a branch](https://github.com/expo/eas-cli/actions/runs/20921082109/job/60106516235).

### DIFF
--- a/packages/worker/.eas/workflows/build-worker.yml
+++ b/packages/worker/.eas/workflows/build-worker.yml
@@ -5,8 +5,8 @@ on:
     inputs:
       upload_artifact:
         description: Whether to upload artifacts
-        type: string
-        default: "true"
+        type: boolean
+        default: true
 
 jobs:
   build_android_worker:
@@ -20,7 +20,7 @@ jobs:
         run: ./package.sh worker-android.tar.gz android
 
       - name: Upload Android worker tarball as artifact
-        if: ${{ inputs.upload_artifact == 'true' }}
+        if: ${{ inputs.upload_artifact }}
         uses: eas/upload_artifact
         with:
           name: worker-android-tarball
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy-worker.yml
```diff
@@ -179,30 +184,18 @@ jobs:
             npx -y eas-cli@latest workflow:cancel "$WORKFLOW_RUN_ID" --non-interactive
           fi
 
-      - uses: ./.github/internal-actions/setup-gcloud
-
       - name: Upload worker tarballs to GCP
-        if: github.ref == 'refs/heads/main' || inputs.environment != ''
+        # Upload when deploying if we built something new
+        # Don't upload PR builds (no environment) or when using cached tarballs
+        if: inputs.environment != '' && steps.check_worker_exists.outputs.worker_exists == 'false'
```
AbbanMustafa, line 190: If we do want to deploy from our branch to staging lets say we will just need to specify the environment correct? Same as before

sjchmiela: Yes! Tested this [here](https://github.com/expo/eas-cli/actions/runs/20921082109/job/60106516235).

PR #3304, opened 2026-01-12
[worker] Build and use single tarball

sjchmiela:
@hSATAC rightly noticed Android and iOS tarballs are not that different (iOS is Android + gem and the gem only weighs 0.1 MB).

We still use `worker-android-staging.tar.gz` and `worker-ios-staging.tar.gz` (this requires changes to Launcher and the Orchestrator), but commit deployments use a no-platform `worker-SHA.tar.gz`.


sjchmiela:
Actually we can't use it because Linux and iOS use different CPU architectures so we don't want to coalesce them.

PR #3312, opened 2026-01-13
[eas-cli] Add --browser option to `eas login`

byronkarlen:
# Why
This is dependent on: https://app.graphite.com/github/pr/expo/universe/24379/%5Bwebsite%5D-Configure-eas-cli-auth-redirect-webpage

And related to: https://github.com/expo/expo/pull/42614

Support for third-party sign in (Apple/Google/GitHub). `eas login` should eventually default to this but it would be a breaking change.

We can also ship this now (although we don't have to). I would opt to ship it, but keep the option hidden from `--help` output until we ship apple sign in.

# How
Mirrored the SSO browser flow but just made it more generic.

Also changed the local web server to redirect to expo.dev on success and failures so that the webpage looks better/more professional.

# Test Plan
Verified locally with `yarn run eas login --browser` and `yarn run eas login -b`.

### DIFF
--- a/packages/eas-cli/src/commands/account/login.ts
+++ b/packages/eas-cli/src/commands/account/login.ts
@@ -21,7 +21,6 @@ export default class AccountLogin extends EasCommand {
       description: 'Login with your browser',
       char: 'b',
       default: false,
-      hidden: true, // soft launch
     }),
   };
 


byronkarlen:
/changelog-entry new-feature Add --browser flag to login command for browser-based authentication

PR #3313, opened 2026-01-14
[worker] bump build-tools

AbbanMustafa:
# Why

https://github.com/expo/eas-build/pull/653


yarn.lock
```diff
@@ -1797,6 +1797,16 @@
     semver "^7.6.2"
     zod "^4.1.3"
 
+"@expo/eas-build-job@1.0.263":
```
sjchmiela, line 1790: need to update other packages too

sjchmiela: maybe we want to copy over the `update-all-eas-dependencies.sh` from `turtle-v2`?

AbbanMustafa: that would be good yes

PR #3340, opened 2026-01-26
[eas-cli] Add multi-account support

brentvatne:
## Summary

- Add multi-account session management, allowing users to be logged into multiple Expo accounts simultaneously
- Add `account:list` command to show all logged-in accounts
- Add `account:switch` command to change the active account
- Update `account:login` to support adding additional accounts
- Update `account:logout` to support logging out specific accounts or all accounts
- Add global `--account` flag for per-command account selection

## Motivation

I've wanted this for a while since I personally switch between Expo accounts fairly often. I've also heard from other folks who switch between company and personal accounts, or accounts across different projects that their agency manages, that this would be helpful.

## Test plan

Set `EAS_EXPERIMENTAL_ACCOUNT_SWITCHER=1` in your env, or prefix all of the commands you run with it.

- [x] `eas account:login` - login to a new account (adds to existing sessions)
- [x] `eas account:list` - shows all logged-in accounts with active indicator
- [x] `eas account:switch` - switch between accounts interactively or by name
- [x] `eas account:logout` - logout current, specific, or all accounts
- [x] `eas account:view` - shows current active account
- [x] `eas <any-command> --account <name>` - uses specified account for that command only (works with any command)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Fix `--account` flag for experimental multi-account switcher to work with oclif argument parsing. ([#3340](https://github.com/expo/eas-cli/pull/3340) by [@brentvatne](https://github.com/brentvatne))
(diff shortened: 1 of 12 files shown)


YashTotale:
@brentvatne This would be a really nice addition. Are there any plans to merge this PR?

PR #3341, opened 2026-01-26
[eas-cli] Fix --source-maps flag compatibility with expo CLI

brentvatne:
## Summary

- Fix compatibility issue with `--source-maps` flag across different SDK versions
- SDK 55+ supports `--source-maps` with a value (e.g., `inline`), but older SDKs only support it as a boolean flag
- Passing a value to older SDKs causes it to be parsed as the project root positional argument, resulting in "Invalid project root" errors

## Test plan

- [x] Test `eas update` with `--source-maps` on SDK 55+ project
- [x] Test `eas update` with `--source-maps` on pre-SDK 55 project (should use boolean flag)
- [x] Unit tests added for SDK version detection

### DIFF
--- a/packages/eas-cli/src/project/publish.ts
+++ b/packages/eas-cli/src/project/publish.ts
@@ -257,17 +257,7 @@ export async function buildBundlesAsync({
         ? ['--platform', 'ios', '--platform', 'android']
         : ['--platform', platformFlag];
 
-    // Use --source-maps if provided, otherwise fall back to --dump-sourcemap.
-    // SDK 55+ supports --source-maps with a value (e.g., 'inline'), older SDKs only support it as
-    // a boolean flag. Passing a value to older SDKs causes it to be parsed as the project root.
-    const supportsSourceMapModes = exp.sdkVersion && semver.satisfies(exp.sdkVersion, '>=55.0.0');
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/project/publish.ts
```diff
@@ -1125,3 +1122,34 @@ export async function getUpdateRolloutInfoGroupAsync(
...
+
+/**
+ * Get the command line arguments for source map generation in expo export.
+ *
+ * Uses --source-maps if provided, otherwise falls back to --dump-sourcemap.
+ * SDK 55+ supports --source-maps with a value (e.g., 'inline'), but older SDKs
+ * only support it as a boolean flag. Passing a value to older SDKs causes it
+ * to be parsed as the project root positional argument.
+ */
+export function getSourceMapExportCommandArgs({
```
brentvatne, line 1134: this is getting... complicated. in order to be able to support all of the edge cases:

- some old versions: support only `--dump-sourcemap`
- `>` old version & `<` sdk 55: `--source-maps` with no params is supported, anything else is considered projectDir
- `>=` sdk 55: `--source-maps` is supported, along with `--source-maps internal`

PR #3342, opened 2026-01-26
[chore] Switch to publishing with OIDC

sjchmiela:
# Why

https://exponent-internal.slack.com/archives/C017N0N99RA/p1769306804484099

# How

Applied a few of the changes we've done to `eas-build`:
- https://github.com/expo/eas-build/commit/784e37c96300b20e3f812083443f4334a77fc17e
- https://github.com/expo/eas-build/commit/ae9dbea0f994d2c43fdf96a885778168a60fa4f7
- https://github.com/expo/eas-build/commit/ec824f1deddef6b5a1459329a3f9273bf9d48832
- https://github.com/expo/eas-build/commit/54ff4e9017beed4e2f98b0794e47612f8f7f6d19

# Test Plan

Merge and try releasing a new version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,14 +4,13 @@ on:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
 
 name: Release workflow
-permissions:
-  contents: write
-  id-token: write
 
 jobs:
   release:
     name: Create a release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     outputs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -165,31 +175,25 @@ jobs:
         run: yarn build
       - name: Publish packages to npm
         run: yarn lerna publish from-package --yes --no-verify-access
-        env:
```
kadikraman, line 168: I think we need to set the provenance env var here? https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools

As in:

```
        env:
          NPM_CONFIG_PROVENANCE: true
```

sjchmiela: I didn't need to do it for `eas-build` so not sure…

kadikraman: I used it for the styleguide but I never tried publishing without so I don't know for sure! Maybe lerna handles it under the hood. Can try without and add it if publish fails 🤷‍♀️

PR #3350, opened 2026-01-27
[worker] Set `DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS` in Xcode 26.0

sjchmiela:
# Why

> Apple's altool in Xcode 26.0 has a bug where it incorrectly selects the wrong app when multiple apps share a common bundle ID prefix (e.g., `com.example`, `com.example.app1`, `com.example.app2`). This causes builds to be uploaded to the wrong app on App Store Connect.

> See: https://github.com/fastlane/fastlane/issues/29698

https://github.com/expo/universe/pull/23310 + https://exponent-internal.slack.com/archives/C06EFBQK3B7/p1765365928911459

# How

If `xcodebuild -version` matches "Xcode 26.0", we'll add `DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS='--use-old-altool'`. This helps if the code does not re-override `$DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS` further, in which case it should do `DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS="overrides $DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS"` or sth like that.

I hope this fixes Launch.

# Test Plan

Tested manually on my computer

```
> spawnSync("xcodebuild", ["-version"]).stdout.includes("Xcode")
true

> spawnSync("xcodebuild", ["-version"]).stdout.includes("Xcode 26.0")
false

(below did not throw)
> spawnSync("abc")
{
  error: Error: spawnSync abc ENOENT
      at Object.spawnSync (node:internal/child_process:1123:20)
      at spawnSync (node:child_process:877:24)
      at REPL12:1:1
      at ContextifyScript.runInThisContext (node:vm:137:12)
      at REPLServer.defaultEval (node:repl:598:22)
      at bound (node:domain:433:15)
      at REPLServer.runBound [as eval] (node:domain:444:12)
      at REPLServer.onLine (node:repl:927:10)
      at REPLServer.emit (node:events:536:35)
      at REPLServer.emit (node:domain:489:12) {
    errno: -2,
    code: 'ENOENT',
    syscall: 'spawnSync abc',
    path: 'abc',
    spawnargs: []
  },
  status: null,
  signal: null,
  output: null,
  pid: 0,
  stdout: null,
  stderr: null
}
```

### DIFF
--- a/packages/worker/src/env.ts
+++ b/packages/worker/src/env.ts
@@ -133,7 +133,7 @@ export function getBuildEnv({
   }
 
   const xcodeVersionResult = spawnSync('xcodebuild', ['-version'], { env });
-  if (xcodeVersionResult.stdout.includes('Xcode 26.0')) {
+  if (xcodeVersionResult.stdout?.includes('Xcode 26.0')) {
     setEnv(env, 'DELIVER_ALTOOL_ADDITIONAL_UPLOAD_PARAMETERS', '--use-old-altool');
   }
 


packages/worker/src/env.ts
```diff
@@ -132,6 +132,11 @@ export function getBuildEnv({
     setEnv(env, 'EXPO_STAGING', '1');
   }
 
+  const xcodeVersionResult = spawnSync('xcodebuild', ['-version'], { env });
+  if (xcodeVersionResult.stdout.includes('Xcode 26.0')) {
```
douglowder:
```suggestion
  if (xcodeVersionResult.stdout.includes('Xcode 26')) {
```

douglowder: No guarantee that Apple will fix this immediately with the next xcodebuild version, so maybe keep this change past Xcode 26.0

sjchmiela: https://github.com/fastlane/fastlane/issues/29698#issuecomment-3484991065 looks like fixed in 26.1

PR #3376, opened 2026-02-04
[eas-cli] Add go command

tchayen:
# Why

We want to give users an option to create custom Expo Go builds in CLI.

# How

Using a mix of existing commands.

Note: the release build of Expo Go for SDK 55 is hardcoded right now, together with version numbers.

# Test Plan

Ran locally several times. Expected behavior: user is authenticated with their apple developer account, repack & TestFlight jobs run, after processing Apple sends a TestFlight invite email.

### DIFF
--- a/packages/eas-cli/src/commands/go.ts
+++ b/packages/eas-cli/src/commands/go.ts
@@ -17,7 +17,11 @@ import { SetUpBuildCredentials } from '../credentials/ios/actions/SetUpBuildCred
 import { SetUpPushKey } from '../credentials/ios/actions/SetUpPushKey';
 import { ensureAppExistsAsync } from '../credentials/ios/appstore/ensureAppExists';
 import { Target } from '../credentials/ios/types';
-import { WorkflowJobStatus, WorkflowProjectSourceType, WorkflowRunStatus } from '../graphql/generated';
+import {
+  WorkflowJobStatus,
+  WorkflowProjectSourceType,
+  WorkflowRunStatus,
+} from '../graphql/generated';
 import { AppMutation } from '../graphql/mutations/AppMutation';
 import { WorkflowRunMutation } from '../graphql/mutations/WorkflowRunMutation';
 import { AppQuery } from '../graphql/queries/AppQuery';
@@ -141,7 +145,6 @@ async function withSuppressedOutputAsync<T>(fn: () => Promise<T>): Promise<T> {
   }
 }
 /* eslint-enable no-console */
-
 
 export default class Go extends EasCommand {
   static override description = 'Create a custom Expo Go and submit to TestFlight';
@@ -524,8 +527,9 @@ export default class Go extends EasCommand {
         return { projectArchiveBucketKey, easJsonBucketKey, packageJsonBucketKey };
       });
 
-    const { id: workflowRunId } =
-      await WorkflowRunMutation.createExpoGoRepackWorkflowRunAsync(graphqlClient, {
+    const { id: workflowRunId } = await WorkflowRunMutation.createExpoGoRepackWorkflowRunAsync(
+      graphqlClient,
+      {
         appId: projectId,
         projectSource: {
           type: WorkflowProjectSourceType.Gcs,
@@ -534,7 +538,8 @@ export default class Go extends EasCommand {
           packageJsonBucketKey,
           projectRootDirectory: '.',
         },
-      });
+      }
+    );
 
     const app = await AppQuery.byIdAsync(graphqlClient, projectId);
     const workflowUrl = `https://expo.dev/accounts/${account.name}/projects/${app.slug}/workflows/${workflowRunId}`;
--- a/packages/eas-cli/src/graphql/mutations/WorkflowRunMutation.ts
+++ b/packages/eas-cli/src/graphql/mutations/WorkflowRunMutation.ts
@@ -122,10 +122,7 @@ export namespace WorkflowRunMutation {
         >(
           /* eslint-disable graphql/template-strings */
           gql`
-            mutation CreateExpoGoRepackWorkflowRun(
-              $appId: ID!
-              $projectSource: WorkflowProjectSourceInput!
-            ) {
+            mutation CreateExpoGoRepackWorkflowRun($appId: ID!, $projectSource: WorkflowProjectSourceInput!) {
               workflowRun {
                 createExpoGoRepackWorkflowRun(appId: $appId, projectSource: $projectSource) {
                   id


packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+        query: {
+          filter: { app: ascApp.id },
+          sort: '-uploadedDate',
+          limit: 1,
+        },
+      });
+
+      if (builds.length > 0) {
+        const build = builds[0];
+        const { processingState, usesNonExemptEncryption } = build.attributes;
```
sjchmiela: I think we can set this in `app.json` and have the build come with this setting set? Then we don't need to do all this?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    console.warn = originalConsoleWarn;
+  }
+}
+/* eslint-enable no-console */
+
+const WORKFLOW_TEMPLATE = `name: Repack Expo Go
+
+jobs:
+  repack:
+    type: build
```
sjchmiela: let's add a name here?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+          mv expo-go-modified.ipa expo-go.ipa
+          set-output ipa_path "$PWD/expo-go.ipa"
+      - uses: eas/repack
+        id: repack
+        with:
+          source_app_path: "\${{ steps.download.outputs.ipa_path }}"
+          platform: ios
+          embed_bundle_assets: false
+          repack_package: "@kudo-chien/repack-app"
+          repack_version: "0.3.0"
```
```suggestion
          repack_version: "0.3.1"
```
as per https://exponent-internal.slack.com/archives/C0ABNHJRWUX/p1770216191688539?thread_ts=1770151373.423329&cid=C0ABNHJRWUX

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+      - uses: eas/upload_artifact
+        with:
+          path: "\${{ steps.repack.outputs.output_path }}"
+
+  testflight:
+    type: testflight
```
sjchmiela: @hSATAC if we're not submitting to any external groups do we need to use `type: testflight`? Maybe we can use `type: submit` if it's only for internal groups?

Moreover, if we guaranteed this workflow uses ASC API Key, we could make it use the submission-from-Linux and thus would **not** use macOS fleet so extensively.

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+  static override description = 'Create a custom Expo Go and submit to TestFlight';
+
+  static override flags = {
+    'bundle-id': Flags.string({
+      description: 'iOS bundle identifier (auto-generated if not provided)',
+      required: false,
+    }),
+    name: Flags.string({
+      description: 'App name',
+      default: 'Custom Expo Go',
```
```suggestion
      default: 'My Expo Go',
```
I liked "My"

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    }),
+  };
+
+  static override contextDefinition = {
+    ...this.ContextOptions.LoggedIn,
+    ...this.ContextOptions.Analytics,
+  };
+
+  async runAsync(): Promise<void> {
+    Log.log(chalk.bold('Creating custom Expo Go for TestFlight...\n'));
```
```suggestion
    Log.log(chalk.bold('Creating your personal Expo Go...\n'));
```
or maybe too personal

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    const { flags } = await this.parse(Go);
+
+    let spinner = ora('Logging in to Expo...').start();
+    const {
+      loggedIn: { actor, graphqlClient },
+      analytics,
+    } = await this.getContextAsync(Go, {
+      nonInteractive: false,
+    });
+    spinner.succeed(`Logged in as ${chalk.cyan(actor.accounts[0].name)}`);
```
sjchmiela: In www I know we use "primary account" concept. Should we use it here too?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+
+    const bundleId = flags['bundle-id'] ?? this.generateBundleId(actor);
+    const appName = flags.name ?? 'Custom Expo Go';
```
```suggestion
    const appName = flags.name ?? 'My Expo Go';
```

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    const slug = bundleId.split('.').pop() || 'custom-expo-go';
```
```suggestion
    const slug = bundleId.split('.').pop() || 'my-expo-go';
```

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    } catch (error) {
+      spinner?.fail();
+      Log.gray(`Project files preserved for debugging: ${projectDir}`);
+      throw error;
+    } finally {
+      process.chdir(originalCwd);
+    }
+  }
+
+  private generateBundleId(actor: Actor): string {
```
sjchmiela, line 280: If a user runs this command two times, do we expect to create two ASC apps for him? I don't think that's good behavior. Shouldn't we use a single: Expo project, ASC App per SDK?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    // Generate 6 random bytes and encode as base64, then make it bundle-ID safe
+    const randomBytes = crypto
+      .randomBytes(6)
+      .toString('base64')
+      .replace(/\+/g, '')
+      .replace(/\//g, '')
+      .replace(/=/g, '')
+      .substring(0, 6)
+      .toLowerCase();
+    return `com.${username}.${randomBytes}`;
```
```suggestion
    return `dev.expo.go.${account_id}.${sdkVersion}`;
```
?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    const extensionBundleId = `${bundleId}.ExpoNotificationServiceExtension`;
+
+    const appJson = {
+      expo: {
+        name: appName,
+        slug,
+        version: '55.0.11',
+        ios: {
+          bundleIdentifier: bundleId,
+          buildNumber: '1017799',
```
sjchmiela: Hmmm I wonder if it makes sense to use real values? If we use "auto increment" maybe we can do 1.0.0 (1)?

If we stick to conforming to Expo Go releases, let's put it close to the URL so when we update the URL we don't forget to change this config too

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    await fs.writeFile(path.join(workflowDir, 'repack.yml'), WORKFLOW_TEMPLATE);
+
+    await spawnAsync('yarn', ['install'], { cwd: projectDir });
+  }
+
+  private async initGitRepoAsync(projectDir: string): Promise<void> {
+    await spawnAsync('git', ['init'], { cwd: projectDir });
+    await spawnAsync('git', ['add', '.'], { cwd: projectDir });
+    await spawnAsync('git', ['commit', '-m', 'Initial commit'], { cwd: projectDir });
+  }
```
sjchmiela, line 366: Is Git necessary for this?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+
+  private async ensureEasProjectAsync(
+    graphqlClient: ExpoGraphqlClient,
+    actor: Actor,
+    projectDir: string,
+    bundleId: string
+  ): Promise<string> {
+    const slug = bundleId.split('.').pop() || 'custom-expo-go';
```
sjchmiela, line 374:
```suggestion
    const slug = bundleId.split('.').pop() || 'my-expo-go';
```

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+
+    await fs.writeJson(path.join(projectDir, 'app.json'), appJson, { spaces: 2 });
+    await fs.writeJson(path.join(projectDir, 'eas.json'), easJson, { spaces: 2 });
+    await fs.writeJson(path.join(projectDir, 'package.json'), packageJson, { spaces: 2 });
+
+    const workflowDir = path.join(projectDir, '.eas', 'workflows');
+    await fs.ensureDir(workflowDir);
+    await fs.writeFile(path.join(workflowDir, 'repack.yml'), WORKFLOW_TEMPLATE);
+
+    await spawnAsync('yarn', ['install'], { cwd: projectDir });
```
sjchmiela: I think this assumes users are required to have `yarn`? Not sure we want that?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+          graphqlClient,
+          accountId: account.id,
+          filePath: packageJsonPath,
+          maxSizeBytes: 1024 * 1024,
+        });
+
+        return { projectArchiveBucketKey, easJsonBucketKey, packageJsonBucketKey };
+      });
+
+    const { id: workflowRunId } = await WorkflowRunMutation.createWorkflowRunAsync(graphqlClient, {
```
sjchmiela: If we use the regular workflow run flow, free users will end up in the free tier queue and paid users will pay for Expo Go build.

I think in v0.1 we may want to have a dedicated mutation for all this (with static workflow YAML managed in `www`, priority = high, waivedAt = NOW()).

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+      {
+        targetName: 'ExpoNotificationServiceExtension',
+        bundleIdentifier: extensionBundleId,
+        parentBundleIdentifier: bundleId,
+        entitlements: {},
+      },
+    ];
+
+    if (!customizeCreds) {
+      prompts.inject(Array(20).fill(true));
```
sjchmiela: wat

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+    const credentialsCtx = new CredentialsContext({
+      projectInfo: { exp, projectId },
+      nonInteractive: false,
+      projectDir,
+      user: actor,
+      graphqlClient,
+      analytics,
+      vcsClient,
+      easJsonCliConfig: {
+        promptToConfigurePushNotifications: false,
```
sjchmiela: Should we prompt for push notifications setup? Do we want this Expo Go to have notifications? Maybe that's not the goal here?

packages/eas-cli/src/commands/go.ts
```diff
@@ -0,0 +1,631 @@
...
+      cli: {
+        version: '>= 5.0.0',
+      },
+      build: {
+        production: {
+          distribution: 'store',
+          credentialsSource: 'remote',
+          ios: {
+            resourceClass: 'm-medium',
+          },
```
```suggestion
```
legacy value lol

sjchmiela (changes requested):

packages/eas-cli/src/commands/go.ts
tchayen: Claude's invention how to use defaults for credentials 🙈 

I will see if I can customize it better

packages/eas-cli/src/commands/go.ts
tchayen: Trying `type: submit` now

packages/eas-cli/src/commands/go.ts
tchayen: Switching to this. There's still the `--bundle-id` param I can use for debugging more builds

Now I will need to add detection if the app already exists

packages/eas-cli/src/commands/go.ts
tchayen: This was to avoid warnings. I moved it closer to top of the file for now

packages/eas-cli/src/commands/go.ts
tchayen: I think it was in my experiments - workflow wouldn't start

packages/eas-cli/src/commands/go.ts
tchayen: Good point. I will see how to use something better (the problem was that without lockfile the workflow wasn't working)

packages/eas-cli/src/commands/go.ts
tchayen: Ideally we would preserve all behavior but if there's any issue with push notifications we are fine dropping them I think

packages/eas-cli/src/commands/go.ts
tchayen: I updated the script to use app.json

```
config: {
  usesNonExemptEncryption: false,
}
```

as in https://docs.expo.dev/versions/latest/config/app/#usesnonexemptencryption

but it doesn't seem to work - apple is again showing me this

Screenshot 2026-02-04 at 21 31 24

while it worked with the previous hacky approach - maybe something about this being repack?

packages/eas-cli/src/commands/go.ts
tchayen: Added names to jobs, hope this is what you meant

packages/eas-cli/src/commands/go.ts
tchayen: Worked

packages/eas-cli/src/commands/go.ts
tchayen: I am not sure. We are probably fine with notifications missing if that would block us otherwise

brentvatne: i'm fine with skipping notifications. although if we already added support for the notification extension when repacking, then maybe it's worth prompting for this

packages/eas-cli/src/commands/go.ts
tchayen: Done! (hopefully)

brentvatne (approved):

PR #3380, opened 2026-02-05
[chore] Switch from Yarn to Bun

sjchmiela:
I fear without https://github.com/lerna/lerna/pull/4264 we're blocked on releasing…


JamesHenry:
@sjchmiela Just happened to see this, I'm a big fan of Expo. Would you accept a small PR to shift you from lerna to nx (that lerna is already using behind the scenes for task runner, graph analysis etc)? It already has battled tested bun support.

Lerna is very much still an important project, but bigger efforts like adding bun to it are just a little slower to get prioritized

sjchmiela:
Oo, thanks for reminding me to come back to this pull request!

I started this before I had upgraded the repo from Yarn v1 to Yarn v4, which also aligns the package manager with the rest of the mainstream Expo workspaces.

As fun as I think migrating to Bun and Nx (which I know a few customers use with Expo with success!) would be, I think Yarn v4 solved most of my problems. I've also considered moving to Bun for a minute too to maybe bundle `eas-cli` or distribute as an executable, but that's not a goal for now.

Thanks for proactively jumping in here, I appreciate that! If we ever come back to this tooling decision I'll keep Nx in mind!

PR #3361, opened 2026-01-30
[infra] Stop preparing tarballs

sjchmiela:
# Why

Our GitHub released executable Oclif bundles of `eas-cli` are… not too popular. v16.28.0 was downloaded 154k times in the last 7 days from NPM. Since its release (2 months ago) it was downloaded from GitHub 82 times.

Zrzut ekranu 2026-01-30 o 15 15 35

# How

Asked Gemini to remove this. This should speed up release process significantly. This release process has been created 5 years ago (https://github.com/expo/eas-cli/pull/54).

Also removed size comparison workflow. It's not going to make comments like https://github.com/expo/eas-cli/pull/3361#issuecomment-3824036144.

Closes https://github.com/expo/eas-cli/pull/3346.

# Test Plan

CI should pass.

PR #3381, opened 2026-02-05
[chore] Fix `pretarball-ci` script

sjchmiela:
# Why

When oclif is building tarballs with executables it does so in a directory separate from the monorepo which requires us to change `eas-cli` dependencies from the monorepo to `file`.

# How

Asked Gemini to update the script to include all local dependencies of `eas-cli`.

Or… we can land https://github.com/expo/eas-cli/pull/3361 …

# Test Plan

Release should pass. Validated locally. (Hmm but on CI it failed? Will see.)

Ok, so size failed because it ran on `main` which is broken because it lacks this pull request.

Zrzut ekranu 2026-02-5 o 19 05 41

### DIFF
--- a/packages/eas-cli/scripts/pretarball-ci.sh
+++ b/packages/eas-cli/scripts/pretarball-ci.sh
@@ -24,7 +24,7 @@ rewrite_deps() {
     KEY="${package%%:*}"
     DIR="${package##*:}"
 
-    # Determine the relative path
+    # Determine the relative path. If IS_EAS_CLI we go one level deeper because oclif builds in `tmp` in current directory.
     if [[ "$IS_EAS_CLI" == "true" ]]; then
       REL_PATH="file:../../../$DIR"
     else


packages/eas-cli/scripts/pretarball-ci.sh
```diff
@@ -5,8 +5,57 @@ set -eo pipefail
...
+# Function to rewrite dependencies in a package.json file
+rewrite_deps() {
+  local PKG_JSON=$1
+  local IS_EAS_CLI=$2
+
+  for package in "${PACKAGES[@]}" ; do
+    KEY="${package%%:*}"
+    DIR="${package##*:}"
+
+    # Determine the relative path
```
sjchmiela:
```suggestion
    # Determine the relative path. If IS_EAS_CLI we go one level deeper because oclif builds in `tmp` in current directory.
```

PR #3384, opened 2026-02-05
[chore] Move to Oxlint

sjchmiela:
# Why

We're moving to Oxc in universe too. The toolset is faster than ESLint / Prettier.

# How

Asked Codex to migrate the repository. I don't think all rules got migrated, unfortunately, but also the lint are in a bit weird state (different rules in `eas-cli` vs the rest of the repo). I hope we can unify the rules soon.

# Test Plan

CI should pass.


sjchmiela:
Oh my, sorry! I was working on this towards the end of one day and submitted for review on the next and didn't realize the PR description is empty. Filling in now.

As for quotes — do you mean quotes in GHA? I think those are the only quotes being changed. I'll try changing the setting for YAML to `"` and see if it reduces changes!

PR #3386, opened 2026-02-06
[worker] - log unknown build error logs in Datadog

AbbanMustafa:
# Why

Record Unknown Errors with Datadog logs so that we can index by build ID, track patterns, and determing regressions/fixes. More detailed here https://github.com/expo/universe/pull/24771

# How

Use Datadogs log ingestion API  https://docs.datadoghq.com/api/latest/logs/#send-logs

### DIFF
--- a/packages/worker/src/service.ts
+++ b/packages/worker/src/service.ts
@@ -317,9 +317,6 @@ export default class BuildService {
       const err = error instanceof errors.BuildError ? error : unknownError;
       const maybeRawError = error instanceof errors.BuildError ? error.innerError : error;
 
-      const rawErrorMessage =
-        maybeRawError instanceof Error ? maybeRawError.message : 'An unknown error occurred';
-
       sentry.handleError(err.message, maybeRawError, {
         tags: {
           ...(err.buildPhase ? { buildPhase: err.buildPhase } : {}),
@@ -334,9 +331,20 @@ export default class BuildService {
       });
 
       if (err.errorCode === errors.ErrorCode.UNKNOWN_ERROR) {
+        let rawErrorMessage: string;
+        if (maybeRawError instanceof Error) {
+          rawErrorMessage = maybeRawError.message;
+        } else {
+          try {
+            rawErrorMessage = JSON.stringify(maybeRawError);
+          } catch {
+            rawErrorMessage = String(maybeRawError);
+          }
+        }
+
         const robotAccessToken = job.secrets?.robotAccessToken;
         if (robotAccessToken) {
-          turtleFetch(
+          await turtleFetch(
             new URL('turtle-builds/error-logs', config.wwwApiV2BaseUrl).toString(),
             'POST',
             {
@@ -351,7 +359,7 @@ export default class BuildService {
               },
               shouldThrowOnNotOk: false,
             }
-          ).catch(() => {});
+          );
         }
       }
 


packages/worker/src/service.ts
```diff
@@ -328,6 +329,17 @@ export default class BuildService {
           ...(maybeRawError.stderr ? { stderr: getLastNLines(100, maybeRawError.stderr) } : {}),
         },
       });
+
+      if (err.errorCode === errors.ErrorCode.UNKNOWN_ERROR) {
+        const rawMessage =
+          maybeRawError instanceof Error ? maybeRawError.message : 'An unknown error occurred';
+        void datadogLogs.send({
+          message: `Unknown build error: ${rawMessage}`,
```
sjchmiela: is "unknown build error:" helpful here?

packages/worker/src/service.ts
```diff
@@ -328,6 +329,17 @@ export default class BuildService {
...
+          level: 'error',
+          tags: { build_id: this.buildId },
```
sjchmiela: Is the cardinality on this ok? I've heard high cardinality is where we would pay big bucks

packages/worker/src/service.ts
```diff
@@ -328,6 +329,17 @@ export default class BuildService {
           ...(maybeRawError.stderr ? { stderr: getLastNLines(100, maybeRawError.stderr) } : {}),
         },
       });
+
+      if (err.errorCode === errors.ErrorCode.UNKNOWN_ERROR) {
+        const rawMessage =
+          maybeRawError instanceof Error ? maybeRawError.message : 'An unknown error occurred';
+        void datadogLogs.send({
```
sjchmiela:
```suggestion
        datadogLogs.send({
```
?

packages/worker/src/external/datadogLogs.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (!this.apiKey) {
+      return;
+    }
+
+    try {
+      await fetch(this.endpoint, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'DD-API-KEY': this.apiKey,
```
sjchmiela: If worker has access to DataDog API key we could as well set it here — users are going to have access to it.

For step metrics we decided to relay these through www. Why can we not do it for this?

packages/worker/src/external/datadogLogs.ts
```diff
@@ -0,0 +1,57 @@
+import { bunyan } from '@expo/logger';
+
+interface DatadogLogsOptions {
+  apiKey: string | null;
+  site: string;
+  service: string;
+  logger: bunyan;
+}
```
sjchmiela: Why does AI like to extract these interfaces so much instead of inlining in the constructor…

packages/worker/src/service.ts
sjchmiela: Also note technically we might finish the build before we get to send the request. Might be ok but worth mentioning.

packages/worker/src/external/datadogLogs.ts
```diff
@@ -0,0 +1,57 @@
...
+            ddtags: Object.entries(log.tags)
+              .map(([k, v]) => `${k}:${v}`)
+              .join(','),
+            status: log.level,
+            service: this.service,
+          },
+        ]),
+      });
+    } catch (error) {
+      this.logger.error(error, 'Failed to send log to Datadog');
```
sjchmiela:
```suggestion
      this.logger.error({ err: error }, 'Failed to send log to Datadog');
```
This is bunyan nomenclature

sjchmiela (changes requested):

packages/worker/src/service.ts
AbbanMustafa: That doesnt apply to logs fields, that pricing is for custom metrics.

>  Adding a high-cardinality tag like a Build ID to your logs does not directly increase your per-log "Log Management" price, but it can trigger massive costs elsewhere in your Datadog bill. 
> Here is how that high-variable tag affects different parts of your bill:
> 1. Log Management (Ingestion & Indexing)
> Direct Cost: There is no extra charge for adding more tags or high-cardinality attributes to your logs.
> The "Weight" Impact: Adding a long tag string increases the size (bytes) of each log. Since you are billed for ingestion by the gigabyte (~$0.10/GB), a very large number of long tags could slightly increase your ingestion costs over time. 
> 2. The Trap: "Log-Based Metrics"
> The real danger is if you use that tag in a [Log-Based Metric](https://docs.datadoghq.com/logs/log_configuration/indexes/).
> The Cost: Datadog charges for custom metrics based on the number of unique tag combinations (cardinality).
> The Result: If you create a metric from your logs and include the build_id tag, every single build will create a new "unique time series." This can quickly exceed your [custom metric allotment](https://docs.datadoghq.com/account_management/billing/custom_metrics/) and lead to thousands of dollars in overage fees ($0.05 per metric/month).

packages/worker/src/service.ts
```diff
@@ -316,6 +317,9 @@ export default class BuildService {
       const err = error instanceof errors.BuildError ? error : unknownError;
       const maybeRawError = error instanceof errors.BuildError ? error.innerError : error;
 
+      const rawErrorMessage =
+        maybeRawError instanceof Error ? maybeRawError.message : 'An unknown error occurred';
```
sjchmiela: Should we `JSON.stringify(maybeRawError)` if it's not an `Error`?

packages/worker/src/service.ts
```diff
@@ -316,6 +317,9 @@ export default class BuildService {
       const err = error instanceof errors.BuildError ? error : unknownError;
       const maybeRawError = error instanceof errors.BuildError ? error.innerError : error;
 
+      const rawErrorMessage =
```
sjchmiela: nit: move to inside `if (== UNKNOWN)`

packages/worker/src/service.ts
```diff
@@ -328,6 +332,29 @@ export default class BuildService {
...
+                message: rawErrorMessage,
+                buildPhase: err.buildPhase ?? null,
+                errorCode: err.errorCode,
+              },
+              headers: {
+                Authorization: `Bearer ${robotAccessToken}`,
+              },
+              shouldThrowOnNotOk: false,
+            }
+          ).catch(() => {});
```
sjchmiela: What would happen if we didn't catch it? Let's not catch if not necessary.

PR #3387, opened 2026-02-07
Fix non-interactive mode prompting for git initialization

EvanBacon:
## Summary

- Fixes `eas build --non-interactive` prompting interactively when the project doesn't have git initialized
- In non-interactive mode, automatically runs `git init` and creates an initial commit
- Shows informative message: `No git repository found. Auto initializing in /path. Set EAS_NO_VCS=1 to skip.`
- Improved error message when git user.name/email not configured to show exact commands to run

## Test plan

- Run `eas build --non-interactive` in a project without git initialized
- Verify it auto-initializes git instead of prompting
- Verify `EAS_NO_VCS=1 eas build --non-interactive` skips git initialization

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This is the log of notable changes to EAS CLI and related packages.
 
 - Hide progress bar in build credits warning when usage reaches 100%. ([#3371](https://github.com/expo/eas-cli/pull/3371) by [@mackenco](https://github.com/mackenco))
 
+### 🐛 Bug fixes
+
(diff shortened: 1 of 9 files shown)


douglowder (changes requested):
In non-interactive mode, we should definitely not prompt. Thanks for fixing that! However, we should not execute a git operation under the hood. I'd prefer that we throw an error stating that the command requires a git repo.

EvanBacon:
@douglowder my thinking here is that git is pretty easy to delete and users can opt-out of this behavior by: a) having a git repo, or b) setting `EAS_NO_VCS=1` which skips all git checks. I imagine we can pull this back a bit by skipping the auto git setup for commands that don't need git to operate (but I'm not sure which those are).

PR #3388, opened 2026-02-09
[build-tools] Add eas/report_maestro_test_results build function

hSATAC:
# Why

Maestro test jobs produce per-flow results on the build worker (pass/fail, duration, error messages, retries), but this data is lost after the job finishes.

We want to persist these as `WorkflowDeviceTestCaseResult` records to surface test outcomes in the dashboard.

# How

Add a new `eas/report_maestro_test_results` build function that:
  1. Parses JUnit XML reports from the `junit_report_directory` (primary data source)
  2. Enriches with flow file paths from Maestro's `ai-*.json` debug output in `tests_directory`
  3. Calls the `createWorkflowDeviceTestCaseResults` GraphQL mutation via `gql.tada` + `graphqlClient`

Key details:
  - JUnit XML as primary data source — `ai-*.json` only used for `flow_file_path` mapping and retry count
  - `junit_report_directory` added as output on `eas/__maestro_test` (set when `output_format=junit`) and as optional input on `eas/report_maestro_test_results`
  - Retry detection by counting occurrences of the same flow across timestamp directories (works for both `reuse_devices: true` and `false` patterns)
  - Duplicate testcase name detection — skips report when Maestro config-level name collisions make flow-to-file mapping ambiguous
  - Path relativization uses `fs.realpath` to handle symlinks (`/tmp` → `/private/tmp` on macOS)
  - Best-effort reporting — errors are logged but never fail the build step
  - `graphqlClient` exposed on `CustomBuildContext` for step functions to use
  - 39 unit tests covering JUnit parsing, flow metadata, retry detection, properties/tags, duplicate detection, and error resilience

# Test Plan

- Automation tests
- Manual test locally with https://github.com/expo/universe/pull/24814

### DIFF
--- a/packages/build-tools/src/customBuildContext.ts
+++ b/packages/build-tools/src/customBuildContext.ts
@@ -10,6 +10,7 @@ import {
 } from '@expo/eas-build-job';
 import { bunyan } from '@expo/logger';
 import { BuildRuntimePlatform, ExternalBuildContextProvider } from '@expo/steps';
+import { Client } from '@urql/core';
 import assert from 'assert';
 import path from 'path';
 
@@ -53,6 +54,7 @@ export class CustomBuildContext<TJob extends Job = Job> implements ExternalBuild
   public readonly startTime: Date;
 
   public readonly logger: bunyan;
+  public readonly graphqlClient: Client;
   public readonly runtimeApi: BuilderRuntimeApi;
   public job: TJob;
   public metadata?: Metadata;
@@ -65,6 +67,7 @@ export class CustomBuildContext<TJob extends Job = Job> implements ExternalBuild
     this.metadata = buildCtx.metadata;
 
     this.logger = buildCtx.logger.child({ phase: BuildPhase.CUSTOM });
+    this.graphqlClient = buildCtx.graphqlClient;
     this.projectSourceDirectory = path.join(buildCtx.workingdir, 'temporary-custom-build');
     this.projectTargetDirectory = path.join(buildCtx.workingdir, 'build');
     this.defaultWorkingDirectory = buildCtx.getReactNativeProjectDirectory();
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -1,0 +1,264 @@
+import { XMLParser } from 'fast-xml-parser';
+import fs from 'fs/promises';
+import path from 'path';
+import { z } from 'zod';
+
+export interface MaestroFlowResult {
+  name: string;
+  path: string;
+  status: 'passed' | 'failed';
+  errorMessage: string | null;
+  duration: number; // milliseconds
+  retryCount: number;
+  tags: string[];
+  properties: Record<string, string>;
+}
+
+// Maestro's TestDebugReporter creates timestamped directories, e.g. "2024-06-15_143022"
+const TIMESTAMP_DIR_PATTERN = /^\d{4}-\d{2}-\d{2}_\d{6}$/;
+
+export function extractFlowKey(filename: string, prefix: string): string | null {
+  const match = filename.match(new RegExp(`^${prefix}-(.+)\\.json$`));
+  return match?.[1] ?? null;
+}
+
+export interface JUnitTestCaseResult {
+  name: string;
+  status: 'passed' | 'failed';
+  duration: number; // milliseconds
+  errorMessage: string | null;
+  tags: string[];
+  properties: Record<string, string>;
+}
+
+const xmlParser = new XMLParser({
+  ignoreAttributes: false,
+  attributeNamePrefix: '@_',
+  // Ensure single-element arrays are always arrays
+  isArray: name => ['testsuite', 'testcase', 'property'].includes(name),
+});
+
+export async function parseJUnitTestCases(junitDirectory: string): Promise<JUnitTestCaseResult[]> {
+  let entries: string[];
+  try {
+    entries = await fs.readdir(junitDirectory);
+  } catch {
+    return [];
+  }
+
+  const xmlFiles = entries.filter(f => f.endsWith('.xml'));
+  if (xmlFiles.length === 0) {
+    return [];
+  }
+
+  const results: JUnitTestCaseResult[] = [];
+
+  for (const xmlFile of xmlFiles) {
+    try {
+      const content = await fs.readFile(path.join(junitDirectory, xmlFile), 'utf-8');
+      const parsed = xmlParser.parse(content);
+
+      const testsuites = parsed?.testsuites?.testsuite;
+      if (!Array.isArray(testsuites)) {
+        continue;
+      }
+
+      for (const suite of testsuites) {
+        const testcases = suite?.testcase;
+        if (!Array.isArray(testcases)) {
+          continue;
+        }
+
+        for (const tc of testcases) {
+          const name = tc['@_name'];
+          if (!name) {
+            continue;
+          }
+
+          const timeStr = tc['@_time'];
+          const timeSeconds = timeStr ? parseFloat(timeStr) : 0;
+          const duration = Number.isFinite(timeSeconds) ? Math.round(timeSeconds * 1000) : 0;
+
+          // Use @_status as primary indicator (more robust than checking <failure> presence)
+          const status: 'passed' | 'failed' = tc['@_status'] === 'SUCCESS' ? 'passed' : 'failed';
+          // Extract error message from <failure> or <error> elements
+          const failureText =
+            tc.failure != null
+              ? typeof tc.failure === 'string'
+                ? tc.failure
+                : (tc.failure?.['#text'] ?? null)
+              : null;
+          const errorText =
+            tc.error != null
+              ? typeof tc.error === 'string'
+                ? tc.error
+                : (tc.error?.['#text'] ?? null)
+              : null;
+          const errorMessage: string | null = failureText ?? errorText ?? null;
+
+          // Extract properties
+          const rawProperties: { '@_name': string; '@_value': string }[] =
+            tc.properties?.property ?? [];
+          const properties: Record<string, string> = {};
+
+          for (const prop of rawProperties) {
+            const propName = prop['@_name'];
+            const value = prop['@_value'];
+            if (typeof propName !== 'string' || typeof value !== 'string') {
+              continue;
+            }
+            properties[propName] = value;
+          }
+
+          // Extract tags from "tags" property (Maestro 2.2.0+, comma-separated)
+          const tagsValue = properties['tags'];
+          const tags: string[] = tagsValue
+            ? tagsValue
+                .split(',')
+                .map(t => t.trim())
+                .filter(Boolean)
+            : [];
+          delete properties['tags'];
+
+          results.push({ name, status, duration, errorMessage, tags, properties });
+        }
+      }
+    } catch {
+      // Skip malformed XML files
+      continue;
+    }
+  }
+
+  return results;
+}
+
+const FlowMetadataFileSchema = z.object({
+  flow_name: z.string(),
+  flow_file_path: z.string(),
+});
+
+type FlowMetadata = z.output<typeof FlowMetadataFileSchema>;
+
+/**
+ * Parses an `ai-*.json` file produced by Maestro's TestDebugReporter.
+ *
+ * The file contains:
+ * - `flow_name`: derived from the YAML `config.name` field if present, otherwise
+ *   the flow filename without extension.
+ *   See: https://github.com/mobile-dev-inc/Maestro/blob/c0e95fd/maestro-cli/src/main/java/maestro/cli/runner/TestRunner.kt#L70
+ * - `flow_file_path`: absolute path to the original flow YAML file.
+ * - `outputs`: screenshot defect data (unused here).
+ *
+ * Filename format: `ai-(flowName).json` where `/` in flowName is replaced with `_`.
+ * See: https://github.com/mobile-dev-inc/Maestro/blob/c0e95fd/maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt#L67
+ */
+export async function parseFlowMetadata(filePath: string): Promise<FlowMetadata | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const data = JSON.parse(content);
+    return FlowMetadataFileSchema.parse(data);
+  } catch {
+    return null;
+  }
+}
+
+export async function parseMaestroResults(
+  junitDirectory: string,
+  testsDirectory: string,
+  projectRoot: string
+): Promise<MaestroFlowResult[]> {
+  // 1. Parse JUnit XML files (primary source)
+  const junitResults = await parseJUnitTestCases(junitDirectory);
+  if (junitResults.length === 0) {
+    return [];
+  }
+
+  // 2. Parse ai-*.json from debug output for flow_file_path + retryCount
+  const flowPathMap = new Map<string, string>(); // flowName → flowFilePath
+  const flowOccurrences = new Map<string, number>(); // flowName → count
+
+  let entries: string[];
+  try {
+    entries = await fs.readdir(testsDirectory);
+  } catch {
+    entries = [];
+  }
+
+  const timestampDirs = entries.filter(name => TIMESTAMP_DIR_PATTERN.test(name)).sort();
+
+  for (const dir of timestampDirs) {
+    const dirPath = path.join(testsDirectory, dir);
+    let files: string[];
+    try {
+      files = await fs.readdir(dirPath);
+    } catch {
+      continue;
+    }
+
+    for (const file of files) {
+      const flowKey = extractFlowKey(file, 'ai');
+      if (!flowKey) {
+        continue;
+      }
+
+      const metadata = await parseFlowMetadata(path.join(dirPath, file));
+      if (!metadata) {
+        continue;
+      }
+
+      // Track latest path (last timestamp dir wins)
+      flowPathMap.set(metadata.flow_name, metadata.flow_file_path);
+
+      // Count occurrences for retryCount
+      flowOccurrences.set(metadata.flow_name, (flowOccurrences.get(metadata.flow_name) ?? 0) + 1);
+    }
+  }
+
+  // 3. Merge: JUnit results + ai-*.json metadata
+  const results: MaestroFlowResult[] = [];
+
+  for (const junit of junitResults) {
+    const flowFilePath = flowPathMap.get(junit.name);
+    const relativePath = flowFilePath
+      ? await relativizePathAsync(flowFilePath, projectRoot)
+      : junit.name; // fallback: use flow name if ai-*.json not found
+
+    const occurrences = flowOccurrences.get(junit.name) ?? 0;
+    const retryCount = Math.max(0, occurrences - 1);
+
+    results.push({
+      name: junit.name,
+      path: relativePath,
+      status: junit.status,
+      errorMessage: junit.errorMessage,
+      duration: junit.duration,
+      retryCount,
+      tags: junit.tags,
+      properties: junit.properties,
+    });
+  }
+
+  return results;
+}
+
+async function relativizePathAsync(flowFilePath: string, projectRoot: string): Promise<string> {
+  if (!path.isAbsolute(flowFilePath)) {
+    return flowFilePath;
+  }
+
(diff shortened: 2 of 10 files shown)


packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      try {
+        const response = await retryOnDNSFailure(fetch)(graphqlUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${robotAccessToken}`,
+          },
+          body: JSON.stringify({
+            query: UPLOAD_MUTATION,
```
sjchmiela: Let's use https://github.com/expo/eas-cli/blob/789cf62bd6d7b0b4173ad3e790426ac6807b9731/packages/build-tools/src/common/projectSources.ts#L189, it handles errors nicer and is type-safe.

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+    return null;
+  }
+}
+
+export async function parseCommandResults(filePath: string): Promise<CommandResultSummary | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const commands: {
+      metadata: { status: string; duration?: unknown; error?: { message?: string } };
+    }[] = JSON.parse(content);
```
sjchmiela: 1. There's `fs.readFileJson` I learned not long ago, we might want to use that.
2. Let's validate content with Zod so we rely on `any` less?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+ * - `outputs`: screenshot defect data (unused here).
+ *
+ * Filename format: `ai-(flowName).json` where `/` in flowName is replaced with `_`.
+ * See: https://github.com/mobile-dev-inc/Maestro/blob/c0e95fd/maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt#L67
+ */
+export async function parseFlowMetadata(filePath: string): Promise<FlowMetadata | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const data = JSON.parse(content);
+    const suffix = extractSuffix(path.basename(filePath), 'ai');
```
sjchmiela: Can we rename "suffix" to "flowFileName" or something?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+
+export async function parseCommandResults(filePath: string): Promise<CommandResultSummary | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const commands: {
+      metadata: { status: string; duration?: unknown; error?: { message?: string } };
+    }[] = JSON.parse(content);
+
+    const failedCommand = commands.find(c => c.metadata.status !== 'COMPLETED');
+    const duration = commands.reduce((sum, c) => sum + safeDuration(c.metadata.duration), 0);
```
sjchmiela: Let's inline `safeDuration` here

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+  projectRoot: string
+): Promise<MaestroFlowResult[]> {
+  let entries: string[];
+  try {
+    entries = await fs.readdir(testsDirectory);
+  } catch {
+    return [];
+  }
+
+  const timestampDirs = entries.filter(name => TIMESTAMP_DIR_PATTERN.test(name)).sort();
```
sjchmiela, line 187: Would it be easier to, instead of validating dirs with pattern and what not, just grep `.maestro/tests` for all files like `ai-….json`, order by path and thus get all the flows run; then respective commands from `ai-….json`s or maybe from `commands-*.json`?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+ * See: https://github.com/mobile-dev-inc/Maestro/blob/c0e95fd/maestro-cli/src/main/java/maestro/cli/report/TestDebugReporter.kt#L67
+ */
+export async function parseFlowMetadata(filePath: string): Promise<FlowMetadata | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const data = JSON.parse(content);
+    const suffix = extractSuffix(path.basename(filePath), 'ai');
+    return {
+      flowName: data.flow_name ?? suffix ?? path.basename(filePath),
+      flowFilePath: data.flow_file_path ?? suffix ?? path.basename(filePath),
```
sjchmiela: Do you have an example where fallbacks make sense?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,213 @@
...
+    for (const file of files) {
+      const suffix = extractSuffix(file, 'ai');
+      if (suffix && !aiFiles.has(suffix)) {
+        aiFiles.set(suffix, path.join(dirPath, file));
+      }
+    }
+
+    for (const [suffix, aiFilePath] of aiFiles) {
+      // NOTE: Maestro supports sharded test output where commands files are prefixed
+      // with `shard-N-`, e.g. `commands-shard-1-(home).json`. This is not currently
```
sjchmiela: Would be nice if we supported shards too

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,119 @@
+import { BuildFunction, BuildStepInput, BuildStepInputValueTypeName } from '@expo/steps';
+import fetch from 'node-fetch';
+
+import { parseMaestroResults } from './maestroResultParser';
+import { retryOnDNSFailure } from '../../utils/retryOnDNSFailure';
+
+const UPLOAD_MUTATION = `
+  mutation UploadWorkflowTestCaseResults($input: UploadWorkflowTestCaseResultsInput!) {
+    workflowTestCaseResult {
+      uploadWorkflowTestCaseResults(input: $input) {
```
sjchmiela: we can call it create

and have the mutation return IDs of the results

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
hSATAC: Hmm I think this requires adding a `graphqlClient` on the `ExternalBuildContextProvider`. Currently we only have this on `BuildContext`. Let me think about this.

hSATAC:
The `Size / compare` check is failing now due the the `yarn gql` prebuild script. The graphql was not deployed yet so it can't fetch the required new gql schema.

packages/build-tools/src/customBuildContext.ts
```diff
@@ -53,6 +54,7 @@ export class CustomBuildContext<TJob extends Job = Job> implements ExternalBuild
   public readonly startTime: Date;
 
   public readonly logger: bunyan;
+  public readonly graphqlClient: Client;
```
sjchmiela, line 57: Ooo, nice! I didn't realize it was missing here.

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      try {
+        const result = await ctx.graphqlClient
+          .mutation(CREATE_MUTATION, {
+            input: {
+              workflowJobId,
+              testCaseResults: flowResults.map(f => ({
+                name: f.name,
+                path: f.path,
+                status: f.status.toUpperCase() as 'PASSED' | 'FAILED',
```
sjchmiela: I tend to prefer

```
const FlowStatusToTestCareResultStatus: Record<string, WorkflowDeviceTestCaseStatus> = { ... }
```

or maybe even `WorkflowDeviceTestCaseStatus | undefined` so we can handle missing mapping here. Man, I wish you could just report this to Sentry…

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      const flowResults = await parseMaestroResults(junitDirectory, testsDirectory, projectRoot);
+      if (flowResults.length === 0) {
+        logger.info('No maestro test results found, skipping report');
+        return;
+      }
+
+      // Maestro allows overriding flow names via config, so different flow files can share
+      // the same name. JUnit XML only contains names (not file paths), making it impossible
+      // to map duplicates back to their original flow files. Skip and let the user fix it.
```
sjchmiela, line 68: nnnnnooooooo, really??

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+      const projectRoot = inputs.project_root.value as string;
+
+      const flowResults = await parseMaestroResults(junitDirectory, testsDirectory, projectRoot);
+      if (flowResults.length === 0) {
+        logger.info('No maestro test results found, skipping report');
+        return;
+      }
+
+      // Maestro allows overriding flow names via config, so different flow files can share
+      // the same name. JUnit XML only contains names (not file paths), making it impossible
```
sjchmiela, line 67: After having parsed both commands and JUnit, what is your professional opinion on which one should we use?

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+      BuildStepInput.createProvider({
+        id: 'workflow_job_id',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'project_root',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
```
sjchmiela: I think you can get it reliably at `stepsCtx.global.projectTargetDirectory`

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+      BuildStepInput.createProvider({
+        id: 'tests_directory',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'workflow_job_id',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
```
sjchmiela: I usually just do https://github.com/expo/eas-cli/blob/789cf62bd6d7b0b4173ad3e790426ac6807b9731/packages/build-tools/src/steps/functions/createSubmissionEntity.ts#L66-L70

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,106 @@
...
+      BuildStepInput.createProvider({
+        id: 'junit_report_directory',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'tests_directory',
+        required: true,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
```
sjchmiela: Are these ever different than `$HOME/.maestro`? If you think it makes sense to customize them (for testing maybe?) we could do a default value in code sort of like https://github.com/expo/eas-cli/blob/da6703706a1f8cb536461a513997cd7714ca1a35/packages/build-tools/src/steps/functions/repack.ts#L77-L78

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
sjchmiela: Actually, maybe we want https://github.com/expo/eas-cli/blob/da6703706a1f8cb536461a513997cd7714ca1a35/packages/build-tools/src/steps/functions/repack.ts#L74?

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
hSATAC: After looking at both, I’m leaning toward using the original JSON as the source of truth.

It simply has more complete data, and we can support all formats without forcing users to adopt JUnit.

With JUnit, we’d still need JSON anyway (no tags / flowFileName), and duration is only second-level precision.

The only real advantage of JUnit is stability — it’s less likely to change upstream. JSON doesn’t give us that guarantee, so if we go with it, we might want some automation to detect schema changes early.

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
hSATAC: I think one of the goals of steps is to keep them as composable as possible.

For example, if someone runs Maestro via a custom step, our `reportMaestroTestResults` step should still be able to plug in afterwards — as long as the required params are provided.

I’m fine with defaulting both params to `$HOME/.maestro/tests`. That said, `junit_report_directory` is overridden in the `reuse_devices: false` code path — it points to a different directory there:
[https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/steps/functions/internalMaestroTest.ts#L202-L204](https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/steps/functions/internalMaestroTest.ts#L202-L204)

So I’d lean toward keeping them configurable, even if we provide sensible defaults.

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
hSATAC: Yeah, the JUnit is super simple like this:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<testsuites>
  <testsuite name="Test Suite" device="iPhone 16 - iOS 18.6 - 59572B21-FEAF-4BEC-B8EB-5793C9A31034" tests="6" failures="3" time="78.0">
    <testcase id="settings_screen" name="settings_screen" classname="settings_screen" time="22.0" status="ERROR">
      <properties>
        <property name="priority" value="high"/>
        <property name="suite" value="smoke"/>
        <property name="jira" value="APP-101"/>
      </properties>
      <failure>Element not found: Text matching regex: Settings</failure>
    </testcase>
  </testsuite>
</testsuites>
```

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,282 @@
...
+  path: string;
+  status: 'passed' | 'failed';
+  errorMessage: string | null;
+  duration: number; // milliseconds
+  retryCount: number;
+  tags: string[];
+  properties: Record<string, string>;
+}
+
+const TIMESTAMP_DIR_PATTERN = /^\d{4}-\d{2}-\d{2}_\d{6}$/;
```
sjchmiela, line 18: add a comment with example?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,282 @@
...
+import path from 'path';
+import YAML from 'yaml';
+import { z } from 'zod';
+
+export interface FlowMetadata {
+  flowName: string;
+  flowFilePath: string;
+}
+
+export interface MaestroFlowResult {
```
sjchmiela, line 6: should this be a zod schema we would validate against? same for `FlowMetadata`?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,282 @@
+import { XMLParser } from 'fast-xml-parser';
+import fs from 'fs/promises';
+import path from 'path';
+import YAML from 'yaml';
+import { z } from 'zod';
+
+export interface FlowMetadata {
```
sjchmiela: Let's use `z.output`, I think we're ok to use `sth.flow_file_path` instead of `sth.flowFilePath` just for the sake of JS consistency

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -0,0 +1,282 @@
...
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * Reads tags from a Maestro flow YAML file's config section.
+ * Flow files are structured as: config (object) + `---` + commands (array).
+ */
+export async function parseFlowTags(flowFilePath: string): Promise<string[]> {
```
sjchmiela: [Maestro 2.2.0](https://github.com/mobile-dev-inc/maestro/blob/main/CHANGELOG.md#220) supposedly added tags and properties to XML. Maybe we can rely on that? I think that might simplify code a lot?

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,119 @@
...
+    namespace: 'eas',
+    id: 'report_maestro_test_results',
+    name: 'Report Maestro Test Results',
+    __metricsId: 'eas/report_maestro_test_results',
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'junit_report_directory',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+        defaultValue: '${ eas.env.HOME }/.maestro/tests',
```
sjchmiela: Damn I don't trust interpolations in default values haha. Do they work? nice

sjchmiela: Let's not use the old `${ eas.env }` stuff. instead, let's use `${{ env }}`

packages/build-tools/src/steps/functions/reportMaestroTestResults.ts
```diff
@@ -0,0 +1,119 @@
...
+      }),
+    ],
+    fn: async (stepsCtx, { inputs }) => {
+      const { logger } = stepsCtx;
+      const workflowJobId = stepsCtx.global.env.__WORKFLOW_JOB_ID;
+      if (!workflowJobId) {
+        return;
```
sjchmiela, line 47: log why?

packages/build-tools/src/steps/functions/maestroResultParser.ts
sjchmiela: Ah, I see you're leaning towards parsing JSON… hmm

One peculiarity about commands is that it may contain multiple FAILED steps. I have such commands.json where a subflow is marked as FAILED _and_ a single command is marked as failed. Not sure if we want to need to remember and handle these cases… But maybe we do and we just pick the last (by timestamp) FAILED step as the one that triggered the failure…

sjchmiela: I'll let you decide what to do. Maybe we can submit a PR to Maestro to get millisecond precision on durations… https://github.com/mobile-dev-inc/Maestro/blob/e47ba45faf8318d411717ff7a03cc1b5d764fb72/maestro-cli/src/main/java/maestro/cli/report/JUnitTestSuiteReporter.kt#L55

packages/build-tools/src/steps/functions/maestroResultParser.ts
hSATAC: `FlowMetadata` will be switched to z.output as part of the next comment.

For `MaestroFlowResult` — this is an internal type we construct ourselves in parseMaestroResults(), not something parsed from external input. TypeScript already guarantees correctness at compile time here, so adding Zod runtime validation wouldn't provide additional safety. Happy to change if you feel strongly though.

packages/build-tools/src/steps/functions/maestroResultParser.ts
hSATAC: Let's stick with JUnit as the primary source for now. Maestro seems to be investing more in JUnit support, so aligning with that direction should be a good bet.

I checked the Maestro source — as of 2.2.0, tags are included in JUnit XML as `<property name="tags" value="smoke, critical, auth"/>` (comma-separated). Custom properties were added in 2.1.0. I'll update parseJUnitTestCases() to extract tags from properties directly, and drop parseFlowTags().

For older Maestro versions that don't emit tags/properties, these will gracefully degrade to empty values ([] / {}), so no breakage.

We still need `ai-*.json` for `flow_file_path` (to compute relative paths) and retry counting (via timestamp directory occurrences).

PR #3391, opened 2026-02-10
[worker] Add `HttpLogStream`

sjchmiela:
# Why

We want to push logs to a logs service.

# How

Initially I wanted to set up a WebSocket connection between worker and the logs service to reduce the HTTP overhead, but I found having a persistent/reconnecting WebSocket connection is not trivial and HTTP should be ok.

Confirmed failure to upload is not logged in the GCS log.

# Test Plan

Tested manually.

https://github.com/user-attachments/assets/f80f0482-73f5-4dfb-855f-2361ed04aa4f

### DIFF
--- a/packages/worker/src/logger.ts
+++ b/packages/worker/src/logger.ts
@@ -153,6 +153,9 @@ export async function createBuildLoggerWithSecretsFilter(secrets: Job['secrets']
         Authorization: `Bearer ${secrets.robotAccessToken}`,
       },
       logger: buildLogger,
+      bufferRetentionMs: config.loggers.gcs.signedUploadUrlForLogs
+        ? Math.max(30_000, config.loggers.gcs.uploadIntervalMs * 2)
+        : null,
     });
     transformStream.pipe(httpLogStream);
   }
--- a/packages/worker/src/utils/HttpLogStream.ts
+++ b/packages/worker/src/utils/HttpLogStream.ts
@@ -5,7 +5,12 @@ import { Writable } from 'stream';
 
 import { retry } from './retry';
 
-const MAX_BATCH_SIZE = 100;
+const MAX_BATCH_BYTES = 200_000;
+
+interface BufferedLogEntry {
+  enqueuedAt: number;
+  serialized: string;
+}
 
 /**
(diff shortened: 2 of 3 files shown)


packages/worker/src/logger.ts
```diff
@@ -110,7 +112,36 @@ export async function createBuildLoggerWithSecretsFilter(
     level: logger.level(),
   });
 
-  return { logger, stream, logBuffer };
+  let httpLogStreamCleanUp: () => Promise<void>;
+
+  if (config.loggers.http.baseUrl && robotAccessToken) {
+    const httpLogStream = new HttpLogStream({
+      url: new URL(config.buildId, config.loggers.http.baseUrl).toString(),
```
AbbanMustafa, line 151: baseUrl is null in prod https://github.com/expo/eas-cli/pull/3391/files#diff-ea240fa8fbb9f7b0696997c73298fda18d79287b31ea54be85137fbbcef9f196R48

AbbanMustafa: or that is intentional to make this PR noop?

sjchmiela: Yes this is intentional to make the PR noop for now

packages/worker/src/utils/HttpLogStream.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const MAX_BATCH_SIZE = 100;
+
+/**
+ * A bunyan-compatible writable stream for sending logs over HTTP.
+ */
+export default class HttpLogStream extends Writable {
+  public writable = true;
+
+  private readonly buffer: unknown[] = [];
```
hSATAC: I saw we have `MAX_BATCH_SIZE` down below but we don't have `MAX_BUFFER_SIZE`. Do you think it's worth protecting?

hSATAC (approved):
LGTM

packages/worker/src/utils/HttpLogStream.ts
sjchmiela: Implemented trimming of the buffer and also fixed how we batch to align with what the logs service expects.

PR #3397, opened 2026-02-12
[worker] Logs improvements

sjchmiela:
# Why

For the new logs experience we need to identify every log with a unique ID.

Also, in the new logs experience we're going to be handling logs in plaintext, uncompressed, so size starts to matter.

# How

Decorating every log with a unique ID is surprisingly difficult with bunyan. It only runs serializers on the object passed in, I tried adding a `log_id` serializer `serializers: { log_id: () => uuidv7() }` and a non-empty value to build logger, but it was getting serialized in the moment of logger creation.

Instead, I'm adding `log_id` in the `Transform` that also masks secrets. The overall architecture looks like this now:

Bez tytułu-2026-02-04-1436-3

Also removes `service: 'worker'` and `hostname` from logs. `service` was only being used [here](https://github.com/expo/universe/blob/3246198a3f913bf6fd0fd82e8e3df402cb59e82e/server/website/common/logs-utils.ts#L218) which I don't think is a thing right now.

# Test Plan

Added unit test that ensures `Transform` adds `logId`.

Zrzut ekranu 2026-02-12 o 16 02 32

### DIFF
--- a/packages/worker/src/logger.ts
+++ b/packages/worker/src/logger.ts
@@ -70,7 +70,7 @@ function createTransformStream(secrets: EnvironmentSecret[]): Transform {
       let chunk = _chunk;
       if (chunk && typeof chunk === 'object') {
         // Remove hostname from logs since we don't need it.
-        const { hostname, ...rest } = chunk;
+        const { hostname: _hostname, ...rest } = chunk;
         chunk = {
           ...rest,
           // Add logId to each log.


packages/worker/src/__unit__/logger.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it('drains transformed logs even without explicit output consumer', async () => {
+    const { logger, outputStream } = await createBuildLoggerWithSecretsFilter([]);
+
+    logger.info('Test log');
+    logger.info('Test log');
+    logger.info('Test log');
+
+    await waitForStreamFlush();
+
+    expect(outputStream.readableLength).toBe(0);
```
AbbanMustafa, line 82: would be nice to add a test with a filtered secret

sjchmiela: https://github.com/expo/eas-cli/blob/598691bf781c84c81a93fb82b6e66b39f4c45823/packages/worker/src/__unit__/logger.test.ts#L11-L41

packages/worker/src/__unit__/logger.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    logger.info('Test log');
+    logger.info('Test log');
+
+    await waitForStreamFlush();
+
+    expect(logs.length).toBe(2);
+
+    for (const log of logs) {
+      expect(log.logId).toBeDefined();
+      expect(z.uuidv7().parse(log.logId)).toBe(log.logId);
```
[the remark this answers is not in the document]
sjchmiela: This is true for Zod v3. Zod v4 has this https://zod.dev/api?id=uuids

PR #3398, opened 2026-02-12
Use auth code flow

byronkarlen:
# Why
Browser-based login should get a code and then exchange that code for a session_secret. 

# How
Used new auth flow provided by website/www. Uses state, code_challenge and code_verifier like OAuth 2.1.

# Test Plan
Tested locally against expo.test with https://github.com/expo/universe/pull/24960
Made sure:
- works when user already logged in and selects existing account
- works when user already logged in and wants to log into another account
- works when user not already logged in
- the new flow is used
- errors with token exchange would be handled properly

Tested against expo.dev as well. Tried to test SSO but was hard since our stub-idp doesn't save state/users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Use authorization code flow with PKCE for browser-based login. ([#3398](https://github.com/expo/eas-cli/pull/3398) by [@byronkarlen](https://github.com/byronkarlen))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- Use authorization code flow with PKCE for browser-based login. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@byronkarlen](https://github.com/byronkarlen))
```
tchayen: XXXX?

byronkarlen: Whoops thank you! (geez claude)

PR #3399, opened 2026-02-12
Auto-generate bundle identifier and package name in non-interactive mode

EvanBacon:
## Summary

- In non-interactive mode, automatically generate and use a bundle identifier / application ID instead of throwing an error
- Matches the behavior of Expo CLI's `getOrPromptApplicationId.ts`
- The generated ID follows the pattern `com.{accountName}.{slug}` (or uses the other platform's ID if already set)
- Only throws an error if no valid suggestion can be generated (e.g., GraphQL lookup fails and no fallback available)

## Test plan

- Updated existing tests to verify new auto-generation behavior
- Added tests for:
  - Auto-generating bundle identifier from account name + slug
  - Using `android.package` as iOS bundle identifier when available
  - Using `ios.bundleIdentifier` as Android package when available  
  - Still throwing error when no valid suggestion can be generated

🤖 Generated with [Claude Code](https://claude.ai/code)


brentvatne (approved):
on board with the idea, will let @douglowder review code

PR #3402, opened 2026-02-13
[worker] unknown error log std error/output

AbbanMustafa:
# Why

Read and log underlying phase step error from stderror or stdout. This will allow us to group patterns and wildcard match https://app.datadoghq.com/logs?query=environment%3Adevelopment&agg_m=count&agg_m_source=base&agg_q=status&agg_q_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&top_n=10&top_o=top&viz=pattern&x_missing=true&from_ts=1770915636876&to_ts=1771002036876&live=true

### DIFF
--- a/packages/worker/src/service.ts
+++ b/packages/worker/src/service.ts
@@ -330,7 +330,8 @@ export default class BuildService {
         },
       });
 
-      if (err.errorCode === errors.ErrorCode.UNKNOWN_ERROR) {
+      const robotAccessToken = job.secrets?.robotAccessToken;
+      if (robotAccessToken && err.errorCode === errors.ErrorCode.UNKNOWN_ERROR) {
         let rawErrorMessage: string = '';
         if (maybeRawError.stderr) {
           rawErrorMessage += '\n' + getLastNLines(100, maybeRawError.stderr);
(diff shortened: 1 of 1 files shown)


packages/worker/src/service.ts
```diff
@@ -331,15 +331,12 @@ export default class BuildService {
...
-        let rawErrorMessage: string;
-        if (maybeRawError instanceof Error) {
-          rawErrorMessage = maybeRawError.message;
-        } else {
-          try {
-            rawErrorMessage = JSON.stringify(maybeRawError);
-          } catch {
-            rawErrorMessage = String(maybeRawError);
-          }
+        let rawErrorMessage: string = '';
```
sjchmiela, line 335: Are you sure we want to skip `maybeRawError.message`? And `JSON.stringify`?

Can you add `robotAccessToken` check to line 333 and remove one from line 346/343?

Also, I suspect we want to log more than just `UNKNOWN_ERROR` errors? https://github.com/expo/eas-cli/blob/e8da383de469c8e667d9efc916fbd06335dd82ef/packages/build-tools/src/buildErrors/detectError.ts#L83-L91

AbbanMustafa: The message was not useful for unknowns https://github.com/expo/eas-cli/blob/main/packages/eas-build-job/src/errors.ts#L80. Yes I intend to expand this for the other error types

PR #3425, opened 2026-02-24
Retry Node install without NVM mirror after diagnostics

sjchmiela:
Summary
- log `nvm ls-remote` output whenever a custom Node.js install fails so diagnostics are always available
- refactor the worker `installNode` flow to share an install helper, rerun without `NVM_NODEJS_ORG_MIRROR` when present, and keep downstream behavior unchanged
- add unit tests that cover diagnostics logging, the mirror-free retry, and eventual failure when both attempts fail

Testing
- Not run (not requested)


sjchmiela:
Closed in favor of https://github.com/expo/eas-cli/pull/3452

PR #3434, opened 2026-02-26
[eas-build] - provide package lockfile-utils, enable preflight checks to eas build command

AbbanMustafa:
# Why

Our new build error [logging](https://app.datadoghq.com/logs?query=build_phase%3Ainstall_dependencies&agg_m=count&agg_m_source=base&agg_q=status%2Cservice&agg_q_source=base%2Cbase&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&messageDisplay=inline&refresh_mode=sliding&sort_m=%2C&sort_m_source=%2C&sort_t=%2C&storage=hot&stream_sort=time%2Cdesc&top_n=10%2C10&top_o=top%2Ctop&viz=pattern&x_missing=true%2Ctrue&from_ts=1771882209595&to_ts=1772487009595&live=true) has highlighted that nearly ~3k builds fail every week on the `install_dependencies` phase. These are typically from out of sync lockfiles or peer dependency conflicts. These can be avoided by providing a set of preflight checks that ensure the dependency installation is valid before queueing up to build. Quicker feedback is a much better experience for the user, and helps us from running wasteful builds. 

Screenshot 2026-03-02 at 3 31 17 PM

Screenshot 2026-03-02 at 3 29 40 PM

# How

  - Add lockfile preflight checks that run before uploading to the EAS builder, catching common `install_dependencies` failures instantly on the client
  - Define a module-level flag (`lockfileChecked`) so the validation only runs once, even when building for both iOS and Android in the same command. Same
  pattern as the existing sdkVersionChecked and metroConfigValidated flags.
  - Supports npm, yarn, pnpm, and bun (text lockfile); binary bun.lockb gracefully skips parse-dependent checks

The set of checks to be made:
```
  ┌─────┬──────────────────────────────┬──────────┬──────────────────────────────────────────────────────────────────────────────────┐
  │  #  │            Check             │ Severity │                                   Description                                    │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 1   │ Lockfile existence           │ Error    │ No lockfile detected at all                                                      │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 2   │ Lockfile missing for manager │ Error    │ e.g. npm detected but no package-lock.json (checks workspace root for monorepos) │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 3   │ Lockfile ignored by VCS      │ Error    │ Lockfile exists but is gitignored/easignored — won't be uploaded                 │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 4   │ Conflicting lockfiles        │ Error    │ Lockfiles from multiple managers (bun's two variants treated as same manager)    │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 5   │ Lockfile out of sync         │ Error    │ Locked versions don't satisfy package.json specifiers                            │
  ├─────┼──────────────────────────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────┤
  │ 6   │ Peer dependency conflicts    │ Error    │ Direct dependencies have unsatisfied peer requirements (skips optional peers)    │
  └─────┴──────────────────────────────┴──────────┴──────────────────────────────────────────────────────────────────────────────────┘

```

# Test Plan
Example of how a build would fail before: https://expo.dev/accounts/mustafaexpo/projects/wander/builds/c596e1ac-f8ba-4af4-9840-fe1492837ef7

After:
Screenshot 2026-03-02 at 3 51 20 PM

Validated each scenario fails locally without queueing a build and with a helpful error log to provide the resolution step to the user.

```
eas build
✔ Incremented versionCode from 30 to 31.
Lockfile is out of sync with package.json:
  react: package.json requires 19.0.4, but lockfile has 19.0.0
Run "bun install" to update your lockfile, then commit the changes.
```
```
✔ Incremented versionCode from 34 to 35.
Conflicting lockfiles detected from multiple package managers: bun, npm.
Having lockfiles from multiple package managers can cause install failures on the EAS builder. Remove the lockfiles you do not need and commit the change.
```
```
✔ Incremented versionCode from 37 to 38.
No lockfile found in the project directory.
A lockfile is required to ensure deterministic dependency installation on the EAS builder.
Run your package manager's install command (e.g. "npm install") to generate one, then commit it to version control.
```
```
✔ Incremented versionCode from 42 to 43.
Peer dependency conflicts detected:
  react-dom@19.0.0 requires peer react@"^19.1.0", but react@19.0.0 is installed
Update the conflicting packages to compatible versions, then run "bun install" and commit the changes.
```


packages/eas-cli/src/build/runBuildAndSubmit.ts
```diff
@@ -414,6 +416,11 @@ async function prepareAndStartBuildAsync({
     env,
   });
 
+  if (!lockfileChecked) {
```
douglowder, line 419: @AbbanMustafa this is a great PR! I do think that we should provide a way for customers to opt out of this check, in case they have an unusual configuration that is really valid for them, but might not pass these checks. Maybe something like EAS_BUILD_SKIP_LOCKFILE_CHECK.

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,485 @@
...
+interface LockfileInfo {
+  /** Filename of the lockfile (e.g. 'yarn.lock') */
+  filename: string;
+  /** Package manager name for display (e.g. 'yarn') */
+  managerName: string;
+}
+
+const LOCKFILE_BY_MANAGER: Record<string, LockfileInfo> = {
+  npm: { filename: NPM_LOCK_FILE, managerName: 'npm' },
+  yarn: { filename: YARN_LOCK_FILE, managerName: 'Yarn' },
```
douglowder: Curious as to why managerName is capitalized for yarn and bun....

douglowder: Also, in reading through this, it occurred to me that it might be good to package all this code for reuse, in case it's needed elsewhere. I also noticed that there are open source packages for this, e.g. https://github.com/snyk/nodejs-lockfile-parser

packages/eas-cli/src/build/runBuildAndSubmit.ts
AbbanMustafa: thank you! yes I agree an opt out would be nice

packages/eas-cli/src/build/validateLockfile.ts
AbbanMustafa: I like the idea of making this a package. I took a look with Claude on what packages are out there, and what could be viable for this usecase.

> - snyk/nodejs-lockfile-parser — Full dependency graph for vulnerability scanning. No bun support, no peer dep info, heavy, wrong output format for our needs.
> - lockfile-lint — Security-focused linter (checks for HTTPS registry URLs, allowed hosts). Doesn't check version sync or peer deps. npm/yarn only.
> - lock-verify — Deprecated (since npm v7). npm only. Checks package.json sync but not peer deps.
> - @pnpm/lockfile-file — pnpm-only, internal pnpm tooling.
> - package-locks-checks — Detects which package manager is in use. Doesn't do version sync or peer dep validation.

None seem to cover the parsing for all four lockfile formats plus checking version sync and peer dependency conflicts. So I am now extracting our parsing and validation logic into `@expo/lockfile-utils`

AbbanMustafa:
Example where the preflight check is enforced and then skipped via env var

Screenshot 2026-03-11 at 2 37 18 PM
Screenshot 2026-03-11 at 2 38 17 PM


### Additional error scenarios:

Screenshot 2026-03-11 at 2 56 21 PM

Screenshot 2026-03-11 at 2 56 52 PM
Screenshot 2026-03-11 at 2 53 56 PM

douglowder (approved):
Looks good. Adding Phil to get his opinions on the lockfile and dependency checks.

packages/eas-cli/src/build/runBuildAndSubmit.ts
sjchmiela, line 419: I think the check for the opt-out should be here, not in the `checkLockfileAsync`.

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+    return;
+  }
+
+  const { projectDir, requiredPackageManager, vcsClient } = ctx;
+
+  // --- 1. Lockfile existence ---
+  if (!requiredPackageManager) {
+    Log.error('No lockfile found in the project directory.');
+    Log.error(
+      'A lockfile is required to ensure deterministic dependency installation on the EAS builder.'
```
sjchmiela, line 109:
```suggestion
      'A lockfile is required to ensure deterministic dependency installation in EAS.'
```

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+    );
+    Log.error(
+      'Run your package manager\'s install command (e.g. "npm install") to generate one, then commit it to version control.'
```
sjchmiela, line 112: committing is not necessary

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+  parseLockfile,
+} from '@expo/lockfile-utils';
+import fs from 'fs-extra';
+import path from 'path';
+
+import { CommonContext } from './context';
+import Log from '../log';
+
+const SKIP_HINT =
+  'To skip this check, set the EAS_BUILD_SKIP_LOCKFILE_CHECK=1 environment variable.';
```
sjchmiela, line 23: "to skip this check, run this command with EAS_BUILD_SKIP_LOCKFILE_CHECK=1"?

maybe slightly more clear as to how to set the variable

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  const lockInfo = LOCKFILE_BY_MANAGER[requiredPackageManager];
+  if (!lockInfo) {
+    // Unknown package manager — nothing we can validate
+    return;
+  }
+
+  // For bun, also check the text lockfile variant
+  const lockfilePath =
+    requiredPackageManager === 'bun'
```
sjchmiela, line 127: instead of this i think we should either use `ALL_LOCKFILES` or make `LOCKFILES_BY_MANAGER` (plural). this if is weird

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+  'To skip this check, set the EAS_BUILD_SKIP_LOCKFILE_CHECK=1 environment variable.';
+
+interface LockfileInfo {
+  /** Filename of the lockfile (e.g. 'yarn.lock') */
+  filename: string;
+  /** Package manager name for display (e.g. 'yarn') */
+  managerName: string;
+}
+
+const LOCKFILE_BY_MANAGER: Record<string, LockfileInfo> = {
```
sjchmiela, line 32: should we key this off of some package manager's enum?

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/**
+ * Pre-flight lockfile validation for `eas build`. Catches common lockfile
+ * problems before uploading to the builder.
+ */
+export async function checkLockfileAsync<T extends Platform>(
+  ctx: CommonContext<T>
+): Promise<void> {
+  if (process.env.EAS_BUILD_SKIP_LOCKFILE_CHECK) {
```
sjchmiela, line 98: as mentioned in the otehr comment, move to callsite

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+    Log.error(
+      'Run your package manager\'s install command (e.g. "npm install") to generate one, then commit it to version control.'
+    );
+    Log.error(SKIP_HINT);
+    process.exit(1);
+    return;
+  }
+
+  const lockInfo = LOCKFILE_BY_MANAGER[requiredPackageManager];
+  if (!lockInfo) {
```
sjchmiela, line 120: if `LOCKFILE_BY_MANAGER` is guaranteed to have values for all `NodePackageManager['name']` we don't need this check

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+      ? (await findLockfileAsync(projectDir, BUN_LOCK_FILE)) ??
+        (await findLockfileAsync(projectDir, BUN_TEXT_LOCK_FILE))
+      : await findLockfileAsync(projectDir, lockInfo.filename);
+
+  if (!lockfilePath) {
+    Log.error(
+      `${lockInfo.managerName} lockfile (${lockInfo.filename}) not found.`
+    );
+    Log.error(
+      `Run "${requiredPackageManager} install" to generate a lockfile, then commit it to version control.`
```
sjchmiela, line 137: no need to commit (unless requireCommit: true)

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+    process.exit(1);
+    return;
+  }
+
+  // --- 2. Lockfile tracked in VCS ---
+  const rootDir = path.normalize(await vcsClient.getRootPathAsync());
+  const relativeLockfilePath = path.relative(rootDir, lockfilePath);
+  if (await vcsClient.isFileIgnoredAsync(relativeLockfilePath)) {
+    Log.error(
+      `${lockInfo.managerName} lockfile (${relativeLockfilePath}) is ignored by your version control system and won't be uploaded to the EAS builder.`
```
sjchmiela, line 149: eas supports no version control system existing, this is misleading

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+  if (!lockfilePath) {
+    Log.error(
+      `${lockInfo.managerName} lockfile (${lockInfo.filename}) not found.`
+    );
+    Log.error(
+      `Run "${requiredPackageManager} install" to generate a lockfile, then commit it to version control.`
+    );
+    Log.error(SKIP_HINT);
+    process.exit(1);
+    return;
```
sjchmiela, line 141: why log.error and process.exit instead of throwing?

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  // --- 4 & 5. Parse lockfile for dependency sync and peer dep checks ---
+  // For bun, prefer the text lockfile for parsing (bun.lockb is binary)
+  const parsableLockfilePath =
+    requiredPackageManager === 'bun'
+      ? (await findLockfileAsync(projectDir, BUN_TEXT_LOCK_FILE)) ?? lockfilePath
+      : lockfilePath;
```
sjchmiela, line 183: again, weird special check for bun

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  const parsed = await readLockfileDataAsync(
+    projectDir,
+    parsableLockfilePath,
+    requiredPackageManager
+  );
+
+  if (!parsed) {
+    return; // Can't parse — skip remaining checks
+  }
```
sjchmiela, line 193: i think it would make more sense for `readLockfileDataAsync` to throw if can't parse and we should catch the error here

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  if (mismatches.length > 0) {
+    Log.error('Lockfile is out of sync with package.json:');
+    for (const m of mismatches) {
+      Log.error(
+        `  ${m.name}: package.json requires ${m.specifier}, but lockfile has ${m.lockedVersion}`
+      );
+    }
+    Log.error(
+      `Run "${requiredPackageManager} install" to update your lockfile, then commit the changes.`
```
sjchmiela, line 206: commit

packages/eas-cli/src/build/validateLockfile.ts
```diff
@@ -0,0 +1,231 @@
...
+    );
+    Log.error(SKIP_HINT);
+    process.exit(1);
+    return;
+  }
+
+  // --- 5. Peer dependency conflicts ---
+  const peerConflicts = findPeerDependencyConflicts(parsed.packageJson, parsed.lockfileData);
```
sjchmiela, line 214: does frozen lockfile fail if peer dependencies are mismatched?

packages/lockfile-utils/src/validators.ts
```diff
@@ -0,0 +1,92 @@
+import semver from 'semver';
+
+import { DependencyMismatch, LockfileData, PeerDependencyConflict } from './types';
+
+function isNonRegistrySpecifier(specifier: string): boolean {
```
sjchmiela, line 5: should this be "if contains `:`"?

packages/lockfile-utils/src/validators.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    if (!semver.satisfies(pkg.version, specifier)) {
+      mismatches.push({ name, specifier, lockedVersion: pkg.version });
+    }
+  }
+
+  return mismatches;
+}
+
+export function findPeerDependencyConflicts(
```
sjchmiela, line 51: the function, to me, suggests it returns cofnlicnts on `packageJson.peerDependencies`, not… peer dependencies of `packageJson.dependencies`?

packages/lockfile-utils/src/validators.ts
```diff
@@ -0,0 +1,92 @@
...
+): DependencyMismatch[] {
+  const allDeps: Record<string, string> = {
+    ...(packageJson.dependencies as Record<string, string> | undefined),
+    ...(packageJson.devDependencies as Record<string, string> | undefined),
+    ...(packageJson.optionalDependencies as Record<string, string> | undefined),
+  };
+
+  const mismatches: DependencyMismatch[] = [];
+  for (const [name, specifier] of Object.entries(allDeps)) {
+    if (isNonRegistrySpecifier(specifier)) {
```
sjchmiela, line 34: what would `semver.satisfies` do if we passed in a non-registry specifier? maybe we can handle the error somehow instead of whitelisting some set of known prefixes of dep versions?

packages/lockfile-utils/src/validators.ts
```diff
@@ -0,0 +1,92 @@
...
+      continue;
+    }
+
+    const pkg = lockfileData.packages.get(name);
+    if (!pkg) {
+      continue; // Package not found in lockfile — may be a new dep or hoisted differently
```
sjchmiela, line 40: i assumed one of the reasons builds fail is because users add a new dep to package.json but don't install deps… (the "new dep" case which we're skipping here, i think?)

maybe this is not the case?

sjchmiela (changes requested):

PR #3442, opened 2026-02-28
[cli] CI check and environment flag override for update

douglowder:
# Why

Provide a way to override the new requirement for the `--environment` flag in `eas update`.

# How

Modify the check to skip when the `EAS_UPDATE_SKIP_ENVIRONMENT_CHECK` variable is set to a boolish value (1 or true).

If the command is run interactively, we now prompt the user for the environment, instead of throwing an error. This is skipped if the `--auto` flag is set. A missing environment flag already throws an error if `--non-interactive` is set.

Add the `CI` flag to the tests for whether we should skip the check in a CI environment.

# Test Plan

New unit tests added.

### DIFF
--- a/packages/eas-cli/src/commands/update/index.ts
+++ b/packages/eas-cli/src/commands/update/index.ts
@@ -264,7 +264,6 @@ export default class UpdatePublish extends EasCommand {
         environment: environmentFromFlags,
       })
     ) {
-
       environment = await promptVariableEnvironmentAsync({
         multiple: false,
         graphqlClient,


douglowder:
* **#3442** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

douglowder:
Thanks @quinlanj I'll take a look at making the flow look like this and adding some tests

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -253,10 +254,24 @@ export default class UpdatePublish extends EasCommand {
...
+
+    // Environment handling
+    if (
+      !autoFlag &&
+      environmentFlagNeededForSdk550OrGreater({
+        sdkVersion: expPossiblyWithoutEasUpdateConfigured.sdkVersion,
+        environment: environmentFromFlags,
+      })
+    ) {
+      console.warn('Prompting for environment...');
```
quinlanj: Since the prompt for environment already asks to 'Select environment'
```suggestion
```

packages/eas-cli/src/commands/update/index.ts
```diff
@@ -253,10 +254,24 @@ export default class UpdatePublish extends EasCommand {
...
+
+    // Environment handling
+    if (
+      !autoFlag &&
+      environmentFlagNeededForSdk550OrGreater({
+        sdkVersion: expPossiblyWithoutEasUpdateConfigured.sdkVersion,
+        environment: environmentFromFlags,
+      })
+    ) {
+
```
douglowder:
```suggestion
```

douglowder:
### Merge activity

* **Mar 6, 2:55 PM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/eas-cli/3442).

PR #3444, opened 2026-03-03
[build-tools] Improve ASC build upload error handling and diagnostics

sjchmiela:
## Why
App Store Connect build upload failures were often surfaced as generic errors, which made remediation unclear for common cases like duplicate build numbers. We also needed stricter and safer response parsing so JSON parsing failures don’t leak raw parse errors.

## How
- Added typed `AscApiRequestError` in `AscApiClient` for structured ASC error payloads only.
- Kept non-structured ASC failures as regular `Error`.
- Hardened response parsing in `AscApiClient`:
  - parse response text with `asyncResult` + `JSON.parse`,
  - convert invalid success JSON into a controlled `Malformed JSON response...` error.
- Added `AscApiUtils.createBuildUploadAsync` and switched `uploadToAsc` to call it.
- Added duplicate build-number recognition for `/v1/buildUploads` failures when all ASC error codes are `ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE`.
- Added specific `UserFacingError` (`EAS_UPLOAD_TO_ASC_VERSION_DUPLICATE`) with concrete remediation:
  - increment `ios.buildNumber` (or set `"autoIncrement": true`),
  - rebuild and resubmit,
  - include docs URL.
- Inlined closed-train constants in `uploadToAsc` as requested.

## Test Plan
- `yarn --cwd packages/build-tools jest-unit src/steps/utils/ios/__tests__/AscApiClient.test.ts`
- `yarn --cwd packages/build-tools jest-unit src/steps/utils/ios/__tests__/AscApiUtils.test.ts`
- `yarn --cwd packages/build-tools jest-unit src/steps/functions/__tests__/uploadToAsc.test.ts`

All commands above pass on this branch.

PR #3445, opened 2026-03-03
[worker] Preserve `UserFacingError` instead of `UNKNOWN_ERROR`

sjchmiela:
## Why
Generic/custom workflow steps can throw `UserFacingError`, but worker-level error handling currently maps any non-`BuildError` to `UnknownBuildError` / `UnknownCustomBuildError`. That causes actionable step errors to be persisted and exposed as `UNKNOWN_ERROR`.

## How
- add a `toBuildError` conversion in worker service catch path
- preserve existing `BuildError` values
- convert `UserFacingError` into `BuildError` with the same code/message/docsUrl
- keep unknown fallbacks only for truly unstructured errors
- add integration coverage in `stateSync` to assert generic job `UserFacingError` is exposed in `externalBuildError` and `internalErrorCode`

We should make a bigger refactor to error handling, but it's more involved — https://github.com/expo/eas-cli/pull/3447.

## Test Plan

CI should pass.

```
 PASS  src/__integration__/stateSync.test.ts
  State sync mechanism
    query new worker
      ✓ should return 'new' state (210 ms)
    query worker after dispatch
      ✓ should return 'in-progress' state (170 ms)
    query worker after build
      ✓ should return 'success' state (2062 ms)
    query worker after error
      ✓ should return 'error' state with artifacts from error (88 ms)
      ✓ should expose user-facing error from generic job (60 ms)
```

### DIFF
--- a/packages/worker/src/service.ts
+++ b/packages/worker/src/service.ts
@@ -373,12 +373,13 @@ function toBuildError(error: unknown, job: Job): errors.BuildError {
   }
 
   if (error instanceof errors.UserFacingError) {
+    const innerError = error.cause instanceof Error ? error.cause : error;
     return new errors.BuildError(error.message, {
       errorCode: error.errorCode,
       userFacingErrorCode: error.errorCode,
       userFacingMessage: error.message,
       docsUrl: error.docsUrl,
-      innerError: error,
+      innerError,
     });
   }
 


packages/worker/src/service.ts
AbbanMustafa, line 328: Whats an example of a user facing error? We will now no longer be logging them here. I think thats what we want

sjchmiela: https://github.com/expo/eas-cli/pull/3444 I too think that's what we want!

sjchmiela:
NICE https://expo.dev/accounts/sdom/projects/sdomcliente/workflows/019cb810-60c1-74c9-9aa3-be40ff433aab

PR #3459, opened 2026-03-05
[build-tools] Add `--include=dev` for `npm install` and `--production false` for Yarn classic

kitten:
# Why

Resolves https://github.com/expo/expo/issues/43664

`npm install` omits `devDependencies` when `NODE_ENV === 'production'`. The same actually seems to go for yarn classic. Yarn Berry is unaffected, Bun is unaffected, and pnpm is also unaffected

# How

- Add `--include=dev` for `npm` consistently
- Add `--production false` for Yarn classic

# Test Plan

- Manually run `yarn install` and `npm install`, altering `NODE_ENV` to reproduce the problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Add missing `--include=dev` for `npm install` commands ([#3459](https://github.com/expo/eas-cli/pull/3459) by [@kitten](https://github.com/kitten))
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/common/installDependencies.ts
```diff
@@ -63,7 +64,10 @@ export async function installDependenciesAsync({
       cwd,
       logger,
       infoCallbackFn,
-      env,
+      env: {
+        ...env,
+        NODE_ENV: 'development',
```
kadikraman: This could have unintended consequences. E.g. in observe we currently show events as development vs production based on the node env during the build. It's likely other services/libraries could be relying on this as well.

PR #3476, opened 2026-03-07
Auto-enable --non-interactive when --json is passed

krystofwoldrich:
## Summary

Make the CLI easier to use by automatically enabling non-interactive mode when JSON output is requested. Users can now pass just `--json` instead of both `--json --non-interactive`.

## Changes

- Add `resolveNonInteractiveAndJsonFlags()` helper that derives `nonInteractive = flags['non-interactive'] || flags.json`
- Remove `dependsOn` constraint from json flag that previously required --non-interactive
- Remove manual validation in build command


brentvatne:
i'm ok with the spirit of this change assigning @douglowder for review

PR #3478, opened 2026-03-08
feat(cli): add flag to prevent uploading server sourcemaps

hassankhan:
# Why

EAS Hosting has a 10MB (gzipped) size limit, and we've had user requests to add a way to disable uploading source maps as a way to keep under this limit.

# How

Added a `--source-maps` boolean flag to the `eas deploy` command, and have it default to `true`. When the flag is `false`, source maps for both client and server assets are skipped.

# Test Plan

1. Create a new Expo project that has SSR/API routes/middleware
2. Ran `eas deploy --dry-run` and verified the tarball contains `.map` files
3. Ran `eas deploy --dry-run --no-source-maps` and verified the tarball does not contain `*.map` files

### DIFF
--- a/packages/eas-cli/src/commands/deploy/index.ts
+++ b/packages/eas-cli/src/commands/deploy/index.ts
@@ -52,6 +52,7 @@ interface DeployFlags {
   deploymentIdentifier?: string;
   exportDir: string;
   dryRun: boolean;
+  sourceMaps: boolean;
 }
 
 interface RawDeployFlags {
@@ -63,6 +64,7 @@ interface RawDeployFlags {
   id?: string;
   'export-dir': string;
   'dry-run': boolean;
+  'source-maps': boolean;
 }
 
 interface UploadAssetBatchInstruction {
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/deploy/index.ts
```diff
@@ -291,6 +300,7 @@ export default class WorkerDeploy extends EasCommand {
       const assetPath = projectDist.type === 'server' ? projectDist.clientPath : projectDist.path;
       assetFiles = await WorkerAssets.collectAssetsAsync(assetPath, {
         maxFileSize: MAX_UPLOAD_SIZE,
+        skipSourceMaps: !flags.sourceMaps,
```
kitten: I reckon we don't want to skip the asset source maps, but only the server ones, since that's where the size limitation applies

hassankhan: Fair enough, I was trying to match `@expo/cli`'s semantics and was additionally considering maybe allowing `'client' | 'server'` as values for the `--source-maps` flag.

PR #3479, opened 2026-03-09
[build-tools][steps] Report step metrics after each step

hSATAC:
# Why

Step metrics are currently batched and uploaded once during the `COMPLETE_JOB` phase. If the job crashes or the process exits early, all metrics are lost. Reporting metrics immediately after each step completes improves observability and resilience.
# How

  - Add `onStepMetricCollected` callback on `BuildStepGlobalContext` that fires each time a step metric is collected
  - Wire up the callback in `generic.ts` to upload each metric via `POST /v2/workflows/{jobId}/metrics` immediately
  - Implement smart retry with a retryable status allowlist (`408, 429, 500, 502, 503, 504`); non-retryable errors fail silently
  - Drain all in-flight uploads with `Promise.allSettled` in `COMPLETE_JOB` (via `try/finally`) to ensure delivery before `process.exit`
  - Remove the old batch upload path

# Test Plan

- Automation tests
- Manual test locally

### DIFF
--- a/packages/build-tools/src/generic.ts
+++ b/packages/build-tools/src/generic.ts
@@ -15,6 +15,8 @@ import { retryAsync } from './utils/retry';
 export async function runGenericJobAsync(
   ctx: BuildContext<Generic.Job>
 ): Promise<{ runResult: Result<void>; buildWorkflow: BuildWorkflow }> {
+  // expoApiV2BaseUrl is empty when running a local build, but generic jobs
+  // are never executed locally, so it is always present here.
   const expoApiV2BaseUrl = nullthrows(
     ctx.expoApiV2BaseUrl,
     'expoApiV2BaseUrl is required for generic jobs'
@@ -64,13 +66,15 @@ export async function runGenericJobAsync(
   const runResult = await asyncResult(workflow.executeAsync());
 
   await ctx.runBuildPhase(BuildPhase.COMPLETE_JOB, async () => {
-    try {
-      await uploadJobOutputsToWwwAsync(globalContext, {
+    const results = await Promise.allSettled([
+      uploadJobOutputsToWwwAsync(globalContext, {
         logger: ctx.logger,
         expoApiV2BaseUrl,
-      });
-    } finally {
-      await customBuildCtx.drainPendingMetricUploads();
+      }),
+      customBuildCtx.drainPendingMetricUploads(),
+    ]);
+    if (results[0].status === 'rejected') {
+      throw results[0].reason;
     }
   });
 
--- a/packages/build-tools/src/utils/stepMetrics.ts
+++ b/packages/build-tools/src/utils/stepMetrics.ts
@@ -1,42 +1,34 @@
 import { bunyan } from '@expo/logger';
 import { StepMetric } from '@expo/steps';
 
-import { sleepAsync } from './retry';
 import { turtleFetch } from './turtleFetch';
 
-export async function uploadStepMetricToWwwAsync({
+export async function uploadStepMetricsToWwwAsync({
   workflowJobId,
   robotAccessToken,
   expoApiV2BaseUrl,
-  stepMetric,
+  stepMetrics,
   logger,
 }: {
   workflowJobId: string;
   robotAccessToken: string;
   expoApiV2BaseUrl: string;
-  stepMetric: StepMetric;
+  stepMetrics: StepMetric[];
   logger: bunyan;
 }): Promise<void> {
-  const maxAttempts = 3;
-  for (let attempt = 0; attempt < maxAttempts; attempt++) {
-    try {
-      await turtleFetch(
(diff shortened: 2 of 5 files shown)


packages/steps/src/BuildStepContext.ts
```diff
@@ -188,7 +189,9 @@ export class BuildStepGlobalContext {
   }
 
   public addStepMetric(metric: StepMetricInput): void {
-    this._stepMetrics.push({ ...metric, platform: this.runtimePlatform });
+    const stepMetric: StepMetric = { ...metric, platform: this.runtimePlatform };
+    this._stepMetrics.push(stepMetric);
```
sjchmiela: Do we want to remove that?

packages/build-tools/src/generic.ts
```diff
@@ -37,6 +37,23 @@ export async function runGenericJobAsync(
 
   const globalContext = new BuildStepGlobalContext(customBuildCtx, false);
 
+  const workflowJobId = customBuildCtx.env.__WORKFLOW_JOB_ID;
+  const robotAccessToken = ctx.job.secrets?.robotAccessToken;
+  const pendingMetricUploads: Promise<void>[] = [];
+
+  if (workflowJobId && robotAccessToken) {
+    globalContext.onStepMetricCollected = metric => {
+      const p = uploadStepMetricToWwwAsync({
```
sjchmiela: should the context be the one doing all this? so managed build jobs upload metrics too?

packages/build-tools/src/utils/stepMetrics.ts
```diff
@@ -39,3 +40,43 @@ export async function uploadStepMetricsToWwwAsync({
...
+        {
+          json: { stepMetrics: [stepMetric] },
+          headers: { Authorization: `Bearer ${robotAccessToken}` },
+          timeout: 20000,
+          logger,
+        }
+      );
+      return;
+    } catch (err) {
+      const RETRYABLE_STATUSES = [408, 429, 500, 502, 503, 504];
```
sjchmiela: let's… not? maybe it's not that important that we need to maintain a list like this?

packages/steps/src/BuildStepContext.ts
hSATAC: Yes we do. Thanks for pointing it out.

packages/build-tools/src/generic.ts
```diff
@@ -10,12 +11,15 @@ import { getEasFunctionGroups } from './steps/easFunctionGroups';
...
 
 export async function runGenericJobAsync(
-  ctx: BuildContext<Generic.Job>,
-  { expoApiV2BaseUrl }: { expoApiV2BaseUrl: string }
+  ctx: BuildContext<Generic.Job>
 ): Promise<{ runResult: Result<void>; buildWorkflow: BuildWorkflow }> {
+  const expoApiV2BaseUrl = nullthrows(
+    ctx.expoApiV2BaseUrl,
+    'expoApiV2BaseUrl is required for generic jobs'
+  );
```
sjchmiela, line 23: Let's add a comment that `expoApiV2BaseUrl` is empty if we're running a local build and the check is "safe" because we don't run generic jobs in local build and can expect URL to always be present

packages/build-tools/src/generic.ts
hSATAC: Tried to move it to context in this commit https://github.com/expo/eas-cli/pull/3479/commits/2f9c9296f13b4d87c7431052649d42be7e6d2d34

What do you think?

packages/build-tools/src/utils/stepMetrics.ts
```diff
@@ -1,41 +1,42 @@
...
-      'POST',
-      {
-        json: { stepMetrics },
-        headers: {
-          Authorization: `Bearer ${robotAccessToken}`,
-        },
-        timeout: 20000,
-        logger,
+  const maxAttempts = 3;
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
```
sjchmiela: `turtleFetch` supports retries

Also, maybe we don't need to have such a long timeout?

nit: I think we could also have the input be `StepMetric[]` and we can just call it with `[stepMetric]`. This way if we ever wanted to send batches (which by itself would be a nice thing) we wouldn't need to change the API here

packages/build-tools/src/generic.ts
```diff
@@ -60,22 +64,13 @@ export async function runGenericJobAsync(
...
-    const workflowJobId = globalContext.env.__WORKFLOW_JOB_ID;
-    const robotAccessToken = ctx.job.secrets?.robotAccessToken;
-
-    if (workflowJobId && robotAccessToken) {
-      await uploadStepMetricsToWwwAsync({
-        workflowJobId,
-        robotAccessToken,
-        expoApiV2BaseUrl,
-        stepMetrics: globalContext.stepMetrics,
+    try {
```
sjchmiela: this can be promise.all?

PR #3481, opened 2026-03-09
[build-tools] Allow custom build phase labels

sjchmiela:
# Why

We want `runBuildPhase` to accept a custom string label so internal phases can show a more helpful name in logs without needing to extend the shared `BuildPhase` enum for each case.

Stanley: I don't think I like the fact build phase is currently an enum. It doesn't convey uniqueness, it's not extensible, it doesn't work well in my opinion. I think we'll move off of it in the future, but for now I just want to 

# How

- widen `BuildContext.runBuildPhase()` and internal phase tracking to accept `BuildPhase | string`
- add `isBuildPhase()` in `@expo/eas-build-job` so enum boundaries can still be enforced centrally
- keep non-enum phase names internal by omitting them from serialized `BuildError` / `ExternalBuildError`
- allow `UnknownError` to use the custom label only for its human-facing message
- skip build phase stats for custom labels
- add focused tests for logger metadata and for omitting non-enum `buildPhase` from returned build errors

# Test Plan

- added unit coverage in `packages/build-tools/src/__tests__/context.test.ts`
- added unit coverage in `packages/build-tools/src/buildErrors/__tests__/detectError.test.ts`
- ran `git diff --check`
- could not run Jest in this environment because `yarn` is not installed on `PATH`

### DIFF
--- a/packages/build-tools/src/__tests__/context.test.ts
+++ b/packages/build-tools/src/__tests__/context.test.ts
@@ -6,7 +6,7 @@ import { createMockLogger } from './utils/logger';
 import { BuildContext } from '../context';
 
 jest.mock('fs');
-jest.mock('fs-extra', () => jest.requireActual('fs-extra'));
+jest.mock('fs-extra');
 
 describe('BuildContext', () => {
   it('should merge secrets', async () => {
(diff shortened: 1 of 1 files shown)


sjchmiela:
Closing this in favor of an enum-backed  build phase PR. I’m replacing the string-based phase approach with a new stacked pair.

Issue #1201, opened 2022-07-08
Provisioning profile "*[expo] XXX AdHoc XXX" doesn't include signing certificate "Apple Distribution: XXX"


rcarter:
### Build/Submit details page URL

https://expo.dev/accounts/XXXXXXX/projects/XXXXXXXX/builds/253dc20a-ffb0-40f9-9707-cff2c0f6575d

### Summary

All of a sudden, with no changes to my certs/profiles when I run: `eas build --platform ios`

I’m running into:
`:x: error: Provisioning profile "*[expo] XXX AdHoc XXX" doesn't include signing certificate "Apple Distribution: XXX".`

I’ve deleted all my provisioning profiles and certs and am letting eas handle creating new ones. But still the same error (both when running `--local` as well as running in the cloud). I recently upgraded to `eas-cli/0.54.1` and I think it started happening after that upgrade. I can’t get any build past this.

### Managed or bare?

Managed

### Environment

```expo-env-info 1.0.4 environment info:
    System:
      OS: macOS 12.2
      Shell: 5.8 - /bin/zsh
    Binaries:
      Node: 16.14.2 - /opt/homebrew/bin/node
      Yarn: 1.22.10 - /opt/homebrew/bin/yarn
      npm: 8.5.0 - /opt/homebrew/bin/npm
      Watchman: 2022.03.21.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
    SDKs:
      iOS SDK:
        Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
      Android SDK:
        API Levels: 29, 30
        Build Tools: 28.0.3, 29.0.2, 30.0.2, 30.0.3
        System Images: android-30 | ARM 64 v8a, android-30 | Google APIs Intel x86 Atom_64, android-S | Google APIs ARM 64 v8a, android-S | Google Play ARM 64 v8a
    IDEs:
      Android Studio: 2020.3 AI-203.7717.56.2031.7784292
      Xcode: 13.2.1/13C100 - /usr/bin/xcodebuild
    npmPackages:
      expo: ^44.0.0 => 44.0.6 
      react: 17.0.1 => 17.0.1 
      react-dom: 17.0.1 => 17.0.1 
      react-native: 0.64.3 => 0.64.3 
      react-native-web: 0.17.1 => 0.17.1 
    npmGlobalPackages:
[...]

PR #3484, opened 2026-03-09
fix(build-tools): check all certificates in provisioning profile during verification

qwertey6:
# Why

Apple provisioning profiles can contain multiple developer certificates (e.g., when a team has multiple developers). The current code only checks `DeveloperCertificates[0]`, causing false **"Provisioning profile and distribution certificate don't match"** errors when the signing certificate is present in the profile but not at index 0.

This is especially painful for teams using `credentialsSource: "local"` with manually created profiles (e.g., for Family Controls entitlement which requires manual Development profiles).

Related issues: https://github.com/expo/eas-cli/issues/3192, https://github.com/expo/eas-cli/issues/1201

# How

- In `load()`, store **all** developer certificates from the plist into a `developerCertificates` array (the `developerCertificate` field in `ProvisioningProfileData` still points to `[0]` for backward compatibility)
- Replace `genDerCertFingerprint()` (single cert) with `getAllDerCertFingerprints()` (all certs)
- In `verifyCertificate()`, check if the provided fingerprint matches **any** certificate in the profile instead of only the first one
- Applied the same fix to both copies of the file:
  - `packages/build-tools/src/ios/credentials/provisioningProfile.ts`
  - `packages/build-tools/src/steps/utils/ios/credentials/provisioningProfile.ts`

# Test Plan

- Added a test case that initializes a provisioning profile, then prepends a dummy certificate to `developerCertificates` so the real cert moves to index 1. Verifies that:
  - The original cert fingerprint still matches (was previously only checked at index 0)
  - A non-matching fingerprint still correctly throws
  - The dummy cert's fingerprint also matches
- All 269 existing tests continue to pass (`yarn test` in `packages/build-tools`)
- `yarn typecheck` passes across all 13 packages
- `yarn lint` and `yarn fmt` pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- Check all certificates in provisioning profile during verification instead of only the first one. ([#3484](https://github.com/expo/eas-cli/pull/3484) by [@qwertey6](https://github.com/qwertey6))
 - Provide an override for the new --environment flag requirement in the update command. ([#3442](https://github.com/expo/eas-cli/pull/3442) by [@douglowder](https://github.com/douglowder))
 - Add missing `--include=dev` for `npm install` commands ([#3459](https://github.com/expo/eas-cli/pull/3459) by [@kitten](https://github.com/kitten))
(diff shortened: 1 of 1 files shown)


quinlanj (approved):
lgtm, thanks for the fix! @qwertey6 please add an entry in CHANGELOG.md

qwertey6:
@quinlanj , done!

PR #3486, opened 2026-03-09
Auto-detect non-interactive mode in CI / non-TTY environments

EvanBacon:
# Why

`eas build -p android` prompts for keystore generation and fails in CI or non-TTY environments:

```
Input is required, but stdin is not readable. Failed to display prompt: Generate a new Android Keystore?
```

Passing `--non-interactive` works around this, but it should be auto-detected when `CI=1` or stdin is not a TTY.

# How

Set a `default` function on both `--non-interactive` flag definitions (`EasNonInteractiveAndJsonFlags` and `EASNonInteractiveFlag`) in `packages/eas-cli/src/commandUtils/flags.ts`. The default checks `CI` env var (via `boolish` from `getenv`) and `process.stdin.isTTY`.

- oclif evaluates `default` only when the user does NOT pass `--non-interactive` on CLI
- `noCacheDefault: true` prevents baking the developer's environment into `oclif.manifest.json`
- All ~60 commands using these flag definitions automatically get CI/TTY auto-detection with no other changes needed

# Test Plan

- Added 5 unit tests for `isNonInteractiveByDefault()` covering: TTY without CI (false), CI=1 (true), CI=true (true), non-TTY stdin (true), both CI and non-TTY (true)
- All existing `resolveNonInteractiveAndJsonFlags` tests pass unchanged
- `yarn typecheck` passes
- `yarn lint` passes (0 errors)

```
PASS src/commandUtils/__tests__/flags-test.ts
  isNonInteractiveByDefault
    ✓ returns false when stdin is a TTY and CI is not set
    ✓ returns true when CI=1
    ✓ returns true when CI=true
    ✓ returns true when stdin is not a TTY
    ✓ returns true when both CI is set and stdin is not a TTY
  resolveNonInteractiveAndJsonFlags
    ✓ returns both false when no flags are set
    ✓ --json implies --non-interactive
    ✓ --non-interactive alone does not enable json
    ✓ both flags explicitly set
    ✓ both flags explicitly false
    ✓ --json false with --non-interactive true

Test Suites: 1 passed, 1 total
Tests:       11 passed, 11 total
```

### DIFF
--- a/packages/eas-cli/src/commandUtils/flags.ts
+++ b/packages/eas-cli/src/commandUtils/flags.ts
@@ -71,6 +71,7 @@ export const EASEnvironmentVariableScopeFlag = {
 export const EASNonInteractiveFlag = {
   'non-interactive': Flags.boolean({
     description: 'Run the command in non-interactive mode.',
+    // eslint-disable-next-line async-protect/async-suffix
     default: async () => {
       return isNonInteractiveByDefault();
     },
--- a/packages/eas-cli/src/commandUtils/__tests__/EasCommand-test.ts
+++ b/packages/eas-cli/src/commandUtils/__tests__/EasCommand-test.ts
@@ -42,7 +42,7 @@ beforeEach(() => {
   jest.mocked(createAnalyticsAsync).mockResolvedValue(analytics);
 });
 
-const createTestEasCommand = () => {
+const createTestEasCommand = (): any => {
   class TestEasCommand extends EasCommand {
     override async runAsync(): Promise<void> {}
   }


douglowder:
@EvanBacon I fixed merge conflicts, and found a change was needed in Jest setup to fix failing tests.

packages/eas-cli/src/commandUtils/flags.ts
```diff
@@ -1,4 +1,9 @@
 import { Flags } from '@oclif/core';
+import { boolish } from 'getenv';
+
+export function isNonInteractiveByDefault(): boolean {
+  return boolish('CI', false) || !process.stdin.isTTY;
```
brentvatne, line 5: should we use `detectAgent()`? https://www.npmjs.com/package/@vercel/detect-agent

douglowder: That's a good suggestion for the future -- Claude also suggested taking a look at https://www.npmjs.com/package/ci-info for this check.

PR #3491, opened 2026-03-10
[build-tools] - support Xcode DerivedData pod products caching for iOS builds

AbbanMustafa:
# Why

 iOS builds currently recompile all CocoaPods from scratch on every build. Unlike local development where Xcode's DerivedData persists compiled products between builds, CI builds start with a clean DerivedData every time. 
  
  The xcodebuild step spends the majority of its time rebuilding pod dependencies that haven't changed between builds. This adds Xcode build product caching to skip that redundant recompilation. Gated behind `EAS_XCODE_CACHE=1`.
  
# How
Cache pre-built pod products (.a l ibraries, .modulemap files, headers) from DerivedData after a successful build, keyed by lockfile hash. On subsequent builds, restore the cached products and patch the Xcodeproject to skip pod recompilation.

**Restore (RESTORE_CACHE)**: Extract cached products to `/tmp/pods-cache/ `and rewrite absolute paths in `.modulemap` files. Products live outside DerivedData because `xcodebuild archive` wipes `ArchiveIntermediates/` at the start of each run.

**Patch (INSTALL_PODS, after pod install):** Placing pre-built products on disk isn't enough. Xcode still sees pod targets in `Pods.xcodeproj` and will rebuild them. Provide a Ruby script removes non-umbrella pod targets from
   the project entirely, then patches `PODS_CONFIGURATION_BUILD_DIR` in the xcconfig files to point at `/tmp/pods-cache/ `so the linker resolves to cached .a files instead of empty DerivedData.

  **Save (SAVE_CACHE):** Locate build products in `ArchiveIntermediates/.../BuildProductsPath/Release-*` (archive) or `Build/Products/Release-*` (simulator) and upload. derived_data_path is set to ./build in the Gymfile so
   products land in a known location. Each save replaces the entire cache, so stale products from removed pods are naturally pruned.

Cache key is `ios-xcode-cache-{device|sim}-{lockfile_hash}` — separate keys for device (Release-iphoneos, arm64) and simulator (Debug-iphonesimulator, x86_64/arm64) builds to prevent architecture mismatches.

### Performance: Archive builds go from 🙌~3 min to ~30 sec on cache hit (6x speedup)🙌

# Test Plan

  - Ran iOS archive build with XCODE_CACHE=1 — first build misses, saves cache after completion
  - Ran second build — cache hit, pod targets removed from xcodeproj, xcodebuild skips recompilation
  - Ran simulator build — uses separate sim cache key, no architecture mismatch with device cache
  - Ran without XCODE_CACHE=1 — all cache phases no-op
  - Verified modulemap paths are correctly rewritten to /tmp/pods-cache/
  
  
Screenshot 2026-03-17 at 1 26 59 AM


AbbanMustafa:
Screenshot 2026-03-18 at 1 12 10 AM

packages/build-tools/src/utils/xcodeCacheKey.ts
```diff
@@ -0,0 +1,23 @@
...
+import { hashFiles } from '@expo/steps';
+import path from 'path';
+
+import { findPackagerRootDir } from './packageManager';
+
+export const XCODE_CACHE_KEY_PREFIX = 'ios-xcode-cache-';
+
+export async function generateXcodeCacheKeyAsync(
+  workingDirectory: string,
+  simulator?: boolean
```
sjchmiela, line 11: Is simulator cache helpful for device build and/or vice versa? It seems it's not based on PR descirption. Let's make the parameter required and an inline enumerator `"simulator" | "device"` or sth like that.

packages/build-tools/src/utils/xcodeCacheKey.ts
```diff
@@ -0,0 +1,23 @@
+import * as PackageManagerUtils from '@expo/package-manager';
+import { hashFiles } from '@expo/steps';
+import path from 'path';
+
+import { findPackagerRootDir } from './packageManager';
+
+export const XCODE_CACHE_KEY_PREFIX = 'ios-xcode-cache-';
```
sjchmiela, line 7: Let's not export it like that if it should never restore "just" `ios-xcode-cache-` cache?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +201,191 @@ export async function restoreCcacheAsync({
...
+    const replaced = content.replace(/"[^"]*\/Release-[a-z]+\//g, `"${cacheDir}/`);
+    if (replaced !== content) {
+      await fs.promises.writeFile(filePath, replaced);
+      patchCount++;
+    }
+  }
+  logger.info(`Rewrote paths in ${patchCount} modulemap files`);
+}
+
+const PATCH_RUBY_SCRIPT = `
```
sjchmiela, line 293: Is this achievable using Expo config plugins code?

sjchmiela:
I'll also let myself add @lukmccall as a reviewer — he has much more experience in the area.

I also wonder how does https://github.com/software-mansion/rnrepo do it for the recently added iOS support. If this pull request's main thing is to cache Pod artifacts and they do build Pod artifacts for reuse…

packages/build-tools/src/utils/xcodeCacheKey.ts
```diff
@@ -0,0 +1,23 @@
...
+  workingDirectory: string,
+  simulator?: boolean
+): Promise<string> {
+  const packagerRunDir = findPackagerRootDir(workingDirectory);
+  const manager = PackageManagerUtils.createForProject(packagerRunDir);
+  const lockPath = path.join(packagerRunDir, manager.lockFile);
+  const variant = simulator ? 'sim' : 'device';
+
+  try {
+    return `${XCODE_CACHE_KEY_PREFIX}${variant}-${hashFiles([lockPath])}`;
```
lukmccall, line 19: The lockfile probably doesn't change if we change the XCode version. So, we probably need to include it in the key.

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +201,191 @@ export async function restoreCcacheAsync({
...
+require 'xcodeproj'
+
+project = Xcodeproj::Project.open('ios/Pods/Pods.xcodeproj')
```
lukmccall, line 296: Consider searching for the xcodeproj rather than relying on the hardcoded path.

sjchmiela:
> Are we sure that we're not leaking any personal information by caching DerivedData? I'm not sure if this directory was intended to be cached.

Oh so this is not going to be public, this cache is scoped per-app

PR #3493, opened 2026-03-11
[eas cli] Builds: wait before showing the concurrency limit message

sswrk:
# Why

This issue: https://linear.app/expo/issue/ENG-19399/fix-concurrency-upsell-showing-when-running-builds-on-orchestrator

The current behavior is correct for Turtle builds (without orchestrator) – they appeared `IN_QUEUE` as soon as they were created unless the concurrency limit was indeed reached. It's different with the orchestrator – new builds need to be picked up by the "enqueue" task that runs periodically, so they stay in the `NEW` status for a few seconds even if the concurrency limit isn't reached.

# How

Added a grace period (15s) for the "NEW" status display in the CLI. After this times out, the concurrency limit message is shown.

# Test Plan

Tested this with staging backend (it uses the orchestrator):

https://github.com/user-attachments/assets/c85323d9-b115-477c-bb73-ceaf440433e3

### DIFF
--- a/packages/eas-cli/src/build/build.ts
+++ b/packages/eas-cli/src/build/build.ts
@@ -485,10 +485,6 @@ async function handleSingleBuildProgressAsync(
     spinner.start('Build is about to start');
   }
 
-  if (statusNewSetAt !== null && build.status !== BuildStatus.New) {
-    statusNewSetAt = null;
-  }
-
   switch (build.status) {
     case BuildStatus.Finished:
       spinner.succeed('Build finished');
@@ -599,10 +595,8 @@ async function handleMultipleBuildsProgressAsync(
       ].includes(build.status)
     ).length === buildCount;
 
-  if (someNew) {
-    statusNewSetAt ??= Date.now();
-  } else {
-    statusNewSetAt = null;
+  if (someNew && statusNewSetAt === null) {
+    statusNewSetAt = Date.now();
   }
 
   if (allSettled) {


packages/eas-cli/src/build/build.ts
```diff
@@ -455,6 +455,9 @@ const queueProgressBar = new cliProgress.SingleBar(
   cliProgress.Presets.rect
 );
 
+let newStatusTimestamp: number | null = null;
```
sjchmiela: nit: rename to `statusNewSetAt` or sth like that? this way we:
- don't say "new status" which is confusing
- "set at" also makes it more clear

packages/eas-cli/src/build/build.ts
```diff
@@ -482,15 +485,29 @@ async function handleSingleBuildProgressAsync(
     spinner.start('Build is about to start');
   }
 
+  if (newStatusTimestamp !== null && build.status !== BuildStatus.New) {
```
sjchmiela: do we need this?

packages/eas-cli/src/build/build.ts
```diff
@@ -482,15 +485,29 @@ async function handleSingleBuildProgressAsync(
...
-        formatAccountBillingUrl(accountName)
-      )}.`;
+    case BuildStatus.New: {
+      const now = Date.now();
+      if (newStatusTimestamp === null) {
+        newStatusTimestamp = now;
+      }
+      const newStatusDurationMs = now - newStatusTimestamp;
+      if (newStatusDurationMs < NEW_STATUS_GRACE_PERIOD_MS) {
+        spinner.text = 'Waiting for build to be queued...';
```
sjchmiela:
```suggestion
        spinner.text = 'Waiting for build to get enqueued…';
```

packages/eas-cli/src/build/build.ts
```diff
@@ -482,15 +485,29 @@ async function handleSingleBuildProgressAsync(
...
       spinner.succeed('Build finished');
       return { refetch: false };
-    case BuildStatus.New:
-      spinner.text = `Build concurrency limit reached for your account. Build will enter queue once a concurrency becomes available. Add additional concurrencies at ${link(
-        formatAccountBillingUrl(accountName)
-      )}.`;
+    case BuildStatus.New: {
+      const now = Date.now();
+      if (newStatusTimestamp === null) {
+        newStatusTimestamp = now;
```
sjchmiela: `newStatusTimestamp ??= now`?

packages/eas-cli/src/build/build.ts
sswrk: It's an extra defensive clean-up, but yeah, we don't really need it – status never goes back to NEW, and the CLI process exits when build is done.

PR #3495, opened 2026-03-11
[eas-cli] add skip-build-if-not-found option to eas build:dev

mmichels-brex:
## Summary
- add `--skip-build-if-not-found` to `eas build:dev`
- skip triggering a new build when no successful build exists for the computed fingerprint
- apply the same skip behavior when a matching build exists but its artifacts have expired

## Testing
- CI

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           script: |
             core.setFailed('Please add a changelog entry!')
-      - uses: mshick/add-pr-comment@v2.1.0
+      - uses: mshick/add-pr-comment@v2.8.2
         if: always()
         with:
           message-id: changelog-entry-check


packages/eas-cli/src/commands/build/dev.ts
```diff
@@ -128,10 +134,17 @@ export default class BuildDev extends EasCommand {
         await this.startDevServerAsync({ projectDir, platform });
         return;
       } else {
-        Log.warn('Artifacts for this build expired. New build will be started.');
```
mmichels-brex, line 131: >  New build will be started

"new build" info is already covered in [L144](https://github.com/expo/eas-cli/pull/3495/files#diff-3fc58de2cc119317f24abb2ef15a125966b5f2bb71f5fe21f6bc8fab2f8639c6R144)

mmichels-brex:
> Looks good, could you add a changelog entry ?

Just did!

PR #3498, opened 2026-03-12
[build-tools] Store per-attempt results for Maestro test retries

hSATAC:
# Why

Maestro test retries currently overwrite JUnit reports and device logs on each attempt, so only the final attempt's results are stored. This means we lose visibility into earlier failed attempts, making it harder to diagnose flaky tests.

Depends on: expo/universe#25691 and expo/universe#25692 (must be deployed first — the new unique index on `(turtle_job_run_id, path, retry_count)` is required to accept multiple results per test path).

# How

  Three changes across `build-tools`:

  1. **`internalMaestroTest.ts`** — Move `outputPath` and device log path inside the retry loop with an `attempt-N` suffix, so each attempt produces a separate
  file instead of overwriting. Removed the now-unnecessary `fs.promises.rm` call.

  2. **`maestroResultParser.ts`** — Extract an internal `parseJUnitFile` helper, then rewrite the merge logic in `parseMaestroResults` to group JUnit results by
   flow name. When multiple files exist for the same flow, the attempt index is extracted from the filename (`attempt-N`). For single-file flows (backward
  compat with `reuse_devices: true` old format), `retryCount` is still derived from `ai-*.json` occurrence count.

  3. **`reportMaestroTestResults.ts`** — Update the duplicate-name guard to check `(name, retryCount)` pairs instead of just names. Same name with different
  `retryCount` is expected (per-attempt results), while same `(name, retryCount)` indicates a real naming conflict.

# Test Plan

  - Added test for per-attempt report filenames in `internalMaestroTest.test.ts`
  - Added 4 tests in `maestroResultParser.test.ts`: per-attempt results, `reuse_devices=true` per-attempt, sharding with multiple testsuites, and backward compat with old single JUnit file
  - Added test in `reportMaestroTestResults.test.ts` for same name with different `retryCount` being sent (not skipped)
  - All 49 tests across the 3 test files pass

### DIFF
--- a/packages/build-tools/src/steps/functions/internalMaestroTest.ts
+++ b/packages/build-tools/src/steps/functions/internalMaestroTest.ts
@@ -214,18 +214,17 @@ export function createInternalEasMaestroTestFunction(ctx: CustomBuildContext): B
       for (const [flowIndex, flowPath] of flowPathsToExecute.entries()) {
         stepCtx.logger.info('');
 
-        // If output_format is empty or noop, we won't use this.
-        const outputPath = path.join(
-          maestroReportsDir,
-          [
-            `${output_format ? output_format + '-' : ''}report-flow-${flowIndex + 1}`,
-            MaestroOutputFormatToExtensionMap[output_format ?? 'noop'],
-          ]
-            .filter(Boolean)
-            .join('.')
-        );
-
         for (let attemptCount = 0; attemptCount < retries; attemptCount++) {
+          // Generate unique report path per attempt (not overwritten on retry)
+          const outputPath = path.join(
+            maestroReportsDir,
+            [
+              `${output_format ? output_format + '-' : ''}report-flow-${flowIndex + 1}-attempt-${attemptCount}`,
+              MaestroOutputFormatToExtensionMap[output_format ?? 'noop'],
+            ]
+              .filter(Boolean)
+              .join('.')
+          );
           const localDeviceName = `eas-simulator-${flowIndex}-${attemptCount}` as
             | IosSimulatorName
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/steps/functions/internalMaestroTest.ts
```diff
@@ -276,12 +275,11 @@ export function createInternalEasMaestroTestFunction(ctx: CustomBuildContext): B
           if (logsResult?.ok) {
             try {
               const extension = path.extname(logsResult.value.outputPath);
-              const destinationPath = path.join(deviceLogsDir, `flow-${flowIndex}${extension}`);
+              const destinationPath = path.join(
+                deviceLogsDir,
+                `flow-${flowIndex}-attempt-${attemptCount}${extension}`
```
sjchmiela, line 280: This is technically a breaking change for the `reuse_devices: false` branch in regards to where / what JUnit test reports are available, right?

hSATAC: I think it should be fine.

Both the producer (`createInternalEasMaestroTestFunction`) and the consumer (`parseMaestroResults` / `reportMaestroTestResults`) are resolved and executed on the worker from the same @expo/build-tools version, so there shouldn’t be any version skew between them.

The parser also supports both formats: single-file runs fall back to counting `ai-*.json`, while multi-file runs read the `attempt-N` from filenames.

Worst case, if parsing fails `reportMaestroTestResults` just skips reporting — it doesn’t affect the Maestro execution or pass/fail result; we’d only miss recording the test case results.

PR #3499, opened 2026-03-12
[build-tools] report app and build version from final archive

kadikraman:
## Why

For projects that don't use CNG, the app version and build number can be evaluated dynamically in the native code. However we detect and save the value when the build is first created (so the actual resolved value is never shown).

Screenshot 2026-03-12 at 13 26 31

See https://exponent-internal.slack.com/archives/C02123T524U/p1772205677890089

## How

After the build is finished, read the actual versions for app and build from the archive and updater the eas metadata.

## Test Plan

Unit and integration tests.

Also tested locally when possible:
- **APK**: `aapt2 dump badging ~/Downloads/application-ba69caff-3ddd-45b3-b766-c74eed33aea3.apk`  (& parse versionName / versionCode from output)
- **AAB**: `bundletool dump manifest --bundle ~/Downloads/application-083f9788-b177-4094-8f06-7b52eb2ef709.aab`  (& parse versionName / versionCode from output)
- **IPA**: use existing `readIpaInfoAsync` (NOT tested locally)
- **iOS Simulator app**: `plutil -convert xml1 -o - <path-to>.app/Info.plist | grep -A1 'CFBundleShortVersionString\|CFBundleVersion'`

Screenshot 2026-03-12 at 16 14 11

### DIFF
--- a/packages/build-tools/src/steps/functions/__tests__/reportResolvedVersion.test.ts
+++ b/packages/build-tools/src/steps/functions/__tests__/reportResolvedVersion.test.ts
@@ -1,3 +1,4 @@
+import { spawnAsync } from '@expo/steps';
 import fs from 'node:fs';
 import path from 'node:path';
 
@@ -6,8 +7,20 @@ jest.unmock('node:fs');
 jest.unmock('fs/promises');
 jest.unmock('node:fs/promises');
 
+jest.mock('@expo/steps', () => ({
+  ...jest.requireActual('@expo/steps'),
+  spawnAsync: jest.fn(),
+}));
+
+const mockedSpawnAsync = jest.mocked(spawnAsync);
+
 import { parseInfoPlistBuffer } from '../readIpaInfo';
-import { parseAaptOutput, parseManifestXml } from '../reportResolvedVersion';
+import {
+  extractAndroidVersionAsync,
+  extractIosVersionAsync,
+  parseAaptOutput,
+  parseManifestXml,
+} from '../reportResolvedVersion';
 
 describe(parseAaptOutput, () => {
   it('extracts versionName and versionCode from aapt2 output', () => {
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/steps/functions/reportResolvedVersion.ts
```diff
@@ -0,0 +1,235 @@
...
+  const versionNameMatch = output.match(/versionName='([^']+)'/);
+  const versionCodeMatch = output.match(/versionCode='([^']+)'/);
+
+  return {
+    appVersion: versionNameMatch?.[1],
+    appBuildVersion: versionCodeMatch?.[1],
+  };
+}
+
+export function parseManifestXml(xml: string): { appVersion?: string; appBuildVersion?: string } {
```
sjchmiela, line 155: Ash added XML parser to `build-tools` recently if this was something you'd like to use.

packages/build-tools/src/steps/functions/reportResolvedVersion.ts
```diff
@@ -0,0 +1,235 @@
...
+  if (iosJob.simulator) {
+    return await extractSimulatorAppVersionAsync(iosJob, workingDirectory, logger);
+  }
+
+  const artifactPattern = iosJob.applicationArchivePath ?? 'ios/build/*.ipa';
+  const artifacts = await findArtifacts({
+    rootDir: workingDirectory,
+    patternOrPath: artifactPattern,
+    logger,
+  });
```
sjchmiela: Instead of looking for artifacts again, could we use value from the previous step?

Right above these functions we usually do `createFindAndUploadBuildArtifactsBuildFunction`. We could add output to it that would point to the first (?) uploaded application archive. Then you would use that output as an input to this, sort of like we do in https://github.com/expo/eas-cli/blob/789cf62bd6d7b0b4173ad3e790426ac6807b9731/packages/build-tools/src/steps/functionGroups/build.ts#L207-L210.

packages/build-tools/src/steps/functions/reportResolvedVersion.ts
kadikraman: Yeah that's better, thanks!

packages/build-tools/src/utils/artifacts.ts
```diff
@@ -136,6 +136,8 @@ export async function uploadApplicationArchive(
     },
     logger,
   });
+
+  return applicationArchives[0];
```
sjchmiela, line 140: nit: let's return all archives and pick only the first one in the consumer?

packages/build-tools/src/steps/functions/reportResolvedVersion.ts
```diff
@@ -0,0 +1,216 @@
...
+            ) {
+              id
+            }
+          }
+        }
+      `),
+      {
+        buildId,
+        appVersion: appVersion ?? null,
+        appBuildVersion: appBuildVersion ?? null,
```
sjchmiela, line 208: honest question — if `appVersion` is missing here, but it _is_ present in a metadata, is this going to clear the `appVersion` or leave it as it is?

packages/build-tools/src/steps/functions/reportResolvedVersion.ts
```diff
@@ -0,0 +1,216 @@
...
+
+export async function reportResolvedVersionAsync(
+  graphqlClient: Client,
+  buildId: string,
+  {
+    appVersion,
+    appBuildVersion,
+  }: {
+    appVersion?: string;
+    appBuildVersion?: string;
```
sjchmiela, line 181:
```suggestion
    appVersion: string | undefined;
    appBuildVersion: string | undefined;
```
Less easy to forget

packages/build-tools/src/builders/ios.ts
```diff
@@ -175,14 +179,32 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
...
-    await uploadApplicationArchive(ctx, {
+  const archivePath = await ctx.runBuildPhase(BuildPhase.UPLOAD_APPLICATION_ARCHIVE, async () => {
+    return await uploadApplicationArchive(ctx, {
       patternOrPath: resolveArtifactPath(ctx),
       rootDir: ctx.getReactNativeProjectDirectory(),
       logger: ctx.logger,
     });
   });
 
+  if (archivePath) {
```
sjchmiela, line 190: We need it to be inside a build phase, otherwise if the users are oging to see "Unnamed build phase" logs which I've seen people complain about before. Maybe we can move it inside upload application archive?

packages/build-tools/src/builders/android.ts
```diff
@@ -174,14 +178,32 @@ async function buildAsync(ctx: BuildContext<Android.Job>): Promise<void> {
...
-    await uploadApplicationArchive(ctx, {
+  const archivePath = await ctx.runBuildPhase(BuildPhase.UPLOAD_APPLICATION_ARCHIVE, async () => {
+    return await uploadApplicationArchive(ctx, {
       patternOrPath: ctx.job.applicationArchivePath ?? 'android/app/build/outputs/**/*.{apk,aab}',
       rootDir: ctx.getReactNativeProjectDirectory(),
       logger: ctx.logger,
     });
   });
 
+  if (archivePath) {
```
sjchmiela, line 189: same

PR #3510, opened 2026-03-17
[build-tools] support Gradle build cache for Android builds

AbbanMustafa:
# Why

  Android builds currently recompile all Kotlin/Java code from scratch on every build. We already persist ccache between builds to speed up native C/C++ compilation (NDK), but ccache doesn't help with the Gradle/Kotlin/Java layer — which is where the majority of Android build time is spent. This adds Gradle build cache support to close that gap, leading to more optimal Android builds.

  
# How

  All behavior is gated behind `GRADLE_CACHE=1` — when unset, everything is a no-op.

  Cache key is android-gradle-cache-{lockfile_hash}, derived from the project's package manager lockfile. Changing dependencies invalidates the cache; code-only changes reuse it. On miss, falls back to prefix match
   (android-gradle-cache-) to find the closest available cache.

  Two new build phases are added to the Android builder:

```
  RESTORE_CACHE              (ccache)
  RESTORE_GRADLE_CACHE       ← new
  POST_INSTALL_HOOK
  ...
  RUN_GRADLEW                (now with --build-cache flag)
  ...
  UPLOAD_APPLICATION_ARCHIVE
  SAVE_GRADLE_CACHE          ← new
  SAVE_CACHE                 (ccache)
```

  Restore downloads and extracts the cache archive into `~/.gradle/caches/`. Save compresses and uploads two specific subdirectories from `~/.gradle/caches/:`
```
  ┌───────────────┬───────────────────────────────────────────────────────┬────────────────────┐
  │   Directory   │                       Contents                        │ Precompressed size │
  ├───────────────┼───────────────────────────────────────────────────────┼────────────────────┤
  │ build-cache-1 │ Gradle build cache — compiled classes, dex, resources │ ~130 MB            │
  ├────────────────────────────────────────────────────────────────────────────────────────────┤
```
 We selectively cache `build-cache-1` rather than the full ~/.gradle/caches/ directory. This subdirectory covers the highest-impact layers: compiled task outputs (classes, dex, resources). The remaining contents — transforms, journal, configuration-cache, etc. — are either derived from modules-2 (transforms are regenerated from cached dependencies on
  first build) or are Gradle bookkeeping with negligible build time impact.
  
  

# Test Plan

Validated locally, observed build time improvement of 4min -> 40sec

  - Ran Android build with GRADLE_CACHE=1 — first build misses, saves cache after completion
  - Ran second build — restores cache, skips save, ~130/386 Gradle tasks served FROM-CACHE
  - Ran across variants (assembleDebug, assembleRelease, bundleRelease) — all hit on the same cache
  - Ran without GRADLE_CACHE=1 — all cache phases no-op, no --build-cache flag passed

Screenshot 2026-03-16 at 7 01 22 PM

Screenshot 2026-03-17 at 12 33 55 AM

### DIFF
--- a/packages/build-tools/src/steps/functions/restoreBuildCache.ts
+++ b/packages/build-tools/src/steps/functions/restoreBuildCache.ts
@@ -228,7 +228,7 @@ export async function restoreGradleCacheAsync({
         '        settings.caches {',
         '            cleanup = Cleanup.ALWAYS',
         '            buildCache {',
-        '                setRemoveUnusedEntriesAfterDays(7)',
+        '                setRemoveUnusedEntriesAfterDays(3)',
         '            }',
         '        }',
         '        println "[EAS] Configured Gradle cache cleanup via init script"',


packages/build-tools/src/steps/functions/restoreGradleCache.ts
```diff
@@ -0,0 +1,90 @@
...
+import os from 'os';
+import path from 'path';
+
+import { downloadCacheAsync } from './restoreCache';
+import { GRADLE_CACHE_KEY_PREFIX, generateGradleCacheKeyAsync } from '../../utils/gradleCacheKey';
+import { TurtleFetchError } from '../../utils/turtleFetch';
+
+const GRADLE_CACHES_DIR = '.gradle/caches';
+
+export function createRestoreGradleCacheFunction(): BuildFunction {
```
sjchmiela: I don't see reason to add a new function? What is your reasoning re: having a single vs having multiple cache functions?

We talked before about introducing config like
```yaml
cache: true

cache:
  node_modules: true
  xcode: false
  gradle: false
```

or sth like that to the workflow YAML. If we had cache config in one object I don't think it makes sense to require user to set up different functions in custom jobs?

Shouldn't this be part of `{restore,save}_build_cache`?

packages/build-tools/src/android/gradle.ts
```diff
@@ -31,19 +31,24 @@ export async function runGradleCommand(
   logger.info(`Running 'gradlew ${gradleCommand}' in ${androidDir}`);
   await fs.chmod(path.join(androidDir, 'gradlew'), 0o755);
   const verboseFlag = ctx.env['EAS_VERBOSE'] === '1' ? '--info' : '';
+  const buildCacheFlag = ctx.env['GRADLE_CACHE'] === '1' ? '--build-cache' : '';
```
sjchmiela: So sad Gradle can't be ordered to use build cache with an environment variable…

packages/build-tools/src/steps/functions/restoreGradleCache.ts
```diff
@@ -0,0 +1,90 @@
...
+      paths: ['gradle-caches'],
+      key: cacheKey,
+      keyPrefixes: [GRADLE_CACHE_KEY_PREFIX],
+      platform: Platform.ANDROID,
+    });
+
+    await fs.promises.mkdir(gradleCachesPath, { recursive: true });
+
+    const { execSync } = require('child_process');
+    execSync(`tar xzf ${archivePath} -C ${gradleCachesPath}`);
```
sjchmiela: AI slop?

packages/build-tools/src/steps/functions/saveGradleCache.ts
```diff
@@ -0,0 +1,120 @@
...
+import nullthrows from 'nullthrows';
+import os from 'os';
+import path from 'path';
+
+import { compressCacheAsync, uploadCacheAsync } from './saveCache';
+import { formatBytes } from '../../utils/artifacts';
+import { generateGradleCacheKeyAsync } from '../../utils/gradleCacheKey';
+
+const GRADLE_CACHES_DIR = '.gradle/caches';
+const CACHE_SUBDIRS = ['build-cache-1', 'modules-2'];
```
sjchmiela: I think we shouldn't cache `modules-2`. As you mentioned, those are Maven/Gradle artifacts and dependencies. To help with that we have the Maven cache proxy. If we start caching these for all builds we'll go from storing one artifact in the Maven cache for all the builds to use to storing one artifact for every build separately. This is going to inflate network bandwidth and cache size a lot.

Can you confirm if caching just the `build-cache-1` helps too? https://docs.gradle.org/current/userguide/build_cache.html does not mention `modules-2`, it does however talk about `build-cache`.

packages/build-tools/src/steps/functions/saveGradleCache.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const jobId = nullthrows(env.EAS_BUILD_ID, 'EAS_BUILD_ID is not set');
+    const robotAccessToken = nullthrows(
+      secrets?.robotAccessToken,
+      'Robot access token is required for cache operations'
+    );
+    const expoApiServerURL = nullthrows(env.__API_SERVER_URL, '__API_SERVER_URL is not set');
+
+    logger.info(`Compressing Gradle caches (${existingDirs.join(', ')})...`);
+    const { archivePath } = await compressCacheAsync({
```
sjchmiela: What about configuring cache pruning?

https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home:configure_cache_cleanup

packages/build-tools/src/android/gradle.ts
sjchmiela: `EAS_GRADLE_CACHE`?

AbbanMustafa: ah yes we can enable with gradle.properties

sjchmiela: Oh I didn't mean `gradle.properties` would necessarily be better. I like the explicitness of `--build-cache` next to `./gradlew`. Or maybe it would? Can we put `gradle.properties` anywhere, or only in the project? https://github.com/expo/eas-cli/blob/25d945b63fb0b2d6f920933b9a07da12c78136f4/packages/worker/src/external/analytics.ts#L144 Would just appending `\n...` be sufficient?

packages/build-tools/src/steps/functions/saveGradleCache.ts
AbbanMustafa: I saw speed benefits when restoring modules-2, but in terms of the gradle's reported cache hit, it performs the same. So yes I was just observing locally the benefits of what our maven repo provides for the worker. So I will remove `modules-2 `

packages/build-tools/src/android/gradle.ts
AbbanMustafa: yes we can do so via the project properties. I write to the file if we find the property has not been set https://github.com/expo/eas-cli/pull/3510/changes/daa38736f845ab109181cdc290e861daf7237664#diff-db8c2375feedbfe1a1b25cbc4ef8375f0d592509438f4babcc5db95d6f28cdc1R57-R69

packages/build-tools/src/steps/functions/restoreGradleCache.ts
AbbanMustafa: I don't feel strongly about separate functions, can definitely fold this into `{restore,save}_build_cache`. I mainly wanted this PR to land the functionality so we can get feedback on how this works in prod, and then follow up with this and Xcode caches to set up introducing the proper config as a next step.

AbbanMustafa:
with EAS_GRADLE_CACHE: '1', consolidate caches in restore/save step
Screenshot 2026-03-17 at 6 01 14 PM


with no or EAS_GRADLE_CACHE: '0', no-op

Screenshot 2026-03-17 at 6 18 05 PM

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +192,82 @@ export async function restoreCcacheAsync({
...
+  logger,
+  workingDirectory,
+  env,
+  secrets,
+}: {
+  logger: bunyan;
+  workingDirectory: string;
+  env: Record<string, string | undefined>;
+  secrets?: { robotAccessToken?: string };
+}): Promise<boolean> {
```
sjchmiela:
```suggestion
}): Promise<{ restored: boolean }> {
```
or `cacheHit` or sth like that?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +192,82 @@ export async function restoreCcacheAsync({
...
+  if (!properties.some(p => p.type === 'property' && p.key === 'org.gradle.cache.cleanup')) {
+    properties.push({ type: 'property', key: 'org.gradle.cache.cleanup', value: 'ALWAYS' });
+    modified = true;
+  }
+  if (modified) {
+    await fs.promises.writeFile(
+      gradlePropertiesPath,
+      AndroidConfig.Properties.propertiesListToString(properties)
+    );
+  }
```
sjchmiela: Should we allow custom properties for `org.gradle.caching` and `org.gradle.cache.cleanup` if a user uses `EAS_GRADLE_CACHE`? To put it differently, shouldn't `EAS_GRADLE_CACHE` override whatever the user has set?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +192,82 @@ export async function restoreCcacheAsync({
...
+    secrets?.robotAccessToken,
+    'Robot access token is required for cache operations'
+  );
+  const expoApiServerURL = nullthrows(env.__API_SERVER_URL, '__API_SERVER_URL is not set');
+
+  try {
+    const cacheKey = await generateGradleCacheKeyAsync(workingDirectory);
+    logger.info(`Restoring Gradle cache key: ${cacheKey}`);
+
+    const jobId = nullthrows(env.EAS_BUILD_ID, 'EAS_BUILD_ID is not set');
```
sjchmiela: Why do we do `nullthrows` sometimes inside `try` and sometimes outside `try`?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +192,82 @@ export async function restoreCcacheAsync({
...
+      expoApiServerURL,
+      robotAccessToken,
+      paths: ['gradle-caches'],
+      key: cacheKey,
+      keyPrefixes: [GRADLE_CACHE_KEY_PREFIX],
+      platform: Platform.ANDROID,
+    });
+
+    await fs.promises.mkdir(gradleCachesPath, { recursive: true });
+    await spawn('tar', ['xzf', archivePath, '-C', gradleCachesPath], { logger });
```
sjchmiela: Why can't we use the regular cache utilities?

packages/build-tools/src/steps/functions/saveBuildCache.ts
```diff
@@ -135,3 +150,91 @@ export async function saveCcacheAsync({
...
+  env: Record<string, string | undefined>;
+  secrets?: { robotAccessToken?: string };
+  cacheHit: boolean;
+}): Promise<void> {
+  if (env.EAS_GRADLE_CACHE !== '1') {
+    return;
+  }
+
+  if (cacheHit) {
+    logger.info('Gradle cache was restored — skipping save');
```
sjchmiela: Why are we skipping save if restored?

packages/build-tools/src/steps/functions/saveBuildCache.ts
```diff
@@ -135,3 +150,91 @@ export async function saveCcacheAsync({
...
+  const gradleCachesPath = path.join(os.homedir(), '.gradle', 'caches');
+
+  try {
+    await fs.promises.access(gradleCachesPath);
+  } catch {
+    logger.warn('No Gradle caches directory found, skipping save');
+    return;
+  }
+
+  const existingDirs = [];
```
sjchmiela: can we simplify?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +192,82 @@ export async function restoreCcacheAsync({
...
+    const gradleCachesPath = path.join(os.homedir(), '.gradle', 'caches');
+
+    const { archivePath, matchedKey } = await downloadCacheAsync({
+      logger,
+      jobId,
+      expoApiServerURL,
+      robotAccessToken,
+      paths: ['gradle-caches'],
+      key: cacheKey,
+      keyPrefixes: [GRADLE_CACHE_KEY_PREFIX],
```
sjchmiela: This does not match `compressCacheAsync` in `saveBuildCache` which means `version` won't match and cache will never get restored? Did you test this?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
AbbanMustafa: Yes, I agree if the user opts in then that should set precedent for the properties. Can also tweak  more how the customization looks once this our

AbbanMustafa:
Screenshot 2026-03-24 at 11 34 10 PM

packages/build-tools/src/steps/functions/restoreBuildCache.ts
AbbanMustafa: Its a bit confusing with the parameter naming, but `paths` in `compressCacheAsync` and `paths` in `uploadCacheAsync`/`downloadCacheAsync` are for different purpose. And we just need it consistent between the up/download functions. I believe you meant upload since that is what determines the version, which was the same https://github.com/expo/eas-cli/pull/3510/changes#diff-b282ed96909cf3405e277859d596a3617f65db34f4f22319efcd062a549a3430R207

I've aligned them all now to just have `['build-cache-1']` as path to be consistent with how ccache does it and avoid confusion.

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +191,87 @@ export async function restoreCcacheAsync({
...
+  }
+
+  try {
+    const gradlePropertiesPath = path.join(workingDirectory, 'android', 'gradle.properties');
+    const gradlePropertiesContent = await fs.promises.readFile(gradlePropertiesPath, 'utf-8');
+    const properties = AndroidConfig.Properties.parsePropertiesFile(gradlePropertiesContent);
+
+    const propsToSet: Record<string, string> = {
+      'org.gradle.caching': 'true',
+      'org.gradle.cache.cleanup': 'ALWAYS',
```
sjchmiela: Looks like we'll need to either handle Gradle 9.0 in a dedicated way or move to init script as per https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_org_gradle_cache_cleanup. We want to support Gradle 9 https://exponent-internal.slack.com/archives/C1QP38NQ5/p1772688272258629

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +191,87 @@ export async function restoreCcacheAsync({
...
+        existing.value = value;
+      } else {
+        properties.push({ type: 'property', key, value });
+      }
+    }
+
+    await fs.promises.writeFile(
+      gradlePropertiesPath,
+      AndroidConfig.Properties.propertiesListToString(properties)
+    );
```
sjchmiela: Is this better than ⬇️ ? (sth like)

```suggestion
    await fs.promises.writeFile(
      gradlePropertiesPath,
      `${gradlePropertiesContent}\n\norg.gradle.caching=true\norg.gradle.cache.cleanup=ALWAYS\n`
    );
```
?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -176,6 +191,87 @@ export async function restoreCcacheAsync({
...
+    logger.info(`Restoring Gradle cache key: ${cacheKey}`);
+
+    const gradleCachesPath = path.join(os.homedir(), '.gradle', 'caches');
+
+    const { archivePath, matchedKey } = await downloadCacheAsync({
+      logger,
+      jobId,
+      expoApiServerURL,
+      robotAccessToken,
+      paths: ['build-cache-1'],
```
sjchmiela: Why not use `path.join(gradleCachesPath, 'build-cache-1')`?

packages/build-tools/src/steps/functions/restoreBuildCache.ts
sjchmiela: Also, I think the default expiration times are too high for us?

AbbanMustafa:
Gradle 9 no longer supports the `org.gradle.cache.cleanup` property in gradle.properties, so we provide an init script at `~/.gradle/init.d/eas-cache-cleanup.gradle` to configure cache cleanup programmatically. The script uses the settings.caches {} API to set cleanup = Cleanup.ALWAYS, which runs cleanup at the start of every build, and [setRemoveUnusedEntriesAfterDays(3)](https://docs.gradle.org/current/javadoc/org/gradle/api/cache/CacheResourceConfiguration.html#setRemoveUnusedEntriesAfterDays(int)) to prune build cache entries unused for 3+ days. A try-catch ensures it falls back silently on older Gradle versions that don't support this API.

before
Screenshot 2026-03-26 at 10 43 01 AM

build start
Screenshot 2026-03-26 at 11 32 19 AM

build finish
Screenshot 2026-03-26 at 11 31 55 AM

packages/build-tools/src/steps/functions/restoreBuildCache.ts
AbbanMustafa: Thank you for catching that. I set up this now with an init script https://github.com/expo/eas-cli/pull/3510#issuecomment-4136481454

PR #3533, opened 2026-03-24
fix(@expo/apple-utils): patch getNextUrl() to fix pagination cookie domain mismatch (#3392)

### LINKED ISSUE #3392: Cookie domain mismatch in @expo/apple-utils pagination (developer-mdn.apple.com vs developer.apple.com)
### Summary

The `Cookie not in this host's domain. Cookie:developer-mdn.apple.com Request:developer.apple.com` error continues to occur in eas-cli 16.32.0 with `@expo/apple-utils` v2.1.13. This is the same bug reported in #1672, #2224, and #2881 — all closed without a root-cause fix.

### Root Cause Analysis

The bug is in `ConnectResponse.fetchNextPageAsync()` inside `@expo/apple-utils`.

**How requests normally work:**

1. `ConnectClientAPI.requestAsync()` sets `baseURL: await this.getHostnameAsync(u)` which resolves to `https://developer-mdn.apple.com`
2. Cookies are stored for the `developer-mdn.apple.com` domain
3. Relative URL paths (e.g., `devices`, `profiles`) are resolved against this baseURL — works fine

**How pagination breaks:**

1. Apple's JSON:API responses include `links.next` with **absolute URLs** like `https://developer.apple.com/services-account/v1/...`
2. `ConnectResponse.getNextUrl()` returns this raw absolute URL from `this.data.links.next`
3. `fetchNextPageAsync()` passes this absolute URL to `this.getRequestAsync(this.context, {method:"get", url: e})`
4. Inside `requestAsync`, axios sees the URL is already absolute and **ignores the `baseURL`** (`developer-mdn.apple.com`)
5. The request goes to `developer.apple.com` instead of `developer-mdn.apple.com`
6. `tough-cookie` rejects the `developer-mdn.apple.com` cookies for the `developer.apple.com` domain
7. Build fails with `Cookie not in this host's domain`

**Relevant code path (from minified build):**

```js
// ConnectResponse class
getNextUrl() {
  return this.data?.links?.next ?? null;  // Returns absolute Apple URL
}

async fetchNextPageAsync() {
  const url = this.getNextUrl();
  return url ? this.getRequestAsync(this.context, { method: "get", url }) : null;
  // ^ Passes absolute URL directly — baseURL is ignored by axios
}

// ConnectClientAPI
async requestAsync(context, options, retryOptions) {
  return await this.requestFromAPIAsync(context, {
    baseURL: await this.getHostnameAsync(context),  // developer-mdn.apple.com
    ...options,  // options.url is absolute, so baseURL is ignored
  }, retryOptions);
}
```

### Proposed Fix

In `getNextUrl()`, convert the absolute pagination URL to a relative path so `baseURL` handles domain resolution:

```typescript
getNextUrl() {
  const next = this.data?.links?.next ?? null;
  if (!next) return null;
  try {
    const parsed = new URL(next);
    return parsed.pathname + parsed.search;  // Strip the domain, keep the path
  } catch {
    return next;
  }
}
```

This ensures pagination requests go through the same domain (`developer-mdn.apple.com`) as all other requests, respecting the `baseURL` set in `requestAsync`.

### Environment

```
eas-cli/16.32.0 darwin-arm64 node-v25.2.1
@expo/apple-utils v2.1.13
macOS 15.4 (Darwin 25.2.0)
```

### Error Output

```
✔ Logged in New session
Cookie not in this host's domain. Cookie:developer-mdn.apple.com Request:developer.apple.com
    Error: build command failed.
```

This occurs during `Handling Apple ad hoc provisioning profiles` — authentication succeeds, but the subsequent paginated API call to list devices/profiles fails.

### Steps to Reproduce

```bash
eas build --platform ios --profile development
```

Using Apple ID login (not ASC API key). The error occurs when EAS CLI successfully authenticates but then paginates through devices or provisioning profiles.

### Why This Keeps Recurring

The underlying bug in `ConnectResponse.fetchNextPageAsync()` has never been fixed. Previous issues were resolved by:
- Apple temporarily changing their API to not redirect
- Users switching to ASC API keys as a workaround
- Stale bot closing issues

The `@expo/apple-utils` pagination code needs to normalize URLs to the working domain before making follow-up requests.

adityapsbisht:
## Summary

Fixes the `Cookie not in this host's domain. Cookie:developer-mdn.apple.com Request:developer.apple.com` error that causes iOS builds to fail at the capability/provisioning profile sync step.

Closes #3392

## Root Cause

Apple's paginated API responses include `links.next` with **absolute URLs** like `https://developer.apple.com/services-account/v1/...`.

In `ConnectResponse.fetchNextPageAsync()` inside `@expo/apple-utils`, these absolute URLs are passed directly to axios. When an axios request URL is already absolute, **axios ignores the configured `baseURL`** (`https://developer-mdn.apple.com`). The request hits `developer.apple.com` instead, and `tough-cookie` correctly rejects the `developer-mdn.apple.com` session cookies for that domain → build fails.

## TypeScript Fix Needed in `@expo/apple-utils`

```typescript
// packages/apple-utils/src/... (ConnectResponse class)

// Before
getNextUrl(): string | null {
  return this.data?.links?.next ?? null;
}

// After — strip domain from absolute pagination URLs so axios uses baseURL
getNextUrl(): string | null {
  const next = this.data?.links?.next ?? null;
  if (!next) return null;
  try {
    const parsed = new URL(next);
    return parsed.pathname + parsed.search;
  } catch {
    return next;
  }
}
```

## This PR

Since `@expo/apple-utils` source is not in this repo, this PR adds a `postinstall` script (`scripts/patch-apple-utils-pagination.js`) that applies the same fix to the compiled `build/index.js` after install.

The script:
- Is a no-op if already patched (idempotent)
- Warns gracefully if the pattern is not found (different version)
- Self-documents exactly what TypeScript change is needed upstream

**Remove this script once `@expo/apple-utils` publishes a version containing the fix.**

## Verification

Fix confirmed working by community members in #3392 (reported Feb 2026, confirmed Mar 17 2026).

Error before fix:
```
✖ Failed to sync capability identifiers com.example.app
Cookie not in this host's domain. Cookie:developer-mdn.apple.com Request:developer.apple.com
    Error: build command failed.
```

After fix: capability sync and provisioning profile steps complete successfully.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- [eas-cli] Fix iOS build failure caused by Apple Developer Portal pagination returning absolute URLs that bypass the authenticated session domain, resulting in `Cookie not in this host's domain` errors during capability/provisioning profile sync. ([#3533](https://github.com/expo/eas-cli/pull/3533) by [@adityapsbisht](https://github.com/adityapsbisht))
 - [eas-cli] Fix workflow:logs for builds using built-in EAS build steps. ([#3523](https://github.com/expo/eas-cli/pull/3523) by [@douglowder](https://github.com/douglowder))
 
 ### 🧹 Chores
--- a/packages/eas-cli/scripts/patch-apple-utils-pagination.js
+++ b/packages/eas-cli/scripts/patch-apple-utils-pagination.js
@@ -39,7 +39,15 @@
 const fs = require('fs');
 const path = require('path');
 
-const TARGET = path.join(__dirname, '..', 'node_modules', '@expo', 'apple-utils', 'build', 'index.js');
+const TARGET = path.join(
+  __dirname,
+  '..',
+  'node_modules',
+  '@expo',
+  'apple-utils',
+  'build',
+  'index.js'
+);
 
 const OLD =
   'getNextUrl(){var e,t,r;return null!==(r=null===(t=null===(e=this.data)||void 0===e?void 0:e.links)||void 0===t?void 0:t.next)&&void 0!==r?r:null}';
@@ -60,9 +68,13 @@ if (content.includes(NEW)) {
 }
 
 if (!content.includes(OLD)) {
-  console.warn('[patch-apple-utils] Target pattern not found. @expo/apple-utils may have changed — patch may no longer be needed or needs updating.');
+  console.warn(
+    '[patch-apple-utils] Target pattern not found. @expo/apple-utils may have changed — patch may no longer be needed or needs updating.'
+  );
   process.exit(0);
 }
 
 fs.writeFileSync(TARGET, content.replace(OLD, NEW), 'utf8');
-console.log('[patch-apple-utils] ✔ Applied getNextUrl() pagination domain fix to @expo/apple-utils.');
+console.log(
+  '[patch-apple-utils] ✔ Applied getNextUrl() pagination domain fix to @expo/apple-utils.'
+);


quinlanj:
hey @adityapsbisht , thanks for putting up a fix! i've gone ahead and ported your fix to our internal apple-utils repo so we dont need to have a postinstall script to modify the minified code :) in the meantime, i'll leave this pr open until we've merged the changes upstream and bumped the deps!

quinlanj:
landed https://github.com/expo/eas-cli/pull/3545

christopherwxyz:
Thanks @quinlanj!

PR #3527, opened 2026-03-21
[eas-cli][build-tools][worker] bump @expo/config and @expo/config-plugins

quinlanj:
# Why

Upgrades Expo configuration dependencies to their latest versions to ensure compatibility and access to the latest features and bug fixes.

# How

Updated the following dependencies across multiple packages:
- `@expo/config` from `10.0.6` to `55.0.10`
- `@expo/config-plugins` from `9.0.12` to `55.0.7` 
- `@expo/config-types` from `52.0.1` to `55.0.5`

Added type casting to resolve TypeScript compatibility issues with the updated Expo config types:
- Cast `ctx.exp.notification?.icon` to `Record<string, any>` in build validation
- Cast `appConfig` objects to `Record<string, unknown>` when accessing `newArchEnabled` property in analytics
- Reformatted `spawnAsync` call parameters for better readability

# Test Plan

all tests pass

PR #3536, opened 2026-03-25
[build-tools] Read app config with `expo config` CLI invocation before falling back to `@expo/config`

kitten:
# Why

The `@expo/config` package isn't a public API and isn't stable across SDK releases. When upgrading this package (#3527), we noticed that this package's SDK 55 version has failure paths when used for SDK <=54 projects. This is generally expected, since the config packages aren't meant to be stable across versions, so there's likely not much we can immediately do to remedy that.

Additionally, since Node.js version requirements change and we may not always bump Node versions retroactively, since `@expo/config` or a sub-package might expect a newer version of Node, behaviour of older Node versions with `@expo/config` (or its transitive dependencies) might break.

Instead, the public API for app config reading is the `expo config` command. We should (like in `eas-cli` and `expo-doctor`) try to invoke it first before attempting to fall back to `@expo/config`.

Older ref link for this issue: https://github.com/expo/eas-build/pull/464#pullrequestreview-2434988655

# How

- Add Expo CLI invocation wrapper, similar to others in the `build-tools` codebase
  - **Note:** Since we read `.env` potentially, we have to apply `EXPO_DEBUG=0` as a safe-guard
- Add two code paths to `readAppConfig` to invoke `expo config` before trying to read with `@expo/config`
  - **Note:** I added a warn output when `expo config` fails, but not sure if we surface it properly via the `logger` in all cases
- Update rest of the codebase to account for `readAppConfig` to be async
  - Side-note: We should probably plan for most "external" calls to be async. `getConfig` being synchronous isn't ideal and more accidental than intentional at this point

# Test Plan

- I think this goes through mocks a lot, so unclear how to best test this; that said, this logic is very similar to what we already do in other places

### DIFF
--- a/packages/build-tools/src/utils/__tests__/appConfig.test.ts
+++ b/packages/build-tools/src/utils/__tests__/appConfig.test.ts
@@ -1,0 +1,108 @@
+import { bunyan } from '@expo/logger';
+
+import { readAppConfig } from '../appConfig';
+
+jest.mock('@expo/env', () => ({
+  load: jest.fn(() => ({ FROM_DOTENV: 'true' })),
+}));
+
+jest.mock('@expo/config', () => ({
+  getConfig: jest.fn(() => ({
+    exp: { name: 'fallback-app', slug: 'fallback-app' },
+  })),
+}));
+
+jest.mock('../expoCli');
+
+const { expoCommandAsync } = jest.requireMock('../expoCli') as {
+  expoCommandAsync: jest.Mock;
+};
+
+const { getConfig } = jest.requireMock('@expo/config') as {
+  getConfig: jest.Mock;
+};
+
+const { load: loadEnv } = jest.requireMock('@expo/env') as {
+  load: jest.Mock;
+};
+
+const logger = { warn: jest.fn(), info: jest.fn(), error: jest.fn() } as unknown as bunyan;
+
+const baseParams = {
+  projectDir: '/project',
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/utils/appConfig.ts
```diff
@@ -3,30 +3,79 @@ import { Env } from '@expo/eas-build-job';
...
+  try {
+    parsed = JSON.parse(result.stdout);
+  } catch {
+    throw new Error(
+      `Failed to parse JSON output from 'expo config'.\nOutput: ${result.stdout.slice(0, 500)}`
+    );
+  }
+
+  if (!('exp' in parsed)) {
+    throw new Error(`Unexpected output from 'expo config': missing 'exp' field.`);
```
sjchmiela, line 59: We can use `SystemError` class to get the builds waived automatically if we want to.

kitten: They wouldn't propagate, since we fall back to the `@expo/config` logic. I think we can still re-wrap this, but it's hard to tell if this will always be our fault (?)
I guess it's cheap enough to waive, but invalid app configs could fail this too

kitten:
Test staging build ID: fa469497-7541-4d2b-8927-7edf15a2d891 (staging deploy: https://github.com/expo/eas-cli/actions/runs/23548061545/job/68553924302)

PR #3541, opened 2026-03-25
[steps] Use counter-based generated step ids

sjchmiela:
## Summary
- replace UUID-based autogenerated step ids with a counter in `BuildStep.getNewId()`
- keep explicit user-defined ids unchanged
- rename the step-id test helper so tests no longer talk about UUIDs

## Testing
- yarn --cwd packages/steps test --runInBand BuildStep-test.ts BuildFunction-test.ts BuildConfigParser-test.ts StepsConfigParser-test.ts AbstractConfigParser-test.ts
- yarn --cwd packages/steps typecheck

### DIFF
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -143,7 +143,7 @@ export class BuildStep extends BuildStepOutputAccessor {
   protected executed = false;
 
   public static getNewId(userDefinedId?: string): string {
-    return userDefinedId ?? `step-${BuildStep.nextGeneratedId++}`;
+    return userDefinedId ?? `step-${String(BuildStep.nextGeneratedId++).padStart(3, '0')}`;
   }
 
   constructor(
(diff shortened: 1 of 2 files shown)


packages/steps/src/BuildStep.ts
```diff
@@ -143,7 +143,7 @@ export class BuildStep extends BuildStepOutputAccessor {
   protected executed = false;
 
   public static getNewId(userDefinedId?: string): string {
-    return userDefinedId ?? uuidv4();
+    return userDefinedId ?? `step-${BuildStep.nextGeneratedId++}`;
```
douglowder: Since these are strings, maybe left pad `nextGeneratedId` with up to 5 zeroes, so that the strings are the same length and sort correctly.

sjchmiela: [`49064bc` (this PR)](https://github.com/expo/eas-cli/pull/3541/commits/49064bc9c80d2957f6d6cf0b219b4ec3682177b6)

PR #3549, opened 2026-03-26
[build-tools] resolve repack version

Kudo:
# Why

nice to have feature requested by customers

# How

showing the resolved repack version in build log, e.g. `Using repack from: @expo/repack@latest (1.2.3)`

# Test Plan

add unit test for the `resolvePackageVersionAsync`


packages/build-tools/src/steps/functions/repack.ts
sjchmiela: Let's then use `resolvedRepackVersion`?

PR #3555, opened 2026-03-27
[auth] logout changes

byronkarlen:
# Why
Instead of waiting for sessions to expire server-side, logout should immediately invalidate the session. Users often have too many sessions at once, most of them unused.

Sessions used to be shared with the website, but are now separate with internal OAuth flow (https://github.com/expo/eas-cli/pull/3542), so logging out from CLI won't affect website session. This would affect existing cli sessions though.

# How
Hit /v2/auth/logout endpoint on `eas logout`. If it fails, just remove credential locally (existing behavior) and credential will soon expire.

# Test Plan
Verified logout API call deleted session server-side. Website session remained. Also ran added unit tests.

### DIFF
--- a/packages/eas-cli/src/user/SessionManager.ts
+++ b/packages/eas-cli/src/user/SessionManager.ts
@@ -90,8 +90,9 @@ export default class SessionManager {
       const apiV2Client = new ApiV2Client({ accessToken: null, sessionSecret });
       try {
         await apiV2Client.postAsync('auth/logout', { body: {} });
-      } catch {
+      } catch (e) {
         // Best-effort: clear the local session even if the server request fails
+        Log.debug('Failed to invalidate session secret on server:', e);
       }
     }
     this.currentActor = undefined;


packages/eas-cli/src/user/SessionManager.ts
```diff
@@ -85,6 +85,15 @@ export default class SessionManager {
   }
 
   public async logoutAsync(): Promise<void> {
+    const sessionSecret = this.getSessionSecret();
+    if (sessionSecret) {
+      const apiV2Client = new ApiV2Client({ accessToken: null, sessionSecret });
+      try {
+        await apiV2Client.postAsync('auth/logout', { body: {} });
+      } catch {
+        // Best-effort: clear the local session even if the server request fails
```
ide, line 94: Would be good to print the error to the debug log. This way most users don't see it but we also don't suppress the error when debugging.

byronkarlen:
(some how lost track of this PR - merging now)

PR #3556, opened 2026-03-27
Add precompiled iOS dependency preparation support

sjchmiela:
## Summary
- export precompiled modules helpers from `@expo/build-tools` and start preparation when flagged builds enable precompiled modules
- wait for precompiled module preparation before running `pod install` in both build-tools pod flows
- set worker env and config for precompiled module downloads, including proxy-aware download handling and coverage for the new behavior

## Testing
Zrzut ekranu 2026-03-28 o 22 26 35

Zrzut ekranu 2026-03-28 o 22 27 27

### DIFF
--- a/packages/build-tools/src/ios/__tests__/pod.test.ts
+++ b/packages/build-tools/src/ios/__tests__/pod.test.ts
@@ -1,6 +1,6 @@
 import downloadFile from '@expo/downloader';
 import spawn from '@expo/turtle-spawn';
-import { mkdirp, mkdtemp, remove } from 'fs-extra';
+import { mkdirp, mkdtemp, move, remove } from 'fs-extra';
 import StreamZip from 'node-stream-zip';
 
 import { createTestIosJob } from '../../__tests__/utils/job';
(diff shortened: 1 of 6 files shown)


chrfalch:
I’m not yet experienced with the build tools, but I’ve gone through the PR and as far as i can see it looks good. This is also a test for supporting precompiled frameworks on iOS in EAS build so there will definitely be room for some improvements when we get decent packaging and versioning of these modules.

packages/worker/src/config.ts
```diff
@@ -98,6 +98,10 @@ export default {
     transform: createBase64EnvTransformer('cocoapodsCacheUrl'),
     defaultValue: null,
   }),
+  precompiledModulesUrls: [
+    'https://storage.googleapis.com/turtle-v2/precompiled-modules/db65b4afac835ff71269ce53937fb20627b133c0/xcframeworks-Debug.zip',
+    'https://storage.googleapis.com/turtle-v2/precompiled-modules/db65b4afac835ff71269ce53937fb20627b133c0/xcframeworks-Release.zip',
+  ],
```
AbbanMustafa, line 104: Just leaving note that this is for sake of testing and POC. We will later need to ensure we publish and download the proper versions

sjchmiela: Yes, from what I heard the plan is to distribute individual xcframeworks inside NPM packages. This solves both distribution and versioning!

PR #3562, opened 2026-03-31
[eas-cli] Fix `hasIgnoredIosProjectAsync`

sjchmiela:
## Why

`hasIgnoredIosProjectAsync()` currently always returns `false` because its `finally` block unconditionally returns `false`, overriding the value from the `try` block.

That means callers never see a positive ignored result even when the Xcode project is actually ignored.

## How

- replace the unconditional `finally` return with a `catch` fallback
- add a unit test that covers the ignored iOS project case

## Test Plan

CI should pass. Verified the test would fail if not for this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,8 +65,6 @@ This is the log of notable changes to EAS CLI and related packages.
 - Remove --environment flag requirement for update:configure. ([#3440](https://github.com/expo/eas-cli/pull/3440) by [@douglowder](https://github.com/douglowder))
 - Fix login spinner interfering with prompts in `eas go` command when not logged in. ([#3451](https://github.com/expo/eas-cli/pull/3451) by [@byronkarlen](https://github.com/byronkarlen))
 
-### 🧹 Chores
-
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -64,6 +65,8 @@ This is the log of notable changes to EAS CLI and related packages.
 - Remove --environment flag requirement for update:configure. ([#3440](https://github.com/expo/eas-cli/pull/3440) by [@douglowder](https://github.com/douglowder))
 - Fix login spinner interfering with prompts in `eas go` command when not logged in. ([#3451](https://github.com/expo/eas-cli/pull/3451) by [@byronkarlen](https://github.com/byronkarlen))
 
+### 🧹 Chores
```
byCedric: Is this addition of the header correct? Otherwise looks good to me!

CHANGELOG.md
```diff
@@ -64,6 +65,8 @@ This is the log of notable changes to EAS CLI and related packages.
...
+
```
sjchmiela:
```suggestion
```

PR #3566, opened 2026-04-01
[eas-build-job] Add `app_store_connect.build_upload` workflow interpolation context

sswrk:
# Why

In workflows triggered by App Store Connect `BUILD_UPLOAD_STATE_UPDATED` it would be handy for users to access the build upload context such as the ID or the state.

# How

Extended the `StaticWorkflowInterpolationContextZ` schema by adding optional `build_upload` to `app_store_connect`.

# Test Plan

Added unit tests.

### DIFF
--- a/packages/eas-build-job/src/__tests__/common.test.ts
+++ b/packages/eas-build-job/src/__tests__/common.test.ts
@@ -116,38 +116,6 @@ describe('StaticWorkflowInterpolationContextZ', () => {
     expect(() => StaticWorkflowInterpolationContextZ.parse(context)).toThrow();
   });
 
-  it('rejects app_store_connect build_upload context with non-UUID id', () => {
-    const context = {
-      after: {},
-      needs: {},
-      workflow: {
-        id: 'workflow-id',
-        name: 'workflow-name',
-        filename: 'workflow.yml',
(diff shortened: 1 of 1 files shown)


packages/eas-build-job/src/common.ts
```diff
@@ -264,6 +264,12 @@ export const StaticWorkflowInterpolationContextZ = z.object({
       app: z.looseObject({
         id: z.string(),
       }),
+      build_upload: z
+        .looseObject({
+          id: z.uuid(),
```
sjchmiela:
```suggestion
          id: z.string(),
```
wdyt?

packages/eas-build-job/src/common.ts
```diff
@@ -264,6 +264,12 @@ export const StaticWorkflowInterpolationContextZ = z.object({
...
+          state: z.enum(['awaiting_uload', 'processing', 'failed', 'complete']),
```
sjchmiela:
```suggestion
          state: z.enum(['awaiting_upload', 'processing', 'failed', 'complete']),
```
hehe

packages/eas-build-job/src/common.ts
sswrk: Makes sense for consistency with other `id` fields in the schema.

PR #3578, opened 2026-04-07
[build-tools] add parseXcactivitylogFunction

hSATAC:
# Why

iOS builds currently have no visibility into per-module compile timing. Developers can't easily identify which targets are slow. Xcode generates `.xcactivitylog` files with detailed build timing data, but since Xcode 11 the CLI (`xcodebuild`) only produces them when `-resultBundlePath` is provided — which our Fastlane templates didn't include.

Linear: ENG-20491

# How

  - Add `result_bundle(true)` and `result_bundle_path` to both Gymfile templates (archive + simulator) to trigger xcactivitylog generation. Also add `derived_data_path` to the archive template so the xcactivitylog lands in a known location (`ios/build`).
  - Add `parseAndReportXcactivitylog()` utility that downloads a pre-built [XCLogParser](https://github.com/MobileNativeFoundation/XCLogParser) binary, parses the xcactivitylog into JSON, and logs a per-module compile metrics table. Never fails the build — all errors are caught and logged as warnings.
  - Register `eas/parse_xcactivitylog` as a built-in step function (opt-in for custom workflows).
  - Also call `parseAndReportXcactivitylog()` in the traditional iOS builder flow under a new `PARSE_XCACTIVITYLOG` build phase, gated behind `EXPERIMENTAL_EAS_XCACTIVITYLOG=1` env var.
  - Add `PARSE_XCACTIVITYLOG` to the `BuildPhase` enum in `@expo/eas-build-job` (requires publishing a new version and updating the dependency in universe for the display name to render correctly on the website).

# Test Plan

- Automation tests
- Tested locally
- To enable in traditional builds, set `EXPERIMENTAL_EAS_XCACTIVITYLOG=1` in build profile env


packages/eas-build-job/src/logs.ts
```diff
@@ -108,6 +109,7 @@ export const buildPhaseDisplayName: Record<BuildPhase, string> = {
   [BuildPhase.INSTALL_PODS]: 'Install pods',
   [BuildPhase.CONFIGURE_XCODE_PROJECT]: 'Configure Xcode project',
   [BuildPhase.RUN_FASTLANE]: 'Run fastlane',
+  [BuildPhase.PARSE_XCACTIVITYLOG]: 'Parse xcactivitylog',
```
sjchmiela: Let's make it more user-friendly? Like… read/report build stats or something? We can also then reuse this step for Android?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
+import spawn from '@expo/turtle-spawn';
+import fs from 'fs-extra';
+import os from 'os';
+import path from 'path';
+import { pipeline as streamPipeline } from 'stream/promises';
+
+const DEFAULT_XCLOGPARSER_VERSION = 'v0.2.46';
+const XCLOGPARSER_DOWNLOAD_URL =
+  'https://github.com/MobileNativeFoundation/XCLogParser/releases/download';
```
sjchmiela: Let's not download from github.com. Let's put it into our GCS and ideally proxy through the CocoaPods cache which can cache arbitrary URLs. See how I do https://github.com/expo/eas-cli/pull/3573

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
...
+
+interface Logger {
```
sjchmiela: can we use `bunyan` instead?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
sjchmiela: Let's also make a PR to `turtle-v2` to install the program to new images we prepare

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
...
+async function downloadXclogparser(
+  tempDir: string,
+  version: string,
+  logger: { info: (msg: string) => void }
+): Promise<string> {
+  const zipName = `XCLogParser-macOS-x86-64-arm64-${version}.zip`;
+  const url = `${XCLOGPARSER_DOWNLOAD_URL}/${version}/${zipName}`;
+  const zipPath = path.join(tempDir, zipName);
+
+  logger.info(`Downloading XCLogParser ${version}...`);
```
sjchmiela: We have this `downloader` thing, not sure if we should use it, but maybe we should

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
...
+}): Promise<void> {
+  let tempDir: string | undefined;
+  try {
+    tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'xclogparser-'));
+    const xclogparserPath = await downloadXclogparser(tempDir, xclogparserVersion, logger);
+
+    logger.info('Parsing xcactivitylog...');
+    const jsonOutput = await runXclogparser(xclogparserPath, derivedDataPath, workspacePath);
+
+    const data = JSON.parse(jsonOutput);
```
sjchmiela: Do we want to report stuff to Datadog too?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
...
+    const report = formatReport(data);
+    logger.info(report);
+  } catch (err: unknown) {
+    logger.warn({ err }, 'Failed to parse xcactivitylog. This does not affect the build.');
+  } finally {
+    if (tempDir) {
+      await fs.rm(tempDir, { force: true, recursive: true }).catch(() => {});
```
sjchmiela: we can do `await asyncResult(fs.rm...)` and we won't need to do this weird `.catch`

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,326 @@
...
+  const binaryPath = path.join(tempDir, 'xclogparser');
+  await fs.chmod(binaryPath, 0o755);
+  return binaryPath;
+}
+
+async function runXclogparser(
+  binaryPath: string,
+  derivedDataPath: string,
+  workspacePath: string
+): Promise<string> {
```
sjchmiela: Would it make sense to write a Zod schema and parse the output here? Then we don't rely on "interface"s we "expect" but have the actual parsed objects. I think theoretically you could also filter out steps you're not interested in (like in an array they may fail validation and get removed? not sure)

packages/build-tools/src/ios/fastlane.ts
```diff
@@ -142,6 +146,8 @@ async function ensureGymfileExists<TJob extends Ios.Job>(
       scheme,
       buildConfiguration,
       outputDirectory: './build',
+      derivedDataPath: './build',
```
sjchmiela, line 149: I presume putting `derivedDataPath` in the same directory as `outputDirectory` is not going to be a problem?

sjchmiela (approved):
Love it!!

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: I think this is out of scope for this PR. For now I want to focus on getting the architecture in place and validating the end-to-end flow. Metrics / telemetry integration seems useful, but I’d rather handle that in a follow-up once we’ve aligned on the shape and destination of the data.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: Let me rethink this part a bit. After taking another look, I realized the current approach (reading stdout into memory and doing a JSON.parse) could be problematic.

The JSON can get quite large (hundreds of MB), which would put significant pressure on memory.

I’m thinking I may need to switch to writing it to disk and then reading it in a buffered/streaming way instead.

I’m not sure if Zod would still fit well under these constraints — I’ll need to look into it — but I think performance should probably be the primary concern here.

sjchmiela: Hmm interesting. Maybe there's a way to tell xclogparser to be smaller about the output? Also, I think computers are good at computering so maybe hundreds of MBs isn't a problem?

hSATAC: I tried locally with `flatJson` reporter it only reduces from 250MB -> 248 MB, it's not helping.

packages/build-tools/src/ios/fastlane.ts
hSATAC: I tested both simulator and archive builds with this setup and did not run into any issues, so I’m planning to keep it as-is for now.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+    [
+      'parse',
+      '--derived_data',
+      derivedDataPath,
+      '--workspace',
+      workspacePath,
+      '--reporter',
+      'json',
+      '--output',
+      outputPath,
```
sjchmiela, line 150: man, i wish eas-cli would work like this too

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+  duration?: number;
+  startTimestamp?: number;
+  endTimestamp?: number;
+  subSteps?: XcactivitylogStep[];
+}
+
+interface XcactivitylogData {
+  schema?: { name?: string } | string;
+  subSteps?: XcactivitylogStep[];
+}
```
sjchmiela, line 171: now that we use zod schema we can use `z.infer` instead of writing two declarations

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+      signature: z.string().optional(),
+      duration: z.number().optional(),
+      startTimestamp: z.number().optional(),
+      endTimestamp: z.number().optional(),
+      subSteps: z.array(XcactivitylogStepSchema).optional(),
+    })
+    .passthrough()
+);
+
+const XcactivitylogDataSchema: z.ZodType<XcactivitylogData> = z
```
sjchmiela, line 187:
```suggestion
const XcactivitylogDataSchemaZ = z
```

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+  endTimestamp?: number;
+  subSteps?: XcactivitylogStep[];
+}
+
+interface XcactivitylogData {
+  schema?: { name?: string } | string;
+  subSteps?: XcactivitylogStep[];
+}
+
+const XcactivitylogStepSchema: z.ZodType<XcactivitylogStep> = z.lazy(() =>
```
sjchmiela, line 173:
```suggestion
const XcactivitylogStepSchemaZ =
```

Why use `z.lazy`?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+  start: number;
+  end: number;
+}
+
+const COMPILE_DETAIL_TYPES = new Set(['cCompilation', 'compileAssetsCatalog', 'compileStoryboard']);
+
+const COMPILE_SIGNATURE_PREFIXES = ['SwiftCompile ', 'SwiftGeneratePch '];
+
+function getXclogparserZipName(version: string): string {
+  return `XCLogParser-macOS-x86-64-arm64-${version}.zip`;
```
sjchmiela, line 213: > x86-64-arm64

that's kinda weird?

packages/build-tools/src/steps/functions/parseXcactivitylog.ts
```diff
@@ -0,0 +1,51 @@
...
+export function parseXcactivitylogFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'parse_xcactivitylog',
+    name: 'Analyze build performance',
+    __metricsId: 'eas/parse_xcactivitylog',
+    supportedRuntimePlatforms: [BuildRuntimePlatform.DARWIN],
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'derived_data',
```
sjchmiela, line 20:
```suggestion
        id: 'derived_data_path',
```
nit, maybe dismissable

packages/build-tools/src/steps/functions/parseXcactivitylog.ts
```diff
@@ -0,0 +1,51 @@
...
+        required: false,
+        defaultValue: 'ios/build',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'workspace',
```
sjchmiela, line 26:
```suggestion
        id: 'workspace_path',
```

packages/build-tools/src/steps/functions/parseXcactivitylog.ts
```diff
@@ -0,0 +1,51 @@
...
+        required: false,
+        defaultValue: 'ios',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+      BuildStepInput.createProvider({
+        id: 'xclogparser_version',
+        required: false,
+        defaultValue: 'v0.2.46',
```
sjchmiela, line 34: this defualt value is defined in multiple places

packages/build-tools/src/builders/ios.ts
```diff
@@ -171,6 +173,16 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     });
   }
 
+  if (ctx.env.EXPERIMENTAL_EAS_XCACTIVITYLOG === '1') {
+    await ctx.runBuildPhase(BuildPhase.PARSE_XCACTIVITYLOG, async () => {
+      await parseAndReportXcactivitylog({
+        derivedDataPath: path.join(ctx.getReactNativeProjectDirectory(), 'ios/build'),
+        workspacePath: path.join(ctx.getReactNativeProjectDirectory(), 'ios'),
```
sjchmiela, line 180: nit: i think it would be nicer to return these paths from wherever they're set (like, from `ensureGymfileExists` or something)

packages/build-tools/src/steps/functions/parseXcactivitylog.ts
```diff
@@ -0,0 +1,51 @@
...
+      }),
+    ],
+    fn: async (stepCtx, { inputs }) => {
+      const derivedData = inputs.derived_data.value as string;
+      const workspace = inputs.workspace.value as string;
+      const version = inputs.xclogparser_version.value as string;
+
+      await parseAndReportXcactivitylog({
+        derivedDataPath: path.resolve(stepCtx.workingDirectory, derivedData),
+        workspacePath: path.resolve(stepCtx.workingDirectory, workspace),
```
sjchmiela, line 45: in builder/ios you used `path.join`. Here you use `path.resolve`. why?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -0,0 +1,431 @@
...
+  tempDir: string,
+  version: string,
+  logger: bunyan
+): Promise<string> {
+  const zipName = getXclogparserZipName(version);
+  const zipPath = path.join(tempDir, zipName);
+  const directUrl = `${XCLOGPARSER_DOWNLOAD_URL}/${zipName}`;
+  const proxiedUrl = getProxiedDownloadUrl({
+    directUrl,
+    proxyBaseUrl: process.env.EAS_BUILD_COCOAPODS_CACHE_URL,
```
sjchmiela, line 66: i don't think env is being exposed as `process.env.`? I think we need to use `env` from context? https://github.com/expo/eas-cli/blob/93280635471658fbac5b317d654a8ca516470952/packages/worker/src/env.ts#L54

sjchmiela (changes requested):

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: The `x86-64-arm64` naming isn't ours; it's how XCLogParser publishes the universal binary on their releases page. Kept it verbatim so the URL maps 1:1 to upstream — swapping in a different name scheme would just add an indirection to maintain.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: Checked again looks like it's a zod v3 pattern we don't need this anymore.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: Thanks. Good catch.

packages/build-tools/src/builders/ios.ts
```diff
@@ -171,6 +173,17 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     });
   }
 
+  if (ctx.env.EXPERIMENTAL_EAS_XCACTIVITYLOG === '1' && fastlaneResult) {
```
sjchmiela: I think `fastlaneResult` is only going to be set if `RUN_FASTLANE` succeeds and yet it looks like the code is structured in a way that wants to parse xcactivitylog even if it fails.

packages/build-tools/src/ios/fastlane.ts
```diff
@@ -28,7 +29,13 @@ export async function runFastlaneGym<TJob extends Ios.Job>(
     entitlements: object | null;
     extraEnv?: Env;
   }
-): Promise<void> {
+): Promise<{
+  derivedDataPath: string;
+  workspacePath: string;
+}> {
+  const workspacePath = path.join(ctx.getReactNativeProjectDirectory(), 'ios');
+  const derivedDataPath = path.join(workspacePath, 'build');
```
sjchmiela, line 37: We might want to make this an argument to `ensureGymfileExists` or have it be returned from `ensureGymfileExists`.

Also, this is an "optimistic" derived data path because theoretically a user may come with their own Gymfile I think?

sjchmiela (approved):
Thank you!!

packages/build-tools/src/ios/fastlane.ts
hSATAC: Good catch — we did think about the user-provided Gymfile case, but decided to skip it for now. Supporting it properly would mean adding a full Gymfile parser just to extract derivedDataPath, which feels like a lot of complexity for a relatively rare case.

The xcactivitylog parser is also fault-tolerant — if this optimistic path is wrong (e.g. custom Gymfile points elsewhere), it’ll just no-op instead of failing the build. So worst case is missing insights, not breaking anything.

Happy to revisit if we see real usage for custom Gymfiles.

PR #3579, opened 2026-04-07
[worker] Read worker capabilities and set EAS_ prefixed envs

gwdp:
# Why

Please describe the motivation for this PR, and link to relevant GitHub issues, forums posts, or feature requests.

# How

How did you build this feature or fix this bug and why?

# Test Plan

Please describe how you tested this change and how a reviewer could reproduce your test, especially if this PR does not include automated tests! If possible, please also provide terminal output and/or screenshots demonstrating your test/reproduction.

Ref ENG-17446

PR #3580, opened 2026-04-07
[build-tools] Early stop `eas/start_android_emulator` when Android emulator host setup is invalid

gwdp:
# Why

Starting an Android emulator without nested virtualization support always fails, but today that failure happens later and with less actionable context.
Adding nested virtualization detection to start_android_emulator allows us to fail fast with a clearer error message when the VM does not support emulator hardware acceleration.

# How

Add an early preflight check in start_android_emulator to detect whether emulator acceleration prerequisites are present (including VMX/SVM support).
If acceleration is not available, fail fast with a clear, user-friendly error explaining that nested virtualization is required and how to enable it.

# Test Plan

- CI checks (with new tests)
- Running `-accel-check` on nested and non nested virtualization to confirm command behavior https://staging.expo.dev/accounts/gwdp/projects/coin-flip/workflows/019d73d1-261b-793a-9db7-98cfa335c197
- Local run for failure format and checking UX
Screenshot 2026-04-13 at 12 33 34 PM
Screenshot 2026-04-13 at 12 33 30 PM

Ref ENG-17446


gwdp:
~~x-ref https://github.com/expo/workflow-orchestration/pull/738~~
 ~~x-ref https://github.com/expo/eas-cli/pull/3579~~

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -43,7 +81,18 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
...
+      const isNestedVirtualizationEnabled = await getIsNestedVirtualizationEnabledAsync(
+        env,
+        global.runtimePlatform
+      );
+      if (!isNestedVirtualizationEnabled) {
+        throw new UserError(
+          ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR_CODE,
+          ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR
+        );
+      }
```
sjchmiela: It might be slightly easier to have a function like
```
async function assertNestedVirtualizationIsAvailableAsync({ env, platform }) {
  // do the check
  // throw the error
}
```

I think then you don't need the global scope consts which is slightly nicer?

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -11,6 +17,38 @@ import { retryAsync } from '../../utils/retry';
...
+ */
+async function getIsNestedVirtualizationEnabledAsync(
+  env: NodeJS.ProcessEnv,
+  runtimePlatform: BuildRuntimePlatform
+): Promise<boolean> {
+  if (runtimePlatform !== BuildRuntimePlatform.LINUX) {
+    return false;
+  }
+  try {
+    await spawn('grep', ['-Eq', '(vmx|svm)', '/proc/cpuinfo'], { env });
```
sjchmiela: I wonder if `emulator -accel-check` wouldn't be simpler? I would appreciate an example workflow run which ran the check command on nested-virt-enabled and not. you can just do

```
jobs:
  test_1:
    runs_on: linux-medium-nested-virtualization
    steps:
      - run: ...
  test_2:
    runs_on: linux-medium
    steps:
      - run: ...
```

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -11,6 +17,38 @@ import { retryAsync } from '../../utils/retry';
...
+ * We detect support by checking `/proc/cpuinfo` for `vmx` (Intel) or `svm` (AMD).
+ * Non-Linux hosts are treated as unsupported for this step.
+ */
+async function getIsNestedVirtualizationEnabledAsync(
+  env: NodeJS.ProcessEnv,
+  runtimePlatform: BuildRuntimePlatform
+): Promise<boolean> {
+  if (runtimePlatform !== BuildRuntimePlatform.LINUX) {
+    return false;
+  }
```
sjchmiela: not sure if necessary?

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -11,6 +17,38 @@ import { retryAsync } from '../../utils/retry';
 
 const ANDROID_STARTUP_ATTEMPT_TIMEOUT_MS = [60_000, 120_000, 180_000];
 const ANDROID_STARTUP_RETRIES_COUNT = ANDROID_STARTUP_ATTEMPT_TIMEOUT_MS.length - 1;
+const ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR_CODE =
+  'EAS_ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_UNAVAILABLE';
```
sjchmiela:
```suggestion
  'ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_UNAVAILABLE';
```

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: thank you for the suggestion, cleaner approach ➕

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Agreed and your proposal is the [documented way](https://developer.android.com/studio/run/emulator-acceleration#accel-check) to check for emulator acceleration, good call. 

Ran sample workflow here with both cmds just to keep as reference: https://staging.expo.dev/accounts/gwdp/projects/coin-flip/workflows/019d73d1-261b-793a-9db7-98cfa335c197

Moved to `emulator -accel-check` as it provided the same outcome, with more focused and recommended approach check for this purpose.

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Misunderstanding from my end, scanned through a few errors and only saw that prefix, but can see that prefix is related to where the error is coming from other than always AES prefixed. Thank you for catching that

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -11,6 +17,18 @@ import { retryAsync } from '../../utils/retry';
...
+
+const ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR = [
+  'Could not start the Android emulator on this runner.',
+  '',
+  'Android emulator requires nested virtualization on Linux. This job does not have the required virtualization support.',
+  '',
+  'Update your workflow YAML to use a nested-virtualization Linux runner, for example:',
+  '  runs_on: linux-medium-nested-virtualization',
+  '  runs_on: linux-large-nested-virtualization',
+].join('\n');
```
sjchmiela: Why do you find it helpful to extract these `const`s to global scope? (An honest curious question.)

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -253,3 +273,30 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
     },
   });
 }
+
+async function assertNestedVirtualizationIsAvailableAsync({
```
sjchmiela: question: now that I'm thinking about it maybe this check should _not_ force-fail the job, but instead only print a helpful warning? We can then add an annotation on the log to make it even more prominent. https://expo.dev/admin/annotations

What happens if a user does `start_android_emulator` on non-virtualized VM? Does `emulator` return a non-zero code or do we just "fail, waiting"?

EDIT: Ok looking more, judging by https://staging.expo.dev/accounts/gwdp/projects/coin-flip/workflows/019d73d1-261b-793a-9db7-98cfa335c197 we don't even install emulator on non-nested VMs which means `emulator` should fail which means start_android_emulator step fails with ENOENT or sth like that which is not too helpful. so maybe your solution is good, but maybe let's rename stuff to indicate that we're checking emulator is set up correctly or what

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -44,6 +62,8 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
       }),
     ],
     fn: async ({ logger }, { inputs, env }) => {
+      await assertNestedVirtualizationIsAvailableAsync({ env });
```
sjchmiela: Maybe it makes sense to add some kind of EAS_NO_NESTED_VIRTUALIZATION_CHECK=1 escape hatch

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -253,3 +273,30 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
...
+  const isNestedVirtualizationEnabled = await getIsNestedVirtualizationEnabledAsync(env);
+  if (!isNestedVirtualizationEnabled) {
+    throw new UserError(
+      ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR_CODE,
+      ANDROID_EMULATOR_LINUX_HARDWARE_VIRT_ERROR
+    );
+  }
+}
+
+async function getIsNestedVirtualizationEnabledAsync(env: NodeJS.ProcessEnv): Promise<boolean> {
```
sjchmiela: now that it's returning false if we don't have ANDROID_HOME set let's rename it to something different?

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: I usually find that useful so we offload cognitive load from the function; Understand how that can be slightly harder to review and maintain, will take a note and get those inline to follow other functions standard.

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Good call, added a escape hatch. `EAS_NO_NESTED_VIRTUALIZATION_CHECK`

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Actually, reasoning further, failure pattern when/if `ANDROID_HOME` is missing would be quite obscure, since spawn would not produce any logs and just assertion would be raised. 

If you are aligned, I modified it to use empty sdk path and actually execute with it, so failure is more visible to us if that ever happen. Thoughts?

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Took this path of the early failure since [my reproduction](https://staging.expo.dev/accounts/gwdp/projects/coin-flip/workflows/019d650b-f709-7bfe-ad19-82091dab2a4a) did show 9m+ of execution into the start emulator function until failure on non nested environment and thought about potential savings with the early path. Renamed things slightly, let me now if aligned

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
```diff
@@ -253,3 +267,31 @@ export function createStartAndroidEmulatorBuildFunction(): BuildFunction {
...
+}): Promise<void> {
+  const isAndroidEmulatorHostSupported = await isAndroidEmulatorHostSupportedAsync(env);
+  if (!isAndroidEmulatorHostSupported) {
+    throw new UserError(
+      'ANDROID_EMULATOR_ACCEL_CHECK_UNAVAILABLE',
+      'Could not start the Android emulator on this runner.\n\n' +
+        'Android emulator requires nested virtualization on Linux. This job does not have the required virtualization support.\n\n' +
+        'Update your workflow YAML to use a nested-virtualization Linux runner, for example:\n' +
+        '  runs_on: linux-medium-nested-virtualization\n' +
+        '  runs_on: linux-large-nested-virtualization'
```
sjchmiela: let's also point to https://docs.expo.dev/eas/workflows/syntax/#jobsjob_idruns_on

sjchmiela (approved):

packages/build-tools/src/steps/functions/startAndroidEmulator.ts
gwdp: Thank you for the link, added and noted that `docsUrl` is not being saved and neither show to users. Handling that as a followup on ENG-20575

gwdp: Added inline error (tmp solution) and will scale back within followup, which handles doscUrl from step errors

PR #3598, opened 2026-04-09
[build-tools] Add eas/deploy function

gwdp:
# Why

Introduce `eas/deploy` function to streamline export+deploy usual combination within single unit. Includes `deploy_json` output from `eas deploy --json` for anything that comes after.

Ref ENG-14737

# How

New function with configurable export command, deploy alias, prod and source_maps

# Test Plan

Local run with the following workflows

✅ Success path (alias did not play nicely locally, so skipped this validation)
```
name: deploy-web-custom

on:
  workflow_dispatch: {}

jobs:
  deploy_custom:
    name: Deploy web with custom steps
    type: custom
    runs_on: linux-medium
    environment: production
    outputs:
      deploy_json: ${{ steps.deploy.outputs.deploy_json }}
      deploy_url: ${{ steps.deploy.outputs.deploy_url }}
      deploy_deployment_url: ${{ steps.deploy.outputs.deploy_deployment_url }}
      deploy_identifier: ${{ steps.deploy.outputs.deploy_identifier }}
      deploy_dashboard_url: ${{ steps.deploy.outputs.deploy_dashboard_url }}
      deploy_alias_url: ${{ steps.deploy.outputs.deploy_alias_url }}
    env:
      APP_VARIANT: production
    steps:
      - uses: eas/checkout
      - uses: eas/use_npm_token
      - uses: eas/install_node_modules
      - uses: eas/export
      - uses: eas/deploy
        id: deploy
      - run: echo "Deploy output - ${{ steps.deploy.outputs.deploy_json }}"
      - run: echo "Deploy URL - ${{ steps.deploy.outputs.deploy_url }}"
      - run: echo "Deployment URL - ${{ steps.deploy.outputs.deploy_deployment_url }}"
      - run: echo "Deployment identifier - ${{ steps.deploy.outputs.deploy_identifier }}"
      - run: echo "Dashboard URL - ${{ steps.deploy.outputs.deploy_dashboard_url }}"
      - run: echo "Alias URL - ${{ steps.deploy.outputs.deploy_alias_url }}"

  read_output:
    name: Read deploy output from another job
    type: custom
    runs_on: linux-medium
    needs: [deploy_custom]
    steps:
      - run: echo "Deploy JSON from previous job - ${{ needs.deploy_custom.outputs.deploy_json }}"
      - run: echo "Deploy URL from previous job - ${{ needs.deploy_custom.outputs.deploy_url }}"
      - run: echo "Deployment URL from previous job - ${{ needs.deploy_custom.outputs.deploy_deployment_url }}"
      - run: echo "Deployment identifier from previous job - ${{ needs.deploy_custom.outputs.deploy_identifier }}"
      - run: echo "Dashboard URL from previous job - ${{ needs.deploy_custom.outputs.deploy_dashboard_url }}"
      - run: echo "Alias URL from previous job - ${{ needs.deploy_custom.outputs.deploy_alias_url }}"
```

Screenshot 2026-04-17 at 1 06 25 PM
Screenshot 2026-04-17 at 1 06 34 PM

❌  Failure pattern
```
name: deploy-web-custom-simple-no-setup

on:
  workflow_dispatch: {}

jobs:
  deploy_custom:
    name: Deploy web without setup steps (expected to fail)
    type: custom
    runs_on: linux-medium
    environment: production
    env:
      APP_VARIANT: production
    steps:
      - run: echo "Before eas/deploy - no npm token, or node_modules"
      - uses: eas/checkout
      # MISSING: use_npm_token, install_node_modules here to simulate failure
      - uses: eas/export
      - uses: eas/deploy
      - run: echo "After eas/deploy stub"
```
Screenshot 2026-04-17 at 1 07 22 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - [build-tools] Add `eas/read_package_json` and `eas/read_app_config` functions ([#3585](https://github.com/expo/eas-cli/pull/3585) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Add eas/deploy function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
 
 ### 🐛 Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - [build-tools] Add `eas/read_package_json` and `eas/read_app_config` functions ([#3585](https://github.com/expo/eas-cli/pull/3585) by [@gwdp](https://github.com/gwdp))
-- [build-tools] Add `eas/deploy` function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Add `eas/deploy` function for EAS Hosting web deployments. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Add `eas/export` function for Expo web exports. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
 
 ### 🐛 Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - [build-tools] Add `eas/read_package_json` and `eas/read_app_config` functions ([#3585](https://github.com/expo/eas-cli/pull/3585) by [@gwdp](https://github.com/gwdp))
-- [build-tools] Add eas/deploy function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Add `eas/deploy` function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
 
 ### 🐛 Bug fixes
 
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -111,7 +111,8 @@ export function createEasDeployBuildFunction(): BuildFunction {
           'Deploy command failed.\n\n' +
             `This can happen when deploy inputs are invalid, the export directory "${exportDir}" is missing, or EAS CLI authentication/network checks fail.\n` +
             'Check the deploy step logs, verify your deploy configuration and exported files, then retry.\n' +
-            `Original error: ${errorMessage}`
+            `Original error: ${errorMessage}`,
+          { cause: error }
         );
       }
     },
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -118,10 +118,8 @@ function getDeployCommand({
     deployCommand.push('--prod');
   }
 
-  if (sourceMaps === true) {
+  if (sourceMaps) {
     deployCommand.push('--source-maps');
-  } else if (sourceMaps === false) {
-    deployCommand.push('--no-source-maps');
   }
 
   return deployCommand;
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -168,7 +168,7 @@ function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; log
   } catch (error: unknown) {
     logger.warn(
       { err: error },
-      'Failed to parse deploy JSON. DEPLOY_URL, DEPLOY_DEPLOYMENT_URL, DEPLOY_IDENTIFIER, DEPLOY_DASHBOARD_URL, DEPLOY_ALIAS_URL will be unavailable on this build.'
+      'Failed to parse "eas deploy" JSON output. Some outputs expected from this step will be undefined.'
     );
     return null;
   }
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -75,7 +75,6 @@ export function createEasDeployBuildFunction(): BuildFunction {
         exportDir,
         alias,
         prod,
-        environment: stepsCtx.global.staticContext.metadata?.environment,
         sourceMaps,
       });
       stepsCtx.logger.info(`Running deploy command: eas ${deployCommand.join(' ')}`);
@@ -97,13 +96,11 @@ export function createEasDeployBuildFunction(): BuildFunction {
           deployJson,
           logger: stepsCtx.logger,
         });
-        if (parsedDeploymentOutput) {
-          outputs.deploy_url.set(parsedDeploymentOutput.deploy_url);
-          outputs.deploy_deployment_url.set(parsedDeploymentOutput.deploy_deployment_url);
-          outputs.deploy_identifier.set(parsedDeploymentOutput.deploy_identifier);
-          outputs.deploy_dashboard_url.set(parsedDeploymentOutput.deploy_dashboard_url);
-          outputs.deploy_alias_url.set(parsedDeploymentOutput.deploy_alias_url);
-        }
+        outputs.deploy_url.set(parsedDeploymentOutput.deploy_url);
+        outputs.deploy_deployment_url.set(parsedDeploymentOutput.deploy_deployment_url);
+        outputs.deploy_identifier.set(parsedDeploymentOutput.deploy_identifier);
+        outputs.deploy_dashboard_url.set(parsedDeploymentOutput.deploy_dashboard_url);
+        outputs.deploy_alias_url.set(parsedDeploymentOutput.deploy_alias_url);
       } catch (error) {
         const errorMessage = error instanceof Error ? error.message : 'unknown error';
         throw new UserError(
@@ -123,19 +120,14 @@ function getDeployCommand({
   exportDir,
   alias,
   prod,
-  environment,
   sourceMaps,
 }: {
   exportDir: string;
   alias?: string;
   prod?: boolean;
-  environment?: string;
   sourceMaps?: boolean;
 }): string[] {
   const deployCommand = ['deploy', '--non-interactive', '--json', '--export-dir', exportDir];
-  if (environment) {
-    deployCommand.push('--environment', environment);
-  }
   if (alias) {
     deployCommand.push('--alias', alias);
   }
@@ -154,7 +146,7 @@ function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; log
   deploy_identifier?: string;
   deploy_dashboard_url?: string;
   deploy_alias_url?: string;
-} | null {
+} {
   try {
     const deployObject = JSON.parse(deployJson);
     return {
@@ -170,6 +162,6 @@ function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; log
       { err: error },
       'Failed to parse "eas deploy" JSON output. Some outputs expected from this step will be undefined.'
     );
-    return null;
+    return {};
   }
 }
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -152,11 +152,12 @@ function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; log
     // TODO: improve typing here; Look into WorkerDeploymentData
     const deployObject = JSON.parse(deployJson);
     return {
-      deploy_url: deployObject.production.url || deployObject.aliases[0].url || deployObject.url,
+      deploy_url:
+        deployObject.production?.url || deployObject.aliases?.[0]?.url || deployObject.url,
       deploy_deployment_url: deployObject.url,
       deploy_identifier: deployObject.identifier,
       deploy_dashboard_url: deployObject.dashboardUrl,
-      deploy_alias_url: deployObject.aliases[0].url,
+      deploy_alias_url: deployObject.aliases?.[0]?.url,
     };
   } catch (error: unknown) {
     const message = error instanceof Error ? error.message : String(error);
@@ -164,7 +165,7 @@ function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; log
       'Failed to parse deploy JSON: ' +
         `${message}\n` +
         'DEPLOY_URL, DEPLOY_DEPLOYMENT_URL, DEPLOY_IDENTIFIER, DEPLOY_DASHBOARD_URL, DEPLOY_ALIAS_URL will be unavailable on this build.',
-      { error }
+      { error, deployJson }
     );
     return null;
   }
--- a/packages/build-tools/src/steps/functions/deploy.ts
+++ b/packages/build-tools/src/steps/functions/deploy.ts
@@ -98,9 +98,11 @@ export function createEasDeployBuildFunction(): BuildFunction {
           logger: stepsCtx.logger,
         });
         if (parsedDeploymentOutput) {
-          for (const [key, value] of Object.entries(parsedDeploymentOutput)) {
-            outputs[key].set(value);
-          }
+          outputs.deploy_url.set(parsedDeploymentOutput.deploy_url);
+          outputs.deploy_deployment_url.set(parsedDeploymentOutput.deploy_deployment_url);
+          outputs.deploy_identifier.set(parsedDeploymentOutput.deploy_identifier);
+          outputs.deploy_dashboard_url.set(parsedDeploymentOutput.deploy_dashboard_url);
+          outputs.deploy_alias_url.set(parsedDeploymentOutput.deploy_alias_url);
         }
       } catch (error) {
         const errorMessage = error instanceof Error ? error.message : 'unknown error';
@@ -146,14 +148,13 @@ function getDeployCommand({
 }
 
 function parseDeploymentOutput({ deployJson, logger }: { deployJson: string; logger: bunyan }): {
-  deploy_url: string;
-  deploy_deployment_url: string;
-  deploy_identifier: string;
-  deploy_dashboard_url: string;
-  deploy_alias_url: string;
+  deploy_url?: string;
+  deploy_deployment_url?: string;
+  deploy_identifier?: string;
+  deploy_dashboard_url?: string;
+  deploy_alias_url?: string;
 } | null {
   try {
-    // TODO: improve typing here; Look into WorkerDeploymentData
     const deployObject = JSON.parse(deployJson);
     return {
       deploy_url:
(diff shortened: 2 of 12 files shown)


packages/build-tools/src/utils/easCli.ts
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      cmd: 'npx',
+      args: [...npxArgsPrefix, `eas-cli@${EasCliNpmTags.STAGING}`],
+      extraEnv: {},
+    };
+  } else if (process.env.ENVIRONMENT === 'staging') {
+    return {
+      cmd: 'npx',
+      args: [...npxArgsPrefix, `eas-cli@${EasCliNpmTags.STAGING}`],
+      extraEnv: { EXPO_STAGING: '1' },
```
sjchmiela, line 41: fyi i think `extraEnv` may not be necessary thanks to https://github.com/expo/eas-cli/blob/190a9daa2eba56f4df3c8febcf8bae7a2f4abc2f/packages/worker/src/env.ts#L138-L142 

I know you're only moving though so leaving it as fyi

packages/build-tools/src/utils/easCli.ts
```diff
@@ -0,0 +1,45 @@
...
+import { isAtLeastNpm7Async } from './packageManager';
+
+export async function resolveEasCommandPrefixAndEnvAsync(): Promise<{
+  cmd: string;
+  args: string[];
+  extraEnv: Env;
+}> {
+  const npxArgsPrefix = (await isAtLeastNpm7Async()) ? ['-y'] : [];
+  if (process.env.ENVIRONMENT === 'development') {
+    return {
```
sjchmiela, line 32: another fyi is that in development it's sometimes worth it to try to use [`easd`](https://github.com/expo/eas-cli/blob/37165428d120582195d00a220a522fc12932be2d/CONTRIBUTING.md#L16-L22)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🎉 New features
 
 - [build-tools] Add `eas/read_package_json` and `eas/read_app_config` functions ([#3585](https://github.com/expo/eas-cli/pull/3585) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Add eas/deploy function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
```
sjchmiela:
```suggestion
- [build-tools] Add `eas/deploy` function group. ([#3598](https://github.com/expo/eas-cli/pull/3598) by [@gwdp](https://github.com/gwdp))
```

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  if (prod) {
+    deployCommand.push('--prod');
+  }
+
+  if (sourceMaps === true) {
+    deployCommand.push('--source-maps');
+  } else if (sourceMaps === false) {
+    deployCommand.push('--no-source-maps');
+  }
```
sjchmiela: let's set this only if `sourceMaps` is no undefined. Let's skip if undefined

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,128 @@
...
+      });
+      stepsCtx.logger.info(`Running deploy command: eas ${deployCommand.join(' ')}`);
+
+      try {
+        const result = await runEasCliCommand({
+          args: deployCommand,
+          options: {
+            cwd: stepsCtx.workingDirectory,
+            env,
+            logger: stepsCtx.logger,
```
sjchmiela, line 88: In current `type: deploy` job through the magic of bash scripting we hide the stdout from the user logs (`ABC=$(echo abc)` prints nothing and we do `DEPLOY_JSON_OUTPUT=$(...)`).

Let's achieve the same with https://github.com/expo/eas-cli/blob/789cf62bd6d7b0b4173ad3e790426ac6807b9731/packages/build-tools/src/common/easBuildInternal.ts#L73-L74

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,128 @@
...
+        id: 'source_maps',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'export_command',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+        required: false,
+        defaultValue: 'npx expo export --platform web',
+      }),
```
sjchmiela, line 40: wdyt about making it be a separate function? then the input to this function would be the `export_dir` and output of `eas/[expo_]export` would be the `dist_dir`?

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,128 @@
...
+          args: deployCommand,
+          options: {
+            cwd: stepsCtx.workingDirectory,
+            env,
+            logger: stepsCtx.logger,
+          },
+        });
+        outputs.deploy_json.set(result.stdout.toString());
+      } catch (error) {
+        throw new UserError(
```
sjchmiela, line 106: i wonder if it would be nice if we made `eas-cli deploy` return different exit codes for different issues and based on that figure out whether it was a system or user error?

I have https://exponent-internal.slack.com/archives/C06EFBQK3B7/p1773417343042269 in my memory

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,128 @@
...
+      BuildStepInput.createProvider({
+        id: 'export_command',
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+        required: false,
+        defaultValue: 'npx expo export --platform web',
+      }),
+    ],
+    outputProviders: [
+      BuildStepOutput.createProvider({
+        id: 'deploy_json',
```
sjchmiela, line 44: i think we should have deploy_json but i also think we can parse the json in this function already and expose identifier, url

sjchmiela (changes requested):

CHANGELOG.md
gwdp: Thank you, will keep a note for recurrences :)

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Fixed, thank you for the feedback

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Thank you for the pointers, fixed 🙇‍♂️

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Great idea, allows a little more space for inputs from the export side. Let me know if you have any additional thoughts around the inputs of export function.

Around the name, not 100% sure if you suggested eas/expo_export, would that be the case? I can see prebuild which similarly, calls expo command, not having the prefix although it would make sense imo. export only could be too broad for people new to EAS, as me?

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Nice addition, thanks for the input!

packages/build-tools/src/utils/easCli.ts
gwdp: Thanks for the fyi. That was not under my radar, added a probe/early route to easd when available on development.

packages/build-tools/src/utils/easCli.ts
gwdp: Thank you! Keeping a note here to come back on a possible cleanup in near future

packages/build-tools/src/steps/functions/deploy.ts
gwdp: That’s a great idea, thanks for raising it. I’m digging into the deploy command to see how we can improve error signalling and bubble that up here. 

A retryable flag could also let us retry once or twice for transient deployment flukes. This likely won’t solve the auth issue from the Slack thread directly, but it should make outcomes much more actionable in logs.

I made a small improvement to the error message in this PR. If you’re aligned, I’d like to follow up in ENG-20594 with improved `eas deploy` exit code/error handling. Thoughts?

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import { resolvePackageManager } from '../../utils/packageManager';
+import { runExpoCliCommand } from '../../utils/project';
+
+/** Matches `npx expo export -p` / `--platform` (see `npx expo export --help`). */
+const EXPO_EXPORT_PLATFORMS = new Set(['android', 'ios', 'web', 'all']);
+
+export function createEasExportBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
```
sjchmiela, line 14: I'm kinda curious if it would make sense to call it `expo/export` but maybe let's not haha. `eas/export` 👍

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,150 @@
...
+      const noSsg = inputs.no_ssg.value as boolean | undefined;
+      const apiOnly = inputs.api_only.value as boolean | undefined;
+      const platform = inputs.platform.value as string;
+
+      if (!EXPO_EXPORT_PLATFORMS.has(platform)) {
+        throw new UserError(
+          'EXPO_EXPORT_INVALID_PLATFORM',
+          `Invalid export platform "${platform}". Expected one of: android, ios, web, all.`
+        );
+      }
```
sjchmiela:
```suggestion
```
I guess `expo export` is going to throw an understandable error when invalid platform is provided?

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,150 @@
...
+        required: false,
+        defaultValue: 'dist',
+      }),
+      BuildStepInput.createProvider({
+        id: 'dev',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'no_minify',
```
sjchmiela: should this be `minify: true/false`?

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,150 @@
...
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'assetmap',
```
sjchmiela: dump_assetmap?

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,150 @@
...
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'no_ssg',
```
sjchmiela: `ssg: true/false`? idk…

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,174 @@
...
+        deployObject.production?.url || deployObject.aliases?.[0]?.url || deployObject.url,
+      deploy_deployment_url: deployObject.url,
+      deploy_identifier: deployObject.identifier,
+      deploy_dashboard_url: deployObject.dashboardUrl,
+      deploy_alias_url: deployObject.aliases?.[0]?.url,
+    };
+  } catch (error: unknown) {
+    logger.warn(
+      { err: error },
+      'Failed to parse deploy JSON. DEPLOY_URL, DEPLOY_DEPLOYMENT_URL, DEPLOY_IDENTIFIER, DEPLOY_DASHBOARD_URL, DEPLOY_ALIAS_URL will be unavailable on this build.'
```
sjchmiela:
```suggestion
      'Failed to parse "eas deploy" JSON output. Some outputs expected from this step will be undefined.'
```
or maybe we want to iterate "deploy_url, deploy_deployment_url, ..." like you did but in lowercase

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,174 @@
...
+    );
+    return null;
```
sjchmiela:
```suggestion
    return {};
```
I think it's simpler if they're all optional either way?

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,174 @@
...
+      const alias = inputs.alias.value as string | undefined;
+      const prod = inputs.prod.value as boolean | undefined;
+      const sourceMaps = inputs.source_maps.value as boolean | undefined;
+      const exportDir = inputs.export_dir.value as string;
+
+      const deployCommand = getDeployCommand({
+        exportDir,
+        alias,
+        prod,
+        environment: stepsCtx.global.staticContext.metadata?.environment,
```
sjchmiela: We shouldn't need this, the environment should come from the whole workflows setup

packages/build-tools/src/steps/functions/deploy.ts
```diff
@@ -0,0 +1,174 @@
...
+          outputs.deploy_alias_url.set(parsedDeploymentOutput.deploy_alias_url);
+        }
+      } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : 'unknown error';
+        throw new UserError(
+          'EAS_DEPLOY_FAILED',
+          'Deploy command failed.\n\n' +
+            `This can happen when deploy inputs are invalid, the export directory "${exportDir}" is missing, or EAS CLI authentication/network checks fail.\n` +
+            'Check the deploy step logs, verify your deploy configuration and exported files, then retry.\n' +
+            `Original error: ${errorMessage}`
```
sjchmiela:
```suggestion
            `Original error: ${errorMessage}`,
          { cause: error }
```

sjchmiela (changes requested):
This is great! A few small changes and will be good to go!

packages/build-tools/src/steps/functions/export.ts
gwdp: Good catch, thank you for highlighting it

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Get rid of a few lines, good simplification, thank you

packages/build-tools/src/steps/functions/export.ts
gwdp: Happy to go with it although drifts a little from the CLI args. Also making `minify` defaults to true since its the default export behaviour. Let me know if any thoughts against it

packages/build-tools/src/steps/functions/export.ts
gwdp: Opsie, that was a drift from the CLI, my bad, thank you for catching it

packages/build-tools/src/steps/functions/export.ts
gwdp: Same as https://github.com/expo/eas-cli/pull/3598#discussion_r3101888256, updated, let me know if any thoughts against the default `true` value for consistency with `eas deploy` cmd

packages/build-tools/src/steps/functions/export.ts
gwdp: Mind reader 🤓 I honestly started with `eas/expo_export`, questioned quite a few times if that was the right call until I saw, already stablished `prepack` function, same pattern -> calls `eas_cli` command. 

Imo, `expo/export`, or anything that makes clear, its just a `eas_cli` wrapper, it would improve general understanding from folks not too familiar with the ecosystem but also dont want go outside of stablished pattern.

packages/build-tools/src/steps/functions/deploy.ts
gwdp: Updated, thank you

packages/build-tools/src/utils/easCli.ts
```diff
@@ -0,0 +1,72 @@
...
+    const result = await spawn('easd', ['--help'], {
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+    return result.status === 0;
+  } catch {
+    return false;
+  }
+}
+
+export async function resolveEasCommandPrefixAndEnvAsync(options?: { logger?: bunyan }): Promise<{
```
sjchmiela: I think the `logger` is a fine addition, I wouldn't have think of it. I guess having it optional makes things weird because it's easy to _not_ provide it?

Maybe we can drop it and _not_ log the `easd` thing?

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (dev) {
+    exportCommand.push('--dev');
+  }
+  if (!minify) {
+    exportCommand.push('--no-minify');
+  }
+  if (dumpAssetmap) {
+    exportCommand.push('--dump-assetmap');
+  }
+  if (!ssg) {
```
sjchmiela:
```suggestion
  if (ssg === false) {
```

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,142 @@
...
+  dumpAssetmap?: boolean;
+  ssg?: boolean;
+  apiOnly?: boolean;
+  platform: string;
+}): string[] {
+  const exportCommand = ['export', '--output-dir', outputDir, '--platform', platform];
+  if (dev) {
+    exportCommand.push('--dev');
+  }
+  if (!minify) {
```
sjchmiela:
```suggestion
  if (minify === false) {
```

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,142 @@
...
+      BuildStepInput.createProvider({
+        id: 'dev',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'minify',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+        defaultValue: true,
```
sjchmiela:
```suggestion
```

packages/build-tools/src/steps/functions/export.ts
```diff
@@ -0,0 +1,142 @@
...
+      BuildStepInput.createProvider({
+        id: 'dump_assetmap',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+      }),
+      BuildStepInput.createProvider({
+        id: 'ssg',
+        allowedValueTypeName: BuildStepInputValueTypeName.BOOLEAN,
+        required: false,
+        defaultValue: true,
```
sjchmiela:
```suggestion
```

packages/build-tools/src/steps/functions/export.ts
sjchmiela: I think `eas-cli` might be a better place to define the defaults ("default" defaults)? Or maybe workflow ("purpose-led" defaults)? This function is just a configurable, "invisible" tool?

sjchmiela (approved):

packages/build-tools/src/steps/functions/export.ts
gwdp: Agreed, the defaults were shadowing `eas-cli deploy` but removing it and leveraging entirely on the cli defaults makes sense to me. Thanks for the feedback.

packages/build-tools/src/steps/functions/export.ts
gwdp: Updated, thanks!

packages/build-tools/src/steps/functions/export.ts
gwdp: Updated, thanks!

packages/build-tools/src/steps/functions/export.ts
gwdp: Updated, thanks!

packages/build-tools/src/utils/easCli.ts
gwdp: My line was around being easy to forget some local env set that is making `easd` usage, so better warn folks but happy to drop it and make it simpler. Appreciate the review!

PR #3608, opened 2026-04-15
[build-tools] Add EAS_FALLBACK_PACKAGE_MANAGER env var support to package manager resolver

hSATAC:
# Why

This PR implements the `eas-cli` / build-tools side of [ENG-20326](https://linear.app/expo/issue/ENG-20326).

Today, when package manager detection cannot determine a package manager from the project, build-tools falls back to `yarn`. That fallback is hardcoded. For example, if Launch wants to use `bun` as the fallback package manager, it currently has to fork steps such as `install_node_modules` to get that behavior.

We want to support a way for Launch (and any job) to customize this fallback. After review discussion (thanks @sjchmiela), we landed on an env-var-based approach: setting `EAS_FALLBACK_PACKAGE_MANAGER` at the job env level is enough to override the `yarn` default. This keeps the surface area minimal — no changes to `@expo/eas-build-job` schemas or the `www` workflow compilation are needed.

The env var propagates through the existing env chain (`jobs.<job>.env` → `job.builderEnvironment.env` → `BuildContext.env` → step `env` parameter), so Launch sets it once per job and all steps that resolve package managers pick it up automatically.

# How

- `resolvePackageManager()` gets two new optional parameters, `env` and `logger`.
- When lockfile detection returns no known package manager, the resolver reads `env?.EAS_FALLBACK_PACKAGE_MANAGER`, validates it against the `PackageManager` enum, and returns the chosen manager. Invalid values log a warning and fall back to `yarn` — user typos don't fail the build.
- Five call sites updated to forward the `env` / `logger` they already hold:
  - `BuildContext.packageManager` getter
  - `eas/install_node_modules`, `eas/prebuild`, `eas/eager_bundle`, `eas/export` step functions

# Test Plan

Automation tests in `packages/build-tools/src/utils/__tests__/packageManager.test.ts` cover:
- detection still wins when a lockfile exists (env var is ignored even if invalid)
- empty / unset env var falls back to `yarn` silently
- valid env var values (`bun`, `pnpm`, …) are honored
- invalid env var values fall back to `yarn` and log a warning with the offending value and env var name
- no logger passed → falls back silently without throwing

### DIFF
--- a/packages/build-tools/src/utils/packageManager.ts
+++ b/packages/build-tools/src/utils/packageManager.ts
@@ -1,5 +1,6 @@
 import { type bunyan } from '@expo/logger';
 import * as PackageManagerUtils from '@expo/package-manager';
+import { type BuildStepEnv } from '@expo/steps';
 import spawnAsync from '@expo/turtle-spawn';
 import semver from 'semver';
 import { z } from 'zod';
@@ -11,7 +12,16 @@ export enum PackageManager {
   BUN = 'bun',
 }
 
-export function resolvePackageManager(directory: string): PackageManager {
+export function resolvePackageManager(
+  directory: string,
+  {
+    env,
+    logger,
+  }: {
+    env?: BuildStepEnv;
+    logger?: bunyan;
+  } = {}
+): PackageManager {
   try {
     const manager = PackageManagerUtils.resolvePackageManager(directory);
     if (manager === 'npm') {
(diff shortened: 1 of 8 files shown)


sjchmiela (changes requested):
If I'm reading all the context correctly, we mostly want to do this for Launch — we don't expect users to actually WANT to run workflows with no lockfiles and provide the fallback package manager, right?

If so, I think it could be just an environment variable? Far less code to maintain

packages/build-tools/src/utils/packageManager.ts
```diff
@@ -20,12 +30,23 @@ export function resolvePackageManager(directory: string): PackageManager {
...
+  } catch {}
+
+  const fallback = env?.EAS_FALLBACK_PACKAGE_MANAGER;
+  if (fallback) {
+    const parsed = z.enum(PackageManager).safeParse(fallback);
+    if (parsed.success) {
+      return parsed.data;
+    }
+    const allowed = Object.values(PackageManager).join(', ');
+    logger?.warn(
```
sjchmiela: I think we can throw a validation error? We only expect it to be used by us so I think we can be strict. This also lets us skip passing in the `logger`.

packages/build-tools/src/utils/packageManager.ts
```diff
@@ -11,7 +12,16 @@ export enum PackageManager {
...
 }
 
-export function resolvePackageManager(directory: string): PackageManager {
+export function resolvePackageManager(
+  directory: string,
+  {
+    env,
+    logger,
+  }: {
+    env?: BuildStepEnv;
```
sjchmiela: Let's make this be required so we don't forget to pass it ever?

sjchmiela (approved):

packages/build-tools/src/utils/packageManager.ts
sjchmiela: ~validation error~ — user error

hSATAC: That's a smart way to omit the logger passing. Will do.

PR #3611, opened 2026-04-16
[build-tools] Reuse target_names to avoid duplicate iOS credentials preparation

hSATAC:
# Why

`eas/configure_ios_version` was preparing iOS credentials again just to recover target names, even when `eas/configure_ios_credentials` had already done that work.

This PR removes that duplicate work so credentials-backed iOS builds can move through the version step faster.

Related: `ENG-20324`

# How

- add an optional `target_names` input to `eas/configure_ios_version`
- make `target_names` take priority so the step can skip the second credentials preparation pass
- add a `target_names` output to `eas/configure_ios_credentials`
- wire the built-in iOS credentials workflow to pass `configure_ios_credentials.target_names` into `configure_ios_version`
- preserve the existing credentials fallback behavior when `target_names` is not provided

# Test Plan

- `cd packages/build-tools && yarn test src/steps/functions/__tests__/configureIosVersion.test.ts`
- `cd packages/build-tools && yarn test src/steps/functions/__tests__/configureIosCredentials.test.ts`
- `cd packages/build-tools && yarn test src/steps/functionGroups/__tests__/build.test.ts`

### DIFF
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -255,7 +255,6 @@ function createStepsForIosBuildWithCredentials({
       workingDirectory,
     }),
     configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx, {
-      id: 'configure_ios_version',
       workingDirectory,
       callInputs: {
         target_names: '${ steps.configure_ios_credentials.target_names }',
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -251,10 +251,15 @@ function createStepsForIosBuildWithCredentials({
     installPods,
     configureEASUpdate,
     configureIosCredentialsFunction().createBuildStepFromFunctionCall(globalCtx, {
+      id: 'configure_ios_credentials',
       workingDirectory,
     }),
     configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx, {
+      id: 'configure_ios_version',
       workingDirectory,
+      callInputs: {
+        target_names: '${ steps.configure_ios_credentials.target_names }',
+      },
     }),
     ...(shouldUseEagerBundle(globalCtx.staticContext.metadata)
       ? [
--- a/packages/build-tools/src/steps/functionGroups/build.ts
+++ b/packages/build-tools/src/steps/functionGroups/build.ts
@@ -257,7 +257,7 @@ function createStepsForIosBuildWithCredentials({
     configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx, {
       workingDirectory,
       callInputs: {
-        target_names: '${ steps.configure_ios_credentials.target_names }',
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/steps/functions/configureIosVersion.ts
```diff
@@ -71,6 +63,31 @@ export function configureIosVersionFunction(): BuildFunction {
         return;
       }
 
+      const targetNamesInput = inputs.target_names.value;
+      let targetNames: string[];
+      if (targetNamesInput !== undefined) {
+        if (
+          !Array.isArray(targetNamesInput) ||
+          targetNamesInput.some(targetName => typeof targetName !== 'string')
```
sjchmiela, line 71: Would it be easier to use Zod for this?

packages/build-tools/src/steps/functions/configureIosVersion.ts
```diff
@@ -71,6 +63,31 @@ export function configureIosVersionFunction(): BuildFunction {
...
+        ) {
+          throw new Error('"target_names" input must be an array of strings.');
```
sjchmiela, line 73: Let's throw a system error?

packages/build-tools/src/steps/functions/configureIosVersion.ts
```diff
@@ -60,7 +52,7 @@ export function configureIosVersionFunction(): BuildFunction {
         (inputs.app_version.value as string | undefined) ?? job.version?.appVersion;
       if (appVersion && !semver.valid(appVersion)) {
         throw new Error(
```
sjchmiela, line 54: is this a user error?

packages/build-tools/src/steps/functionGroups/build.ts
```diff
@@ -251,10 +251,14 @@ function createStepsForIosBuildWithCredentials({
     installPods,
     configureEASUpdate,
     configureIosCredentialsFunction().createBuildStepFromFunctionCall(globalCtx, {
+      id: 'configure_ios_credentials',
       workingDirectory,
     }),
     configureIosVersionFunction().createBuildStepFromFunctionCall(globalCtx, {
       workingDirectory,
+      callInputs: {
+        target_names: '${ steps.configure_ios_credentials.target_names }',
```
sjchmiela, line 260:
```suggestion
        target_names: '${{ steps.configure_ios_credentials.outputs.target_names }}',
```
would this work too? i think we tend to prefer the new interpolation now

sjchmiela (approved):
Awesome!!

packages/build-tools/src/steps/functions/configureIosVersion.ts
hSATAC: This input is also configurable by users, so a malformed value can come from user-authored YAML, not only internal wiring. Happy to make it UserError to match semantic, but SystemError would mis-classify user misconfig.

sjchmiela: Hmm yea… I mostly wanted to use SystemError to waive the build charge. Do what you think makes sense!

PR #3612, opened 2026-04-16
[build-errors] - log all build errors with enriched metadata

AbbanMustafa:
# Why

Error logs sent from build workers currently only fire for `UNKNOWN_ERROR` builds and include minimal context (build ID, message, phase, error code). This means we have no visibility into known error types like `INVALID_PODFILE` or `XCODE_BUILD_ERROR`, and can't easily group or filter errors by platform, SDK version, or app.

# How

  Two changes to the error logging in BuildService:

  1. Removed the `UNKNOWN_ERROR` gate — all build errors are now logged.
  2. Added metadata fields to the error log payload:
    - `appId` — link errors to specific projects
    - `buildProfile` — distinguish production vs preview vs development
    - `appName` — human-readable app name
    - `appIdentifier` — bundle ID (iOS) / package name (Android)
    - `distribution` — store vs internal vs simulator
    - `cliVersion` — detect CLI version-specific issues

Datadog cost impact: Logging all errors (not just unknown) adds one entry per failed build, which is a small fraction of total builds. The additional metadata fields have no extra cost. Datadog pricing is per log event, not per attribute. Adding more attributes to an existing log entry is free.

# Test Plan

  - The downstream consumer (www `turtle-builds/error-logs` endpoint) ignores unknown fields, so the new fields are safe to ship before www is updated to index them
  - Once deployed, verify in Datadog that error logs contain the new fields by querying `service:eas-build` and inspecting log attributes


douglowder (approved):

sjchmiela (approved):

packages/worker/src/service.ts
```diff
@@ -350,6 +353,18 @@ export default class BuildService {
...
               errorCode: err.errorCode,
+              tags: {
+                platform: job.platform,
+                workflow: job.type,
+                sdk_version: metadata?.sdkVersion ?? null,
+                react_native_version: metadata?.reactNativeVersion ?? null,
+                app_id: job.appId ?? null,
+                build_profile: metadata?.buildProfile ?? null,
+                app_name: metadata?.appName ?? null,
+                app_identifier: metadata?.appIdentifier ?? null,
```
sjchmiela, line 364: should we log these?

sjchmiela (changes requested):

packages/worker/src/service.ts
sjchmiela: ah ok i see the description. do we need to worry about PII or something?

sjchmiela (approved):

PR #3629, opened 2026-04-24
[build-tools] - Gradle build profile report with task execution breakdown

AbbanMustafa:
# Why

  We want to give developers visibility into which Gradle tasks are consuming the most build time. This helps them and us identify optimization opportunities, especially when evaluating Gradle caching effectiveness and which modules we can improve. 

# How

  - Added `--profile` to the gradlew command, which tells Gradle to generate an HTML profile report. We parse this report and display the data in a build phase, broken down by module.
  - Added `parseGradleProfile()` which reads the HTML report and extracts task path, duration, and result from the Task Execution table
  - Added a `GRADLE_BUILD_PROFILE` build phase that logs a formatted table with:
    - Tasks grouped by module, with individual tasks nested underneath using tree characters
    - Duration, % of total time, result (executed/up-to-date/skipped/etc.), and a proportional bar chart
    - Tasks under 1 second filtered out to reduce noise
    - Summary line with total task count, cached/up-to-date count, and total time
  - Removed the previous custom trace init script and trace parsing/upload plumbing
  - Simplified `BuildResult` to return `Artifacts` directly since `gradleProfileTasks` is no longer passed up
  - Everything is gated behind `EXPERIMENTAL_GRADLE_PROFILE=1` env var

# Test Plan
- Validated locally with `EXPERIMENTAL_GRADLE_PROFILE=1` — profile table shows in build logs with correct task grouping and durations
- Validated with `EAS_GRADLE_CACHE=1` — module totals and executed tasks display correctly

Screenshot 2026-05-07 at 5 53 00 PM


byCedric:
Big fan of this, after testing Android builds in Launch and seeing it's roughly 5x as slow for the default Expo template (SDK 54) - amounting for a total of 15m12s _just for the gradle step_

thank-you

packages/worker/src/build.ts
```diff
@@ -83,6 +91,19 @@ export async function build({
 
     analytics.logEvent(Event.WORKER_BUILD_SUCCESS, {});
 
+    if (job.platform === Platform.ANDROID && ctx.env.EXPERIMENTAL_GRADLE_PROFILE === '1') {
```
sjchmiela: i've been thinking we may not need to prefix flags with EXPERIMENTAL -- until we document flags i think we can consider them experimental?

packages/worker/src/build.ts
```diff
@@ -83,6 +91,19 @@ export async function build({
...
+      await ctx.runBuildPhase(BuildPhase.GRADLE_BUILD_PROFILE, async () => {
+        const androidDir = path.join(ctx.getReactNativeProjectDirectory(), 'android');
+        const profileTasks = await parseGradleProfile(androidDir, logger);
+        if (profileTasks && profileTasks.length > 0) {
+          const report = formatGradleProfileReport(profileTasks);
+          for (const line of report.split('\n')) {
+            logger.info(line);
```
sjchmiela: we could also `logger.info(report)`? or no?

packages/build-tools/src/steps/utils/android/gradle.ts
```diff
@@ -23,18 +23,29 @@ export async function runGradleCommand({
...
-      if (!line || /^\.+$/.exec(line)) {
-        return null;
-      } else {
-        return line;
-      }
-    },
-    env: { ...env, ...extraEnv, LC_ALL: 'C.UTF-8' },
-  });
+
+  const profileFlag = env['EXPERIMENTAL_GRADLE_PROFILE'] === '1' ? '--profile' : '';
```
sjchmiela: does this impact build performance? could we add it to all builds?

packages/build-tools/src/android/gradle.ts
```diff
@@ -103,6 +113,218 @@ function resolveVersionOverridesEnvs(ctx: BuildContext<Job>): Env {
   return extraEnvs;
 }
 
+export interface GradleProfileTask {
+  path: string;
+  durationMs: number;
+  result: string;
+}
+
+export async function parseGradleProfile(
```
sjchmiela: let's have a separate utility for parsing gradle profiles?

packages/build-tools/src/android/gradle.ts
```diff
@@ -103,6 +113,218 @@ function resolveVersionOverridesEnvs(ctx: BuildContext<Job>): Env {
...
+
+  if (!htmlFile) {
+    logger?.info('No Gradle profile HTML found in %s', profileDir);
+    return null;
+  }
+
+  const html = await fs.readFile(path.join(profileDir, htmlFile), 'utf8');
+
+  // Find the "Task Execution" section - it's the last <table> in the "Task Execution" tab
+  const taskExecMatch = html.match(/<h2[^>]*>\s*Task Execution\s*<\/h2>([\s\S]*?)(?:<\/div>)/i);
```
sjchmiela: wouldn't it be nicer to parse the profile as an xml?

packages/build-tools/src/android/gradle.ts
```diff
@@ -103,6 +113,218 @@ function resolveVersionOverridesEnvs(ctx: BuildContext<Job>): Env {
...
+  }
+
+  // Compute totals from individual tasks only (avoid double-counting)
+  const totalMs = individualTasks.reduce((sum, t) => sum + t.durationMs, 0);
+  const maxMs = totalMs || 1;
+
+  const nameWidth = Math.max(4, ...rows.map(r => r.displayName.length)) + 2;
+  const barMaxWidth = 20;
+
+  const header =
```
sjchmiela: @hSATAC was formatting a table manually for xclogparsing too… i wonder if there's some kind of utility we could pull in / abstract some logic…

sjchmiela (changes requested):

packages/worker/src/build.ts
AbbanMustafa: bunyan escapes newlines, it will all log on one line.

sjchmiela: what i meant is that i think we might be ok with this looking like this (package.json has newlines and prints in one… log entry)

Zrzut ekranu 2026-05-18 o 17 53 57

AbbanMustafa: ok if thats fine, we can do that

packages/build-tools/src/android/gradle.ts
AbbanMustafa: thanks I moved over to using fast-xml-parser!

packages/build-tools/src/steps/utils/android/gradle.ts
AbbanMustafa: it does not, yes we can and would be nice to enable by default.

packages/worker/src/build.ts
```diff
@@ -83,6 +91,17 @@ export async function build({
...
     analytics.logEvent(Event.WORKER_BUILD_SUCCESS, {});
 
+    if (job.platform === Platform.ANDROID) {
+      await ctx.runBuildPhase(BuildPhase.GRADLE_BUILD_PROFILE, async () => {
+        const androidDir = path.join(ctx.getReactNativeProjectDirectory(), 'android');
+        const profileTasks = await parseGradleProfile(androidDir, logger);
+        if (profileTasks && profileTasks.length > 0) {
+          const report = formatGradleProfileReport(profileTasks);
+          logger.info(report);
+        }
```
sjchmiela: shall we wrap with try-catch so if something fails here we can log the error and mark the phase as skipped? bonus points for logging the error with sentry

sjchmiela: haven't looked it yet but i bet `parseGradleProfile` is probably a big try-catch, but i think maybe it's nicer to have the caller catch the error?

packages/build-tools/src/android/gradle.ts
AbbanMustafa: I agree. Will most likely prefer to keep it this way for now and then we can later agree how we would like to parse and show these in the UI

packages/build-tools/src/android/gradleProfile.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  if (!htmlFile) {
+    logger?.info('No Gradle profile HTML found in %s', profileDir);
+    return null;
+  }
+
+  const html = await fs.readFile(path.join(profileDir, htmlFile), 'utf8');
+
+  // Locate the <h2>Task Execution</h2> heading and extract its <table>
+  const headingMatch = html.match(/<h2[^>]*>\s*Task Execution\s*<\/h2>/i);
```
sjchmiela, line 30: ugh i thought fast-xml-parser would give us what cheerio would do…

i'll let you decide what to do

packages/build-tools/src/android/gradleProfile.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  const files = await fs.readdir(profileDir);
+  const htmlFile = files
+    .filter(f => f.startsWith('profile-') && f.endsWith('.html'))
+    .sort()
+    .pop();
+
+  if (!htmlFile) {
+    logger?.info('No Gradle profile HTML found in %s', profileDir);
```
sjchmiela: so if we allowed `parseGradleProfile` to throw we wouldn't need to pass in `logger` and optionally call it. instead we could throw `SystemError` and catch it and log it once

packages/build-tools/src/android/gradleProfile.ts
```diff
@@ -0,0 +1,237 @@
...
+    isArray: name => name === 'tr' || name === 'td',
+    trimValues: true,
+  });
+
+  const parsed = parser.parse(tableHtml);
+  const rows: any[] = parsed?.table?.tbody?.tr ?? parsed?.table?.tr ?? [];
+
+  const tasks: GradleProfileTask[] = [];
+  for (const row of rows) {
+    const cells: any[] = row?.td;
```
sjchmiela:
```suggestion
    const cells: unknown[] = row?.td;
```

sjchmiela (approved):

PR #3162, opened 2025-09-04
CLI debug log should respect DEBUG=eas:*

douglowder:
# Why

Please describe the motivation for this PR, and link to relevant GitHub issues, forums posts, or feature requests.

# How

How did you build this feature or fix this bug and why?

# Test Plan

Please describe how you tested this change and how a reviewer could reproduce your test, especially if this PR does not include automated tests! If possible, please also provide terminal output and/or screenshots demonstrating your test/reproduction.

PR #3630, opened 2026-04-24
[eas-cli] add missing debug runtime dependency

mariusud:
# Why
 A dependency on debug was introduced in https://github.com/expo/eas-cli/pull/3162 but not added to `packages/eas-cli/package.json`.

 Since `packages/eas-cli/src/log.ts` now requires `debug` at runtime, this causes a `MODULE_NOT_FOUND` error.

# How

Modify package.json to include debug

# Test Plan


mariusud:
@douglowder

douglowder:
@mariusud thanks for flagging this. I didn't hit this issue, and our CI tests didn't either. I'll investigate.

pataar:
@douglowder Any update?

douglowder:
> @mariusud thanks for flagging this. I didn't hit this issue, and our CI tests didn't either. I'll investigate.

We are not seeing any issue at the moment. Inspection of the yarn lockfile shows that the debug package is included. It is required by babel and many other packages. So I'm not sure why you are seeing a MODULE_NOT_FOUND. Did you use yarn v4, or some other package manager? I suspect there is some difference from our environment, as your yarn lockfile diff is very large.

I do agree that the dependency should be added, so I'll have a look at doing that in the Expo environment and make sure tests pass and there are no regressions.

douglowder:
Closing, this will be fixed in #3741

PR #3631, opened 2026-04-24
[worker] - wrap error log fetch in try-catch to prevent unhandled rejections

AbbanMustafa:
# Why

The `turtleFetch` call to `turtle-builds/error-logs` can throw a network-level `FetchError` (e.g. DNS failure). Since `startBuildInternal` is called without `await`, this becomes an unhandled promise rejection. 

This was surfaced after removing the `UNKNOWN_ERROR` gate — now that all errors are logged, the fetch runs in more cases (including integration tests where the CI can't reach `api.expo.dev`).

# How

Wrapped the `turtleFetch` call in a try-catch. The error log to Datadog is best-effort — a failed fetch should never block `finishError` or crash the worker.

# Test Plan

  - CI integration tests should pass (no more unhandled `FetchError`)

### DIFF
--- a/packages/worker/jest/integration-setup.ts
+++ b/packages/worker/jest/integration-setup.ts
@@ -14,3 +14,4 @@ if (!process.env.LISTENING_TO_UNHANDLED_REJECTION) {
 
 // Always mock:
 jest.mock('../src/CacheManager');
+jest.mock('../src/utils/turtleFetch');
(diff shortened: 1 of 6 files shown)


sjchmiela (approved):
Can we also make sure tests don't do real http requests? Also let's make sure the test url is api.expo.test not dev

PR #3634, opened 2026-04-27
[eas-cli] stop device run session when time out happens

szdziedzic:
# Why

When the readiness polling loop in `eas simulator:start` timed out (15 min) or threw, the device run session was left running on the server. Users had to remember to clean it up manually with `eas simulator:stop --id <id>`.

# How

On timeout or any failure during the readiness polling loop, call the server-side `ensureDeviceRunSessionStopped` mutation. The mutation is idempotent (no-op if the session is already stopped), so we don't need a client-side liveness flag — addresses [@sjchmiela's review feedback](https://github.com/expo/eas-cli/pull/3634#discussion_r3155123755).

The call is wrapped in a best-effort helper that catches transport-level errors and logs them as warnings, so cleanup failures don't mask the original error being thrown to the user.

`eas simulator:stop` is migrated to the same mutation for consistency.

# Test Plan

- [ ] `eas simulator:start --platform ios --type agent-device` succeeds, then `eas simulator:stop --id <id>` returns a stopped session
- [ ] Force the polling loop to time out (e.g. point readiness check at logs that never produce the daemon URL) and confirm the session ends up stopped on the server
- [ ] Trigger a failure mid-poll (e.g. session errored) and confirm the original error is surfaced while cleanup still runs
- [ ] `yarn typecheck`, `yarn lint`, `yarn fmt:check` pass

### DIFF
--- a/packages/eas-cli/src/commands/simulator/start.ts
+++ b/packages/eas-cli/src/commands/simulator/start.ts
@@ -2,6 +2,7 @@ import { Flags } from '@oclif/core';
 
 import { getBareJobRunUrl } from '../../build/utils/url';
 import EasCommand from '../../commandUtils/EasCommand';
+import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
 import { EASNonInteractiveFlag } from '../../commandUtils/flags';
 import {
   AppPlatform,
@@ -76,6 +77,7 @@ export default class SimulatorStart extends EasCommand {
 
     const createSpinner = ora('🚀 Creating device run session').start();
     let deviceRunSessionId: string;
+    let jobRunUrl: string;
     try {
       const session = await DeviceRunSessionMutation.createDeviceRunSessionAsync(graphqlClient, {
         appId: projectId,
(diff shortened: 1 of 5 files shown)


szdziedzic:
* **#3634** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -152,6 +162,22 @@ export default class SimulatorStart extends EasCommand {
...
 
+async function stopDeviceRunSessionSafelyAsync(
+  graphqlClient: ExpoGraphqlClient,
+  deviceRunSessionId: string
+): Promise<void> {
+  try {
+    await DeviceRunSessionMutation.stopDeviceRunSessionAsync(graphqlClient, deviceRunSessionId);
+  } catch (err) {
+    // Cleanup is best-effort; surface the failure but don't mask the original error.
+    Log.warn(
```
sjchmiela, line 174: maybe we should make a mutation like "ensure stopped" so we don't throw if it's already stopped? or maybe we should throw an error we can catch and handle here as noop if we try to stop a stopped session

then we wouldn't need to worry about `maybeIsAlive`

sjchmiela (approved):

packages/eas-cli/src/commands/simulator/start.ts
szdziedzic: added ensureDeviceRunSessionStopped mutation

szdziedzic: 🫡

PR #3635, opened 2026-04-27
[build-tools] Add eas/run_maestro_tests step with smart retry

hSATAC:
# Why

After porting the Maestro step to TS (#3641), we want to stop re-running the entire flow set on every retry. With dozens of flows a single flaky one masks other failures and inflates the whole run. Smart retry re-runs only the flows that actually failed.

Linear: [ENG-19927](https://linear.app/expo/issue/ENG-19927). Stacked on #3641.

# How

After each failed attempt, parse the per-attempt JUnit XML, find the failing testcase names, map them back to flow file paths, and pass only those paths to the next `maestro test` invocation.

- `parseFailedFlowsFromJUnit` returns the failed-flow subset, or `null` when the result can't be trusted (malformed XML, duplicate testcase names, unknown flow). On `null` the loop falls back to dumb retry, preserving the previous behavior.
- Replaces "copy latest attempt's JUnit" with `mergeJUnitReports`, which keeps the latest result *per flow* across attempts. The merged final report reflects each flow's best outcome (a flow that flaked once and passed on retry shows as passed).

The `flow_name → flow_file_path` map is built from Maestro's `ai-*.json` debug output. PR #3639 (next in stack) replaces this with a self-built map after Maestro 2.5.0 broke this assumption.

# Test Plan

- Unit tests in `maestroResultParser.test.ts` for `parseFailedFlowsFromJUnit` (happy path, all-passing, collisions, malformed XML, partial parse) and `mergeJUnitReports` (per-flow latest, fallback on data errors).
- New cases in `runMaestroTests.test.ts` exercising the smart-retry loop.
- Local end-to-end smoke test against the playground.

### DIFF
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -1,4 +1,4 @@
-import { XMLParser } from 'fast-xml-parser';
+import { XMLBuilder, XMLParser, XMLValidator } from 'fast-xml-parser';
 import fs from 'fs/promises';
 import path from 'path';
 import { z } from 'zod';
@@ -41,11 +41,9 @@ const xmlParser = new XMLParser({
   isArray: name => ['testsuite', 'testcase', 'property'].includes(name),
 });
 
-// Internal helper — not exported. Parses a single JUnit XML file.
-async function parseJUnitFile(filePath: string): Promise<JUnitTestCaseResult[]> {
+function parseJUnitContent(content: string): JUnitTestCaseResult[] {
   const results: JUnitTestCaseResult[] = [];
   try {
-    const content = await fs.readFile(filePath, 'utf-8');
     const parsed = xmlParser.parse(content);
 
     const testsuites = parsed?.testsuites?.testsuite;
@@ -109,9 +107,18 @@ async function parseJUnitFile(filePath: string): Promise<JUnitTestCaseResult[]> 
       }
     }
   } catch {
-    // Skip malformed XML files
+    // Malformed XML — return whatever we collected before the parser bailed.
   }
   return results;
+}
+
+async function parseJUnitFile(filePath: string): Promise<JUnitTestCaseResult[]> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    return parseJUnitContent(content);
+  } catch {
+    return [];
+  }
 }
 
 export async function parseJUnitTestCases(junitDirectory: string): Promise<JUnitTestCaseResult[]> {
@@ -127,11 +134,10 @@ export async function parseJUnitTestCases(junitDirectory: string): Promise<JUnit
     return [];
   }
 
-  const results: JUnitTestCaseResult[] = [];
-  for (const xmlFile of xmlFiles) {
-    results.push(...(await parseJUnitFile(path.join(junitDirectory, xmlFile))));
-  }
-  return results;
+  const perFile = await Promise.all(
+    xmlFiles.map(f => parseJUnitFile(path.join(junitDirectory, f)))
+  );
+  return perFile.flat();
 }
 
 const FlowMetadataFileSchema = z.object({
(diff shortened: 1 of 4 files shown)


hSATAC:
* **#3639** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3639?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3635** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3641** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -241,20 +261,44 @@ export function createRunMaestroTestsBuildFunction(): BuildFunction {
...
             sourceDir: junitReportDirectory,
             outputPath: finalReportPath,
           });
-        } catch (copyErr: any) {
-          throw new SystemError('Failed to produce final_report_path', { cause: copyErr });
+        } catch (mergeErr: any) {
+          if (isFilesystemError(mergeErr)) {
+            throw new SystemError('Failed to write final_report_path', { cause: mergeErr });
+          }
+          logger.warn({ err: mergeErr }, 'Smart merge failed; falling back to copy-latest.');
```
sjchmiela, line 272: let's make it more user-understandable

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -361,3 +593,25 @@ export async function copyLatestAttemptXml(args: {
...
+  if (!path.isAbsolute(flowFilePath)) {
+    return flowFilePath;
+  }
+
+  // Resolve symlinks (e.g., /tmp -> /private/tmp on macOS) for consistent comparison
+  let resolvedRoot = projectRoot;
+  let resolvedFlow = flowFilePath;
+  try {
+    resolvedRoot = await fs.realpath(projectRoot);
+  } catch {}
```
sjchmiela, line 595: i like `asyncResult(fs.realpath(projectRoot))` for stuff like this

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -301,33 +307,258 @@ export async function parseMaestroResults(
...
+    testcasesByName: Map<string, unknown[]>;
+  }
+  const contents = await Promise.all(
+    xmlFiles.map(async f => ({
+      filename: f,
+      content: await fs.readFile(path.join(args.sourceDir, f), 'utf-8'),
+    }))
+  );
+  const fileGroups: FileGroup[] = [];
+  const skippedFiles: string[] = [];
```
sjchmiela: should we just throw an error as soon as we have to skip a file or find an error? does this happen often or is this just preventive

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -301,33 +307,258 @@ export async function parseMaestroResults(
...
+  for (const { filename, content } of contents) {
```
sjchmiela, line 452: what would your parser do if instead of merging in this smart way you did

```
<testsuites>
  <testsuite>
    {...testsuitesfromfirstfile}
    {...testsuitesfromsecondfile}
    {...testsuitesfromthirdfile}
```
?

sjchmiela (approved):
let's make this opt-in?

packages/build-tools/src/steps/functions/maestroTests.ts
hSATAC: I think this is covered in the next PR: https://github.com/expo/eas-cli/pull/3639/changes#diff-fa4d0db2be5715542edd5cf992cc4aaedbc5349a1663bd70164767b85707f2f7R262-R267

packages/build-tools/src/steps/functions/maestroResultParser.ts
hSATAC: This is removed in the next PR as well. I'll skip it here.

packages/build-tools/src/steps/functions/maestroResultParser.ts
hSATAC: This is just preventive programming. Let me simply it here.

packages/build-tools/src/steps/functions/maestroResultParser.ts
hSATAC: One thing to clarify first — this merge isn’t for our own parsers. All of our internal parsers read the per-attempt `attempt-*.xml` files directly from `junit_report_directory`. The merged `final_report_path` only exists as a human-facing artifact for customers who download it and run their own analysis.

Given that, two reasons I’d prefer to keep the smart merge:

1. **JUnit shape** — Concatenating attempt-N `<testsuite>`s under one `<testsuites>` creates an ambiguous shape (e.g. `flow-a, flow-b, flow-c, flow-b, flow-b` for retries). It’s not a shape that Maestro emits.

2. **Backwards compatibility** — Changing this to concat would likely break existing customer aggregation logic that depends on the current `final_report_path` shape.

hSATAC:
Tested locally with 2.4 & 2.5.1 
Confirmed it works even without smart retry. 
Merging this now.

PR #3637, opened 2026-04-27
add simulator flag to build dev and run

mmichels-brex:
## Summary
- add a --simulator flag to eas build:dev and eas build:run for iOS dev-client/simulator installs
- support both --simulator <name-or-udid> and bare --simulator, which prompts for an available simulator
- reuse the same optional-value flag preprocessing and shared simulator run target path for both commands

## Validation

https://github.com/user-attachments/assets/da061a03-59b0-45c1-ab80-dc528f8c4a62


packages/eas-cli/src/commands/build/dev.ts
```diff
@@ -51,6 +51,9 @@ export default class BuildDev extends EasCommand {
       description: 'Install and run the development build without starting the bundler server.',
       default: false,
     }),
+    'simulator-id': Flags.string({
```
kadikraman: Could we add this flag to the `build:run` command as well for parity?

mmichels-brex: done!

douglowder (approved):

kadikraman (approved):

PR #3638, opened 2026-04-27
[eas-cli] Add observe:logs command for viewing custom events

douglowder:
# Why

Add new command to retrieve custom events for Expo Observe.

```
USAGE
  $ eas observe:logs [EVENTNAME] [--platform android|ios] [--after <value>] [--limit <value>] [--start <value> |
    --days <value>] [--end <value> | ] [--app-version <value>] [--update-id <value>] [--session-id <value>]
    [--all-events] [--project-id <value>] [--json] [--non-interactive]

ARGUMENTS
  [EVENTNAME]  Custom event name to filter by

FLAGS
  --after=<value>        Cursor for pagination. Use the endCursor from a previous query to fetch the next page.
  --all-events           When no event name argument is provided, list all events across all event names instead of a
                         summary of event names + counts.
  --app-version=<value>  Filter by app version
  --days=<value>         Show events from the last N days (mutually exclusive with --start/--end)
  --end=<value>          End of time range (ISO date)
  --json                 Enable JSON output, non-JSON messages will be printed to stderr. Implies --non-interactive.
  --limit=<value>        The number of items to fetch each query. Defaults to 10 and is capped at 100.
  --non-interactive      Run the command in non-interactive mode.
  --platform=<option>    Filter by platform
                         <options: android|ios>
  --project-id=<value>   EAS project ID (defaults to the project ID of the current directory)
  --session-id=<value>   Filter by session ID
  --start=<value>        Start of time range (ISO date)
  --update-id=<value>    Filter by EAS update ID

DESCRIPTION
  display individual custom events (logs) emitted by the app, filtered by the event name in the argument. With no
  arguments, a list of the available event names and associated event counts is returned.


```

# How

- Add new GraphQL query and fragment, and new formatter for the retrieved custom events
- New command `observe:logs`, designed to be similar to `observe:events`
- Did some refactoring so that observe commands reuse methods for formatting results

The command can return a list of the different event names found and their counts, a list of events for a specific name, or a list of all events found:

```
  ┌───────────────────────────┬───────────────────────────────────────────┐                                             
  │        invocation         │                  result                   │                                             
  ├───────────────────────────┼───────────────────────────────────────────┤                                             
  │ observe:logs              │ event-name summary table (names + counts) │                                             
  ├───────────────────────────┼───────────────────────────────────────────┤                                             
  │ observe:logs --all-events │ full list of events across all names      │                                             
  ├───────────────────────────┼───────────────────────────────────────────┤                                             
  │ observe:logs <event>      │ full list of events filtered by name      │                                             
  └───────────────────────────┴───────────────────────────────────────────┘  
```

# Test Plan

- New unit tests added
- Will test against Kadi's ingestion endpoint data

### DIFF
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -13,6 +13,7 @@ import {
   buildObserveCustomEventsJson,
   buildObserveCustomEventsTable,
 } from '../../observe/formatCustomEvents';
+import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
 import { resolveTimeRange } from '../../observe/startAndEndTime';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
@@ -88,22 +89,13 @@ export default class ObserveLogs extends EasCommand {
   async runAsync(): Promise<void> {
     const { flags, args } = await this.parse(ObserveLogs);
 
-    let projectId: string;
-    let graphqlClient;
-    if (flags['project-id']) {
-      projectId = flags['project-id'];
-      const ctx = await this.getContextAsync(
-        { contextDefinition: ObserveLogs.loggedInOnlyContextDefinition },
-        { nonInteractive: flags['non-interactive'] }
-      );
-      graphqlClient = ctx.loggedIn.graphqlClient;
-    } else {
-      const ctx = await this.getContextAsync(ObserveLogs, {
-        nonInteractive: flags['non-interactive'],
-      });
-      projectId = ctx.projectId;
-      graphqlClient = ctx.loggedIn.graphqlClient;
-    }
+    const { projectId, graphqlClient } = await resolveObserveCommandContextAsync({
+      command: this,
+      commandClass: ObserveLogs,
+      loggedInOnlyContextDefinition: ObserveLogs.loggedInOnlyContextDefinition,
+      projectIdOverride: flags['project-id'],
+      nonInteractive: flags['non-interactive'],
+    });
 
     if (flags.json) {
       enableJsonOutput();
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -64,6 +64,11 @@ export default class ObserveLogs extends EasCommand {
     'session-id': Flags.string({
       description: 'Filter by session ID',
     }),
+    'all-events': Flags.boolean({
+      description:
+        'When no event name argument is provided, list all events across all event names instead of a summary of event names + counts.',
+      default: false,
+    }),
     'project-id': Flags.string({
       description: 'EAS project ID (defaults to the project ID of the current directory)',
     }),
@@ -113,7 +118,7 @@ export default class ObserveLogs extends EasCommand {
         : AppObservePlatform.Ios
       : undefined;
 
-    if (!args.eventName) {
+    if (!args.eventName && !flags['all-events']) {
       const { names, isTruncated } = await ObserveQuery.customEventNamesAsync(graphqlClient, {
         appId: projectId,
         startTime,
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -89,6 +89,12 @@ export default class ObserveLogs extends EasCommand {
   async runAsync(): Promise<void> {
     const { flags, args } = await this.parse(ObserveLogs);
 
+    if (args.eventName && flags['all-events']) {
+      throw new Error(
+        '--all-events cannot be combined with an event name argument. Pass an event name to filter by it, or pass --all-events to list all events across all event names.'
+      );
+    }
+
     const { projectId, graphqlClient } = await resolveObserveCommandContextAsync({
       command: this,
       commandClass: ObserveLogs,
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -20,7 +20,8 @@ const DEFAULT_EVENTS_LIMIT = 10;
 
 export default class ObserveLogs extends EasCommand {
   static override hidden = true;
-  static override description = 'display individual custom events (logs) emitted by the app';
+  static override description =
+    'display individual custom events (logs) emitted by the app, filtered by the event name in the argument. With no arguments, a list of the available event names and associated event counts is returned.';
 
   static override args = {
     eventName: Args.string({
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -9,6 +9,8 @@ import { fetchObserveCustomEventsAsync } from '../../observe/fetchCustomEvents';
 import {
   buildObserveCustomEventNamesJson,
   buildObserveCustomEventNamesTable,
+  buildObserveCustomEventsEmptyWithSuggestionsJson,
+  buildObserveCustomEventsEmptyWithSuggestionsTable,
   buildObserveCustomEventsJson,
   buildObserveCustomEventsTable,
 } from '../../observe/formatCustomEvents';
@@ -148,6 +150,32 @@ export default class ObserveLogs extends EasCommand {
       updateId: flags['update-id'],
       sessionId: flags['session-id'],
     });
+
+    if (args.eventName && events.length === 0) {
+      const { names, isTruncated } = await ObserveQuery.customEventNamesAsync(graphqlClient, {
+        appId: projectId,
+        startTime,
+        endTime,
+        platform,
+      });
+
+      if (flags.json) {
+        printJsonOnlyOutput(
+          buildObserveCustomEventsEmptyWithSuggestionsJson(args.eventName, names, isTruncated)
+        );
+      } else {
+        Log.addNewLineIfNone();
+        Log.log(
+          buildObserveCustomEventsEmptyWithSuggestionsTable(args.eventName, names, {
+            daysBack,
+            startTime,
+            endTime,
+            isTruncated,
+          })
+        );
+      }
+      return;
+    }
 
     if (flags.json) {
       printJsonOnlyOutput(buildObserveCustomEventsJson(events, pageInfo));
--- a/packages/eas-cli/src/commands/observe/logs.ts
+++ b/packages/eas-cli/src/commands/observe/logs.ts
@@ -5,8 +5,11 @@ import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
 import { getLimitFlagWithCustomValues } from '../../commandUtils/pagination';
 import { AppObservePlatform } from '../../graphql/generated';
 import Log from '../../log';
+import { ObserveQuery } from '../../graphql/queries/ObserveQuery';
 import { fetchObserveCustomEventsAsync } from '../../observe/fetchCustomEvents';
 import {
+  buildObserveCustomEventNamesJson,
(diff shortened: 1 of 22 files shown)


douglowder:
* **#3638** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3638?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3638?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

kadikraman (commented):
Nice thanks for the PR, it works well!

I also added a resolver for `observe.customEventNames` which returns `{ eventName, count }` for the custom events for the given time period. I think we should include this because it will help with discoverability since the event names are user-chosen. Otherwise right now you need to run `observe:logs` with no event name to get an idea of what event names exist, but this is less useful if the event you're looking for is not in the first few pages.

So I would expect the actual usage to be more like: "list all the event names and counts in the given time period" and then you can dig deeper into the events you're interested in.

douglowder:
> So I would expect the actual usage to be more like: "list all the event names and counts in the given time period" and then you can dig deeper into the events you're interested in.

Thanks for the new GraphQL! I've adjusted the command to emit event names and counts by default. See the PR description for the new help menu and a table of the different behaviors.

packages/eas-cli/src/observe/formatUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+export function formatTimestamp(isoString: string): string {
+  const date = new Date(isoString);
+  return date.toLocaleDateString('en-US', {
+    year: 'numeric',
+    month: 'short',
+    day: 'numeric',
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    fractionalSecondDigits: 3,
```
tsapeta, line 31: [🏗️ **design**] Adding `second: '2-digit'` and `fractionalSecondDigits: 3` here changes the user-visible timestamp output for the already-shipped `observe:events` and `observe:metrics` commands — the snapshot diffs in `formatEvents.test.ts` and `formatMetrics.test.ts` show timestamps going from `Jan 15, 2025, 10:30 AM` to `Jan 15, 2025, 10:30:00.000 AM`.

Millisecond precision makes sense for individual log entries in `observe:logs`, but it's a lot of noise in the performance-metric event list. Consider either:
- keeping `formatTimestamp` at minute precision and adding a separate `formatLogTimestamp` for the logs table, or
- calling out this UX change explicitly in the PR description so it isn't a surprise.

packages/eas-cli/src/observe/formatUtils.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Format an ISO timestamp for display in event tables, including
+ * seconds and milliseconds.
+ */
+export function formatTimestamp(isoString: string): string {
+  const date = new Date(isoString);
+  return date.toLocaleDateString('en-US', {
```
tsapeta: [💡 suggestion] Hardcoding `'en-US'` is inconsistent with how the rest of eas-cli formats dates. A grep across `packages/eas-cli/src` shows ~15 call sites using `toLocaleString()` / `toLocaleDateString()` with no locale argument (e.g. `build/utils/formatBuild.ts`, `webhooks/formatWebhook.ts`, `submit/utils/summary.ts`, `commandUtils/builds.ts`) — those respect the user's system locale. Only `commandUtils/usageUtils.ts:60` and these observe formatters pin `'en-US'`.

Two reasonable directions:
- Drop the `'en-US'` argument here and in `formatDate` so observe matches the rest of the CLI and respects the user's locale.
- Or leave it for now, but note that this came from the existing `formatEvents.ts` — i.e. the PR is propagating the inconsistency, not introducing it. In that case a follow-up issue to align all the date formatting in one pass might be worth filing.

Not blocking — but since this PR is the moment the helper becomes shared, it's a cheap time to pick a direction.

packages/eas-cli/src/commands/observe/logs.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    const { daysBack, startTime, endTime } = resolveTimeRange(flags);
+
+    const platform = flags.platform
+      ? flags.platform === 'android'
+        ? AppObservePlatform.Android
+        : AppObservePlatform.Ios
+      : undefined;
+
+    if (!args.eventName && !flags['all-events']) {
```
tsapeta, line 118: [💡 suggestion] When both a positional `eventName` and `--all-events` are passed, `--all-events` is silently ignored — the code routes to the filtered events list. Given that the flag's description explicitly says "When no event name argument is provided, list all events…", consider rejecting the combination with an `EasCommandError` so users don't get unexpected behavior. (Alternatively, document it as "--all-events has no effect when an event name is provided" — but a hard error is friendlier.)

packages/eas-cli/src/commands/observe/logs.ts
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      Log.warn('EAS Observe is in preview and subject to breaking changes.');
+    }
+
+    const { daysBack, startTime, endTime } = resolveTimeRange(flags);
+
+    const platform = flags.platform
+      ? flags.platform === 'android'
+        ? AppObservePlatform.Android
+        : AppObservePlatform.Ios
```
tsapeta: [🏗️ **design**] This `=== 'android' ? Android : Ios` pattern silently maps any non-`'android'` value to `Ios`. Today the oclif `options` list (line 36) constrains input to the enum values lowercased, so the bug is latent — but the moment a third value is added to `AppObservePlatform` (e.g. `Web`), `--platform web` will be accepted by oclif and then mismapped to iOS here.

Safer to construct a `lower → enum` map once from `Object.values(AppObservePlatform)`, or do an explicit switch with an exhaustive default. Worth fixing in one place since `events.ts` has the same pattern.

packages/eas-cli/src/observe/formatCustomEvents.ts
```diff
@@ -0,0 +1,175 @@
...
+  events: AppObserveCustomEvent[],
+  pageInfo: PageInfo,
+  options?: BuildCustomEventsTableOptions
+): string {
+  if (events.length === 0) {
+    return chalk.yellow('No custom events found.');
+  }
+
+  const showEventName = !options?.eventName;
+  const hasSeverity = events.some(e => e.severityText);
```
tsapeta: [💡 suggestion] `hasSeverity` only checks `severityText`, but the row body falls back to `'-'` when `severityText` is missing. If an event has `severityNumber` set but no `severityText`, the severity column will be hidden entirely even though severity data is present. Either include `severityNumber` in the `hasSeverity` check, or render the number as a fallback.

packages/eas-cli/src/observe/formatCustomEvents.ts
```diff
@@ -0,0 +1,175 @@
...
+  return lines.join('\n');
+}
+
+export function buildObserveCustomEventNamesJson(
+  names: AppObserveCustomEventName[],
+  isTruncated: boolean
+): { names: Array<{ eventName: string; count: number }>; isTruncated: boolean } {
+  return {
+    names: names.map(n => ({ eventName: n.eventName, count: n.count })),
+    isTruncated,
```
tsapeta, line 235: [💡 suggestion] No `formatCustomEvents.test.ts` accompanies this new file, even though sibling `formatEvents.ts` has comprehensive table/JSON snapshot tests. The command-level `logs.test.ts` mocks these formatters, so the actual table layout — column visibility logic, severity fallback, totals header, truncation message — is currently untested. Worth adding parity with `formatEvents.test.ts`.

packages/eas-cli/src/observe/fetchCustomEvents.ts
```diff
@@ -0,0 +1,48 @@
...
+    ...(options.sessionId && { sessionId: options.sessionId }),
+  };
+
+  return await ObserveQuery.customEventListAsync(graphqlClient, {
+    appId,
+    filter,
+    first: options.limit,
+    ...(options.after && { after: options.after }),
+  });
+}
```
tsapeta, line 48: [🐧 nit] No `fetchCustomEvents.test.ts` either — `fetchMetrics.test.ts` exists for the parallel module. Probably low risk since the function is mostly filter-shape massaging, but a small test verifying the conditional spread (e.g. that `appUpdateId` isn't sent when `updateId` is undefined) would document the contract.

packages/eas-cli/src/commands/observe/logs.ts
```diff
@@ -0,0 +1,172 @@
...
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectId,
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  private static loggedInOnlyContextDefinition = {
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
```
tsapeta, line 91: [🐧 nit] The `project-id` / context branching here (lines 91-106) is duplicated almost verbatim from `events.ts`. Not blocking for this PR, but as `observe:*` commands keep growing, this is a good candidate to extract into a helper on `EasCommand` (e.g. `resolveProjectAndClientAsync`).

packages/eas-cli/src/commands/observe/logs.ts
douglowder: Done

packages/eas-cli/src/commands/observe/logs.ts
douglowder: Done

packages/eas-cli/src/commands/observe/logs.ts
douglowder: Done

packages/eas-cli/src/observe/fetchCustomEvents.ts
douglowder: Done

packages/eas-cli/src/observe/formatCustomEvents.ts
douglowder: Done

packages/eas-cli/src/observe/formatCustomEvents.ts
douglowder: Done

packages/eas-cli/src/observe/formatUtils.ts
douglowder: Done

packages/eas-cli/src/observe/formatUtils.ts
douglowder: Done

douglowder:
@tsapeta I addressed the issues you found. I also added a usability improvement -- if no events are returned by observe:logs, the command does the event names/counts query so the user can see what event names actually exist in the data (this allows correction of spelling errors, etc.).

```
$ easd observe:logs debug_warn_button_pressed
EAS Observe is in preview and subject to breaking changes.

No events found matching "debug_warn_button_pressed" for the last 60 days.

Available event names in this time range:

Event Name                  Count
--------------------------  -----
debug.warn_button_pressed   13   
debug.info_button_pressed   8    
debug.error_button_pressed  7    
debug.debug_button_pressed  6    
debug.fatal_button_pressed  6    
debug.trace_button_pressed  3    

```

tsapeta (approved):

douglowder:
### Merge activity

* **May 4, 8:24 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/expo/eas-cli/3638).
* **May 4, 8:24 PM UTC**: @douglowder merged this pull request with [Graphite](https://app.graphite.com/github/pr/expo/eas-cli/3638).

PR #3639, opened 2026-04-28
[build-tools] Fix ai-*.json is not guarenteed in Maestro 2.5

hSATAC:
# Why

Maestro 2.5.0 (released 2026-04-27) changed `TestDebugReporter.saveSuggestions` to early-return when a flow has no AI screen outputs. As a result, `ai-${flow_name}.json` is now only emitted for flows using AI commands (e.g. `assertWithAI`).

Two callers in `@expo/build-tools` relied on `ai-*.json` for the `flow_name → flow_file_path` mapping:

- **Smart retry** (`parseFailedFlowsFromJUnit`): lookup misses every time on 2.5.0+, so smart retry silently degrades to dumb retry — a functional regression of #3635 against the latest Maestro.
- **Test result reporting** (`parseMaestroResults`): the `path` on uploaded testCaseResults falls back to flow name. Cosmetic-only — `path` is read in the workflow detail UI as a hover tooltip.

Linear: [ENG-19927](https://linear.app/expo/issue/ENG-19927). Stacked on #3635.

# How

Replace the `ai-*.json` walk with a self-built `flow_name → flow_file_path` map (`buildFlowNameToPathMap`). The helper walks the user's `flow_paths` input, parses each flow YAML's top-level `name` (or falls back to file basename without extension), and produces a `name → path` map. Both `parseFailedFlowsFromJUnit` and `parseMaestroResults` consume it.

- Map values are project-root-relative paths (preserving today's GraphQL `path` semantics).
- Duplicate flow names produce a warning and disable smart retry for that run; consumers receive `null` and fall back to the previous degraded behavior.
- Adds a new optional `flow_paths` input to `eas/report_maestro_test_results`. Universe-side wiring (passing `job.params.flow_path` into this input) is a separate PR.

Behavior change for `eas/report_maestro_test_results`: single-attempt `retryCount` is no longer derived from `ai-*.json` occurrence count, falls back to `0`. Production impact: zero — no readers of `retryCount`, and the only caller of the single-attempt branch (the bash step) is retired by #3641.

Backward compatibility during deployment: between this PR landing and the universe wiring landing, old job specs (without `flow_paths` on the report step) run against the new code. The step silently falls back to `null` map → flow-name-as-path, no warnings.

# Test Plan

[...]

PR #3641, opened 2026-04-28
[build-tools] Port Maestro Tests bash step to TypeScript

hSATAC:
# Why

The Maestro test execution currently lives as an inline bash shellscript step in `WorkflowMaestroTestJob` (universe). Porting to a typed build function gives us type checking, unit tests, and a base for smart retry (next PR in the stack), which is painful to express in bash.

Linear: [ENG-19927](https://linear.app/expo/issue/ENG-19927).

# How

New `eas/run_maestro_tests` build function with behavior identical to the bash version: runs `maestro test ...` with the full input flow set, retries the whole set on failure, copies the latest attempt's JUnit XML to `final_report_path`. Adds a small `parseMaestroResults` / `parseJUnitFile` helper shared with `eas/report_maestro_test_results`.

Universe-side wiring (replacing the bash step with `uses: eas/run_maestro_tests`) is a separate PR.

# Test Plan

- New unit tests in `runMaestroTests.test.ts` covering single-attempt success, retry-on-failure, attempt-N JUnit naming, copy-latest behavior, env propagation, error paths.
- Local end-to-end smoke test against the playground.

### DIFF
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -17,6 +17,9 @@ export interface MaestroFlowResult {
 // Maestro's TestDebugReporter creates timestamped directories, e.g. "2024-06-15_143022"
 const TIMESTAMP_DIR_PATTERN = /^\d{4}-\d{2}-\d{2}_\d{6}$/;
 
+// Per-attempt JUnit XML files use `*-attempt-N.xml` names; this extracts N.
+const ATTEMPT_PATTERN = /attempt-(\d+)/;
+
 export function extractFlowKey(filename: string, prefix: string): string | null {
   const match = filename.match(new RegExp(`^${prefix}-(.+)\\.json$`));
   return match?.[1] ?? null;
@@ -239,9 +242,6 @@ export async function parseMaestroResults(
   // 3. Merge: JUnit results + ai-*.json metadata
   const results: MaestroFlowResult[] = [];
 
-  // Parse attempt index from filename pattern: *-attempt-N.*
-  const ATTEMPT_PATTERN = /attempt-(\d+)/;
-
   // Group results by flow name
   const resultsByName = new Map<string, JUnitResultWithSource[]>();
   for (const entry of junitResultsWithSource) {
@@ -321,4 +321,43 @@ async function relativizePathAsync(flowFilePath: string, projectRoot: string): P
     return flowFilePath;
   }
   return relative;
+}
+
+/**
+ * Copies the highest-attempt-index *.xml file from sourceDir to outputPath.
+ * After the maestro retry loop completes, this produces a single canonical
+ * JUnit report at final_report_path matching the bash step's "cp latest
+ * attempt" semantics.
+ *
+ * Throws if sourceDir contains no *.xml files or if the copy fails.
+ */
+export async function copyLatestAttemptXml(args: {
+  sourceDir: string;
+  outputPath: string;
+}): Promise<void> {
+  const entries = await fs.readdir(args.sourceDir);
+  const xmlFiles = entries.filter(f => f.endsWith('.xml')).sort();
+  if (xmlFiles.length === 0) {
+    throw new Error(`No *.xml files found in ${args.sourceDir}`);
+  }
+
+  // Pick the file with the highest attempt index. Files without the marker are
+  // treated as attempt 0. Ties are broken by sorted filename — later wins.
+  let winner = xmlFiles[0];
+  let winnerAttempt = (() => {
+    const m = winner.match(ATTEMPT_PATTERN);
+    return m ? parseInt(m[1], 10) : 0;
+  })();
+  for (let i = 1; i < xmlFiles.length; i++) {
+    const candidate = xmlFiles[i];
+    const match = candidate.match(ATTEMPT_PATTERN);
+    const attempt = match ? parseInt(match[1], 10) : 0;
+    if (attempt >= winnerAttempt) {
+      winner = candidate;
+      winnerAttempt = attempt;
+    }
+  }
+
+  const content = await fs.readFile(path.join(args.sourceDir, winner), 'utf-8');
+  await fs.writeFile(args.outputPath, content);
 }
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -358,6 +358,5 @@ export async function copyLatestAttemptXml(args: {
     }
   }
 
-  const content = await fs.readFile(path.join(args.sourceDir, winner), 'utf-8');
-  await fs.writeFile(args.outputPath, content);
+  await fs.copyFile(path.join(args.sourceDir, winner), args.outputPath);
 }
--- a/packages/build-tools/src/steps/functions/runMaestroTests.ts
+++ b/packages/build-tools/src/steps/functions/runMaestroTests.ts
@@ -1,0 +1,273 @@
+import { SystemError, UserError } from '@expo/eas-build-job';
+import {
+  BuildFunction,
+  BuildRuntimePlatform,
+  BuildStepInput,
+  BuildStepInputValueTypeName,
+  BuildStepOutput,
+} from '@expo/steps';
+import spawn from '@expo/turtle-spawn';
+import fs from 'fs/promises';
+import path from 'path';
+
+import { copyLatestAttemptXml } from './maestroResultParser';
+import { sleepAsync } from '../../utils/retry';
+
+// `outputPath: null` means "let maestro pick" (no --output flag). Junit and
+// other declared formats pass an explicit path so downstream upload steps
+// know where to find the result.
+function buildMaestroArgs(args: {
+  flow_paths: string[];
+  outputPath: string | null;
+  output_format: string | undefined;
+  shards: number | undefined;
+  include_tags: string | undefined;
+  exclude_tags: string | undefined;
+}): string[] {
+  const out: string[] = ['test'];
+  if (args.output_format) {
+    out.push(`--format=${args.output_format.toUpperCase()}`);
+  }
+  if (args.outputPath) {
+    out.push(`--output=${args.outputPath}`);
+  }
+  if (args.shards !== undefined) {
+    out.push(`--shard-split=${args.shards}`);
+  }
+  if (args.include_tags) {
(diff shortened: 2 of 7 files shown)


hSATAC:
* **#3639** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3639?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3635** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3641** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3641?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3641?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+      // this env and writes debug output under $HOME/.maestro/tests; the step
+      // must read from the same place or stale files leak across runs.
+      const home = env.HOME;
+      if (!home) {
+        throw new SystemError('HOME env var is not set');
+      }
+      const testsDirectory = path.join(home, '.maestro', 'tests');
+      const junitReportDirectory = path.join(testsDirectory, 'junit-reports');
+      const finalReportPath =
+        outputFormat === 'junit' ? path.join(testsDirectory, `${platform}-maestro-junit.xml`) : '';
```
sjchmiela:
```suggestion
        outputFormat === 'junit' ? path.join(testsDirectory, `${platform}-maestro-junit.xml`) : null;
```

it might be semantically more correct to set `null` if empty?

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+      if (!Array.isArray(flowPaths) || flowPaths.length === 0) {
+        throw new UserError(
+          'ERR_MAESTRO_NO_FLOW_PATHS',
+          'No flow_paths provided to maestro test step.'
+        );
+      }
+      if (!flowPaths.every(p => typeof p === 'string' && p.length > 0)) {
+        throw new UserError(
+          'ERR_MAESTRO_INVALID_FLOW_PATHS',
+          'All flow_paths entries must be non-empty strings.'
```
sjchmiela:
```suggestion
          'You have to provide at least one flow_path entry.'
```
I think this error message better describes the condition?

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+
+export function createRunMaestroTestsBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'run_maestro_tests',
+    name: 'Run Maestro Tests',
+    __metricsId: 'eas/run_maestro_tests',
+    inputProviders: [
+      BuildStepInput.createProvider({
+        id: 'flow_paths',
```
sjchmiela: let's use `flow_path` like we do in the job input. while it may be less correct, it will make for better error messages -- we won't suddenly speak of `flow_paths` in error messages while the only thing the user can configure is `flow_path` param to the job
```suggestion
        id: 'flow_path',
```

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+      }
+      if (!Number.isInteger(retries) || retries < 0) {
+        throw new UserError(
+          'ERR_MAESTRO_INVALID_RETRIES',
+          'retries must be a non-negative integer.'
+        );
+      }
+      if (shards !== undefined && (!Number.isInteger(shards) || shards < 1)) {
+        throw new UserError('ERR_MAESTRO_INVALID_SHARDS', 'shards must be a positive integer.');
+      }
```
sjchmiela: what we might want to do is instead of writing out the manual validation, have a zod schema, parse and throw a `UserError` if something is wrong. I think we do it in one or two places. We don't have a nice `validateWithZodSchema` utility like we do in www though so you need to handle the zod error manually or add a utility

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      try {
+        await fs.mkdir(junitReportDirectory, { recursive: true });
+      } catch (err) {
+        throw new SystemError('Failed to create JUnit report directory', { cause: err });
+      }
+
+      // Wipe stale *.xml from prior runs (cross-platform, since this dir is
```
sjchmiela: this only makes sense for our internal testing, right? or is it future-proofing?

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -322,3 +322,42 @@ async function relativizePathAsync(flowFilePath: string, projectRoot: string): P
...
+    const match = candidate.match(ATTEMPT_PATTERN);
+    const attempt = match ? parseInt(match[1], 10) : 0;
+    if (attempt >= winnerAttempt) {
+      winner = candidate;
+      winnerAttempt = attempt;
+    }
+  }
+
+  const content = await fs.readFile(path.join(args.sourceDir, winner), 'utf-8');
+  await fs.writeFile(args.outputPath, content);
```
sjchmiela: `fs` has a `copyFile`, would it be easier to use?

sjchmiela (approved):
amazing

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+    out.push(`--exclude-tags=${args.exclude_tags}`);
+  }
+  out.push(...args.flow_paths);
+  return out;
+}
+
+export function createRunMaestroTestsBuildFunction(): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'run_maestro_tests',
```
sjchmiela: let's call it `maestro_tests`? plays well with `after_maestro_tests` hook

packages/build-tools/src/steps/functions/runMaestroTests.ts
```diff
@@ -0,0 +1,273 @@
...
+      // Copy the latest attempt's JUnit to the final report path so downstream
+      // upload/report steps have a single canonical file.
+      if (outputFormat === 'junit') {
+        try {
+          await copyLatestAttemptXml({
+            sourceDir: junitReportDirectory,
+            outputPath: finalReportPath,
+          });
+        } catch (copyErr: any) {
+          throw new SystemError('Failed to produce final_report_path', { cause: copyErr });
```
sjchmiela: in bash we would swallow this error. if we don't, an early failing maestro test (bad yaml) that does not write any xml is going to say "failed to produce final_report_path" instead of "maestro tests failed"

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: `BuildStepOutput.set` accepts `string | undefined` only — there's no `null` in the type (`BuildStepOutput.ts:21-23, 65-73`), and `final_report_path` is `required: true` so it has to be a `string`.

Switching it to `required: false` and skipping `.set()` doesn't help either: `interpolate` does `result.replace(match, value)` (`template.ts:126`), so `${{ steps.maestro_tests.outputs.final_report_path }}` would render the literal string `"undefined"` for non-junit formats, which is worse than `''`. Happy to keep `''` here unless we want to extend `BuildStepOutput` to support `null`.

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: Hmm this error message means the array content was malformed. I assume you were talking about the error message above? 
> 'No flow_paths provided to maestro test step.'

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: Yeah we don't really need this in our environment. This was more for local testing. Removing it.

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/steps/functionGroups/maestroTest.ts#L20
I chose to use `run_maestro_tests` since we already have a `maestro_test` function group.

We *could* name it `maestro_tests`, but I’m a bit worried that might be confusing to users. `run_maestro_tests` makes it clearer that this step is actually responsible for running Maestro tests, which I think is a good signal.

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: Good catch. I didn't think about that scenario. Will fix here.

packages/build-tools/src/steps/functions/runMaestroTests.ts
sjchmiela: > We could name it maestro_tests, but I’m a bit worried that might be confusing to users.

we should remove `maestro_test` function group along with ability to run maestro tests in custom builds. see https://linear.app/expo/issue/ENG-15179/remove-disable-new-android-vm

if there was no pre-existing `eas/maestro_test`, what would you think of `eas/maestro_tests`? it _sort of_ resembles `eas/deploy` (no `eas/run_deploy`) and others?

packages/build-tools/src/steps/functions/runMaestroTests.ts
sjchmiela: oh right, hmm. i know `undefined` pops up from time to time and i have generally been ok with that (like, maybe a thing being _actually_ `undefined` is a signal (vs `''`)? wdyt?

hSATAC: ok let's go with `undefined`. I am ok with that.

packages/build-tools/src/steps/functions/runMaestroTests.ts
hSATAC: If we don't have `eas/maestro_test` then I am ok with `eas/maestro_tests`.

I'll modify the step here per your suggestion.

PR #3655, opened 2026-04-29
[eas-cli] Allow eas build:download to accept a build ID

douglowder:
# Why

For certain workflows, it is useful to be able to download build artifacts for a known build ID.

# How

Added a new flag `--build-id` to the `eas build:download` command. When this flag is used,`--fingerprint`, `--dev-client`, and `--platform` are excluded.

Also added the `--all-artifacts`flag. When set, all available artifacts are downloaded. Otherwise, the previous behavior applies (only application archive URL is downloaded).

# Test Plan

- Tested in a real project
- New unit tests added

### DIFF
--- a/packages/eas-cli/src/commands/build/download.ts
+++ b/packages/eas-cli/src/commands/build/download.ts
@@ -111,6 +111,12 @@ export default class Download extends EasCommand {
     if (build.artifacts?.applicationArchiveUrl) {
       buildArtifactPath = await this.getPathToBuildArtifactAsync(build, selectedPlatform);
     } else if (!allArtifacts) {
+      const availableArtifacts = listAvailableExtraArtifactNames(build);
+      if (availableArtifacts.length > 0) {
+        throw new Error(
+          `Build does not have an application archive url. Other artifacts are available (${availableArtifacts.join(', ')}); re-run with --all-artifacts to download them.`
+        );
+      }
       throw new Error('Build does not have an application archive url');
     }
 
@@ -275,6 +281,23 @@ async function resolvePlatformAsync({
     ],
   });
   return selectedPlatform;
+}
+
+function listAvailableExtraArtifactNames(build: BuildFragment): string[] {
+  const names: string[] = [];
+  if (build.artifacts?.buildArtifactsUrl) {
+    names.push('buildArtifacts');
+  }
+  if (build.artifacts?.xcodeBuildLogsUrl) {
+    names.push('xcodeBuildLogs');
+  }
+  if (
+    build.artifacts?.buildUrl &&
+    build.artifacts.buildUrl !== build.artifacts.applicationArchiveUrl
+  ) {
+    names.push('build');
+  }
+  return names;
 }
 
 function getFileNameFromUrl(url: string, fallbackName: string): string {
--- a/packages/eas-cli/src/commands/build/download.ts
+++ b/packages/eas-cli/src/commands/build/download.ts
@@ -19,20 +19,29 @@ import { downloadAndMaybeExtractAppAsync } from '../../utils/download';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
 export default class Download extends EasCommand {
-  static override description = 'download simulator/emulator builds for a given fingerprint hash';
+  static override description =
+    'download a simulator/emulator build by build ID or fingerprint hash';
+
 
   static override flags = {
+    'build-id': Flags.string({
+      description:
+        'ID of the build to download. Mutually exclusive with --fingerprint, --platform, and --dev-client; the platform is derived from the build itself.',
+      exclusive: ['fingerprint', 'platform', 'dev-client'],
+    }),
     fingerprint: Flags.string({
       description: 'Fingerprint hash of the build to download',
-      required: true,
+      exclusive: ['build-id'],
     }),
     platform: Flags.option({
       char: 'p',
       options: [Platform.IOS, Platform.ANDROID] as const,
+      exclusive: ['build-id'],
     })(),
     'dev-client': Flags.boolean({
       description: 'Filter only dev-client builds.',
       allowNo: true,
+      exclusive: ['build-id'],
     }),
     ...EasNonInteractiveAndJsonFlags,
   };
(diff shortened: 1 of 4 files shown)


douglowder:
* **#3655** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3655?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3655?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -64,24 +84,107 @@ export default class Download extends EasCommand {
...
+    graphqlClient,
+    buildId,
+  }: {
+    graphqlClient: ExpoGraphqlClient;
+    buildId: string;
+  }): Promise<BuildFragment> {
+    try {
+      return await BuildQuery.byIdAsync(graphqlClient, buildId);
+    } catch (error: any) {
+      throw new Errors.CLIError(`Could not find build with ID ${buildId}: ${error.message}`);
```
sjchmiela: in one place we throw `new Error`, here we throw a specific `Errors.CLIError` — is there a difference?

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -64,24 +84,107 @@ export default class Download extends EasCommand {
...
+
+    let buildArtifactPath: string | null = null;
+    if (build.artifacts?.applicationArchiveUrl) {
+      buildArtifactPath = await this.getPathToBuildArtifactAsync(build, selectedPlatform);
+    } else if (!buildId) {
+      throw new Error('Build does not have an application archive url');
+    }
+
+    let extraArtifactPaths: Record<string, string> = {};
+    if (buildId) {
```
sjchmiela: I presume… we believe downloading by fingerprint is for dev client and stuff which never uses extra artifacts and by build ID is general download which may use extra artifacts?

This difference in behavior in the same command feels somewhat weird to me. Maybe we want to add a flag to decide whether to download the extra artifacts or not?

packages/eas-cli/src/commands/build/download.ts
```diff
@@ -64,24 +84,107 @@ export default class Download extends EasCommand {
...
+    }
+    if (artifacts.buildUrl && artifacts.buildUrl !== artifacts.applicationArchiveUrl) {
+      extraArtifacts.push({ name: 'build', url: artifacts.buildUrl });
+    }
+
+    if (extraArtifacts.length === 0) {
+      return {};
+    }
+
+    const outputDir = path.join(getEasBuildRunCacheDirectoryPath(), `${build.id}-artifacts`);
```
sjchmiela, line 165: not sure if we want to put extra artifacts (or build-id-downloaded app archive) into cache dir? i thought cache dir is meant to keep reusable dev client apps and stuff. do uses have some neat way of pruning the cache or is it going to grow indefinitely?

sjchmiela (approved):

packages/eas-cli/src/commands/build/download.ts
douglowder: That is a good idea! I'll do that. I thought it might even make sense to have the build ID behavior in a separate command, but could not think of a good name other than download.

packages/eas-cli/src/commands/build/download.ts
douglowder: In the workflow that I'm using this for, the artifact is downloaded only once, used in building a test app, and then not needed any more. On my local, the artifact is going to a folder inside the tmp directory, so that seems ok. Is the cache dir persistent over different jobs in workers?

packages/eas-cli/src/commands/build/download.ts
douglowder: Good catch, it should throw the same error as it does if it finds no matching fingerprint.

douglowder: It does also throw CLIError for no matching fingerprint. I do see that it throws Error and CLIError for different things. I'll take a look and try to make the usage more consistent.

packages/eas-cli/src/commands/build/download.ts
douglowder: In thinking about this more, it seems to me that we do want all these build artifacts cached. In my workflow, the same artifact could potentially be used downstream by multiple build jobs.

packages/eas-cli/src/commands/build/download.ts
douglowder: Removed all CLIError usage, since it is not used elsewhere in commands.

packages/eas-cli/src/commands/build/download.ts
douglowder: Done.

packages/eas-cli/src/commands/build/download.ts
sjchmiela: > On my local, the artifact is going to a folder inside the tmp directory

oohh sorry i was certain it goes into `~/.expo/...`. i may have been thinking about `~/.expo/ios-simulator-app-cache`. i see `getEasBuildRunCacheDirectoryPath` is in tmp so it's ok

> Is the cache dir persistent over different jobs in workers?

never -- separate job, separate vm, separate filesystem

> In thinking about this more, it seems to me that we do want all these build artifacts cached. In my workflow, the same artifact could potentially be used downstream by multiple build jobs.

i don't think i agree -- if we cache (i.e. upload the cache dir in one job and automatically download in another job) the directory with downloaded builds we don't improve anything:
- if we need the build in a subsequent job it doesn't matter if we download the build cache or the individual build
- if we don't need the build in a subsequent job we download the build cache in vain
- if we need less builds in subsequent jobs than in upstream jobs, we download more builds than we need to
- we use additional CPU time to manage the cache archive, bandwidth to upload/download and storage to store the cache

---

i expected this to be more useful in local development, in workflows i expect people to use https://docs.expo.dev/eas/workflows/syntax/#easdownload_build which does not require eas-cli call, so it should be faster

douglowder: I would like to use download_build, but it looks like it only gets application archive. I’ll have a look at the source for that.

PR #3657, opened 2026-04-30
[eas-cli] eas go server-side project setup and SDK version selection

gwdp:
# Why:

Remove local project setup and npm install from `eas go`, enable multiple Expo Go SDK versions server-side, and reduce surface area in the CLI.

# How:

- Calls `expoGoRepackConfiguration` to get server-synthesized project files per SDK version
- Packs them in-memory into a tar.gz and uploads tarball + eas.json + package.json to GCS
- Calls `createExpoGoRepackWorkflowRun` with bucket keys + optional `sdkVersion`
- Adds optional `--sdk-version` flag to select the Expo Go SDK version

# Test Plan:

- `yarn tsc` and `yarn lint` pass
- Ran local eas-cli against prod -> https://expo.dev/accounts/gwdp/projects/expogo/workflows/019e1833-6330-7b65-9849-41984bd59768

Base: https://github.com/expo/eas-cli/pull/3514
x-ref: https://github.com/expo/universe/pull/27053 (merged)


packages/eas-cli/src/utils/workflowProjectSourceUtils.ts
```diff
@@ -0,0 +1,77 @@
+import { createGzip } from 'node:zlib';
+import { pack as tarPack } from 'tar-stream';
+
+import { ExpoGraphqlClient } from '../commandUtils/context/contextUtils/createGraphqlClient';
+import { AccountUploadSessionType } from '../graphql/generated';
+import { uploadAccountScopedBufferToGCSAsync } from '../uploads';
+
+export async function packProjectTarGzInMemoryAsync(
```
sjchmiela: Why would we want to introduce a completely new way of packaging project for three files? Why not just create real files in os.tmpdir, package that and remove the temp dir?

I feel relatively strongly the effort, maintenance cost and complexity burden is not worth it.

packages/eas-cli/src/graphql/queries/WorkflowRunQuery.ts
```diff
@@ -16,7 +16,57 @@ import {
 import { WorkflowJobFragmentNode } from '../types/WorkflowJob';
 import { WorkflowRunFragmentNode } from '../types/WorkflowRun';
 
+export type ExpoGoRepackInput = {
```
sjchmiela: Can these be imported from generated GraphQL instead of being redeclared manually?

sjchmiela: Can these be imported from generated GraphQL instead of being redeclared manually?

packages/eas-cli/src/commands/go.ts
```diff
@@ -187,190 +183,97 @@ export default class Go extends EasCommand {
...
+    const { workflowUrl, workflowRunId } = await this.dispatchWorkflowAsync(
+      graphqlClient,
+      projectId,
+      actor,
+      bundleId,
+      appName,
+      ascApp.id,
+      sdkVersion
+    );
+    Log.withTick(`Workflow started: ${chalk.cyan(workflowUrl)}`);
```
sjchmiela: nit: while we're here maybe we can think of a nicer user messages? they don't need to know that it runs as a workflow?

packages/eas-cli/src/commands/go.ts
```diff
@@ -492,58 +385,53 @@ export default class Go extends EasCommand {
...
-          accountId: account.id,
-          filePath: path.join(projectDir, 'package.json'),
-          maxSizeBytes: 1024 * 1024,
-        });
+    if (!repackConfig.files.some(f => f.fileName === 'eas.json')) {
+      throw new Error('Server response is missing eas.json. Please try again.');
+    }
+    if (!repackConfig.files.some(f => f.fileName === 'package.json')) {
+      throw new Error('Server response is missing package.json. Please try again.');
+    }
```
sjchmiela: i don't think we need to manually validate that the files are present here? i'd rather have www decide all the way

sjchmiela (changes requested):

packages/eas-cli/src/commands/go.ts
gwdp: Agreed, thanks for raising that! Updated

packages/eas-cli/src/graphql/queries/WorkflowRunQuery.ts
gwdp: 100%, missed that from when changes werent merged on `universe`, my bad 🙇‍♂️

packages/eas-cli/src/utils/workflowProjectSourceUtils.ts
gwdp: I initially thought the serialization layer wasn’t needed, but I see your point about the extra code, complexity, and maintenance my approach brought. I’ve switched to using the existing layer instead. Thanks for your input!

packages/eas-cli/src/commands/go.ts
gwdp: Updated to more build-focused messages, let me know how do you feel about it?

packages/eas-cli/src/commands/go.ts
```diff
@@ -159,6 +155,10 @@ export default class Go extends EasCommand {
       description: 'App name',
       default: 'My Expo Go',
     }),
+    'sdk-version': Flags.string({
+      description: 'Expo Go SDK version to repack (server uses latest supported if not specified)',
```
```suggestion
      description: 'Expo Go SDK version to prepare (default: latest)',
```

packages/eas-cli/src/commands/go.ts
```diff
@@ -187,190 +187,97 @@ export default class Go extends EasCommand {
...
-      await fs.remove(projectDir);
-    } catch (error) {
-      Log.gray(`Project files preserved for debugging: ${projectDir}`);
-      throw error;
-    } finally {
-      process.chdir(originalCwd);
+    try {
+      await setupTestFlightAsync(ascApp);
+    } catch {
+      // Non-fatal: TestFlight group setup failure shouldn't block the user
```
sjchmiela: should we log the error?

packages/eas-cli/src/commands/go.ts
```diff
@@ -187,190 +187,97 @@ export default class Go extends EasCommand {
...
-  }
-
-  private async initGitRepoAsync(projectDir: string): Promise<void> {
-    await spawnAsync('git', ['init'], { cwd: projectDir });
-    await spawnAsync('git', ['add', '.'], { cwd: projectDir });
-    await spawnAsync('git', ['commit', '-m', 'Initial commit'], { cwd: projectDir });
+      .replace(/[^a-z0-9-]/g, '-')
+      .replace(/-{2,}/g, '-')
+      .replace(/^-+|-+$/g, '');
+    return `com.${sanitizedUsername || 'app'}.expogo`;
```
sjchmiela: i believe apple requires bundle identifiers to be unique across their server. we can't have a static fallback

packages/eas-cli/src/commands/go.ts
```diff
@@ -187,190 +187,97 @@ export default class Go extends EasCommand {
...
-        graphqlClient,
-        projectDir,
-        projectId,
-        actor,
-        vcsClient
+    const setupSpinner = ora('Setting up project...').start();
+    let projectId: string;
+    try {
+      projectId = await withSuppressedOutputAsync(() =>
+        this.ensureEasProjectAsync(graphqlClient, actor, bundleId)
```
sjchmiela: `slug` should be an argument to `ensureEasProjectAsync`, we shouldn't derive it twice

packages/eas-cli/src/commands/go.ts
```diff
@@ -380,42 +287,50 @@ export default class Go extends EasCommand {
...
     actor: Actor,
     analytics: Analytics,
-    vcsClient: VcsClient,
     customizeCreds: boolean,
     onBeforeAppleAuth?: () => void
   ): Promise<App> {
-    const exp = await getPrivateExpoConfigAsync(projectDir);
     const extensionBundleId = `${bundleId}.ExpoNotificationServiceExtension`;
 
+    const exp: ExpoConfig = {
```
sjchmiela: shouldn't this come from the server? like the config thing you were building?

packages/eas-cli/src/commands/go.ts
```diff
@@ -492,58 +389,75 @@ export default class Go extends EasCommand {
...
-          graphqlClient,
-          accountId: account.id,
-          filePath: path.join(projectDir, 'package.json'),
-          maxSizeBytes: 1024 * 1024,
+    try {
+      await fs.ensureDir(tmpDir);
+      await Promise.all(
+        repackConfig.files.map(f => fs.writeFile(path.join(tmpDir, f.fileName), f.fileContents))
+      );
+      process.chdir(tmpDir);
```
sjchmiela: this feels weird. would it help to know `NoVcsClient` has a `cwdOverride` in the constructor?

sjchmiela: also, i think it would be nicer to share a single `NoVcsClient` instance across the command?

packages/eas-cli/src/commands/go.ts
gwdp: Updated, thank you

packages/eas-cli/src/commands/go.ts
gwdp: Added, thanks for the suggestion

packages/eas-cli/src/commands/go.ts
gwdp: Removed fallback, thank you for catching that

packages/eas-cli/src/commands/go.ts
gwdp: Updates, ty!

packages/eas-cli/src/commands/go.ts
gwdp: `NoVcsClient` had some unusual behavior with the `cwdOverride` and `getRootPathAsync`, and I didn’t want to risk causing issues with this edge case, to be fair. 

I made a quick patch there and double checked for regressions

packages/eas-cli/src/commands/go.ts
gwdp: The challenge is that we need the ASC app ID from the credential setup before we can move forward. I could split `expoGoRepackConfigurationAsync` into 2 parts, but complexity/win appears low. Let me know if you have different thoughts or any other way out 🙇

packages/eas-cli/src/commands/go.ts
```diff
@@ -492,58 +390,67 @@ export default class Go extends EasCommand {
...
     actor: Actor,
-    vcsClient: VcsClient
+    bundleId: string,
+    appName: string,
+    ascAppId: string,
+    sdkVersion: string | undefined,
+    tmpDir: string,
+    vcsClient: Client
   ): Promise<{ workflowUrl: string; workflowRunId: string }> {
     const account = actor.accounts[0];
```
sjchmiela: I don't know if we guarantee that the first account is the designated primary account and I suspect we want to always use the primary account so might want to double check that

sjchmiela (approved):
Can't wait to test it!!!

packages/eas-cli/src/commands/go.ts
gwdp: > I don't know if we guarantee that the first account is the designated primary account and I suspect we want to always use the primary account so might want to double check that

Thanks for calling this out, added `ensureActorHasPrimaryAccount` gate for allowing only primary

gwdp:
final sanity check https://expo.dev/accounts/gwdp/projects/expogo/workflows/019e1df1-97ca-7954-8c45-5aa9103853a9

PR #3660, opened 2026-04-30
[worker] warn on invalid resource class config when setting `GRADLE_OPTS`

sswrk:
# Why

This issue https://github.com/expo/eas-cli/issues/3682 is reporting that the max heap size in Gradle options is set to an incorrect value when the resource class is `large` (should be `-Xmx8g`, is `-Xmx4g`).

I suspect the worker is receiving an empty or incorrect `resourceClass` config value, as we silently fallback to `4g` in such cases. This PR adds Sentry reporting to enable us to easily track such cases. Reporting is turned off in `Environment.TEST`.

# How

Report a sentry warning on:
- No `config.resourceClass` set
- Unknown `config.resourceClass`

The reporting is plugged just before determining the Gradle options `getBuildEnv` in `packages/worker/src/env.ts`.

# Test Plan

CI passes.

### DIFF
--- a/packages/worker/src/env.ts
+++ b/packages/worker/src/env.ts
@@ -10,6 +10,7 @@ import {
   ResourceClassToPlatform,
   androidImagesWithJavaVersionLowerThen11,
 } from './external/turtle';
+import sentry from './sentry';
 import { getAccessedEnvs } from './utils/env';
 
 // keep in sync with local-build-plugin env vars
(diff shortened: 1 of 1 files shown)


sjchmiela (changes requested):
Sssso resource classes is an outdated concept, or a concept we're trying to move away from. For instance, workflow-orchestration has never and will never provide the `resourceClass` in `config`.

Maybe we want to instead of relying on resource class, execute some command to learn the free memory on the instance and adjust `GRADLE_OPTS` from there?

sswrk:
Superseded by https://github.com/expo/eas-cli/pull/3782

PR #3674, opened 2026-05-05
[build-tools] Log Gradle cache restore events to Datadog

AbbanMustafa:
# Why

We want to track Gradle cache restore events in Datadog to gain visibility into time benefits across builds/users.

# How

  - Added a turtleFetch call in restoreGradleCacheAsync that sends a log to the new /turtle-builds/logs endpoint when Gradle cache is successfully restored
  - Tags the log with build_id, cache_hit_type (direct hit vs prefix match), cache_key, and matched_key
  - Updated `service.ts` to use the new `/turtle-builds/logs` route with level: 'error' (replaces /turtle-builds/error-logs)
  - Updated corresponding tests

# Test Plan

  - Deploy the [www route](https://github.com/expo/universe/pull/27125) first so both /logs and /error-logs are available
  - Deploy this change and verify Gradle cache restore logs appear in Datadog with correct build ID tags
  - Clean up legacy /error-logs route in www after rollout

### DIFF
--- a/packages/build-tools/src/steps/functions/restoreBuildCache.ts
+++ b/packages/build-tools/src/steps/functions/restoreBuildCache.ts
@@ -283,24 +283,20 @@ export async function restoreGradleCacheAsync({
     );
 
     try {
-      await turtleFetch(
-        new URL('turtle-builds/logs', expoApiServerURL).toString(),
-        'POST',
-        {
-          json: {
-            buildId: jobId,
-            message: `Gradle cache restored (${hitType})`,
-            tags: {
-              event: 'gradle_cache_restored',
-              cache_hit_type: hitType,
-            },
+      await turtleFetch(new URL('turtle-builds/logs', expoApiServerURL).toString(), 'POST', {
+        json: {
+          buildId: jobId,
+          message: `Gradle cache restored (${hitType})`,
+          tags: {
+            event: 'gradle_cache_restored',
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/functions/restoreBuildCache.ts
```diff
@@ -277,9 +277,29 @@ export async function restoreGradleCacheAsync({
...
     });
 
+    const hitType = matchedKey === cacheKey ? 'direct_hit' : 'prefix_match';
     logger.info(
-      `Gradle cache restored to ${gradleCachesPath} ${matchedKey === cacheKey ? '(direct hit)' : '(prefix match)'}`
+      `Gradle cache restored to ${gradleCachesPath} (${hitType === 'direct_hit' ? 'direct hit' : 'prefix match'})`
     );
+
+    try {
+      await turtleFetch(new URL('turtle-builds/logs', expoApiServerURL).toString(), 'POST', {
```
sjchmiela, line 286: shouldn't this be a metric, not a log?

sjchmiela (approved):

packages/build-tools/src/steps/functions/restoreBuildCache.ts
AbbanMustafa: i want to tag by build id 💸💸💸🤫

PR #3676, opened 2026-05-05
[eas-cli] Hold eas simulator:start until session ends or Ctrl+C

szdziedzic:
# 

# Why

Fixes ENG-20929

`eas simulator:start` currently prints the daemon connection env vars and exits, leaving the device run session running on EAS. Users have to remember to call `eas simulator:stop --id ...` to clean up — if they just close their terminal, the session leaks. This aligns the command with how other "session-holder" dev tools work (`kubectl port-forward`, `ngrok`, `vercel dev`): hang for the duration of the session and clean up on Ctrl+C.

# How

After printing the connection env vars, in interactive mode the command now stays attached:

- A poll loop checks the session every 5s and exits cleanly when it reaches a terminal state (`Stopped`/`Canceled`/`Finished`) or fails  
(`Errored`). Errors at either layer (device run session vs. underlying turtle job run) are surfaced with the same user-facing message — the distinction is an implementation detail.
- A `SIGINT` handler flips an `interrupted` flag and resolves an interrupt promise that's raced against `sleepAsync`, so Ctrl+C wakes the loop immediately rather than after the next 5s tick. On wake-up, `ensureDeviceRunSessionStoppedSafelyAsync` cancels the session.
- A second Ctrl+C during cleanup forces `process.exit(130)` as an escape hatch in case the stop mutation hangs.
- Polling errors are downgraded to `Log.debug` and the loop continues — a transient network blip shouldn't tear down a session the user is actively using.
- `--non-interactive` keeps the prior fire-and-forget behavior, so scripts and CI are unaffected.

# Test Plan

- `eas simulator:start --platform ios` → env vars print, spinner stays up. Press Ctrl+C → session is stopped on EAS, command exits.
- `eas simulator:start --platform ios`, then `eas simulator:stop --id <id>` from another shell → spinner resolves with "Device run session
ended" and command exits.
- `eas simulator:start --platform ios --non-interactive` → env vars print and command exits immediately (existing behavior preserved).
- Press Ctrl+C twice during the cleanup mutation → process exits with code 130 right away.
- `yarn typecheck`, `yarn lint`, `yarn fmt:check` pass.

### DIFF
--- a/packages/eas-cli/src/commands/simulator/start.ts
+++ b/packages/eas-cli/src/commands/simulator/start.ts
@@ -154,21 +154,119 @@ export default class SimulatorStart extends EasCommand {
     Log.newLine();
     Log.log(result.message);
     Log.newLine();
-    Log.log(
-      `When you are done, stop the session with: eas simulator:stop --id ${deviceRunSessionId}`
+
+    if (flags['non-interactive']) {
+      Log.log(
+        `When you are done, stop the session with: eas simulator:stop --id ${deviceRunSessionId}`
+      );
+      return;
+    }
+
+    await waitForSessionEndOrInterruptAsync({
+      graphqlClient,
+      deviceRunSessionId,
+      jobRunUrl,
+    });
+  }
+}
+
+async function waitForSessionEndOrInterruptAsync({
+  graphqlClient,
+  deviceRunSessionId,
+  jobRunUrl,
+}: {
+  graphqlClient: ExpoGraphqlClient;
+  deviceRunSessionId: string;
+  jobRunUrl: string;
+}): Promise<void> {
+  const spinner = ora(
+    `Device run session active — press Ctrl+C to stop, or run \`eas simulator:stop --id ${deviceRunSessionId}\` from another shell`
+  ).start();
+
+  const abortController = new AbortController();
+  const { signal } = abortController;
+  const abortPromise = new Promise<void>(resolve => {
+    signal.addEventListener(
+      'abort',
+      () => {
+        resolve();
+      },
+      { once: true }
     );
+  });
+  const sigintHandler = (): void => {
+    if (signal.aborted) {
+      // Force exit on a second Ctrl+C in case cleanup is hanging. The session may still be
+      // running on EAS, so tell the user how to make sure it gets terminated.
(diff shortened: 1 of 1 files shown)


szdziedzic:
* **#3676** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3676?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3676?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -154,9 +154,88 @@ export default class SimulatorStart extends EasCommand {
...
+}): Promise<void> {
+  const spinner = ora(
+    `Device run session active — press Ctrl+C to stop, or run \`eas simulator:stop --id ${deviceRunSessionId}\` from another shell`
+  ).start();
+
+  let interrupted = false;
+  let resolveInterrupt: (() => void) | undefined;
+  const interruptPromise = new Promise<void>(resolve => {
+    resolveInterrupt = resolve;
+  });
```
sjchmiela: I think an `AbortController` is a nice thing to use for this type of thing? On the other hand it's nice what you did with the `Promise.race`.

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -154,9 +154,88 @@ export default class SimulatorStart extends EasCommand {
...
+  const sigintHandler = (): void => {
+    if (interrupted) {
+      // Force exit on a second Ctrl+C in case cleanup is hanging.
+      process.exit(130);
```
sjchmiela: Should we print that the device run session has not been stopped so we don't incur unexpected fees for users who just frantically do ctrl+cccccc?

sjchmiela (approved):

packages/eas-cli/src/commands/simulator/start.ts
szdziedzic: yeah, this sounds right

Issue #3645, opened 2026-04-29
[local build] iOS: "0 valid identities found" on macOS Tahoe 26 - Distribution certificate not recognized in temp keychain


JannisP007:
### Summary

EAS local builds on macOS Tahoe 26 (26.4.1 / 25E253) fail at PREPARE_CREDENTIALS 
with "Distribution certificate hasn't been imported successfully". The certificate 
is successfully imported (find-certificate confirms it), but security find-identity 
returns 0 valid identities. This appears to be a macOS Tahoe compatibility issue 
with the temporary keychain mechanism used by EAS local builds.

### Managed or bare?

Managed

### Environment

npx expo-env-info

### Error output

[PREPARE_CREDENTIALS] [DIAG-1] find-identity -v ALLE: 0 valid identities found
[PREPARE_CREDENTIALS] [DIAG-2] find-identity -v -p codesigning: 0 valid identities found
[PREPARE_CREDENTIALS] [DIAG-3] find-certificate (filtered): SHA-1 hash: 3A50325D796B9014D4F4FF06B7D708A42BF20E11 "iPhone Distribution: Jannis Pohl (WRV9RQW49Z)"
[PREPARE_CREDENTIALS] [FIX-1] explicit security import OK
[PREPARE_CREDENTIALS] [FIX-2] set-key-partition-list OK
Error: Distribution certificate with fingerprint 3A50325D796B9014D4F4FF06B7D708A42BF20E11 hasn't been imported successfully

### Reproducible demo or steps to reproduce from a blank project

1. Install any Expo managed project on macOS Tahoe 26.4.1 (25E253)
2. Run: eas build --platform ios --profile preview --local
3. Build fails at PREPARE_CREDENTIALS phase

Root cause:
- EAS creates a temporary keychain
- Imports distribution certificate into it (FIX-1, FIX-2, FIX-3 all report OK)
- find-certificate confirms certificate IS in the keychain
- BUT security find-identity -v -p codesigning returns "0 valid identities found"
- This appears to be a macOS Tahoe 26 security policy change affecting 
  temporary keychains

Workarounds tried (none worked):
- Generating a new distribution certificate via eas credentials
- Downloading credentials.json locally
- Manually importing p12 into login keychain

Note: Cloud builds (without --local) work fine. Issue is specific to 
--local flag on macOS Tahoe 26.

nzakas:
I'm also running into this.

<details>
<legend>Log</legend>

```
[...]

PR #3679, opened 2026-05-06
[build-tools] Fix iOS local build on macOS Tahoe — drop `-v` from find-identity in `findIdentitiesByTeamId`

### LINKED ISSUE #3678: iOS local build: `find-identity -v` in `findIdentitiesByTeamId` falsely fails on macOS Tahoe with `Distribution certificate ... hasn't been imported successfully`
## Summary

`eas build --profile development --platform ios --local` fails at the `Prepare credentials` step on macOS Tahoe (26.x) with `Distribution certificate <fingerprint> hasn't been imported successfully`, even though the certificate and private key were imported into the build keychain successfully. The failure is in a presence-check that uses `security find-identity -v` against an ephemeral build keychain that — by design — does not contain the full Apple trust chain required for `-v`'s "validity" judgment.

## Symptom

```
✖ Prepare credentials
Error: Distribution certificate with fingerprint <FINGERPRINT> hasn't been imported successfully
    at Keychain.ensureCertificateImported (.../@expo/build-tools/dist/ios/credentials/keychain.js:57:19)
```

The cert+key are in fact imported. Verified: running the same `import_certificate` fastlane action manually against the same `.p12` produces `1 identity imported.` and the fingerprint shows up in `security find-identity` (without `-v`) in the build keychain, annotated `(CSSMERR_TP_NOT_TRUSTED)`.

## Reproduction

1. macOS Tahoe (26.x), Xcode 26.x, fastlane 2.x, eas-cli 18.8+ (any version that pulls `eas-cli-local-build-plugin` >= 18.8.0 with `@expo/build-tools` >= 18.8.0). 
2. Project configured for an iOS development build (real-device, not simulator).
3. `npx eas-cli build --profile development --platform ios --local`.
4. EAS downloads the dist .p12 and provisioning profile from Expo's server (or reads from `credentials.json` — same downstream path), creates a fresh ephemeral keychain at `os.tmpdir()/eas-build-<uuid>.keychain`, imports the cert+key into it, and then immediately invokes `findIdentitiesByTeamId` to verify the import. That call fails.

## Root cause

`Keychain.findIdentitiesByTeamId` ([packages/build-tools/src/ios/credentials/keychain.ts#L100-L107](https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/ios/credentials/keychain.ts#L100-L107)):

```ts
private async findIdentitiesByTeamId(teamId: string): Promise<string> {
  const { output } = await spawn(
    'security',
    ['find-identity', '-v', '-s', `(${teamId})`, this.keychainPath],
    { stdio: 'pipe' }
  );
  return output.join('');
}
```

The `-v` flag means "valid identities only". `security`'s validity judgment requires the full trust chain (dist cert → Apple WWDR Intermediate → Apple Root CA) to resolve from the keychain(s) in the search list. The build keychain created in `Keychain.create()` only holds the dist cert + private key. Apple Root CA lives in `/Library/Keychains/System.keychain` and Apple WWDR G3 lives in the user's login keychain.

Critically: passing `System.keychain` (or any other keychain) as a positional argument to `security find-identity` does **not** cause it to aggregate trust resolution across keychains. Trust aggregation across keychains only happens for the user's session-wide search list (`security list-keychains -s`) — which `Keychain.create()` deliberately does not modify (and shouldn't, because that change is session-wide and would surprise the user).

So on a fresh ephemeral keychain holding only the cert+key, `find-identity -v -s "(<teamId>)" <buildKeychainPath>` returns 0 identities, `ensureCertificateImported` throws, and the build dies — even though the cert+key are present and the build would actually succeed if allowed to proceed (because gym/codesign later perform their own trust resolution via `Security.framework`, which **does** aggregate across keychains).

This is fundamentally a misuse of `find-identity -v` as a presence check. `-v`'s "valid" predicate depends on the whole trust environment, not just whether the cert+key are in the named keychain.

## Proposed fix

Drop the `-v` flag. The remaining flags (`-s "(${teamId})" <buildKeychainPath>`) still constrain the match to the right team and keychain, so the `includes(fingerprint)` check in `ensureCertificateImported` correctly answers \"is the cert+key pair present in this keychain?\" The cert will appear in the output annotated `(CSSMERR_TP_NOT_TRUSTED)`, but that's expected and ignored — codesign resolves trust correctly downstream via `Security.framework`.

```diff
  private async findIdentitiesByTeamId(teamId: string): Promise<string> {
    const { output } = await spawn(
      'security',
-     ['find-identity', '-v', '-s', `(${teamId})`, this.keychainPath],
+     ['find-identity', '-s', `(${teamId})`, this.keychainPath],
      {
        stdio: 'pipe',
      }
    );
    return output.join('');
  }
```

This is consistent with what the surrounding code is actually trying to do (verify import succeeded), and matches what works in practice on every macOS version I've tested (a dropped-`-v` find-identity reliably reports the imported cert+key, and the build subsequently succeeds end-to-end).

### Alternative fix (less preferred)

Inject Apple Root CA + WWDR G3 into the build keychain at creation time. This would let `-v` continue to make sense semantically. Trade-off: requires bundling/fetching the Apple intermediates and adds a second `security` call per build. The drop-`-v` fix has none of that overhead, doesn't change observable behavior (codesign still resolves trust), and matches the intent of the call (presence check, not validity check).

## Workaround (currently in use)

Manually edit `~/.npm/_npx/<hash>/node_modules/@expo/build-tools/dist/ios/credentials/keychain.js` to drop the `-v`. Fragile — wiped on `npx clear-cache`, on `eas-cli` upgrade (different `eas-cli-local-build-plugin` version → new npx hash → fresh download), and on fresh-Mac setup. Re-applying it programmatically requires a small idempotent shell script that globs the npx cache; we run such a script from our repo's `scripts/` directory after every cache invalidation event.

`credentialsSource: \"local\"` in `eas.json` does **not** work around this bug — `credentialsSource` only changes where eas-cli sources the .p12 (Expo's server vs. local `credentials.json`); the same `manager.js` → `keychain.js` code path runs downstream regardless. We discovered this the hard way after planning a credentials.json migration; the only way forward is dropping `-v` upstream.

## Environment

- macOS 26.0 (Tahoe)
- Xcode 26.x
- Node 24.15.0
- eas-cli 18.x (verified on 18.8.0 and 18.11.0)
- `eas-cli-local-build-plugin` 18.8.0 / `@expo/build-tools` 18.8.0
- fastlane 2.x

## Related code

- [`packages/build-tools/src/ios/credentials/keychain.ts`](https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/ios/credentials/keychain.ts) — `findIdentitiesByTeamId`, `ensureCertificateImported`
- [`packages/build-tools/src/ios/credentials/manager.ts`](https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/ios/credentials/manager.ts) — `prepareTargetCredentials` (calls `ensureCertificateImported` after import)

Happy to send a PR for the one-line fix if it's welcome.

kearnsm293-afk:
## Why

Closes #3678. On macOS 26 (Tahoe), `eas build --profile development --platform ios --local` fails at `Prepare credentials` with `Distribution certificate ... hasn't been imported successfully` even when the cert+key are correctly imported into the build keychain.

`Keychain.findIdentitiesByTeamId` runs:

```
security find-identity -v -s "(<teamId>)" <buildKeychainPath>
```

The `-v` ("valid identities only") flag requires the full trust chain (dist cert → Apple WWDR Intermediate → Apple Root CA) to resolve from the keychain(s) in the search list. The build keychain (created in `Keychain.create()`) only holds the cert + private key; Apple Root CA lives in `/Library/Keychains/System.keychain` and Apple WWDR G3 lives in the user's login keychain. `security find-identity` does **not** aggregate trust resolution across keychains passed as positional args — only `security list-keychains -s` does, and that change would be session-wide and undesirable.

On macOS 26 (Tahoe), this trust gate causes `find-identity -v -s "(<teamId>)" <buildKeychainPath>` to return 0 identities even when the cert+key were correctly imported, falsely tripping `ensureCertificateImported`.

## How

Drop `-v` from the `find-identity` call. The remaining flags (`-s "(<teamId>)" <buildKeychainPath>`) still constrain the match to the right team and keychain, so the `includes(fingerprint)` check in `ensureCertificateImported` correctly answers \"is the cert+key pair present in this keychain?\" The cert appears in the output annotated `(CSSMERR_TP_NOT_TRUSTED)` — that's expected and ignored. Codesign performs its own trust resolution downstream via `Security.framework` (which does aggregate across keychains), so signing still succeeds end-to-end.

The change is one line in `packages/build-tools/src/ios/credentials/keychain.ts` plus a comment block explaining the reasoning so a future reader doesn't add `-v` back.

## Test Plan

- The existing integration test [`packages/build-tools/src/ios/credentials/__integration-tests__/keychain.test.ios.ts`](https://github.com/expo/eas-cli/blob/main/packages/build-tools/src/ios/credentials/__integration-tests__/keychain.test.ios.ts) (\"shouldn't throw any error if the certificate has been imported successfully\") covers exactly this scenario. It currently fails on macOS 26 (Tahoe) due to this bug; the fix makes it pass.
- Manually verified end-to-end: `eas build --profile development --platform ios --local` against an Expo-managed dist cert + ad-hoc provisioning profile completes successfully on macOS 26.0 (Tahoe) with the patch applied. Produces a signed `.ipa` that installs and runs on a registered iOS 26 device.
- No regression risk on older macOS: dropping `-v` is strictly more permissive in what `find-identity` reports back. On systems where `-v` worked (older macOS, or systems where Apple Root CA happens to be visible in the build keychain's effective search list), the same cert+key pair still appears in the output without `-v`. The `includes(fingerprint)` check passes in both cases.

## Checklist

- [x] CHANGELOG.md updated under `## main` → `### 🐛 Bug fixes`. (PR URL placeholder will be replaced with the real PR number once this is opened.)
- [x] Commit message follows the `[package] Description` format.
- [x] No public API change. No breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
-- [build-tools] Fix `eas build --local` for iOS on macOS 26 (Tahoe) where `Keychain.findIdentitiesByTeamId` falsely reported the dist certificate as not imported. The `-v` flag on `security find-identity` requires the full trust chain to resolve from the build keychain alone, but the build keychain only holds the cert + private key — Apple Root CA lives in `/Library/Keychains/System.keychain` and `find-identity` does not aggregate trust resolution across keychains. Dropped `-v`; presence check now works across macOS versions and codesign continues to resolve trust downstream via `Security.framework`. ([#TODO](https://github.com/expo/eas-cli/pull/TODO) by [@kearnsm293-afk](https://github.com/kearnsm293-afk))
+- [build-tools] Fix `eas build --local` for iOS on macOS 26 (Tahoe) where `Keychain.findIdentitiesByTeamId` falsely reported the dist certificate as not imported. The `-v` flag on `security find-identity` requires the full trust chain to resolve from the build keychain alone, but the build keychain only holds the cert + private key — Apple Root CA lives in `/Library/Keychains/System.keychain` and `find-identity` does not aggregate trust resolution across keychains. Dropped `-v`; presence check now works across macOS versions and codesign continues to resolve trust downstream via `Security.framework`. ([#3679](https://github.com/expo/eas-cli/pull/3679) by [@kearnsm293-afk](https://github.com/kearnsm293-afk))
 
 ### 🧹 Chores
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- [build-tools] Fix `eas build --local` for iOS on macOS 26 (Tahoe) where `Keychain.findIdentitiesByTeamId` falsely reported the dist certificate as not imported. The `-v` flag on `security find-identity` requires the full trust chain to resolve from the build keychain alone, but the build keychain only holds the cert + private key — Apple Root CA lives in `/Library/Keychains/System.keychain` and `find-identity` does not aggregate trust resolution across keychains. Dropped `-v`; presence check now works across macOS versions and codesign continues to resolve trust downstream via `Security.framework`. ([#TODO](https://github.com/expo/eas-cli/pull/TODO) by [@kearnsm293-afk](https://github.com/kearnsm293-afk))
+
 ### 🧹 Chores
 
 ## [18.11.0](https://github.com/expo/eas-cli/releases/tag/v18.11.0) - 2026-05-05
--- a/packages/build-tools/src/ios/credentials/keychain.ts
+++ b/packages/build-tools/src/ios/credentials/keychain.ts
@@ -97,9 +97,23 @@ export default class Keychain<TJob extends Ios.Job> {
   }
 
   private async findIdentitiesByTeamId(teamId: string): Promise<string> {
+    // Note: no `-v` flag. `-v` ("valid identities only") requires the full
+    // trust chain (dist cert -> Apple WWDR Intermediate -> Apple Root CA) to
+    // resolve from the keychain(s) in the search list. The build keychain
+    // created above only holds the dist cert + private key; Apple Root CA
+    // lives in /Library/Keychains/System.keychain. `security find-identity`
+    // does not aggregate trust resolution across keychains passed as
+    // positional args (only `security list-keychains -s` does, and that's
+    // session-wide and undesirable). On macOS 26 (Tahoe), this caused
+    // `find-identity -v -s "(<teamId>)" <buildKeychainPath>` to return 0
+    // identities even when the cert+key were correctly imported, falsely
+    // tripping `ensureCertificateImported`. Without `-v`, the presence
+    // check works correctly across macOS versions; codesign performs its
+    // own trust resolution downstream via Security.framework (which does
+    // aggregate across keychains), so signing still succeeds.
     const { output } = await spawn(
       'security',
-      ['find-identity', '-v', '-s', `(${teamId})`, this.keychainPath],
+      ['find-identity', '-s', `(${teamId})`, this.keychainPath],
       {
         stdio: 'pipe',
       }


jaketoolson:
This has happened to me.

mifi:
Removing `-v` solved the problem for me too. Related #1331

keegzkenz:
Confirming this on macOS 27.0 (Xcode 27.0), and the diagnosis in this PR matches what I measured.

`eas build --local --platform ios` failed every time in `PREPARE_CREDENTIALS`:

```
Distribution certificate with fingerprint F88CCB5E... hasn't been imported successfully
```

fastlane's `import_certificate` reports success. The import genuinely works — it's only the `-v` check that disagrees. Instrumenting `ensureCertificateImported` to dump diagnostics against the live build keychain before it throws:

```
$ security find-identity -v -s "(TEAMID)" <temp.keychain>
Looking for identities matching "(TEAMID)"
     0 valid identities found

$ security find-identity <temp.keychain>          # same keychain, no -v
  1) F88CCB5E06E1BE328F5174C139BC3C36A87B0153 "iPhone Distribution: ... (TEAMID)" (CSSMERR_TP_NOT_TRUSTED)
     1 identities found
```

Cert and private key are both present, with exactly the fingerprint the check is looking for. It fails only the trust-validity filter, because the ephemeral keychain has no Apple trust chain and chain resolution doesn't aggregate from other keychains when an explicit keychain path is passed.

Things I ruled out before landing on `-v`, in case they save anyone else time:

- **Missing WWDR intermediate** — present in `/Library/Keychains/System.keychain`. Not the issue; it isn't consulted for this evaluation.
- **`.keychain` vs `.keychain-db` path mismatch** — the `-db` file doesn't exist here. `security create-keychain` with an explicit path outside `~/Library/Keychains` does not append the suffix.
- **Locked keychain / legacy PKCS#12 encryption** — neither.

`CSSMERR_TP_NOT_TRUSTED` is a red herring. Dropping `-v` produces a correctly signed build: the resulting `.ipa` verifies `valid on disk` / `satisfies its Designated Requirement`, with the full chain resolved at signing time from the provisioning profile:

```
Authority=iPhone Distribution: ... (TEAMID)
Authority=Apple Worldwide Developer Relations Certification Authority
Authority=Apple Root CA
```

So the check is asking a stricter question than it needs. It wants "is the cert+key pair in this keychain?", and `-v` answers "is this identity trust-valid right now in this context?" — which an ephemeral keychain can't satisfy on 26/27 by design.

Workaround I'm running in the meantime, in case it's useful to others: install `eas-cli-local-build-plugin` at the version eas-cli demands (the version of its bundled `@expo/eas-build-job`) into a local cache dir, apply this PR's one-line change there, and point `EAS_LOCAL_BUILD_PLUGIN_PATH` at that copy. Keeps the shared `npx` cache untouched and survives cache refreshes.

Would appreciate this landing so the override can go away. Latest published `@expo/build-tools` (24.7.0) still ships `-v`, so there's no version to upgrade to. Related: #3678, #3645, #4440.

PR #3681, opened 2026-05-06
[build-tools] Re-enable 3rd party precompiled modules

sjchmiela:
## Summary
- Re-enable `EXPO_PRECOMPILED_MODULES_BASE_URL` for iOS pod installs so third-party precompiled modules are used alongside Expo modules.
- Keep the env scoped to `pod install` only, and proxy the base URL through `EAS_BUILD_COCOAPODS_CACHE_URL` when CocoaPods cache is enabled.
- Update pod install unit coverage for the default storage URL and proxied URL behavior.

## Testing
- Added focused unit tests for direct and proxied precompiled modules base URL resolution.
- Ran `yarn exec oxfmt` on the touched files.
- Ran focused `@expo/build-tools` pod tests and package typecheck; both passed.

### DIFF
--- a/packages/build-tools/src/ios/__tests__/pod.test.ts
+++ b/packages/build-tools/src/ios/__tests__/pod.test.ts
@@ -19,9 +19,11 @@ const IOS_DIR = path.join(PROJECT_DIR, 'ios');
 
 function makeIosBuildContext({
   expoVersion,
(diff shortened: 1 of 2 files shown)


chrfalch (approved):
👍

PR #3699, opened 2026-05-08
[eas-cli] Add eas update:upload-embedded command

gwdp:
# Why

EAS Update bundle diffing can't generate patches against the JS bundle embedded in a native build, so the first OTA after install always downloads the full bundle. This command registers the embedded bundle with EAS Update so the diff worker can use it as a patch base.

Ref ENG-21033

# How

- New `eas update:upload-embedded` command with `--platform`, `--bundle`, `--manifest`, `--channel`, and optional `--build-id` / `--json` flags
- Reads `app.manifest` to extract the embedded update UUID and validates it
- Resolves `runtimeVersion` via the same helper `eas update` uses
- Gets a presigned URL via `getSignedEmbeddedUpdateAssetUploadSpecifications`, uploads the bundle
- Calls `uploadEmbeddedUpdate` with the storage key + manifest fields
- Retries on `EMBEDDED_UPDATE_ASSET_NOT_READY` (GCS finalization race) with exponential backoff; exits immediately on `EMBEDDED_UPDATE_CONFLICT`

# Test Plan

- `yarn jest src/commands/update/__tests__/upload-embedded.test.ts src/graphql/mutations/__tests__/EmbeddedUpdateMutation-test.ts` — 31 tests pass
- `yarn tsc && yarn lint` pass


gwdp:
Superseded by #3720 which has the corrected implementation (removed channelId lookup and launchAssetStorageKey, stacked with view-embedded and list-embedded).

PR #3703, opened 2026-05-11
[build-tools] bump xclogparser default version

hSATAC:
# Why

Bump the default xclogparser binary to v0.2.47, which is now uploaded to the GCS bucket and ready to use.

# How

Updated `DEFAULT_XCLOGPARSER_VERSION` in `packages/build-tools/src/steps/utils/ios/xcactivitylog.ts` from `v0.2.46` to `v0.2.47` and updated the matching expectations in the tests.

# Test Plan

- `yarn test` in `packages/build-tools` passes with the updated version string.


sjchmiela (approved):
Is it better? Are we going to need to bump it for every patch?

hSATAC:
> Is it better? Are we going to need to bump it for every patch?

It fixed some format issue so I think it's probably better to bump it.

I subscribed the release to our slack channel, I will check the release note once there's a new release and decide if we need to upgrade or not.

PR #3708, opened 2026-05-12
[build-tools] Route xcactivitylog parser failures to Sentry

hSATAC:
# Why

The xcactivitylog parser silently swallowed all internal failures at `debug` level. Users saw no report and we had no signal to triage. [ENG-21157](https://linear.app/expo/issue/ENG-21157)

These failures (Zod schema, spawn, download) are non-actionable for users and don't affect build artifacts. They should be silent for users, but visible to engineers via Sentry.

# How

- `parseAndReportXcactivitylog` tracks the current phase in a local variable: `creating_temp_directory` / `downloading_xclogparser` / `running_xclogparser` / `parsing_xclogparser_output`.
- On failure: user sees only `Build performance analysis skipped.` (info); the phase identifier + full error details routed to Sentry via `Sentry.capture(msg, err, { tags: { phase } })`.
- Phase values are stored in snake_case from the start (matching the existing Sentry tag convention in `worker/src/service.ts` / `restoreBuildCache.ts`) — no runtime transform at the call site.
- Tests assert the right phase tag fires per failure mode, and the happy path doesn't invoke Sentry.

# Test Plan

- `packages/build-tools/src/steps/utils/ios/__tests__/xcactivitylog.test.ts` — each failure path (invalid JSON, Zod schema mismatch, spawn failure, download failure) asserts `Sentry.capture` invoked with the correct phase tag; happy path asserts not invoked.
- Full monorepo `yarn typecheck`, `yarn run -T lerna run test`, `yarn lint`, `yarn fmt:check` green.

---

⬇️ Stacked on top of #3712 (Sentry singleton). Merge #3712 first.

### DIFF
--- a/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
+++ b/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
@@ -7,17 +7,17 @@ import os from 'os';
 import path from 'path';
 import { z } from 'zod';
 
+import { Sentry } from '../../../sentry';
+
 const DEFAULT_XCLOGPARSER_VERSION = 'v0.2.47';
 const XCLOGPARSER_DOWNLOAD_URL = 'https://storage.googleapis.com/turtle-v2/xclogparser';
 const XCLOGPARSER_DOWNLOAD_TIMEOUT_MS = 20_000;
 const XCLOGPARSER_OUTPUT_FILENAME = 'xcactivitylog.json';
 
 /**
- * Download xclogparser, parse xcactivitylog from derived data, and log a
- * compile metrics report. Never throws — all errors are logged at debug level.
- *
- * Can be called from both the step-based flow (BuildFunction) and the
- * traditional builder flow (runBuildPhase).
+ * Never throws — best-effort observability that does not affect build status.
+ * Failures route to Sentry via `Sentry.capture` for engineering triage;
+ * users see only a generic skip message.
  */
 export async function parseAndReportXcactivitylog({
   derivedDataPath,
@@ -33,26 +33,36 @@ export async function parseAndReportXcactivitylog({
   proxyBaseUrl?: string;
 }): Promise<void> {
   let tempDir: string | undefined;
+  let phase = 'creating_temp_directory';
   try {
     tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'xclogparser-'));
+
+    phase = 'downloading_xclogparser';
     const xclogparserPath = await downloadXclogparser(
       tempDir,
       xclogparserVersion,
       logger,
       proxyBaseUrl
     );
+
+    phase = 'running_xclogparser';
     const jsonOutputPath = await runXclogparser({
       binaryPath: xclogparserPath,
       derivedDataPath,
       workspacePath,
       outputDir: tempDir,
     });
+
+    phase = 'parsing_xclogparser_output';
     const data = XcactivitylogDataSchemaZ.parse(
       JSON.parse(await fs.readFile(jsonOutputPath, 'utf8'))
     );
+
     logger.info(formatReport(data));
-  } catch (err: unknown) {
-    logger.debug({ err }, 'Failed to analyze build performance; continuing without a report');
+  } catch (err: any) {
+    logger.info('Build performance analysis skipped.');
+    const msg = `Build performance analysis failed during "${phase}"`;
+    Sentry.capture(msg, err, { tags: { phase } });
   } finally {
     if (tempDir) {
       await asyncResult(fs.rm(tempDir, { force: true, recursive: true }));
--- a/packages/build-tools/src/steps/utils/ios/__tests__/xcactivitylog.test.ts
+++ b/packages/build-tools/src/steps/utils/ios/__tests__/xcactivitylog.test.ts
@@ -4,6 +4,7 @@ import fs from 'fs-extra';
 import path from 'path';
 
 import { createMockLogger } from '../../../../__tests__/utils/logger';
+import { Sentry } from '../../../../sentry';
 import {
   buildTargetMetrics,
   isCompileStep,
@@ -16,6 +17,13 @@ jest.mock('@expo/downloader');
 jest.mock('@expo/turtle-spawn', () => ({
   __esModule: true,
   default: jest.fn(),
+}));
+jest.mock('../../../../sentry', () => ({
+  Sentry: {
+    setup: jest.fn(),
+    capture: jest.fn(),
+    flush: jest.fn(),
+  },
 }));
 
 // Minimal fixture: two modules with compile steps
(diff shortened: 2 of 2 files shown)


packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+        } catch {
+          serialized = Flatted.stringify(err);
+        }
+        sentryNode.captureMessage(`${msg}\n${serialized}`);
+      }
+    });
+  },
+
+  handleCritical(msg: string, err?: Error, options: HandleErrorOptions = {}): void {
+    Sentry.handleError(`Critical error, might require an urgent action [${msg}]`, err, {
```
sjchmiela: let's not follow existing convention from external/sentry. prefixing the message with superficial "critical error !!11 oen one" doesn't help and amkes it difficult to differentiate between different erros

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
+import { bunyan } from '@expo/logger';
+import * as sentryNode from '@sentry/node';
+import Flatted from 'flatted';
+
+interface SetupOptions {
```
sjchmiela: inline

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+  dsn: string | null;
+  environment: string;
+  tags?: Record<string, string>;
+  logger: bunyan;
+}
+
+interface HandleErrorOptions {
```
sjchmiela: inline, easier when you remove handlecritical

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+let state: SentryState | null = null;
+
+export const Sentry = {
+  setup({ dsn, environment, tags, logger }: SetupOptions): void {
+    if (dsn) {
+      sentryNode.init({ dsn, environment });
+    }
+    state = { defaultTags: tags ?? {}, defaultLogger: logger };
```
sjchmiela: i wonder if there's a built-in way to give all errors of sentry tags…

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  handleError(msg: string, err?: Error, options: HandleErrorOptions = {}): void {
+    if (!state) {
+      return;
+    }
+
+    const tags = { ...state.defaultTags, ...options.tags };
+    const extras = options.extras ?? {};
+    const logger = options.logger ?? state.defaultLogger;
+    logger.error({ tags, extras, err }, msg);
```
sjchmiela: i think handleError hsould only send to sentry, not log?

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+      tags: { ...options.tags, severity: 'critical' },
+    });
+  },
+
+  async flush(timeoutMs: number = 2000): Promise<boolean> {
+    if (!state) {
+      return true;
+    }
+    return sentryNode.flush(timeoutMs);
+  },
```
sjchmiela: this does not appear to be an async function

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+        sentryNode.captureException(err);
+      } else if (err === undefined) {
+        sentryNode.captureMessage(msg);
+      } else {
+        let serialized: string;
+        try {
+          serialized = JSON.stringify(err);
+        } catch {
+          serialized = Flatted.stringify(err);
+        }
```
sjchmiela: looks like we can expect `err` to be an instance of `Error` (by the type) or to be missing?

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+        sentryNode.captureMessage(`${msg}\n${serialized}`);
```
sjchmiela: i don't think this is a good convention. whatever we need attached we should add as `extras`

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const Sentry = {
+  setup({ dsn, environment, tags, logger }: SetupOptions): void {
+    if (dsn) {
+      sentryNode.init({ dsn, environment });
+    }
+    state = { defaultTags: tags ?? {}, defaultLogger: logger };
+  },
+
+  handleError(msg: string, err?: Error, options: HandleErrorOptions = {}): void {
```
sjchmiela: i think it might be easier to have `captureMessage(msg, err?, options)` and attach `err` as an extra? this way msg is alwasy present and categorizes error nicely?

sjchmiela: i think https://github.com/expo/universe/blob/013a4b06bb377bd84f024da6ee12889358245cc6/server/libraries/server-logging/src/sentry.ts#L99-L112 is a good pattern?

sjchmiela (changes requested):

hSATAC:
* **#3708** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3708?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3708?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3712** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3712?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hSATAC:
Will address the comments of this round in https://github.com/expo/eas-cli/pull/3712 because we split into stacked PRs. Sorry for the inconvenience and thanks for the quick review.

packages/build-tools/src/sentry.ts
hSATAC: We probably need to write logs from the callers to keep the behavior identical as it is now. Otherwise we'll lose logs.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -33,26 +33,38 @@ export async function parseAndReportXcactivitylog({
...
       JSON.parse(await fs.readFile(jsonOutputPath, 'utf8'))
     );
+
     logger.info(formatReport(data));
   } catch (err: unknown) {
-    logger.debug({ err }, 'Failed to analyze build performance; continuing without a report');
+    logger.info('Build performance analysis skipped.');
+    const msg = `Build performance analysis failed during "${phase}"`;
+    Sentry.captureMessage(msg, err instanceof Error ? err : undefined, {
+      tags: { phase: phase.replace(/\s+/g, '-') },
```
sjchmiela: I guess we could just write the phase with underscores from the start

sjchmiela (approved):

PR #3710, opened 2026-05-12
[eas-cli] Add eas simulator:get command

szdziedzic:
# Why

We need a way to inspect an existing remote simulator session and retrieve its shell credentials again — for example after losing the original terminal output from `eas simulator:start`. Without this, the only way to re-attach is to start a new session.

# How

- Adds a hidden `eas simulator:get --id <ID>` command that fetches the device run session by ID.
- Always prints session id, type, status, job run id, job run status, and job run URL (blank when the turtle job run is not available yet).
- Prints the shell snippet to attach to the session only when the session is `IN_PROGRESS` and `remoteConfig` is present.
- Extracts `formatRemoteConfigShellSnippet` (and the `DeviceRunSessionRemoteConfig` type) into `src/simulator/utils.ts` so `start.ts` and the new `get.ts` can share it. The helper lives outside `src/commands/` so oclif does not try to treat it as a command.
- Expands `DeviceRunSessionByIdQuery` with the app's `slug` and `ownerAccount.name` so the job-run URL can be built from the session response.

# Test Plan

Ran against a running session:

```
$ easd simulator:get --id 019e1c30-465b-7fd9-98a1-eb0ceaafafb5
✔ Fetched device run session 019e1c30-465b-7fd9-98a1-eb0ceaafafb5

ID:               019e1c30-465b-7fd9-98a1-eb0ceaafafb5
Type:             AGENT_DEVICE
Status:           IN_PROGRESS
Job run ID:       019e1c30-465b-75e1-b746-a528054649e9
Job run status:   IN_PROGRESS
Job run URL:      https://expo.dev/accounts/expo/projects/testapp/job-runs/019e1c30-465b-75e1-b746-a528054649e9

🔑 Run the following in your shell to attach to the session:

export AGENT_DEVICE_DAEMON_BASE_URL='...'
export AGENT_DEVICE_DAEMON_AUTH_TOKEN='...'
```

- `yarn typecheck`, `yarn lint`, and `yarn fmt:check` pass.
- Verified the previous `command simulator:utils not found` warning is gone (helper now lives outside `commands/`).


szdziedzic:
* **#3710** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3710?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/simulator/get.ts
```diff
@@ -0,0 +1,67 @@
...
+      ? getBareJobRunUrl(session.app.ownerAccount.name, session.app.slug, session.turtleJobRun.id)
+      : '';
+
+    Log.newLine();
+    Log.log(`ID:               ${session.id}`);
+    Log.log(`Type:             ${session.type}`);
+    Log.log(`Status:           ${session.status}`);
+    Log.log(`Job run ID:       ${session.turtleJobRun?.id ?? ''}`);
+    Log.log(`Job run status:   ${session.turtleJobRun?.status ?? ''}`);
+    Log.log(`Job run URL:      ${jobRunUrl ? link(jobRunUrl) : ''}`);
```
szdziedzic: I think we should hide it from users in final version but for now it seems to be useful?

sjchmiela: we could call it just "URL"

packages/eas-cli/src/commands/simulator/get.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    if (session.status === DeviceRunSessionStatus.InProgress && session.remoteConfig) {
```
sjchmiela: should we print if remote config is not there yet?

sjchmiela (approved):

packages/eas-cli/src/commands/simulator/get.ts
szdziedzic: good idea

packages/eas-cli/src/commands/simulator/get.ts
szdziedzic: good suggestion

PR #3712, opened 2026-05-12
[build-tools] Move Sentry SDK from worker to build-tools as a singleton

hSATAC:
# Why

The Sentry SDK lives in `packages/worker/` today, so build-tools code only reaches it through the `ctx.reportError` callback wired in `worker/src/context.ts`. Step-based code in `@expo/steps` has no equivalent channel. Hosting the SDK as a module-level singleton inside `@expo/build-tools` lets any caller use `Sentry.capture(...)` directly. [ENG-21166](https://linear.app/expo/issue/ENG-21166)

# How

- New `packages/build-tools/src/sentry.ts` exports `Sentry` with `setup` / `capture` / `flush`. Three overloads on `capture`: `(msg, options?)`, `(msg, err, options?)`, `(err, options?)`.
- Routing: with `err` → `captureException(err)` (Sentry parses `err.stack` into structured frames); the caller `msg` is attached as a `'message'` extra only when distinct from `err.message`. Without `err` → `captureMessage(msg)`. Fingerprinting is left to Sentry defaults. Mirrors www's `SentryReportingStream`.
- 12 worker call sites switch from `sentry.handleError(...)` to `sentry.capture(...)`. Each call site explicitly logs to bunyan before Sentry — the wrapper no longer dual-writes.
- Worker exit paths now `await sentry.flush()` before `process.exit(...)` so terminal events aren't lost (the deleted class had no flush method).
- `@sentry/node` (7.77.0) moves to build-tools' direct deps; worker pulls it transitively. `packages/worker/src/external/sentry.ts` deleted.

# Test plan

- `packages/build-tools/src/__tests__/sentry.test.ts`: covers all three overloads, the conditional `'message'` extra, primitive/non-`Error` coercion at the `catch (err: any)` boundary, and `flush` forwarding.
- `packages/worker/src/__unit__/sentry.test.ts`: asserts module-load wiring.
- Monorepo typecheck / test / lint / fmt green.

---

⬆️ Stacked under #3708 (xcactivitylog parser using the singleton). Merge this first.


hSATAC:
* **#3708** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3708?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3712** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,66 @@
...
+  captureMessage(
+    msg: string,
+    err?: Error,
+    options: {
+      tags?: Record<string, string>;
+      extras?: Record<string, unknown>;
+      level?: sentryNode.SeverityLevel;
+    } = {}
+  ): void {
+    if (!isSetup) {
```
sjchmiela: Does Sentry log or throw errors when calling when not set up? I would expect it to be ok with that

packages/worker/src/service.ts
```diff
@@ -220,23 +220,22 @@ export default class BuildService {
...
-          startBuildTime: String(this.startBuildTime),
-          isSockedClosed: String(!this.ws),
-          wasSocketClosedAtBuildFinish: String(wasSocketClosedAtBuildFinish),
-        },
-      }
-    );
+    const hangingMsg =
+      'Worker still alive 5 minutes after sending BuildSuccess/BuildError message - possibly hanging';
+    logger.error(hangingMsg);
+    sentry.captureMessage(hangingMsg, undefined, {
```
sjchmiela: Maybe it makes sense to export a second function head and support calling captureMessage without undefined for err?

packages/worker/src/service.ts
```diff
@@ -320,7 +319,8 @@ export default class BuildService {
       const maybeRawError =
         error instanceof errors.ExpoError && error.cause instanceof Error ? error.cause : error;
 
-      sentry.handleError(err.message, maybeRawError, {
+      logger.error({ err: maybeRawError }, err.message);
+      sentry.captureMessage(err.message, maybeRawError, {
```
sjchmiela: And maybe another one for (err, { tags })?

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    sentryNode.withScope(scope => {
+      if (options.tags) {
+        scope.setTags(options.tags);
+      }
+      if (options.extras) {
+        scope.setExtras(options.extras);
+      }
+      if (err) {
+        scope.setExtra('err', err);
```
sjchmiela: Is it the same as captureException? I'd expect captureException to do more than just pass err as extra. (Unless proven it doesn't matter I'd suggest using captureException for errors (for stacktrace) and captureMessage for strings)

packages/worker/src/ws.ts
```diff
@@ -53,13 +55,16 @@ function handleConnection(socket: WebSocket, service: BuildService): void {
         );
       }
     } catch (err: any) {
-      sentry.handleError('Unhandled WebSocket message', err);
+      const msg = 'Unhandled WebSocket message';
+      logger.error({ err }, msg);
+      sentry.captureMessage(msg, err);
```
sjchmiela: Or maybe we want to have a sentry.error({ err, msg, tags, extras })...

We can come back to this later

Actually @AbbanMustafa and I were talking today about revisiting how we send stuff to Sentry too and I was independently thinking a few weeks ago how nicer stuff is in www where we just report on logger.error and we don't need to think about Sentry... so maybe there is going to be a sequel to this which is going to remove the manual calls and is going to just hook to logger... though I still think what you're doing here is helpful - for places where WE want to learn about something but we don't necessarily want to log it to the user. In backend there are only developer-facing logs, in worker we have logs for user and, until this pull request, no way to leave traces for us

sjchmiela (approved):
Thank you!! I'm thinking already of all the places I want to put a manual Sentry.capture on for myself

packages/build-tools/src/sentry.ts
hSATAC: You're right. I checked the Sentry/node implementation, `captureException` actually uploads more information then just `captureMessage`.

I'll modify the logic here based on https://github.com/expo/universe/blob/013a4b06bb377bd84f024da6ee12889358245cc6/server/libraries/server-logging/src/sentry.ts#L99-L112

packages/build-tools/src/sentry.ts
```diff
@@ -0,0 +1,81 @@
...
+type CaptureOptions = {
+  tags?: Record<string, string>;
+  extras?: Record<string, unknown>;
+  level?: sentryNode.SeverityLevel;
+};
+
+type SentryAPI = {
+  setup(opts: { dsn: string | null; environment: string; tags?: Record<string, string> }): void;
+  capture(msg: string, options?: CaptureOptions): void;
+  capture(msg: string, err: Error | undefined, options?: CaptureOptions): void;
```
sjchmiela, line 12: Would it be easier to require err to be Error and never undefined?

sjchmiela (approved):

packages/build-tools/src/sentry.ts
hSATAC: That was my previous implementation and then I found myself adding boilerplate code like `if err sentry.capture; else sentry.capture` in multiple places so I think this sugar is worth adding.

PR #3716, opened 2026-05-12
[eas-cli] Add `--refresh-ad-hoc-provisioning-profile` for non-interactive builds

sswrk:
# Why

When a new device is added to an account, and a non-interactive development build is run (e.g. invoked by CI or by a workflow job), even without the `freeze-credentials` flag, the provisioning profile is not refreshed, and the internal build is not available for the newly registered device. See the Linear issue: https://linear.app/expo/issue/ENG-7041/allow-including-all-eas-registered-devices-automatically-in-non

# How

Added an opt-in ad-hoc provisioning profile refresh to non-interactive builds with flag `--refresh-ad-hoc-provisioning-profile`.

When the flag is present, we hit `www` GQL API to get the ASC API key (the submission key), and we invoke the ad-hoc profile sync with all devices selected. `packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts` is where this logic happens.

The flag mustn't be present when `--freeze-credentials` is present, these are conflicting flags.

I'm leaning towards an opt-in flag instead of extending the default behavior because of the extra credentials configuration requirements and side effects of the provisioning profile refresh logic.

# Test Plan

Added/amended tests.

Manual verification:
1. Did `easd build --platform ios --profile development --non-interactive --refresh-ad-hoc-credential`, verified the provisioning profile matches the set of Apple devices on my account.
2. Changed the set of Apple devices on my account.
3. `easd build --platform ios --profile development --non-interactive --refresh-ad-hoc-credential` again, verified the re-generated provisioning profile matches the Apple devices on my account.

### DIFF
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
@@ -6,7 +6,11 @@ import nullthrows from 'nullthrows';
 
 import { resolveAppleTeamIfAuthenticatedAsync } from './AppleTeamUtils';
 import { assignBuildCredentialsAsync, getBuildCredentialsAsync } from './BuildCredentialsUtils';
-import { chooseDevicesAsync, formatDeviceLabel } from './DeviceUtils';
+import {
+  chooseDevicesAsync,
+  filterDevicesForApplePlatform,
+  formatDeviceLabel,
+} from './DeviceUtils';
 import { SetUpDistributionCertificate } from './SetUpDistributionCertificate';
 import DeviceCreateAction, { RegistrationMethod } from '../../../devices/actions/create/action';
 import {
@@ -150,13 +154,13 @@ export class SetUpAdhocProvisioningProfile {
     const provisionedDeviceIdentifiers = (
       currentBuildCredentials?.provisioningProfile?.appleDevices ?? []
     ).map(i => i.identifier);
+    const applePlatform = getApplePlatformFromTarget(target);
     const chosenDevices =
       ctx.nonInteractive && ctx.refreshAdHocProvisioningProfile
-        ? registeredAppleDevices
+        ? filterDevicesForApplePlatform(registeredAppleDevices, applePlatform)
         : await chooseDevicesAsync(registeredAppleDevices, provisionedDeviceIdentifiers);
 
     // 4. Reuse or create the profile on Apple Developer Portal
-    const applePlatform = getApplePlatformFromTarget(target);
     const profileType =
       applePlatform === ApplePlatform.TV_OS
         ? ProfileType.TVOS_APP_ADHOC
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
@@ -429,6 +429,8 @@ export class SetUpAdhocProvisioningProfile {
       ascApiKey: resolvedKey.ascApiKey,
       teamId: resolvedKey.teamId,
       teamName: resolvedKey.teamName,
+      // Provide a non-enterprise team type to avoid interactive team-type resolution.
+      // Ad-hoc profile handling below uses explicit ProfileType and does not branch on team.inHouse.
       teamType: AppleTeamType.COMPANY_OR_ORGANIZATION,
     });
   }
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
@@ -129,9 +129,9 @@ export class SetUpAdhocProvisioningProfile {
       appleTeam
     );
     if (registeredAppleDevices.length === 0) {
-      if (ctx.nonInteractive && ctx.refreshAdHocProvisioningProfile) {
+      if (ctx.nonInteractive) {
         throw new Error(
-          'No devices are registered for this Apple team. Register devices with eas device:create before refreshing ad-hoc provisioning profile.'
+          'No devices are registered for this Apple team. Register devices with eas device:create first.'
         );
       }
       const shouldRegisterDevices = await confirmAsync({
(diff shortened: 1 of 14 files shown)


sswrk:
* **#3716** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3716?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3717](https://github.com/expo/eas-cli/pull/3717) <a href="https://app.graphite.com/github/pr/expo/eas-cli/3717?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3739](https://github.com/expo/eas-cli/pull/3739) <a href="https://app.graphite.com/github/pr/expo/eas-cli/3739?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
```diff
@@ -29,9 +31,13 @@ import {
 import differenceBy from '../../../utils/expodash/differenceBy';
 import { CredentialsContext } from '../../context';
 import { MissingCredentialsNonInteractiveError } from '../../errors';
+import { getAscApiKeyForAppSubmissionsAsync } from '../api/GraphqlClient';
```
sjchmiela, line 38: should we use a different API key than the one "for submissions"?
we also have an API key "for builds" _and_ we also have the App Store Connect connection.

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
```diff
@@ -51,11 +57,23 @@ export class SetUpAdhocProvisioningProfile {
 
   async runAsync(ctx: CredentialsContext): Promise<IosAppBuildCredentialsFragment> {
     const { app } = this.options;
+
+    if (ctx.refreshAdHocProvisioningProfile && ctx.freezeCredentials) {
+      throw new Error(
+        'Cannot refresh ad-hoc provisioning profile when credentials are frozen. Disable --freeze-credentials or --refresh-ad-hoc-provisioning-profile.'
```
sjchmiela:
```suggestion
        'Cannot refresh ad-hoc provisioning profile when credentials are frozen. Remove --freeze-credentials or --refresh-ad-hoc-provisioning-profile.'
```

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
```diff
@@ -371,6 +398,65 @@ export class SetUpAdhocProvisioningProfile {
...
+          '  - EAS credentials service: configure an App Store Connect API Key for submissions on this app'
+      );
+    }
+
+    await ctx.appStore.ensureAuthenticatedAsync({
+      mode: AuthenticationMode.API_KEY,
+      ascApiKey: resolvedKey.ascApiKey,
+      teamId: resolvedKey.teamId,
+      teamName: resolvedKey.teamName,
+      teamType: AppleTeamType.COMPANY_OR_ORGANIZATION,
```
sjchmiela, line 434: I presume this is a requirement for ad-hoc distribution?

sjchmiela (approved):
this looks good to me, but i'd also wait for Quin to take a look

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
quinlanj: i made a column for asc api keys for builds but i never ended finishing the feature so it is unpopulated, unfortunately

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
```diff
@@ -111,6 +129,11 @@ export class SetUpAdhocProvisioningProfile {
       appleTeam
     );
     if (registeredAppleDevices.length === 0) {
+      if (ctx.nonInteractive && ctx.refreshAdHocProvisioningProfile) {
```
quinlanj: this should be the error message for all non interactive calls when there are no registered devices

```suggestion
      if (ctx.nonInteractive) {
```

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
```diff
@@ -127,10 +150,10 @@ export class SetUpAdhocProvisioningProfile {
     const provisionedDeviceIdentifiers = (
       currentBuildCredentials?.provisioningProfile?.appleDevices ?? []
     ).map(i => i.identifier);
-    const chosenDevices = await chooseDevicesAsync(
-      registeredAppleDevices,
-      provisionedDeviceIdentifiers
-    );
+    const chosenDevices =
```
quinlanj, line 158: We should filter the registered devices based on the target type. For example, if an apple tv is registered as a device but the target type is ios, the operation will fail

quinlanj (changes requested):
requesting changes because we should filter devices based on the target type

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
sswrk: Added filtering.

Also realized that `www` Apple devices API has no proper support for apple tv device class. I'm wondering why. I'll look into adding this to `www` API, for now I'll throw when target apple platform is TV_OS or VISION_OS.

packages/eas-cli/src/credentials/ios/actions/SetUpAdhocProvisioningProfile.ts
sswrk: Not a strict requirement, rather a default value to avoid prompting, just like here: https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/metadata/auth.ts#L173-L174

This is then used to derive the value of `authCtx.team.inHouse`, which is not used for anything in ad-hoc provisioning profile creation.

It's a bit confusing, so I'll add a comment. In a longer term, it could be nice to make this not require passing `teamType` in all scenarios.

quinlanj (approved):

PR #3720, opened 2026-05-12
[eas-cli] Add eas update:embedded:upload command

gwdp:
# Why

EAS Update bundle diffing can't generate patches against the JS bundle embedded in a native build, so the first OTA after install always downloads the full bundle. This command registers the embedded bundle with EAS Update so the diff worker can use it as a patch base.

Ref ENG-21033

# How

- New `eas update:embedded:upload` command with `--platform`, `--bundle`, `--manifest`, `--channel`, and optional `--build-id` / `--json` flags
- Uploads the bundle via presigned URL then registers it via `uploadEmbeddedUpdate`; two-phase spinner shows progress for each step
- Retries registration on `EMBEDDED_UPDATE_ASSET_NOT_AVAILABLE` (GCS finalization race) with exponential backoff; other errors bubble up with the server's message (e.g. `EMBEDDED_UPDATE_ALREADY_EXISTS` for re-uploads)

# Test Plan

- `yarn jest src/commands/update/embedded/__tests__/upload.test.ts src/graphql/__tests__/client-test.ts` — 24 tests pass
- `yarn tsc && yarn lint` pass
- Ran locally against a real iOS archive:

```
$ node bin/dev update:embedded:upload \
  --platform ios \
  --bundle ios/CoinFlip.xcarchive/Products/Applications/CoinFlipPreview.app/main.jsbundle \
  --manifest ios/CoinFlip.xcarchive/Products/Applications/CoinFlipPreview.app/EXUpdates.bundle/app.manifest \
  --channel production
✔ Uploaded bundle
✔ Registered ios embedded update (runtimeVersion: 1.0.0, channel: "production")
Embedded update ID: 5f2b4c4b-77a0-45cc-9320-bc8ec2eb36fb
```


gwdp:
* **#3722** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3722?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3721** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3721?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3767** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3720** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3720?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/update/embedded/upload.ts
```diff
@@ -0,0 +1,169 @@
...
+    } = await this.getContextAsync(UpdateEmbeddedUpload, {
+      nonInteractive,
+      withServerSideEnvironment: null,
+    });
+
+    if (jsonFlag) {
+      enableJsonOutput();
+    }
+
+    if (!fs.existsSync(bundlePath)) {
```
quinlanj: use `fs-extra` so you can await instead

packages/eas-cli/src/commands/update/embedded/upload.ts
```diff
@@ -0,0 +1,169 @@
...
+          turtleBuildId: buildId,
+        });
+        break;
+      } catch (e: unknown) {
+        if (!isEmbeddedUpdateAssetNotAvailableError(e)) {
+          registerSpinner.fail('Failed to register embedded update');
+          Log.warn(
+            'This bundle may already be registered. To replace it, delete it manually and re-upload.'
+          );
+          throw e;
```
quinlanj: This code will run on any error that is not an `EmbeddedUpdateAssetNotAvailableError`, so it wouldn't make sense to print this since we don't know if the issue is bundleregistration (ie) network error, etc. 

Also I'm not sure i follow the use case here, if the bundle is already registered (the embedded asset) wouldn't the creation of the embedded update succeed? Or if you meant the EmbeddedUpdateEntity already exists, we can detect that from www and throw a special error to catch that specific case (so we aren't telling users to delete manually when we catch a generic error)

packages/eas-cli/src/update/embeddedManifest.ts
```diff
@@ -0,0 +1,35 @@
+import fs from 'fs';
+import * as uuid from 'uuid';
+
+export function readEmbeddedManifest(manifestPath: string): { id: string } {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(fs.readFileSync(manifestPath, 'utf8'));
```
quinlanj: use `fs-extra` so we can await

quinlanj (changes requested):

packages/eas-cli/src/commands/update/embedded/upload.ts
gwdp: Thanks for catching that @quinlanj, updated!

packages/eas-cli/src/update/embeddedManifest.ts
gwdp: Updated, thanks!

packages/eas-cli/src/commands/update/embedded/upload.ts
gwdp: You're right, I was missing a way to recognize the duplicate explicitly. 

On the universe side, opened https://github.com/expo/universe/pull/27602 to surface it as a dedicated EMBEDDED_UPDATE_ALREADY_EXISTS error code  Here I added isEmbeddedUpdateAlreadyExistsError(e) (same shape as isEmbeddedUpdateAssetNotAvailableError) and the "delete and re-upload" hint now only fires when matched — for any other non-retry error it just bubbles up with no false guess.

Keeping this in draft until the universe PR lands so the CLI and server land in order. 🙇‍♂️

packages/eas-cli/src/commands/update/embedded/upload.ts
```diff
@@ -0,0 +1,166 @@
...
+        embeddedUpdate = await EmbeddedUpdateMutation.uploadEmbeddedUpdateAsync(graphqlClient, {
+          appId: projectId,
+          platform: appPlatform,
+          runtimeVersion,
+          channel: channelName,
+          embeddedUpdateId,
+          turtleBuildId: buildId,
+        });
+        break;
+      } catch (e: unknown) {
```
quinlanj, line 140: let's catch EMBEDDED_UPDATE_ALREADY_EXISTS here now :)

gwdp: Thanks for the review @quinlanj I actually had this and pulled it because with both our hint and the re-thrown server error, the message printed twice (plus a `Request ID:` line). Re-adding it now — but I'll `Errors.error(...)` with a single clean message for the `EMBEDDED_UPDATE_ALREADY_EXISTS` case instead of warn + re-throw, so it shows once without the Request ID noise. Other non-retry errors still bubble up as-is.

quinlanj: oh right, good point. thanks for explaining. you can feel free to remove the catch then, i'm fine either way!

quinlanj (approved):

packages/eas-cli/src/commands/update/embedded/upload.ts
gwdp: No sweat, I honestly liked the improved handling better if you don't mind keeping it. Thanks for the input, always up to make things better!

PR #3730, opened 2026-05-14
[eas-cli] Add observe:routes for navigation metrics by route name

douglowder:
# Why

New command `observe:routes` to get navigation metrics by route.

```
USAGE
  $ eas observe:routes [--platform android|ios] [--metric cold_ttr|warm_ttr|nav_tti...] [--stat
    median|med|p90|count|event_count|eventCount...] [--after <value>] [--limit <value>] [--start <value> | --days
    <value>] [--end <value> | ] [--app-version <value>] [--update-id <value>] [--build-number <value>] [--project-id
    <value>] [--json] [--non-interactive]

FLAGS
  --after=<value>         Cursor for pagination. Use the endCursor from a previous query to fetch the next page.
  --app-version=<value>   Filter by app version
  --build-number=<value>  Filter by app build number
  --days=<value>          Show results from the last N days (mutually exclusive with --start/--end)
  --end=<value>           End of time range (ISO date)
  --json                  Enable JSON output, non-JSON messages will be printed to stderr. Implies --non-interactive.
  --limit=<value>         The number of items to fetch each query. Defaults to 50 and is capped at 200.
  --metric=<option>...    Navigation metric to display (can be specified multiple times). Defaults to all three.
                          <options: cold_ttr|warm_ttr|nav_tti>
  --non-interactive       Run the command in non-interactive mode.
  --platform=<option>     Filter by platform
                          <options: android|ios>
  --project-id=<value>    EAS project ID (defaults to the project ID of the current directory)
  --start=<value>         Start of time range (ISO date)
  --stat=<option>...      Statistic to display per metric (can be specified multiple times)
                          <options: median|med|p90|count|event_count|eventCount>
  --update-id=<value>     Filter by EAS update ID

DESCRIPTION
  display app navigation route metrics (Cold TTR, Warm TTR, TTI) grouped by route name

```

# How

Added a new command that that queries the new navigationRoutes GraphQL field to display Cold TTR, Warm TTR, and TTI, broken down per route name and platform.

# Test Plan

Tested with staging data in `@expokadi/cadence` project.

New unit tests added.

### DIFF
--- a/packages/eas-cli/src/commands/observe/routes.ts
+++ b/packages/eas-cli/src/commands/observe/routes.ts
@@ -1,0 +1,170 @@
+import { Flags } from '@oclif/core';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
+import { getLimitFlagWithCustomValues } from '../../commandUtils/pagination';
+import Log from '../../log';
+import { fetchObserveNavigationRoutesAsync } from '../../observe/fetchNavigationRoutes';
+import {
+  NAVIGATION_METRICS,
+  NavigationMetricKey,
+  NavigationStatKey,
+  buildObserveNavigationRoutesJson,
+  buildObserveNavigationRoutesTable,
+  resolveNavigationMetricKey,
+  resolveNavigationStatKey,
+} from '../../observe/formatNavigationRoutes';
+import { allowedPlatformFlagValues, appPlatformsFromFlag } from '../../observe/platforms';
+import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
+import { resolveTimeRange } from '../../observe/startAndEndTime';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+const DEFAULT_ROUTES_LIMIT = 50;
+
+const DEFAULT_METRIC_KEYS: NavigationMetricKey[] = NAVIGATION_METRICS.map(m => m.key);
+
+const METRIC_ALIAS_OPTIONS = [
+  'cold_ttr',
+  'warm_ttr',
+  'tti',
+  'nav_cold_ttr',
+  'nav_warm_ttr',
+  'nav_tti',
+  'expo.navigation.cold_ttr',
+  'expo.navigation.warm_ttr',
+  'expo.navigation.tti',
+];
+
+const STAT_OPTIONS = ['median', 'med', 'p90', 'count', 'event_count', 'eventCount'];
+
+const DEFAULT_STATS_TABLE: NavigationStatKey[] = ['median', 'count'];
+const DEFAULT_STATS_JSON: NavigationStatKey[] = ['median', 'p90', 'count'];
+
+export default class ObserveRoutes extends EasCommand {
+  static override hidden = true;
+  static override description =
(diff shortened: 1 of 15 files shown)


douglowder:
* **#3731** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3731?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3730** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3730?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3730?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/observe/__tests__/formatNavigationRoutes.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    const output = buildObserveNavigationRoutesTable(
+      routes,
+      ['expo.navigation.cold_ttr', 'expo.navigation.warm_ttr', 'expo.navigation.tti'],
+      ['median', 'count'],
+      { daysBack: 7 }
+    );
+
+    expect(output).toContain('Med values (navigation count) for the last 7 days');
+    expect(output).toContain('Cold TTR');
+    expect(output).toContain('Warm TTR');
```
Ubax, line 106: Shouldn't this be `Nav Cold TTR` and `Nav Warm TTR`?

Can you add one more test that validates that the whole table header has expected shape?

packages/eas-cli/src/observe/__tests__/formatNavigationRoutes.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    expect(output).toContain('Nav TTI');
+    expect(output).toContain('/home');
+    expect(output).toContain('/profile');
+    expect(output).toContain('0.50s (10)');
+    expect(output).toContain('0.70s (5)');
+    expect(output).toContain('iOS');
+    expect(output).toContain('Android');
```
Ubax, line 113: What do you think about splitting the result by columns and validating that the values belong to correct columns?

packages/eas-cli/src/observe/__tests__/formatNavigationRoutes.test.ts
Ubax: WDYT about adding an additional test case where we do `expect(output).toMatchInlineSnapshot`, to validate the whole table output is correct? This would ensure that we will get the correct table shape.

packages/eas-cli/src/commands/observe/routes.ts
```diff
@@ -0,0 +1,155 @@
...
+    }),
+    'update-id': Flags.string({
+      description: 'Filter by EAS update ID',
+    }),
+    'build-number': Flags.string({
+      description: 'Filter by app build number',
+    }),
+    'project-id': Flags.string({
+      description: 'EAS project ID (defaults to the project ID of the current directory)',
+    }),
```
Ubax, line 81: What do you think about adding a filter for routeName(s)?

So that people can get stats only for several routes instead of all of them

Ubax (approved):

packages/eas-cli/src/commands/observe/routes.ts
douglowder: We can certainly do that, but we would first have to add route name to the GraphQL filter on www. I'll have a look at doing that.

PR #3739, opened 2026-05-18
[eas-cli]  Add --refresh-distribution-certificate for non-intractive builds 

sswrk:
# Why

When there's no valid distribution certificate associated with an app, and a non-interactive development build is run (e.g. invoked by CI or by a workflow job), even without the freeze-credentials flag, the distribution certificate is not validated or refreshed, which may fail the internal build and require doing manual `eas build` to create a new certificate. See the Linear issue: https://linear.app/expo/issue/ENG-21330/make-it-possible-to-refresh-a-distribution-certificate-in-non

# How

Added an opt-in distribution certificate refresh to non-interactive builds with flag `--refresh-distribution-certificate`.

When the flag is present, we hit www GQL API to get the ASC API key (the submission key), and we distribution certificate validation or set-up if needed. `packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts` is where this logic happens.

The flag mustn't be present when `--freeze-credentials` is present, these are conflicting flags.

I'm leaning towards an opt-in flag instead of extending the default behavior because of the extra credentials configuration requirements and side effects of the distribution certificate refresh logic.

# Test Plan

Added unit tests.

Manual verification:
- Did `easd build --platform ios --profile development --non-interactive --refresh-distribution-certificate` in the following scenarios, verified it works:
  - current certificate is valid
  - current certificate is invalid
- Did `easd build --platform ios --profile development --non-interactive --refresh-distribution-certificate --refresh-ad-hoc-provisioning-profile`, verified the ASC API key got resolved just once.

### DIFF
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
@@ -1,5 +1,3 @@
-import assert from 'assert';
-
 import { resolveAppleTeamIfAuthenticatedAsync } from './AppleTeamUtils';
 import { tryAuthenticateAppStoreWithEasAscApiKeyAsync } from './AscApiKeyUtils';
 import { CreateDistributionCertificate } from './CreateDistributionCertificate';
@@ -70,8 +68,10 @@ export class SetUpDistributionCertificate {
       AppleTeamType.COMPANY_OR_ORGANIZATION
     );
 
-    if (await this.isCurrentCertificateValidAsync(ctx, currentCertificate)) {
-      assert(currentCertificate, 'currentCertificate is defined here');
+    if (
+      currentCertificate &&
+      (await this.isCurrentCertificateValidAsync(ctx, currentCertificate))
+    ) {
       Log.log('Using existing valid distribution certificate.');
       return currentCertificate;
     }
@@ -104,8 +104,10 @@ export class SetUpDistributionCertificate {
     ctx: CredentialsContext,
     currentCertificate: AppleDistributionCertificateFragment | null
   ): Promise<AppleDistributionCertificateFragment> {
-    if (await this.isCurrentCertificateValidAsync(ctx, currentCertificate)) {
-      assert(currentCertificate, 'currentCertificate is defined here');
+    if (
+      currentCertificate &&
+      (await this.isCurrentCertificateValidAsync(ctx, currentCertificate))
+    ) {
       return currentCertificate;
     }
 
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
@@ -58,6 +58,17 @@ export class SetUpDistributionCertificate {
     ctx: CredentialsContext,
     currentCertificate: AppleDistributionCertificateFragment | null
   ): Promise<AppleDistributionCertificateFragment> {
+    if (!ctx.refreshDistributionCertificate) {
+      Log.addNewLineIfNone();
+      Log.warn(
+        'Using the existing distribution certificate without validating it against Apple servers. Use --refresh-distribution-certificate to validate and refresh if needed.'
+      );
+      if (!currentCertificate) {
+        throw new MissingCredentialsNonInteractiveError();
+      }
+      return currentCertificate;
+    }
+
     if (!currentCertificate) {
       throw new MissingCredentialsNonInteractiveError();
     }
--- a/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
+++ b/packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
@@ -1,7 +1,7 @@
 import assert from 'assert';
 
 import { resolveAppleTeamIfAuthenticatedAsync } from './AppleTeamUtils';
-import { resolveAscApiKeyForAppCredentialsAsync } from './AscApiKeyUtils';
+import { tryAuthenticateAppStoreWithEasAscApiKeyAsync } from './AscApiKeyUtils';
 import { CreateDistributionCertificate } from './CreateDistributionCertificate';
 import { formatDistributionCertificate } from './DistributionCertificateUtils';
 import {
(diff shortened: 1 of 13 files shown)


sswrk:
* **#3739** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3739?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3739?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3805** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3805?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
```diff
@@ -50,13 +53,64 @@ export class SetUpDistributionCertificate {
...
+    if (ctx.refreshDistributionCertificate) {
+      await this.ensureAppStoreAuthenticatedForDistCertRefreshAsync(ctx);
+
+      if (await this.isCurrentCertificateValidAsync(ctx, currentCertificate)) {
+        assert(currentCertificate, 'currentCertificate is defined here');
+        Log.log('Using existing valid distribution certificate.');
+        return currentCertificate;
+      }
+
+      Log.warn('Current distribution certificate is invalid. Creating a new one...');
```
quinlanj: move this log.warn to when we actually create a new cert, since it's possible we could reuse
```suggestion
```

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
```diff
@@ -50,13 +53,64 @@ export class SetUpDistributionCertificate {
...
+      const validDistCerts = await this.getValidDistCertsAsync(ctx);
+      if (validDistCerts.length > 0) {
+        const cert = validDistCerts[0];
+        Log.log(`Reusing distribution certificate with serial number ${cert.serialNumber}`);
+        return cert;
+      }
+      return await this.createNewDistCertAsync(ctx);
```
quinlanj:
```suggestion
       Log.warn('Current distribution certificate is invalid. Creating a new one...');
      return await this.createNewDistCertAsync(ctx);
```

quinlanj (approved):
lgtm. I'm surprised this was a feature people asked for because the number of dist certs apple lets you have in your account is very low (2-3 max)

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
```diff
@@ -50,13 +53,64 @@ export class SetUpDistributionCertificate {
...
+
+    const resolvedKey = await resolveAscApiKeyForAppCredentialsAsync({
+      graphqlClient: ctx.graphqlClient,
+      app: this.app,
+    });
+    if (!resolvedKey) {
+      throw new Error(
+        'No App Store Connect API Key found for distribution certificate refresh. In non-interactive mode, provide one via:\n' +
+          '  - Environment variables: EXPO_ASC_API_KEY_PATH, EXPO_ASC_KEY_ID, EXPO_ASC_ISSUER_ID\n' +
+          '  - EAS credentials service: configure an App Store Connect API Key for submissions on this app'
```
sjchmiela: ugh it's still weird to me we're saying to use key for submissions to generate certificates

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
```diff
@@ -50,13 +53,64 @@ export class SetUpDistributionCertificate {
...
-    _ctx: CredentialsContext,
+    ctx: CredentialsContext,
     currentCertificate: AppleDistributionCertificateFragment | null
   ): Promise<AppleDistributionCertificateFragment> {
-    // TODO: implement validation
+    if (ctx.refreshDistributionCertificate) {
+      await this.ensureAppStoreAuthenticatedForDistCertRefreshAsync(ctx);
+
+      if (await this.isCurrentCertificateValidAsync(ctx, currentCertificate)) {
+        assert(currentCertificate, 'currentCertificate is defined here');
```
sjchmiela:
```suggestion
      if (currentCertificate && await this.isCurrentCertificateValidAsync(ctx, currentCertificate)) {
```
this makes more sense to me?

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
sjchmiela: how likely is that we're going to hit https://github.com/expo/eas-cli/blob/62ab329f3900974641f9dae24a888b74cd0b0bcd/packages/eas-cli/src/credentials/ios/actions/DistributionCertificateUtils.ts#L219-L223 often? can we do something about it?

sjchmiela (approved):

packages/eas-cli/src/credentials/ios/actions/SetUpDistributionCertificate.ts
sswrk: We're going to hit this when a new certificate needs to be created and there are already 3 (non-matching, e.g. for different teams) certificates in the Apple account. I don't think we should do anything automatically in this case, as this would involve revoking unrelated certs. We could make an opt-in flag to "force override" a cert.

sswrk:
Changed this to follow the "best effort" approach like provisioning profiles in https://github.com/expo/eas-cli/pull/3805 instead of being opt-in with a flag, as I don't see any downsides of just attempting to do the validation when an ASC API key is available in EAS. If we were to do something like "force setup dist cert even if some other unrelated (other apple team) cert needs to be revoked", that would be an opt-in flag for special use cases.

@quinlanj Apologies, re-requesting review.

quinlanj (changes requested):
The code looks beautiful, I'd push back on this from a product perspective though. For a lot of developers, distribution certs are often treated like account-wide signing infrastructure, rather than app-local disposable state. I suspect there will be users who will be unpleasantly surprised to discover the submission API key was used as implicit consent to create signing certs in non interactive mode.

My vote is to put this behind the original flag you had unless you felt strongly about this

sswrk:
> [...] My vote is to put this behind the original flag you had unless you felt strongly about this

This is some amazing feedback, thank you. Reverting this to the opt-in approach, I'm happy to lean on your expertise

PR #3750, opened 2026-05-19
[build-tools] Enable build time statistics for xcodebuild

hSATAC:
# Why

The `PARSE_XCACTIVITYLOG` phase was gated behind `EXPERIMENTAL_EAS_XCACTIVITYLOG=1`. The underlying `parseAndReportXcactivitylog` is documented as "Never throws — best-effort observability that does not affect build status" and routes all failures to Sentry, so it is safe to enable by default for cloud iOS builds.

Linear: ENG-21340.

# How

- Removed the `EXPERIMENTAL_EAS_XCACTIVITYLOG` env check in `packages/build-tools/src/builders/ios.ts`.
- Added a `ctx.isLocal` early-return inside the phase so `eas build --local` keeps skipping the parse, matching the existing `SAVE_CACHE` local-skip pattern in the same file.

No server-side or worker coordination is required — the env was only referenced in this single file.

# Test Plan

- `cd packages/build-tools && yarn test` — 516/516 passing.
- `yarn fmt:check` — clean.
- Existing `parseAndReportXcactivitylog` unit tests (`packages/build-tools/src/steps/utils/ios/__tests__/xcactivitylog.test.ts`) continue to cover the parse/download/report paths.
- Behavior after this change:
  - Cloud iOS build → runs `PARSE_XCACTIVITYLOG` and prints the build performance report.
  - `eas build --local` → enters the phase, logs `Local builds skip build performance analysis.`, returns immediately.
  - Any error inside `parseAndReportXcactivitylog` → captured by Sentry, build status unaffected.

### DIFF
--- a/packages/build-tools/src/builders/ios.ts
+++ b/packages/build-tools/src/builders/ios.ts
@@ -173,18 +173,20 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
     });
   }
 
-  if (ctx.env.EXPERIMENTAL_EAS_XCACTIVITYLOG === '1') {
-    const { derivedDataPath, workspacePath } = nullthrows(fastlaneResult);
-    await ctx.runBuildPhase(BuildPhase.PARSE_XCACTIVITYLOG, async () => {
-      await parseAndReportXcactivitylog({
-        derivedDataPath,
-        workspacePath,
-        logger: ctx.logger,
-        proxyBaseUrl: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL,
-        env: ctx.env,
-      });
+  const { derivedDataPath, workspacePath } = nullthrows(fastlaneResult);
+  await ctx.runBuildPhase(BuildPhase.PARSE_XCACTIVITYLOG, async () => {
+    if (ctx.isLocal) {
+      ctx.logger.info('Local builds skip build performance analysis.');
+      return;
+    }
+    await parseAndReportXcactivitylog({
+      derivedDataPath,
+      workspacePath,
+      logger: ctx.logger,
+      proxyBaseUrl: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL,
+      env: ctx.env,
     });
-  }
+  });
 
   await ctx.runBuildPhase(BuildPhase.PRE_UPLOAD_ARTIFACTS_HOOK, async () => {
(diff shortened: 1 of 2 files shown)


hSATAC:
@AbbanMustafa I disabled the xcactivitylog parsing phase for `eas build --local` since it depends on external tooling. For Gradle profiling it's a built-in feature, so I think either approach is reasonable there.

AbbanMustafa:
> @AbbanMustafa I disabled the xcactivitylog parsing phase for `eas build --local` since it depends on external tooling. For Gradle profiling it's a built-in feature, so I think either approach is reasonable there.

Nice! makes sense, I will keep it on by default. Models love this stuff

AbbanMustafa (approved):
I just merged updating schema in www which includes this build phase too https://github.com/expo/universe/pull/27529/changes#diff-e3dc3175010bd503c883a6c8b45b470141528aa5c8cdcf93da22b5e4a75e4e9bR4226

so this should be good to merge once this is done rolling out

packages/build-tools/src/builders/ios.ts
```diff
@@ -173,17 +173,19 @@ async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
...
+      return;
+    }
+    await parseAndReportXcactivitylog({
+      derivedDataPath,
+      workspacePath,
+      logger: ctx.logger,
+      proxyBaseUrl: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL,
     });
-  }
+  });
```
sjchmiela:
```suggestion

  await ctx.runBuildPhase(BuildPhase.PARSE_XCACTIVITYLOG, async () => {
    if (ctx.isLocal) {
      ctx.logger.info('Local builds skip build performance analysis.');
      return;
    }
    try {
    
      const { derivedDataPath, workspacePath } = nullthrows(fastlaneResult);
    await parseAndReportXcactivitylog({
      derivedDataPath,
      workspacePath,
      logger: ctx.logger,
      proxyBaseUrl: ctx.env.EAS_BUILD_COCOAPODS_CACHE_URL,
});
} catch (err) {
// log error
ctx.markBuildPhaseSkipped();
}
  });
```
?

sjchmiela (approved):
i was also talking to Mustafa how if we logged the JSON that we format into a table as a bunyan log property we could parse it with React components in website and display Actually Nice™️ report from xcactivity with bars and colors and stuff. Something we might want to look into down the road!

PR #3768, opened 2026-05-21
[eas-cli] eas go: prompt for SDK version interactively

gwdp:
## Summary

- Adds an interactive SDK version select prompt to `eas go` when `--sdk-version` is not passed.
- Fetches `supportedSdkVersions` from the GraphQL API (expo/universe#27451) — returns only versions with an actual repack target, labeled as Latest, Beta, or by major version number.
- Defaults the cursor to the latest released version; falls back to the last entry if no version is marked latest.
- `--sdk-version` and current project scope bypasses the prompt entirely

## Validation

- Ran `eas go` without `--sdk-version` — prompted with SDK list, latest selected by default.
- Ran `eas go --sdk-version 55.0.0` — no prompt, proceeds directly.
- Ran `eas go` within a 55.x project, auto-selected, no prompt

### DIFF
--- a/packages/eas-cli/src/commands/go.ts
+++ b/packages/eas-cli/src/commands/go.ts
@@ -306,7 +306,7 @@ export default class Go extends EasCommand {
 
   private async selectSdkVersionAsync(
     graphqlClient: ExpoGraphqlClient
-  ): Promise<string | undefined> {
+  ): Promise<{ sdkVersion: string | undefined }> {
     let versions;
     try {
       versions = await WorkflowRunQuery.expoGoSupportedSdkVersionsAsync(graphqlClient);
--- a/packages/eas-cli/src/commands/go.ts
+++ b/packages/eas-cli/src/commands/go.ts
@@ -212,7 +212,7 @@ export default class Go extends EasCommand {
     }
     let sdkVersion = flags['sdk-version'] ?? detectedSdkVersion;
     if (!sdkVersion) {
-      sdkVersion = await this.selectSdkVersionAsync(graphqlClient);
+      ({ sdkVersion } = await this.selectSdkVersionAsync(graphqlClient));
     }
     const bundleId = flags['bundle-id'] ?? this.generateBundleId(actor);
     if (!isBundleIdentifierValid(bundleId)) {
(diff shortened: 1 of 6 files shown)


gwdp:
* **#3768** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3776** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3776?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/go.ts
```diff
@@ -301,6 +304,35 @@ export default class Go extends EasCommand {
     }
   }
 
+  private async selectSdkVersionAsync(
+    graphqlClient: ExpoGraphqlClient
+  ): Promise<string | undefined> {
```
```suggestion
  ): Promise<{ sdkVersion: string | undefined }> {
```
if you do it like this `return;` being a shorthand for `return undefined` won't be allowed (and using a shorthand like this is a bit weird)

you can also just write out `return undefined`

packages/eas-cli/src/commands/go.ts
```diff
@@ -301,6 +304,35 @@ export default class Go extends EasCommand {
...
+    const defaultVersion = selectable.find(v => v.isLatest) ?? selectable.at(-1);
+    return selectAsync(
+      'Select an Expo SDK version',
+      selectable.map(v => {
+        const major = v.sdkVersion.split('.')[0];
+        const title = v.isLatest
+          ? `Latest (SDK ${major})`
+          : v.isBeta
+            ? `Beta (SDK ${major})`
+            : `SDK ${major}`;
```
sjchmiela: not sure what you think of that but maybe "SDK XX" / "SDK XX (latest)" / "SDK XX (beta)" would look better? all "SDK" aligned and then the appendix?

sjchmiela: Image

packages/eas-cli/src/commands/go.ts
```diff
@@ -301,6 +304,35 @@ export default class Go extends EasCommand {
...
+    }
+    const selectable = versions.filter(v => !v.isDeprecated);
+    if (selectable.length === 0) {
+      return;
+    }
+    const defaultVersion = selectable.find(v => v.isLatest) ?? selectable.at(-1);
+    return selectAsync(
+      'Select an Expo SDK version',
+      selectable.map(v => {
+        const major = v.sdkVersion.split('.')[0];
```
sjchmiela: looking at every time we split sdk version by `.` i wonder if maybe sdkversions from graphql shouldn't have have been just
```
key: String!
label: String!
isDefault: Boolean!
```

sjchmiela (approved):

packages/eas-cli/src/commands/go.ts
gwdp: Thank you, 100% my bad, I should have followed that standard to start with 🙇‍♂️

packages/eas-cli/src/commands/go.ts
gwdp: I agree with your point about the structure, but if not missing anything, it wouldn't clean up much of the `.split('.')[0]` because that is only executed once for the GraphQL.

My intention here was to avoid making the server responsible for the presentation layer and instead focus on providing the data needed for the CLI to handle that. I'm happy to bounce more here and come with another iteration if anything we can do better 🙇‍♂️

PR #3772, opened 2026-05-21
[build-tools] keep exit code & stderr when xclogparser failed

hSATAC:
# Why

Sentry shows ~13 events from the iOS build worker where `xclogparser` exits non-zero, but the only info captured is the spawn command line + exit code. The actual stderr is sitting on the rejected Error (`@expo/spawn-async` does `Object.assign(error, { stdout, stderr, status, signal })`), but `Sentry.captureException` only surfaces `err.message` and `err.stack`. Without stderr we can't tell whether `xclogparser` is failing on a missing log, an unsupported SLF version, an internal crash, or something else.

# How

In the `parseAndReportXcactivitylog` catch block, extract `stderr` / `stdout` / `status` / `signal` from the error and pass them as Sentry `extras`. A conditional spread keeps the `extras` key out when no spawn-shaped fields are present, so non-spawn error paths (`RangeError`, `SyntaxError`, `ZodError`) report the same payload as before. `stderr` and `stdout` are capped at the last 4000 chars to bound event payload size.

# Test Plan

- New unit test `surfaces spawn error stderr/stdout/status/signal as Sentry extras` constructs a spawn-shaped Error and asserts the expected extras land in `Sentry.capture`.
- All 37 tests in `xcactivitylog.test.ts` pass. The four existing tests for non-spawn error paths continue to assert `{ tags: { phase } }` exactly and pass unchanged because the conditional spread omits `extras` when nothing is populated.
- Real validation comes from Sentry after deploy: the ongoing `running_xclogparser` events should start showing stderr within 24–48h.

### DIFF
--- a/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
+++ b/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
@@ -82,14 +82,14 @@ export async function parseAndReportXcactivitylog({
   } catch (err: any) {
     logger.info('Build performance analysis skipped.');
     const msg = `Build performance analysis failed during "${phase}"`;
-    const extras: Record<string, unknown> = {};
-    if (typeof err?.stderr === 'string') extras.stderr = err.stderr.slice(-4000);
-    if (typeof err?.stdout === 'string') extras.stdout = err.stdout.slice(-4000);
-    if (err?.status != null) extras.exitStatus = err.status;
-    if (err?.signal != null) extras.signal = err.signal;
     Sentry.capture(msg, err, {
       tags: { phase },
-      ...(Object.keys(extras).length > 0 ? { extras } : {}),
+      extras: {
+        exitStatus: err?.status,
+        signal: err?.signal,
+        stderr: err?.stderr?.slice(-4000),
+        stdout: err?.stdout?.slice(-4000),
+      },
     });
   } finally {
     if (tempDir) {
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -82,7 +82,15 @@ export async function parseAndReportXcactivitylog({
...
     const msg = `Build performance analysis failed during "${phase}"`;
-    Sentry.capture(msg, err, { tags: { phase } });
+    const extras: Record<string, unknown> = {};
+    if (typeof err?.stderr === 'string') extras.stderr = err.stderr.slice(-4000);
+    if (typeof err?.stdout === 'string') extras.stdout = err.stdout.slice(-4000);
+    if (err?.status != null) extras.exitStatus = err.status;
+    if (err?.signal != null) extras.signal = err.signal;
+    Sentry.capture(msg, err, {
+      tags: { phase },
+      ...(Object.keys(extras).length > 0 ? { extras } : {}),
```
sjchmiela: we can just always pass in extras even if empty?

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -82,7 +82,15 @@ export async function parseAndReportXcactivitylog({
   } catch (err: any) {
     logger.info('Build performance analysis skipped.');
     const msg = `Build performance analysis failed during "${phase}"`;
-    Sentry.capture(msg, err, { tags: { phase } });
+    const extras: Record<string, unknown> = {};
+    if (typeof err?.stderr === 'string') extras.stderr = err.stderr.slice(-4000);
+    if (typeof err?.stdout === 'string') extras.stdout = err.stdout.slice(-4000);
+    if (err?.status != null) extras.exitStatus = err.status;
+    if (err?.signal != null) extras.signal = err.signal;
```
sjchmiela: would it work if we just did

```
extras: {
  exit_status: err.status,
  signal: err.signal,
  stderr: err.stderr?.slice(-4000)
```
or sth like that?

sjchmiela (approved):

PR #3777, opened 2026-05-21
feat(simulator): add exec command and config persistence via dotenv

krystofwoldrich:
## Goal
When testing the `simulator:start` I've noticed that agents correctly use the recommended agent-device envs, but because agents can't set envs for all child processes, they must include them in every agent-device call. This fogs up what is the agent doing, see the example below.

This PR proposes an option to save simulator session config to `.env.eas-simulator`, which can later be used by new exec command. This make working with the sim less verbose and also persist the config in case of agent cli crash, new sessions...

This PR also adds `EAS_SIMULATOR_SESSION_ID` env which is saved to the dotenv file for later reference to the session. This let's user call all `simulator:*` commands without specifying the `--id`.

### Before 

Because of the truncation it's not clear what command was executed.

Screenshot 2026-05-21 at 23 15 36

### After

Now it's clearly visible.

Screenshot 2026-05-21 at 23 13 32

### Considered alternative solutions

`.env.eas-simulator` **currently implemented** dedicated donenv file for the simulator session config. The benefit is isolation and easy of prune of the session, no parsing of the common `.env` files. It's used in the industry for example `.env.sentry-sourcemaps`.

`.env.local` already established file for ephemeral envs, contains other user's/project's envs would require parsing to clean up 

`.expo/.eas-simulator-config` new specialized config file for simulator session. Seems like the best next option, but makes less transparent if someone would not want to use `simulator:exec`, `.env.local` is common for ephemeral vals.

`dotenv agent-device <command>` only works out of the box with `.env` not `.env.local`. Other eas command like `convex` use `.env.local` which is better as typically not committed file for tmp session envs.

`./agent-device.json` can have both url and token properties same as when using env, but it's unclear to me if this file is typically committed or not, there are no secrets, so my expectations are it's likely ment to be committed.

## Summary
- add `eas simulator:exec` to run simulator commands with local .env files loaded
- make `simulator:start` write dotenv config by default, including agent-device envs and session id
- keep `--out-config-type env` for print-only shell exports

## Test plan
- tests

### Start command output

Screenshot 2026-05-25 at 17 01 14

### Interactions without `--id`

```bash
eas simulator:start # creates .env.eas-simulator
eas simulator:get # uses id from .env.eas-simulator
eas simulator:stop # uses id from .env.eas-simulator

eas simulator:start # overwrites .env.eas-simulator and logs out the previous id
```

Why start with `--force` by default? This minimizes the number of fails runs where user/agent haven't stopped the prev session and the env file have not cleaned up. It's possible to use `--no-force` if needed. We do only check the session id presence not the session state, to avoid extra network req on every start. This works well with the default force.

### DIFF
--- a/packages/eas-cli/src/commands/simulator/exec.ts
+++ b/packages/eas-cli/src/commands/simulator/exec.ts
@@ -1,18 +1,22 @@
 import spawnAsync from '@expo/spawn-async';
-import pkgDir from 'pkg-dir';
 
 import EasCommand from '../../commandUtils/EasCommand';
-import { loadSimulatorEnvironmentVariablesAsync } from '../../simulator/env';
+import { SIMULATOR_DOTENV_FILE_NAME, loadSimulatorEnvAsync } from '../../simulator/env';
 
 export default class SimulatorExec extends EasCommand {
   static override hidden = true;
-  static override description =
-    '[EXPERIMENTAL] execute a simulator command with local .env files loaded';
+  static override description = `[EXPERIMENTAL] execute a simulator command with ${SIMULATOR_DOTENV_FILE_NAME} environment loaded`;
   static override strict = false;
 
+  static override contextDefinition = {
+    ...this.ContextOptions.ProjectDir,
+  };
+
   async runAsync(): Promise<void> {
-    const projectDir = (await pkgDir(process.cwd())) ?? process.cwd();
-    await loadSimulatorEnvironmentVariablesAsync(projectDir);
+    const { projectDir } = await this.getContextAsync(SimulatorExec, {
+      nonInteractive: true,
+    });
+    await loadSimulatorEnvAsync(projectDir);
 
     const [command, ...args] = this.argv as [string, ...string[]];
     await spawnAsync(command, args, {
--- a/packages/eas-cli/src/commands/simulator/exec.ts
+++ b/packages/eas-cli/src/commands/simulator/exec.ts
@@ -15,8 +15,9 @@ export default class SimulatorExec extends EasCommand {
   private isRunningSubprocess = false;
 
   async runAsync(): Promise<void> {
-    const [command, ...args] = this.argv;
-    if (!command) {
+    const { argv } = await this.parse(SimulatorExec);
+    const [command, ...args] = argv as string[];
+    if (typeof command !== 'string' || command.length === 0) {
       throw new Error('No command provided. Run `eas simulator:exec <command> [args...]`.');
     }
 
--- a/packages/eas-cli/src/commands/simulator/exec.ts
+++ b/packages/eas-cli/src/commands/simulator/exec.ts
@@ -1,8 +1,8 @@
-import { load } from '@expo/env';
 import spawnAsync from '@expo/spawn-async';
 import pkgDir from 'pkg-dir';
 
 import EasCommand from '../../commandUtils/EasCommand';
+import { loadSimulatorEnvironmentVariablesAsync } from '../../simulator/env';
 
 export default class SimulatorExec extends EasCommand {
   static override hidden = true;
@@ -12,7 +12,7 @@ export default class SimulatorExec extends EasCommand {
 
   async runAsync(): Promise<void> {
     const projectDir = (await pkgDir(process.cwd())) ?? process.cwd();
-    load(projectDir, { force: true, silent: true });
+    await loadSimulatorEnvironmentVariablesAsync(projectDir);
 
     const [command, ...args] = this.argv as [string, ...string[]];
     await spawnAsync(command, args, {
--- a/packages/eas-cli/src/commands/simulator/exec.ts
+++ b/packages/eas-cli/src/commands/simulator/exec.ts
@@ -12,13 +12,20 @@ export default class SimulatorExec extends EasCommand {
     ...this.ContextOptions.ProjectDir,
   };
 
+  private isRunningSubprocess = false;
+
   async runAsync(): Promise<void> {
+    const [command, ...args] = this.argv;
+    if (!command) {
+      throw new Error('No command provided. Run `eas simulator:exec <command> [args...]`.');
+    }
+
     const { projectDir } = await this.getContextAsync(SimulatorExec, {
       nonInteractive: true,
     });
     await loadSimulatorEnvAsync(projectDir);
 
-    const [command, ...args] = this.argv as [string, ...string[]];
+    this.isRunningSubprocess = true;
     await spawnAsync(command, args, {
       stdio: 'inherit',
       env: process.env,
@@ -27,7 +34,10 @@ export default class SimulatorExec extends EasCommand {
 
   protected override catch(err: Error): Promise<any> {
     // Propagate wrapped command from spawnAsync rejection
-    process.exitCode = process.exitCode ?? (err as any).status ?? 1;
-    return Promise.resolve();
+    if (this.isRunningSubprocess) {
+      process.exitCode = process.exitCode ?? (err as any).status ?? 1;
+      return Promise.resolve();
+    }
+    return super.catch(err);
   }
 }
--- a/packages/eas-cli/src/commands/simulator/exec.ts
+++ b/packages/eas-cli/src/commands/simulator/exec.ts
@@ -25,8 +25,9 @@ export default class SimulatorExec extends EasCommand {
     });
   }
 
-  // eslint-disable-next-line async-protect/async-suffix
-  protected override async catch(err: Error): Promise<any> {
+  protected override catch(err: Error): Promise<any> {
+    // Propagate wrapped command from spawnAsync rejection
     process.exitCode = process.exitCode ?? (err as any).status ?? 1;
+    return Promise.resolve();
   }
 }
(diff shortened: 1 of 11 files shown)


packages/eas-cli/src/commands/simulator/exec.ts
```diff
@@ -0,0 +1,32 @@
...
+    await spawnAsync(command, args, {
+      stdio: 'inherit',
+      env: process.env,
+    });
+  }
+
+  // eslint-disable-next-line async-protect/async-suffix
+  protected override async catch(err: Error): Promise<any> {
+    process.exitCode = process.exitCode ?? (err as any).status ?? 1;
+  }
```
szdziedzic: do we need this?

packages/eas-cli/src/commands/simulator/stop.ts
```diff
@@ -35,25 +40,33 @@ export default class SimulatorStop extends EasCommand {
...
     } = await this.getContextAsync(SimulatorStop, {
       nonInteractive,
     });
 
-    const stopSpinner = ora(`🛑 Stopping device run session ${flags.id}`).start();
+    await loadSimulatorEnvAsync(projectDir);
+    const flagId = flags.id || process.env[EAS_SIMULATOR_SESSION_ID];
+    if (!flagId) {
+      throw new Error('Missing required flag id');
+    }
```
szdziedzic, line 55: should we also reset dotenv for stop command?

packages/eas-cli/src/simulator/env.ts
```diff
@@ -0,0 +1,49 @@
+import { loadEnvFiles, loadProjectEnv } from '@expo/env';
+import * as fs from 'fs-extra';
+import path from 'path';
+
+export const SIMULATOR_DOTENV_FILE_NAME = '.env.eas-simulator';
+export const EAS_SIMULATOR_SESSION_ID = 'EAS_SIMULATOR_SESSION_ID';
+export const SIMULATOR_DOTENV_FILE_HEADER =
+  '# Do not commit this file.\n# It holds configuration only for the current simulator session.\n\n';
```
szdziedzic: Nice future improvement: maybe we can ask user if they want to add `.env.eas-simulator` to gitignore in start command?

szdziedzic: Maybe we should also mention that user shouldn't modify it manually?

packages/eas-cli/src/simulator/utils.ts
```diff
@@ -21,17 +21,44 @@ export function deviceRunSessionTypeToFlagValue(type: DeviceRunSessionType): str
...
+    case 'ServeSimRunSessionRemoteConfig':
+      return {};
+  }
+}
+
+type RemoteSessionInstructionsConfigType = 'env' | 'dotenv';
+
+export function formatRemoteSessionInstructions(
+  remoteConfig: DeviceRunSessionRemoteConfig,
+  configType: RemoteSessionInstructionsConfigType = 'env'
```
szdziedzic: I don't think we need default here

packages/eas-cli/src/commands/simulator/exec.ts
```diff
@@ -0,0 +1,32 @@
+import spawnAsync from '@expo/spawn-async';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import { SIMULATOR_DOTENV_FILE_NAME, loadSimulatorEnvAsync } from '../../simulator/env';
+
+export default class SimulatorExec extends EasCommand {
```
szdziedzic, line 6: We can also try to think if we can allow users to have N sessions running at once and still be able to use exec. This is not part of this PR.

szdziedzic (commented):
awesome, I think this is useful and good idea overall

I have a few small comments/ideas

some of them we can add as follow ups to not block this PR

packages/eas-cli/src/commands/simulator/exec.ts
krystofwoldrich: I wasn't sure at first, but I checked and we need it for the passthru of the exist code of the wrapped command.

Added comment in 390a6cc3d9eeab3f42a9c355d38c5f79a760fc5d

packages/eas-cli/src/commands/simulator/stop.ts
krystofwoldrich: I originally added that as well.

But I think it's better that the `eas simulator:stop` is idempotent which it would not be if we reset the file. So the second call would fail on missing `--id`.

packages/eas-cli/src/simulator/env.ts
krystofwoldrich: Added in 67bb8718249ccb452963d8b5ad3efadf44aeb65a

packages/eas-cli/src/simulator/utils.ts
krystofwoldrich: Fixed in https://github.com/expo/eas-cli/pull/3777/commits/1a45db5b4a9e4ecb698b9e1d51dca3db47dd6188

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -50,11 +62,23 @@ export default class SimulatorStart extends EasCommand {
...
+      description:
+        '[default: true] Create a new device session even when an existing simulator session is present in the environment.',
+      default: true,
+      allowNo: true,
+    }),
+    'out-config-type': Flags.option({
+      description: `How to output simulator connection configuration. Use "env" to print shell exports, or "dotenv" to write ${SIMULATOR_DOTENV_FILE_NAME}.`,
+      options: Object.values(OUT_CONFIG_TYPE_VALUES),
+      default: OUT_CONFIG_TYPE_VALUES.Dotenv,
+    })(),
```
szdziedzic, line 75: I wonder if this should be the default? Maybe we can make it the default once we polish exec for multiple sessions?

krystofwoldrich: No strong opinion, but I would love it to be the default already. Multiple sessions can still be controlled with --id.

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -68,11 +92,24 @@ export default class SimulatorStart extends EasCommand {
...
 
+    await loadSimulatorEnvAsync(projectDir);
+    const existingDeviceRunSessionId = process.env[EAS_SIMULATOR_SESSION_ID];
+    if (existingDeviceRunSessionId && !flags.force) {
+      throw new Error(
+        `Existing simulator session in environment. Use --force to create a new device session.`
+      );
+    }
+    if (existingDeviceRunSessionId) {
+      Log.warn(`  Overwriting previous simulator session (id: ${existingDeviceRunSessionId}).`);
```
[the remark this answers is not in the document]
szdziedzic: this is also intersting @krystofwoldrich

packages/eas-cli/src/commands/simulator/get.ts
```diff
@@ -42,18 +47,25 @@ export default class SimulatorGet extends EasCommand {
...
       loggedIn: { graphqlClient },
     } = await this.getContextAsync(SimulatorGet, {
       nonInteractive,
     });
 
-    const fetchSpinner = ora(`Fetching device run session ${flags.id}`).start();
+    await loadSimulatorEnvAsync(projectDir);
+    const flagId = flags.id || process.env[EAS_SIMULATOR_SESSION_ID];
+    if (!flagId) {
+      throw new Error('Missing required flag id');
```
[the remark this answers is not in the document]
szdziedzic: this sounds reasonable too @krystofwoldrich

packages/eas-cli/src/commands/simulator/exec.ts
```diff
@@ -0,0 +1,33 @@
...
+    const { projectDir } = await this.getContextAsync(SimulatorExec, {
+      nonInteractive: true,
+    });
+    await loadSimulatorEnvAsync(projectDir);
+
+    const [command, ...args] = this.argv as [string, ...string[]];
+    await spawnAsync(command, args, {
+      stdio: 'inherit',
+      env: process.env,
+    });
```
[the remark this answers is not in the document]
szdziedzic: this looks interesting @krystofwoldrich

packages/eas-cli/src/simulator/env.ts
```diff
@@ -0,0 +1,49 @@
+import { loadEnvFiles, loadProjectEnv } from '@expo/env';
+import * as fs from 'fs-extra';
+import path from 'path';
+
+export const SIMULATOR_DOTENV_FILE_NAME = '.env.eas-simulator';
+export const EAS_SIMULATOR_SESSION_ID = 'EAS_SIMULATOR_SESSION_ID';
+export const SIMULATOR_DOTENV_FILE_HEADER =
+  '# Do not commit this file.\n# Do not modify these values manually. They are managed by eas-cli.\n# It holds configuration only for the current simulator session.\n\n';
```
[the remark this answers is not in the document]
krystofwoldrich: True, users can accidentally commit it, but we can improve the situation later. In template defaults...

szdziedzic (approved):
generally LGTM

PR #3778, opened 2026-05-21
[eas-cli] improve naming for observe commands

kadikraman:
# Why

Part of ENG-21384

The naming of our CLI commands does not match the website.

# How

- Rename `observe:metrics` -> `observe:metrics-summary`
- Rename `observe:events` -> `observe:metrics`
- Rename `observe:logs` -> `observe:events`

# Test Plan

Manual and automated testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- [eas-cli] Rename observe commands: `observe:logs` → `observe:events` (events emitted via `logEvent`), previous `observe:events` → `observe:metrics` (individual performance metric samples), previous `observe:metrics` → `observe:metric-summary` (aggregated stats by app version). ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@kadikraman](https://github.com/kadikraman))
+
 ### 🎉 New features
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -6,14 +6,14 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- [eas-cli] Rename observe commands: `observe:logs` → `observe:events` (events emitted via `logEvent`), previous `observe:events` → `observe:metrics` (individual performance metric samples), previous `observe:metrics` → `observe:metric-summary` (aggregated stats by app version). ([#3778](https://github.com/expo/eas-cli/pull/3778) by [@kadikraman](https://github.com/kadikraman))
```
kadikraman: This _is_ a breaking change, but there isn't really a way around it, and that's why we're in preview, right!
I think adding this here will create a major version will mean a major version bump when publishing - is that ok or should I put it in another section?

packages/eas-cli/src/observe/metricNames.ts
```diff
@@ -23,6 +24,7 @@ export const METRIC_SHORT_NAMES: Record<string, string> = {
   'expo.app_startup.tti': 'Startup TTI',
   'expo.app_startup.ttr': 'Startup TTR',
   'expo.app_startup.bundle_load_time': 'Bundle Load',
+  'expo.updates.download_time': 'Update Download',
```
kadikraman, line 27: Noticed this was missing.

CHANGELOG.md
douglowder: Major bump is fine

douglowder (approved):

PR #3781, opened 2026-05-22
[build-tools] Add max_idle_minutes to agent-device remote session step

szdziedzic:
# Why

The `eas/start_agent_device_remote_session` step keeps the turtle job alive indefinitely (`new Promise<never>(() => {})`) until `stopDeviceRunSession` cancels it. If a client connects, disappears, and never sends `stopDeviceRunSession`, the worker burns until the orchestrator's `max_run_time_seconds` ceiling fires. We want a tighter, activity-aware ceiling so abandoned sessions release their worker quickly.

# How

Adds an optional `max_idle_minutes` build function input. When set, the step:

1. Generates a one-file ESM auth-hook module at `${tmpdir}/agent-device-idle-hook.mjs` that stamps `Date.now()` into `${tmpdir}/agent-device-last-activity` on every authenticated daemon request, then returns `true` so the daemon's own token check remains the source of truth for authorization.
2. Launches the daemon with `AGENT_DEVICE_HTTP_AUTH_HOOK` pointing at that module. This is a documented extension point on agent-device's HTTP server (`src/daemon/http-server.ts` `loadHttpAuthHook`); it fires for `/rpc`, `/upload`, and `/artifacts/*` — every meaningful client call — while skipping `/health`, so cloudflared/health probes don't reset the timer.
3. Races a watchdog (`waitForIdleTimeoutAsync`) against the existing `await new Promise<never>(() => {})`. The watchdog polls the state file every 30s; if `Date.now() - lastActivityAt >= maxIdleMinutes * 60_000`, it throws `SystemError`. That fails the step and cascades to the device run session via the existing turtle-job-status trigger.

When `max_idle_minutes` is absent or non-positive, the step behaves exactly as before (no hook installed, no watchdog, infinite wait).

The state file is pre-seeded with the current time before the daemon starts so the first poll has a baseline; malformed/missing state is treated as fresh activity rather than firing a false-positive timeout.

The complementary www-side change that plumbs this from the GraphQL `createDeviceRunSession` mutation lives in a follow-up PR; until that ships, no caller passes the new input so the step is a no-op for everyone.

# Test Plan

- `yarn typecheck` (build-tools): clean
- `yarn lint` on the changed file: 0 errors
- `yarn fmt:check`: clean
- `yarn jest-unit` (build-tools): 522/522 pass
- Manual verification will follow once the www side is wired up and a real turtle job runs end-to-end.

### DIFF
--- a/packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts
@@ -8,6 +8,7 @@ import {
   BuildStepInputValueTypeName,
 } from '@expo/steps';
 import spawn from '@expo/turtle-spawn';
+import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 
@@ -21,12 +22,17 @@ import {
   waitForFileAsync,
   waitForMatchInOutputAsync,
 } from '../utils/remoteDeviceRunSession';
+import { sleepAsync } from '../../utils/retry';
 
 const AGENT_DEVICE_REPO_URL = 'https://github.com/callstackincubator/agent-device.git';
 const SRC_DIR = '/tmp/agent-device-src';
 const DAEMON_JSON_PATH = path.join(os.homedir(), '.agent-device', 'daemon.json');
 const XCODE_DEVELOPER_DIR = '/Applications/Xcode.app/Contents/Developer';
 const STARTUP_TIMEOUT_MS = 60_000;
+
+const IDLE_HOOK_PATH = path.join(os.tmpdir(), 'agent-device-idle-hook.mjs');
+const IDLE_STATE_PATH = path.join(os.tmpdir(), 'agent-device-last-activity');
+const IDLE_POLL_INTERVAL_MS = 30_000;
 
 export function createStartAgentDeviceRemoteSessionBuildFunction(
   ctx: CustomBuildContext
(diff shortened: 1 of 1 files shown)


szdziedzic:
* **#3781** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts
```diff
@@ -177,3 +213,67 @@ function parseDaemonInfo(raw: string): { port: number; token: string } {
...
+  // request lands; without it, an early poll would compute an unbounded idle
+  // window and false-positive the timeout.
+  await fs.promises.writeFile(IDLE_STATE_PATH, String(Date.now()), 'utf8');
+
+  // The daemon dynamically imports this module
+  // (agent-device src/daemon/http-server.ts loadHttpAuthHook). The default
+  // export must:
+  //   - stamp IDLE_STATE_PATH on every authenticated request,
+  //   - return true so the daemon's own token check stays the source of
+  //     truth for authorization,
```
sjchmiela: i don't see token being validated around https://github.com/callstackincubator/agent-device/blob/5b63b9b427b218f685bb01812a0d04bcbb5514f3/src/daemon/http-server.ts#L569

packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts
```diff
@@ -81,12 +99,21 @@ export function createStartAgentDeviceRemoteSessionBuildFunction(
         logger,
       });
 
+      // When a max-idle window is set, install an auth hook that stamps the
+      // current time into IDLE_STATE_PATH on every authenticated daemon
+      // request. The watchdog below polls that file to detect inactivity.
+      const idleEnv: Record<string, string> = {};
```
sjchmiela, line 102: idleEnv is a bit of a weird name

sjchmiela (approved):

PR #3786, opened 2026-05-22
[build-tools] pass env to repack spawnAsync

Kudo:
# Why

support nvm for repack job

# How

pass context env into repack's spawnAsync

# Test Plan

deploy and test

### DIFF
--- a/packages/build-tools/src/steps/functions/repack.ts
+++ b/packages/build-tools/src/steps/functions/repack.ts
@@ -10,6 +10,7 @@ import {
 } from '@expo/repack-app';
 import {
   BuildFunction,
+  BuildStepEnv,
   BuildStepInput,
   BuildStepInputValueTypeName,
   BuildStepOutput,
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/steps/functions/repack.ts
```diff
@@ -89,8 +90,9 @@ export function createRepackBuildFunction(): BuildFunction {
       }
 
       const repackSpawnAsync = createSpawnAsyncStepAdapter({
-        verbose,
+        env: stepsCtx.global.env,
```
sjchmiela, line 93: we shouldn't use `stepsCtx.global.env`, but `env` that's passed to the step. we already pass it to repack

sjchmiela (approved):
we can _try_ it, but i think we should exchange the `global.env` for the `env` of the step (which includes the global env and also step-specific env) before merging. and ideally repack should adhere to `env` passed to it in the main call, the adapter should receive this `env` in `options`

PR #3790, opened 2026-05-25
[build-tools] handle custom Gymfile err in xcactivitylog

hSATAC:
# Why

Fix Sentry [TURTLE-V2-WORKER-1WT](https://expoio.sentry.io/issues/TURTLE-V2-WORKER-1WT) (ENG-21383) — 48 events of `xclogparser` failing during the `PARSE_XCACTIVITYLOG` phase with stderr:

```
Error: The file "Build" couldn't be opened because there is no such file.
```

`runFastlaneGym` always returns `derivedDataPath = <ios>/build`, but projects with a custom `ios/Gymfile` can set a different `derived_data_path` (or omit it entirely, in which case Xcode uses its default DerivedData location). xclogparser then runs against a directory that has no `.xcactivitylog` files and fails.

# How

Added a pre-flight check in `parseAndReportXcactivitylog` that lists `<derivedDataPath>/Logs/Build/`:

- If no `.xcactivitylog` files are present, log actionable guidance pointing users at the Gymfile lines they need and skip the analysis (no Sentry — this is user config, not an engineering bug).
- If `readdir` fails with anything other than `ENOENT` (e.g. `EACCES`, `EMFILE`), rethrow so it still lands in Sentry under a new `checking_xcactivitylog_existence` phase tag — real infra problems are not masked by the swallow.

Return type changed from `Promise<void>` to `Promise<{ skipped: boolean }>` so the caller in `builders/ios.ts` can call `ctx.markBuildPhaseSkipped()` and reflect the correct phase status in the build summary.

# Test Plan

Unit tests in `xcactivitylog.test.ts` cover the four pre-check branches:

- `Logs/Build` exists but has no `.xcactivitylog` files → skipped + actionable Gymfile log, no Sentry
- `Logs/Build` does not exist (`ENOENT`) → skipped, no Sentry
- `Logs/Build` has only non-`.xcactivitylog` entries (e.g. `LogStoreManifest.plist`) → skipped
- `readdir` fails with `EACCES` → Sentry captured with `checking_xcactivitylog_existence` phase tag, no misleading Gymfile message logged

Existing tests updated to assert the new `{ skipped }` return shape. 41/41 tests pass; typecheck/lint/fmt clean.

### DIFF
--- a/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
+++ b/packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
@@ -16,9 +16,9 @@ const XCLOGPARSER_DOWNLOAD_TIMEOUT_MS = 20_000;
 const XCLOGPARSER_OUTPUT_FILENAME = 'xcactivitylog.json';
 
 /**
- * Never throws — best-effort observability that does not affect build status.
- * Failures route to Sentry via `Sentry.capture` for engineering triage;
- * users see only a generic skip message.
+ * Catches all internal failures (logged + routed to Sentry); returns
+ * `{ skipped: true }` when analysis did not produce a report so callers can
+ * mark the build phase skipped.
  */
 export async function parseAndReportXcactivitylog({
   derivedDataPath,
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -16,9 +16,9 @@ const XCLOGPARSER_DOWNLOAD_TIMEOUT_MS = 20_000;
 const XCLOGPARSER_OUTPUT_FILENAME = 'xcactivitylog.json';
 
 /**
- * Never throws — best-effort observability that does not affect build status.
- * Failures route to Sentry via `Sentry.capture` for engineering triage;
- * users see only a generic skip message.
+ * Catches all internal failures (logged + routed to Sentry); returns
+ * `{ skipped: true }` when analysis did not produce a report so callers can
+ * mark the build phase skipped.
```
sjchmiela, line 21: feels like we could as well throw from this function and mark step as skipped in `catch` above. `{ skipped: boolean }` is good too, clear

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
```diff
@@ -34,10 +34,38 @@ export async function parseAndReportXcactivitylog({
...
+      logger.info(
+        [
+          `Build performance analysis skipped: no .xcactivitylog files found at ${logsBuildDir}.`,
+          '',
+          'This typically happens when your project has a custom ios/Gymfile. To enable',
+          'build performance analysis, add the following lines to your Gymfile:',
+          '',
+          '    derived_data_path("./build")',
+          '    result_bundle(true)',
+          '    result_bundle_path("./build/result-bundle.xcresult")',
```
sjchmiela, line 62: is this suggestion going to work for any project even the ones that are being built on developer's computers?

sjchmiela (approved):

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: Yes and no 🙂

These flags do work on developers’ local machines as well — Xcode needs them enabled in order to generate the xcactivitylog files in the first place.

However, for eas build --local we currently skip the entire parsing/reporting step because it depends on external tooling. So in practice, local users shouldn’t see these messages.

Do you have thoughts on how these messages should be adjusted/improved?

sjchmiela: Oh no I didn't mean `eas build --local`, I was thinking of `expo run:ios`! Like if someone does add it to their non-CNG `Podfile`, are they going to suddenly get… untracked files or something? Like I was thinking if we need to scope these by e.g. `process.env.EAS_BUILD == "1"`, but maybe not.

hSATAC: I think in most cases the `./ios/build/` folder should be included in the .gitignore file so it should be okay. Let's revisit if there's any feedback from users in the future.

packages/build-tools/src/steps/utils/ios/xcactivitylog.ts
hSATAC: That was my initial idea but I went for this approach because I want to make sure the `eas/parse_xcactivitylog` step still preserves the `skip when error, doesn't fail the job` behavior as a standalone step that can be combined in any workflow job.

PR #3792, opened 2026-05-26
Security: Unsafe Buffer Construction from Base64 Credentials

tomaioo:
## Summary

Security: Unsafe Buffer Construction from Base64 Credentials

## Problem

**Severity**: `Medium` | **File**: `packages/build-tools/src/android/credentials.ts:L18`

In `restoreCredentials`, `buildCredentials.keystore.dataBase64` is decoded using `Buffer.from(..., 'base64')` and written to a file. While base64 decoding itself is safe, there's no validation of the decoded data size or format before writing. More critically, the keystore password and key passwords are written in plaintext to `credentials.json`. Additionally, in `injectAndroidCredentialsFunction`, the same pattern is used with Joi validation but the `dataBase64` field is not validated for maximum length, potentially allowing memory exhaustion attacks.

## Solution

Add maximum size validation for base64-decoded data. Use secure temporary file permissions (0600) when writing keystores. Consider using `fs.writeFile` with explicit mode options. For credentials.json, ensure it's written with restrictive permissions and cleaned up promptly after use.

## Changes

- `packages/build-tools/src/android/credentials.ts` (modified)

### DIFF
--- a/packages/build-tools/src/android/credentials.ts
+++ b/packages/build-tools/src/android/credentials.ts
@@ -6,6 +6,8 @@ import { v4 as uuidv4 } from 'uuid';
 
 import { BuildContext } from '../context';
 
+const MAX_KEYSTORE_SIZE = 10 * 1024 * 1024; // 10MB
+
 async function restoreCredentials(ctx: BuildContext<Android.Job>): Promise<void> {
   const { buildCredentials } = nullthrows(
     ctx.job.secrets,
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/android/credentials.ts
```diff
@@ -16,11 +18,12 @@ async function restoreCredentials(ctx: BuildContext<Android.Job>): Promise<void>
     throw new Error('secrets are missing in the job object');
   }
   ctx.logger.info("Writing secrets to the project's directory");
+  const keystoreBuffer = Buffer.from(buildCredentials.keystore.dataBase64, 'base64');
```
AbbanMustafa, line 21: This doesn't seem to prevent the memory use it's trying to guard against. `Buffer.from(dataBase64, 'base64')` will allocate the full buffer in memory before the length check

PR #3794, opened 2026-05-27
[build-tools] Fix iOS Simulator readiness hang on Xcode 26.4

gwdp:
## Why

`eas/start_ios_simulator` never starts on the Xcode 26.4 image — it hangs ~30 min then fails.

`waitForReadyAsync` polls `xcrun simctl io <udid> screenshot /dev/null`. Since Xcode 26.4 `simctl io screenshot` writes the file atomically (temp file + rename in the destination's directory), so `/dev/null` fails — it tries to create a file in `/dev`, which the build user can't:

```
NSCocoaErrorDomain code=513: You don't have permission to save the file "null" in the folder "dev"
```

It fails every attempt, so the readiness retry burns its full 30-minute budget. Worked on Xcode 26.2 (which wrote directly to the path).

## How

Write the readiness screenshot to a real temp file (`/tmp/shot.png`) and remove it after. Works across Xcode versions.

## Test Plan

Reproduced on a custom build with the real `eas/start_ios_simulator`:

- Xcode 26.4 — hangs (cancelled after ~2.5 min): https://expo.dev/accounts/gwdp/projects/coin-flip/workflows/019e6802-de7d-71b7-9bd5-7bf811d6574c
- Xcode 26.2 — succeeds: https://expo.dev/accounts/gwdp/projects/coin-flip/workflows/019e6809-c230-7d68-9b18-6625951038c8
- `screenshot /dev/null` fails on 26.4 but a real path works (diagnostic): https://expo.dev/accounts/gwdp/projects/coin-flip/workflows/019e6802-d14d-763e-bd49-5365208727d1

`yarn typecheck` and `oxlint` pass.

### DIFF
--- a/packages/build-tools/src/utils/IosSimulatorUtils.ts
+++ b/packages/build-tools/src/utils/IosSimulatorUtils.ts
@@ -1,5 +1,5 @@
 import spawn, { SpawnPromise, SpawnResult } from '@expo/turtle-spawn';
-import fs from 'node:fs';
+import fs from 'fs-extra';
 import os from 'node:os';
 import path from 'node:path';
 import { setTimeout } from 'node:timers/promises';
@@ -122,9 +122,10 @@ export namespace IosSimulatorUtils {
     udid: IosSimulatorUuid;
     env: NodeJS.ProcessEnv;
   }): Promise<void> {
+    const readinessScreenshotPath = '/tmp/shot.png';
     await retryAsync(
       async () => {
-        await spawn('xcrun', ['simctl', 'io', udid, 'screenshot', '/dev/null'], {
+        await spawn('xcrun', ['simctl', 'io', udid, 'screenshot', readinessScreenshotPath], {
           env,
         });
       },
@@ -136,6 +137,7 @@ export namespace IosSimulatorUtils {
         },
       }
     );
+    await fs.remove(readinessScreenshotPath);
 
     // Wait for data migration to complete before declaring the simulator ready
     // Based on WebKit's approach: https://trac.webkit.org/changeset/231452/webkit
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/utils/IosSimulatorUtils.ts
```diff
@@ -122,9 +122,10 @@ export namespace IosSimulatorUtils {
     udid: IosSimulatorUuid;
     env: NodeJS.ProcessEnv;
   }): Promise<void> {
+    const readinessScreenshotPath = '/tmp/shot.png';
```
sjchmiela: Usually I think we use os.tmpdir()? Can we use it here too? And some maybe more specific filename?

packages/build-tools/src/utils/IosSimulatorUtils.ts
```diff
@@ -1,5 +1,5 @@
 import spawn, { SpawnPromise, SpawnResult } from '@expo/turtle-spawn';
-import fs from 'node:fs';
+import fs from 'fs-extra';
```
sjchmiela: Can we keep this on node fs? Rm({ force }) should work?

sjchmiela (approved):

packages/build-tools/src/utils/IosSimulatorUtils.ts
gwdp: Thank you for the input! updated

packages/build-tools/src/utils/IosSimulatorUtils.ts
gwdp: Yup, apologies, been using that on updates command side and thought I was supposed to update where not

PR #3795, opened 2026-05-27
[build-tools] disable apsd in iOS simulator

gwdp:
## Why

`apsd` (the push daemon) can't get a push token in the Simulator — a fresh Simulator has no push identity — so it retries forever and floods the system log. That keeps `diagnosticd` busy for the whole life of the Simulator, eating about a CPU core. Simulators can't get real pushes anyway.

Seen on Xcode 16.4, 26.2 and 26.4.

## How

Disable `apsd` as soon as the Simulator boots, before we wait for it to be ready:

```
xcrun simctl spawn <udid> launchctl disable system/com.apple.apsd
xcrun simctl spawn <udid> launchctl bootout system/com.apple.apsd
```

If it fails, we warn and keep going. Same for cloned Simulators.

## Results

Idle Simulator, Xcode 26.4:

| | apsd on | apsd off |
| --- | --- | --- |
| apsd logs/min | ~100k | ~0 |
| diagnosticd CPU | ~30-50% | ~0% |
| Simulator CPU | ~75% | ~0% |

App launch time was the same either way (~210ms).

## Test Plan

- Unit test for `disableApsdAsync`.
- Before/after on a build (Xcode 26.4): https://expo.dev/accounts/gwdp/projects/coin-flip/workflows/019e66ad-11c5-7ae1-a5a8-f4f590b2b8f8
- Watched it stay idle for 3 min with apsd off: https://expo.dev/accounts/gwdp/projects/coin-flip/workflows/019e67e4-9a80-7416-b631-25de268b9b15
- `yarn typecheck`, `oxlint`, tests pass.

## Notes

Still not sure why iOS 26.0 didn't storm. There apsd connected once and went quiet, while 16.4, 26.2 and 26.4 all kept spamming — same missing token, same network. Haven't worked out what's different about 26.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This is the log of notable changes to EAS CLI and related packages.
 ### 🐛 Bug fixes
 
 - [build-tools] Fix `eas/start_ios_simulator` hanging on Xcode 26.4 by writing the readiness screenshot to a temp file instead of `/dev/null`. ([#3794](https://github.com/expo/eas-cli/pull/3794) by [@gwdp](https://github.com/gwdp))
+- [build-tools] Disable `apsd` in the iOS Simulator after boot to stop a push-registration log storm that pegs `diagnosticd` CPU for the lifetime of the Simulator. ([#3795](https://github.com/expo/eas-cli/pull/3795) by [@gwdp](https://github.com/gwdp))
 
 ### 🧹 Chores
 
--- a/packages/build-tools/src/steps/functions/__tests__/startIosSimulator.test.ts
+++ b/packages/build-tools/src/steps/functions/__tests__/startIosSimulator.test.ts
@@ -1,0 +1,76 @@
+import spawn from '@expo/turtle-spawn';
+
+import { createGlobalContextMock } from '../../../__tests__/utils/context';
+import { createMockLogger } from '../../../__tests__/utils/logger';
+import { IosSimulatorUtils } from '../../../utils/IosSimulatorUtils';
+import { createStartIosSimulatorBuildFunction } from '../startIosSimulator';
+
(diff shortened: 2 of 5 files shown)


hSATAC (approved):
LGTM

sjchmiela (changes requested):
are we certain apsd absolutely does not work in simulators or is not needed? see https://exponent-internal.slack.com/archives/C01D24DMP98/p1778175913170849, https://developer.apple.com/forums/thread/795433?answerId&#x3D;881732022#881732022

gwdp:
> are we certain apsd absolutely does not work in simulators or is not needed? see https://exponent-internal.slack.com/archives/C01D24DMP98/p1778175913170849, https://developer.apple.com/forums/thread/795433?answerId&#x3D;881732022#881732022

​Thank you for the review and bubbling some of your the concerns @sjchmiela  :pray:

APNS token registration in the simulator is bound to the host Mac (per the [Xcode 14 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes)), and EAS workers aren't signed into any Apple ID, so apsd can't get a token regardless of the runtime. Measured through "no token found in keychain" ~130/min on iOS `18.x/26.0/26.2/26.4`. Enabling the registration path would mean signing the CI workers into an Apple ID, which we currently don't.

In the other hand,`simctl push` doesn't go through `apsd` (verified — succeeds with the same output whether apsd is running or `launchctl bootout`'d), and that's already how Maestro delivers notifications, so disabling doesn't break anything we use. snorkling session documented at [notion](https://app.notion.com/p/expo/Push-notifications-on-iOS-Simulator-372e5b57352481ebb99ff79786bfff94)

sjchmiela (approved):
AWESOME

PR #3801, opened 2026-05-28
[worker] Use runtime settings from storage

sjchmiela:
# Why

https://github.com/expo/workflow-orchestration/pull/851

# How

Added `RuntimeSettings` namespace in `build-tools`. It is expected to be called in the beginning with `loadAsync` to fetch data from the bucket.

# Test Plan

Once approved deploy to staging, run a few builds with different settings.


packages/worker/src/runtimeEnvironment.ts
```diff
@@ -23,25 +24,28 @@ export async function prepareRuntimeEnvironmentConfigFiles(): Promise<void> {
     return;
   }
 
-  if (config.npmCacheUrl) {
+  const npmCacheUrl = shouldUseCache('npm') ? config.npmCacheUrl : null;
```
AbbanMustafa: nice this is where we'll check for https://github.com/expo/universe/pull/27727

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -0,0 +1,89 @@
...
+    try {
+      const response = await fetch(url, {
+        signal: AbortSignal.timeout(1000),
+      });
+
+      if (!response.ok) {
+        logger.warn(
+          { url, status: response.status },
+          'Failed to fetch worker runtime settings, using defaults'
+        );
```
szdziedzic, line 64: will it be logged to sentry? can we log it to sentry/datadog/somewhere? I think it would be useful to see that there is spike in such err once we broke sth here

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -0,0 +1,89 @@
...
+        return;
+      }
+
+      runtimeSettings = RuntimeSettingsSchema.parse(await response.json());
+      logger.info({ url, settings: runtimeSettings }, 'Loaded worker runtime settings');
+    } catch (err) {
+      logger.warn({ err, url }, 'Failed to load worker runtime settings, using defaults');
```
szdziedzic, line 89: similar question?

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  export function isUsingIosPrecompiledModulesEnabled(): boolean {
+    return runtimeSettings.iosPrecompiledModules ?? false;
+  }
+
+  export function getNpmCacheUrl(): string | null {
+    return runtimeSettings.caches?.[process.platform]?.npm
```
szdziedzic, line 106: this is probably stupid thing to worry about, but shouldn't we just throw an error if `process.platform` is not present in this json file?

like it should never happen, so this way we won't mask our configuration issue, right?

packages/worker/src/env.ts
```diff
@@ -47,8 +53,11 @@ export function getBuildEnv({
 
   const runnerPlatform = job.platform;
   if (runnerPlatform === Platform.IOS) {
-    setEnv(env, 'EAS_BUILD_COCOAPODS_CACHE_URL', config.cocoapodsCacheUrl);
+    setEnv(env, 'EAS_BUILD_COCOAPODS_CACHE_URL', cocoapodsCacheUrl);
     setEnv(env, 'COMPILER_INDEX_STORE_ENABLE', 'NO');
+    if (shouldUsePrecompiledModules(job)) {
```
AbbanMustafa: nit: technically this is always true because we already check its an ios job at line 55 but I get its for rolling back and what not

AbbanMustafa (approved):
I agree with @szdziedzic 's comments to capture the errors when we fail to load settings

packages/worker/src/env.ts
sjchmiela: yea it's for rolling back

packages/build-tools/src/runtimeSettings.ts
sjchmiela: sssure, added a check in `loadAsync`

packages/build-tools/src/runtimeSettings.ts
sjchmiela: added `Sentry.capture`

szdziedzic (approved):

szdziedzic (approved):

PR #3812, opened 2026-05-30
[scripts] Add runtime settings editor

sjchmiela:
I want us to move off of killswitches in Redis coming from `workflow-orchestration` towards fetching killswitches from a simple JSON file.

I picked GCS over R2 because it lets us make objects fully public (no `Authorization` required) and I (Stanley) found no way to achieve the same thing with R2.

This pull request adds a script `scripts/edit-runtime-settings staging | production` that is going to let us edit the file.

### DIFF
--- a/scripts/edit-runtime-settings
+++ b/scripts/edit-runtime-settings
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ $# -ne 1 || ("$1" != "staging" && "$1" != "production") ]]; then
+  echo "Usage: scripts/edit-runtime-settings staging|production" >&2
+  exit 1
+fi
+
+if ! command -v curl >/dev/null 2>&1; then
+  echo "curl is required." >&2
+  exit 1
+fi
+
+for tool in gsutil jq; do
+  if ! command -v "$tool" >/dev/null 2>&1; then
+    echo "$tool is required. Run \`mise install\` from the repository root and try again." >&2
+    exit 1
+  fi
+done
+
+env="$1"
+object_uri="gs://eas-workflows-${env}/runtime-settings.json"
+object_url="https://storage.googleapis.com/eas-workflows-${env}/runtime-settings.json"
+cache_control="no-cache, no-store, must-revalidate"
+editor="${EDITOR:-vi}"
+
+tmpfile="$(mktemp)"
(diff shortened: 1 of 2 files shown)


scripts/edit-runtime-settings
```diff
@@ -0,0 +1,75 @@
...
+    echo "Production upload canceled." >&2
+    exit 1
+  fi
+fi
+
+gsutil \
+  -h "Cache-Control:${cache_control}" \
+  -h "Content-Type:application/json" \
+  cp "$minified_tmpfile" "$object_uri"
+gsutil acl ch -u AllUsers:R "$object_uri"
```
AbbanMustafa: its possible we fail after cp and the object remains private. I think we can just do `gsutil cp -a public-read ...`

scripts/edit-runtime-settings
```diff
@@ -0,0 +1,75 @@
...
+  echo "Usage: scripts/edit-runtime-settings staging|production" >&2
+  exit 1
+fi
+
+if ! command -v curl >/dev/null 2>&1; then
+  echo "curl is required." >&2
+  exit 1
+fi
+
+for tool in gsutil jq; do
```
AbbanMustafa: I see this on the gsutil docs

> Important: gsutil is not the recommended CLI for Cloud Storage. [Use gcloud storage](https://docs.cloud.google.com/storage/docs/discover-object-storage-gcloud) commands in the Google Cloud CLI instead.

scripts/edit-runtime-settings
```diff
@@ -0,0 +1,75 @@
...
+  fi
+fi
+
+gsutil \
+  -h "Cache-Control:${cache_control}" \
+  -h "Content-Type:application/json" \
+  cp "$minified_tmpfile" "$object_uri"
+gsutil acl ch -u AllUsers:R "$object_uri"
+
+curl -fsSL "$object_url" >"$downloaded_tmpfile"
```
AbbanMustafa, line 69: can also do 
  ```suggestion
  gcloud storage cat "$object_uri" >"$downloaded_tmpfile"
  ```

AbbanMustafa (approved):

scripts/edit-runtime-settings
sjchmiela: decided to keep using `curl` cause i want the full "end-to-end" test of what the worker is going to see

PR #3818, opened 2026-06-03
[build-tools] fix worker crash when xcpretty formatter throws on xcodebuild output

hSATAC:
# Why

EAS iOS build workers crash with an uncaught `RangeError: Invalid array length` when `@expo/xcpretty` formats certain compile errors ([TURTLE-V2-WORKER-1D4](https://expoio.sentry.io/issues/TURTLE-V2-WORKER-1D4), 160 events since Feb 2026). The error is thrown inside the `tail` stdout `data` handler, out of reach of the existing try/catch in `startBuildLogger`, so a log-formatting bug takes down the whole worker process and fails the user's build.

Fixes TURTLE-V2-WORKER-1D4

# How

- Bump `@expo/xcpretty` to 4.4.4 (yarn.lock only), which fixes the known crash — `fill()` now guards against negative widths.
- Wrap the formatter call in the `data` handler with a try/catch (in both copies of `XcodeBuildLogger`): fall back to logging the raw chunk, and report the first formatter failure per build to Sentry with the offending chunk, so future formatter bugs surface without crashing builds.

# Test Plan

Added unit tests for `XcodeBuildLogger` covering the formatted happy path, the raw-log fallback when the formatter throws, and Sentry capture deduplication. `yarn test` in `build-tools` passes (69 suites / 558 tests), plus `yarn typecheck` / `yarn lint` / `yarn fmt:check`.


packages/build-tools/src/steps/utils/ios/xcpretty.ts
```diff
@@ -60,9 +63,22 @@ export class XcodeBuildLogger {
...
-          this.logger.info(line);
+        try {
+          const lines = formatter.pipe(data.toString());
+          for (const line of lines) {
+            this.logger.info(line);
+          }
+        } catch (err: any) {
+          // The formatter can throw on unexpected xcodebuild output. Fall back to
+          // raw logs instead of crashing the process with an uncaught exception.
+          this.logger.debug({ err }, 'xcpretty formatter failed, falling back to raw logs');
```
sjchmiela: have you seen this debug being visible anywhere?

sjchmiela (approved):
while you're here may i ask you what's the dfiference between those two xcprettys? would it make sense to unify them?

packages/build-tools/src/steps/utils/ios/xcpretty.ts
hSATAC: You're right, and we already have Sentry here. Let me remove this debug log.

hSATAC:
> while you're here may i ask you what's the dfiference between those two xcprettys? would it make sense to unify them?

I don't really know. The lastest change was [2024](https://github.com/expo/eas-build/commits/90bafbfe63f8d431119ea196ee18fd777cbb1bb4/packages/build-tools/src/ios/xcpretty.ts), this file joined expo before I do 🤣 

My guess is someone decided to copy the whole file back then, let me see if we can unify it a bit.

sjchmiela:
> My guess is someone decided to copy the whole file back then, let me see if we can unify it a bit.

Yes! So the story is that when work towards function-based build process has started (custom builds) it was decided to copy _everything_ in order to _not accidentally break anything in the regular build process_. I think the theoretical plan may have been to move away from the regular build process to function-based build process soon, but it was never completed. I think it's probably ok for the yaml "functions" to be more like entrypoints to existing processes, rather than a completely separate world. Thank you for merging the files!!

PR #3819, opened 2026-06-03
[build-tools] Route Datadog metrics and logs by build vs job-run target

hSATAC:
# Why

The `Datadog` singleton in `build-tools` can only emit to the turtle-build proxy endpoints (`turtle-builds/{buildId}/metrics`, `turtle-builds/logs`). When running a turtle-job-run (Maestro tests, deploys, generic jobs), metrics would be silently lost (posted against a nonexistent turtle-build) and logs would be mislabeled with a `build_id` tag carrying a jobRunId. ENG-21520.

Server-side counterpart (new `turtle-job-runs/*` proxy endpoints): expo/universe#27858.

# How

- `Datadog.setup()` now takes a `target` discriminated union keyed off the execution entity: `{ kind: 'build', turtleBuildId }` | `{ kind: 'jobRun', turtleJobRunId }`.
- `distribution()`/`log()` keep their public signatures and select the endpoint (and the log body id field) from `target.kind`. Sentry failure labels are target-aware; the build-path messages stay byte-identical to preserve existing grouping.
- The worker computes the target at `setup()` from `job.platform` (set only for build jobs); for a jobRun, `this.buildId` is the turtle-job-run id.
- Adjacent cleanup: `flushAsync()` rotates `pendingUploads` so each flush only awaits its own batch.

No new emit call sites: the jobRun metrics path stays dormant until a consumer lands (ENG-21441). The logs path takes effect immediately — a jobRun running `restore_build_cache` now logs with a correct `job_run_id` tag instead of a mislabeled `build_id`.

# Test Plan

- `packages/build-tools` `datadog.test.ts` (13 tests): build/jobRun routing for metrics and logs, target-aware Sentry labels, null-setup no-op, flush rotation.
- `packages/worker` `service.test.ts`: target computation for build vs jobRun jobs, and the missing-token no-op.
- `yarn typecheck` / `yarn lint` / `yarn fmt:check` clean.

### DIFF
--- a/packages/build-tools/src/__tests__/datadog.test.ts
+++ b/packages/build-tools/src/__tests__/datadog.test.ts
@@ -22,13 +22,13 @@ describe('Datadog singleton', () => {
     Datadog.setup(null);
   });
 
-  it('POSTs distribution metrics to the turtle build metrics endpoint', () => {
+  it('POSTs distribution metrics to the turtle build metrics endpoint for build targets', () => {
     turtleFetchMock.mockResolvedValueOnce({} as Response);
 
     Datadog.setup({
       expoApiV2BaseUrl: 'https://api.expo.test/v2/',
-      turtleBuildId: 'build-id',
       robotAccessToken: 'token-abc',
+      target: { kind: 'build', turtleBuildId: 'build-id' },
     });
 
     Datadog.distribution('eas.build.phase_duration', 1234, {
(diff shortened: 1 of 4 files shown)


AbbanMustafa (approved):
Thanks for catching this. I wish we had a single entity and didn't need to set these checks and separate routes. Maybe this can all be cleaned up soon

PR #3820, opened 2026-06-03
[eas-cli] ref(env:*): Validate all required flags at one in non-interactive mode

krystofwoldrich:
# Why

In Expo Agent monitoring (Braintrust) we can see the agent struggles with missing all required flags during envs creation.

From [[link]](https://www.braintrust.dev/app/expo-agent/p/expo-agent/logs) Braintrust's agent: Non-interactive mode confusion (5/15) — the agent keeps hitting "this needs --non-interactive", "this needs --visibility", "this needs --force", "this needs a stdin prompt". A non-interactive cheat-sheet would eliminate most of these.

(Also based on what we have inside of the Expo Agent envs/db [skill](https://github.com/expo/echo-backend/blob/fb37d511c825707244fbb50aed8d5b026f013260/backend/skills/database/SKILL.md))

# How

- add `validateNonInteractiveRequiredInputs` helper which validates all required inputs in the non-interactive mode at once.
- mark all required in non-interactive as required (in interactive mode users are prompted as before)
- add examples with all required flags

# Test Plan

updated existing tests

### DIFF
--- a/packages/eas-cli/src/commands/env/get.ts
+++ b/packages/eas-cli/src/commands/env/get.ts
@@ -9,7 +9,6 @@ import {
   EASNonInteractiveFlag,
   EASVariableFormatFlag,
   EasEnvironmentFlagParameters,
-  extendFlagDescription,
   validateNonInteractiveRequiredInputs,
 } from '../../commandUtils/flags';
 import {
@@ -39,13 +38,11 @@ interface GetFlags {
 }
 
 export default class EnvGet extends EasCommand {
-  static override description = `view an environment variable for the current project or account
-
-In non-interactive mode, provide --variable-name and an environment with ENVIRONMENT or --variable-environment.`;
+  static override description = 'view an environment variable for the current project or account';
 
   static override examples = [
-    '$ eas env:get production --variable-name API_URL --non-interactive',
-    '$ eas env:get --variable-environment production --variable-name API_TOKEN --format long --non-interactive',
+    '$ eas env:get --variable-environment production --variable-name API_URL',
+    '$ eas env:get --variable-environment production --variable-name API_TOKEN --format long',
   ];
 
   static override contextDefinition = {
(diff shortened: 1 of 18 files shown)


packages/eas-cli/src/commands/env/get.ts
```diff
@@ -54,11 +60,11 @@ export default class EnvGet extends EasCommand {
 
   static override flags = {
     'variable-name': Flags.string({
-      description: 'Name of the variable',
+      description: '(required) Name of the variable',
```
gwdp, line 63: Do we want to flag this as required, regardless of whether it's interactive or not? Feels it could be misleading for human users?

packages/eas-cli/src/commands/env/push.ts
```diff
@@ -132,6 +146,12 @@ export default class EnvPush extends EasCommand {
                 )} environment variables already exist in ${displayedEnvironment} environment. Do you want to override them all?`
               : `The ${variableNames[0]} environment variable already exists in ${displayedEnvironment} environment. Do you want to override it?`;
 
+          if (nonInteractive) {
```
gwdp, line 149: Unsure if missing any detail, but noting that the failure mode for this command could be inconsistent with `create/delete/get` set since this one goes through getContextAsync/auth. Could this lead to model confusion?

packages/eas-cli/src/commands/env/update.ts
```diff
@@ -212,15 +221,18 @@ export default class EnvUpdate extends EasCommand {
...
+      nonInteractive: flags['non-interactive'],
+      requiredInputs: [
+        { name: '--variable-name', value: flags['variable-name'] },
+        { if: !!flags.type, name: '--value', value: flags.value },
+      ],
+      helpCommand: 'eas env:update --help',
+    });
+    if (environment && flags['variable-environment']) {
+      throw new Error(
+        "You can't use both --variable-environment flag when environment is passed as an argument. Run `eas env:update --help` for more information."
```
gwdp, line 234:
```suggestion
        "You can't use the --variable-environment flag when the environment is passed as a positional argument. Run `eas env:update --help` for more information."
```

gwdp (approved):
LGTM! Left a few non-blocking comments

PR #3822, opened 2026-06-05
[build-tools] track installed maestro cli version

hSATAC:
# Why

[ENG-21441](https://linear.app/expo/issue/ENG-21441/track-maestro-cli-version-in-datadog): track which Maestro CLI versions are used on EAS, to inform migrations to new Maestro features and removal of deprecated ones.

# How

- `eas/install_maestro` now reports the detected version after version detection on EAS Build VMs: `Datadog.distribution('eas.maestro.install', 1, { maestro_version })`. Every EAS-generated Maestro path runs this step (custom builds via the `eas/maestro_test` function group, workflow `maestro` / `maestro_cloud` jobs), so one call site covers them all.
- The `Datadog` module auto-tags every metric/log with `target: build | job_run`, derived from the target it already routes by — so this metric (and existing ones like `eas.build.phase_duration`) can be split by custom build vs workflow job.
- Fire-and-forget: nothing is written to the user's build log; upload failures go to Sentry via the existing `.catch`.

Companion PR expo/universe#27921 allowlists the new metric name on the turtle-builds metrics endpoint and **must be deployed before this PR is released** (the turtle-job-runs endpoint is free-form and needs no change).

# Test Plan

- New `installMaestro.test.ts` unit test: emits with the detected version on the EAS Build VM path; does not emit on the local-runner path.
- `datadog.test.ts` updated for the automatic `target` tag on all metric/log payloads.
- `yarn test` (build-tools unit suite): 559/559 passing.
- After both PRs are live: verify `expo.www.eas.maestro.install` in Datadog, grouped by `maestro_version` and `target`.

### DIFF
--- a/packages/build-tools/src/__tests__/datadog.test.ts
+++ b/packages/build-tools/src/__tests__/datadog.test.ts
@@ -51,6 +51,7 @@ describe('Datadog singleton', () => {
                 build_phase: 'install_dependencies',
                 platform: 'ios',
                 result: 'success',
+                target: 'build',
               },
             },
           ],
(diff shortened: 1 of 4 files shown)


gwdp (approved):
LGTM :fire: Cool addition to the logs as well!

PR #3824, opened 2026-06-05
[expo-cocoapods-proxy] Bump bundler to fix compatibility with RubyGems 4

gwdp:
# Why

iOS worker tarball builds started failing on the new macOS Tahoe image. Failing run: https://expo.dev/accounts/expo/projects/eas-worker/workflows/019e99b9-9621-72be-bc71-086f6b5fe87c

The new Tahoe image ships `ruby@3.2` with RubyGems 4.x, which removed `Gem::Platform.match`. Bundler 2.1.4 calls it unconditionally:

```
NoMethodError: undefined method `match' for Gem::Platform:Class
  bundler-2.1.4/lib/bundler/index.rb:198
```

The previous Sequoia image had an older `ruby@3.2` where `match` still existed (deprecated), so it passed with warnings.

# How

- Bumped bundler dev dependency from `~> 1.17` to `>= 2.4` in the gemspec — bundler 2.2+ uses `match_gem?` instead
- Regenerated `Gemfile.lock` (`BUNDLED WITH` 2.1.4 → 4.0.13)
- Added `.github/workflows/trigger-worker-build.yml` — builds the worker tarball on PRs touching `packages/worker` or `packages/expo-cocoapods-proxy`, and available as a manual dispatch

Note: `bundle install` only runs at tarball build time on `image: latest`, not on customer machines, so this is safe across all worker images.

# Test Plan

- Reproduced locally — `bundle install` under modern Ruby/RubyGems errors before the fix, passes after
- This PR should trigger the new worker build CI check


.github/workflows/trigger-worker-build.yml
sjchmiela, line 1: Let's not commit this?

(I think you can dispatch build-and-deploy-worker manually for tests too?)

sjchmiela (approved):

hSATAC:
I think [my release](https://github.com/expo/eas-cli/actions/runs/27121520298/job/80039841434) failed due to the same issue.

I'll wait for this to be merged.

.github/workflows/trigger-worker-build.yml
gwdp: Cleaned up, thanks for catching that! Will double-check later and propose a test workflow if any gaps

PR #3828, opened 2026-06-08
[build-tools] fetch Cloudflare TURN ICE servers from www and pass to serve-sim

szdziedzic:
# Why

Part of adding a Cloudflare TURN relay so the remote-simulator WebRTC stream works on networks that block UDP or use symmetric NAT (corporate firewalls) — where STUN-only leaves the **video black even though touch/clicks still work**. The companion change in universe (expo/universe#27973) exposes a worker-facing route that mints short-lived TURN credentials on demand; this PR fetches them and feeds them to serve-sim.

# How

When launching serve-sim, the worker fetches short-lived TURN ICE servers from www's `POST /v2/device-run-sessions/:deviceRunSessionId/turn-ice-servers` route via `turtleFetch` — the same robot-token pattern it already uses to fetch the project clone URL (`download-project-archive`) — keyed on the `DEVICE_RUN_SESSION_ID` it already has, and turns them into serve-sim CLI flags:

- `--stun-url` from the credential-less (STUN) entries, and `--turn-url` / `--turn-username` / `--turn-credential` from the entry that carries the short-lived credentials.
- Wired into the shared `startServeSimWithTunnelAsync`, so it covers all three serve-sim launchers (serve-sim, agent-device, and argent web preview).
- **Best-effort:** any failure (mint error, network, unexpected shape) → no flags; serve-sim falls back to its built-in P2P/STUN behavior, so it never blocks a session.
- The credential is passed as a process arg and deliberately **not logged** (turtle-spawn never logs argv, and the worker is single-tenant).

No GraphQL/schema change — this rides the existing `turtle-job-runs` REST router.

> Requires a serve-sim release that accepts a comma-separated `--turn-url` (the worker runs `serve-sim-szdziedzic@latest`), so publish serve-sim first.

# Test Plan

- New unit tests (`remoteDeviceRunSession.test.ts`):
  - `turnIceServersToServeSimArgs` — a valid Cloudflare ICE-servers payload produces the expected `--stun-url` / `--turn-url` / `--turn-username` / `--turn-credential` flags; TURN-only and STUN-only payloads emit the right subset; an empty list produces no flags.
  - `fetchServeSimTurnArgsAsync` — requests the job-run route and returns args on success; returns `[]` and warns on failure (P2P/STUN fallback).
  - **6/6 passing** (`yarn jest-unit remoteDeviceRunSession`).
- `tsc` and `oxlint` clean for the changed files.

### DIFF
--- a/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
+++ b/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
@@ -4,11 +4,15 @@ import { BuildStepEnv } from '@expo/steps';
 import spawn from '@expo/turtle-spawn';
 import * as ngrok from '@ngrok/ngrok';
 import { graphql } from 'gql.tada';
+import nullthrows from 'nullthrows';
+import { z } from 'zod';
 import { randomBytes } from 'node:crypto';
 import fs from 'node:fs';
 
 import { CustomBuildContext } from '../../customBuildContext';
+import { Sentry } from '../../sentry';
 import { sleepAsync } from '../../utils/retry';
+import { turtleFetch } from '../../utils/turtleFetch';
 
 const START_DEVICE_RUN_SESSION_MUTATION = graphql(`
   mutation StartDeviceRunSession($deviceRunSessionId: ID!, $remoteConfig: JSONObject!) {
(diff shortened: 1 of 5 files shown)


szdziedzic:
* **#3828** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3828?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3828?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -60,6 +63,109 @@ export function getNgrokAuthtokenOrThrow(env: BuildStepEnv): string {
...
+ * worker is single-tenant).
+ */
+export async function fetchServeSimTurnArgsAsync({
+  ctx,
+  env,
+  logger,
+}: {
+  ctx: CustomBuildContext;
+  env: BuildStepEnv;
+  logger: bunyan;
```
sjchmiela: doesn't `ctx` already have `env` and `logger`?

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -60,6 +63,109 @@ export function getNgrokAuthtokenOrThrow(env: BuildStepEnv): string {
...
+      new URL(
+        `/v2/device-run-sessions/${deviceRunSessionId}/turn-ice-servers`,
+        expoApiServerUrl
+      ).toString(),
+      'POST',
+      {
+        headers: {
+          Authorization: `Bearer ${robotAccessToken}`,
+        },
+        timeout: 10000,
```
sjchmiela: that's a long time to wait

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -60,6 +63,109 @@ export function getNgrokAuthtokenOrThrow(env: BuildStepEnv): string {
...
+    if (args.length > 0) {
+      logger.info('Configured serve-sim with Cloudflare TURN ICE servers.');
+    }
+    return args;
+  } catch (err) {
+    logger.warn(
+      `Could not fetch Cloudflare TURN ICE servers; serve-sim will fall back to P2P/STUN.${
+        err instanceof Error ? ` ${err.message}` : ''
+      }`
+    );
```
sjchmiela, line 161: let's log to sentry? also, bunyan expects `{ err }` as the first argument

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -122,16 +228,19 @@ export function spawnDetached({
 
 export async function startServeSimWithTunnelAsync({
   baseDomain,
+  ctx,
   env,
   logger,
   timeoutMs,
 }: {
   baseDomain: string;
+  ctx: CustomBuildContext;
```
sjchmiela: as much as i dont' like contexts, i think if they _are_ passed into a function i think the convention from www for them to be a separate first argument makes the most sense. however see my other comment about `env` and `logger` being already inside `ctx`.

sjchmiela: maybe we don't need to pass in `ctx` at all

sjchmiela (approved):
in general looks good, but comments are worth adhering to

PR #3834, opened 2026-06-09
[build-tools] fix getMaestroVersion parsing error

hSATAC:
# Why

`getMaestroVersion` returned the entire stdout of `maestro --version`, which on some Maestro versions prints an analytics notice before the version number. As a result the `eas.maestro.install` metric recorded polluted `maestro_version` tag values (the whole notice instead of e.g. `2.0.10`); the same value also fed the step output.

ENG-21746 (follow-up to ENG-21682).

# How

Parse the version out of the output instead of using the whole stdout: match all `X.Y.Z` tokens and take the last one, since the real version is printed after the notice (robust even if the notice itself contains a version-like string). Best-effort — falls back to the raw output if none is found, so it never fails the build over a version string.

# Test Plan

Added unit tests in `installMaestro.test.ts`:

- analytics notice + version → reports `2.0.10`
- notice containing an earlier version-like string → still reports `2.0.10`

`yarn test installMaestro` → 4 passing.

### DIFF
--- a/packages/build-tools/src/steps/functions/installMaestro.ts
+++ b/packages/build-tools/src/steps/functions/installMaestro.ts
@@ -134,8 +134,15 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
 }
 
 async function getMaestroVersion({ env }: { env: BuildStepEnv }): Promise<string> {
-  const maestroVersion = await spawn('maestro', ['--version'], { stdio: 'pipe', env });
-  return maestroVersion.stdout.trim();
+  const { stdout } = await spawn('maestro', ['--version'], { stdio: 'pipe', env });
+  // `maestro --version` can print an analytics notice to stdout before the version,
+  // e.g. "Anonymous analytics enabled. To opt out, set MAESTRO_CLI_NO_ANALYTICS...\n2.0.10".
+  // Take the last version-looking token: the real version is printed after the notice, so
+  // this stays correct even if the notice itself contains a version-like string. Keeps the
+  // step output and the eas.maestro.install metric tag clean. Best-effort only: fall back to
+  // the raw output if none is found, so we never fail the build over a version string.
+  const versions = stdout.match(/\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?/g);
+  return versions?.at(-1) ?? stdout.trim();
 }
 
 async function installMaestro({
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -134,8 +134,15 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
 }
 
 async function getMaestroVersion({ env }: { env: BuildStepEnv }): Promise<string> {
-  const maestroVersion = await spawn('maestro', ['--version'], { stdio: 'pipe', env });
-  return maestroVersion.stdout.trim();
+  const { stdout } = await spawn('maestro', ['--version'], { stdio: 'pipe', env });
+  // `maestro --version` can print an analytics notice to stdout before the version,
+  // e.g. "Anonymous analytics enabled. To opt out, set MAESTRO_CLI_NO_ANALYTICS...\n2.0.10".
```
sjchmiela, line 139: weird? https://github.com/expo/eas-cli/blob/36235bf8cf8ba4421306befcca7e4cddbad38229/packages/worker/src/env.ts#L33

packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -134,8 +134,15 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
...
+  // Take the last version-looking token: the real version is printed after the notice, so
+  // this stays correct even if the notice itself contains a version-like string. Keeps the
+  // step output and the eas.maestro.install metric tag clean. Best-effort only: fall back to
+  // the raw output if none is found, so we never fail the build over a version string.
+  const versions = stdout.match(/\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?/g);
```
sjchmiela, line 144: maybe we can just take the last non-empty line as output?

sjchmiela (approved):
i think last line is probably going to give us good results too and is much simpler, but leaving you the decision what to do

also maybe we want to move this to datadog logs (@gwdp suggested that)?

packages/build-tools/src/steps/functions/installMaestro.ts
hSATAC: Yeah you can see from the datadog metrics, most of the parsed versions are valid. Only a few of them coming with these warnings. I think there must be some path bypassing/overriding these envs but I have no clue now.

packages/build-tools/src/steps/functions/installMaestro.ts
hSATAC: I feel more confident about the regex and this is simple enough, I'll keep it this way.

PR #3835, opened 2026-06-10
[eas-cli] Add PostHog GraphQL layer and command utils

gwdp:
## Why

Plumbing for the `posthog` commands — GraphQL layer + shared utils.

## How

- `PostHogQuery` + `PostHogMutation` + fragments/types
- `commandUtils/posthog.ts` (dashboard URL, formatter)
- `generated.ts` from the universe schema

## Test Plan

- CI
- Live integration needs the universe PostHog backend deployed

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -2129,6 +2129,18 @@
             "deprecationReason": null
           },
           {
+            "name": "posthogOrganizationConnection",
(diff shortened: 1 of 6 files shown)


gwdp:
Folding the GraphQL layer into #3836. We don't ship GraphQL-only PRs (it's dead code until `connect` uses it), so the queries/mutations/types now live with the command. #3836's base is retargeted to `main` and its diff includes this plumbing.

PR #3847, opened 2026-06-11
[build-tools] Log user-provided custom build functions

sjchmiela:
I want to learn how much are custom functions being used.


hSATAC (approved):
Nice

PR #3848, opened 2026-06-11
[worker] Remove legacy upload code, add retries, auto-waive builds

sjchmiela:
## What changed

- Removed `gcsSignedUploadUrlForApplicationArchive` and `gcsSignedUploadUrlForBuildArtifacts` from worker runtime config and runtime config typing.
- Removed launcher signed-upload fallback behavior for managed application archive and build artifacts uploads.
- Classified managed artifact upload failures as `SystemError`s with internal tracking codes while keeping the public error code as `SERVER_ERROR`.
- Added retry options to worker upload API `turtleFetch` calls for upload-session creation and artifact metadata saving.

## Why

These signed upload URL fields are no longer expected, so keeping fallback paths made the worker handle an obsolete runtime contract. Upload failures in these managed paths are infrastructure/API/storage failures rather than user-actionable project issues.

## Validation

- `yarn oxfmt packages/worker/src/config.ts packages/worker/src/upload.ts packages/worker/src/external/turtle.ts packages/worker/src/__unit__/upload.test.ts`
- `yarn workspace @expo/worker test:unit src/__unit__/upload.test.ts --runInBand`
- `yarn workspace @expo/worker typecheck`

### DIFF
--- a/packages/worker/src/upload.ts
+++ b/packages/worker/src/upload.ts
@@ -27,6 +27,22 @@ class ErrorWithMetadata extends Error {
   }
 }
 
+function sanitizeUploadErrorForLogs(err: unknown): Error {
+  if (!(err instanceof Error)) {
+    return new Error(String(err));
+  }
+
+  const safeMessage = err.message
+    .replace(/^(Unexpected response from server \(\d+\)): [\S\s]*$/u, '$1: <redacted>')
+    .replace(/^(Failed to upload file: status: .*), body: [\S\s]*$/u, '$1, body: <redacted>');
+  const safeError = new Error(safeMessage);
+  safeError.name = err.name;
+  safeError.stack = err.stack
+    ? `${safeError.name}: ${safeMessage}\n${err.stack.split('\n').slice(1).join('\n')}`
+    : undefined;
+  return safeError;
+}
+
 export async function uploadApplicationArchiveAsync(
   ctx: BuildContext,
   {
@@ -69,7 +85,7 @@ export async function uploadApplicationArchiveAsync(
   } catch (err: any) {
     // Otherwise, we log the error and proceed to upload to Launcher's upload URL.
     const msg = 'Upload to upload session failed';
-    logger.error({ err, filename, size }, msg);
+    logger.error({ err: sanitizeUploadErrorForLogs(err), filename, size }, msg);
 
     throw new errors.SystemError('Failed to upload application archive.', {
       trackingCode: 'EAS_BUILD_UPLOAD_APPLICATION_ARCHIVE_FAILED',
@@ -125,7 +141,7 @@ export async function uploadBuildArtifactsAsync(
     return { filename: null };
   } catch (err: any) {
     const msg = 'Upload to upload session failed';
-    logger.error({ err, filename, size }, msg);
+    logger.error({ err: sanitizeUploadErrorForLogs(err), filename, size }, msg);
 
     throw new errors.SystemError('Failed to upload build artifacts.', {
       trackingCode: 'EAS_BUILD_UPLOAD_BUILD_ARTIFACTS_FAILED',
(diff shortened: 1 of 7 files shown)


packages/worker/src/upload.ts
```diff
@@ -308,6 +272,9 @@ async function createUploadSessionAsync(
             Authorization: `Bearer ${robotAccessToken}`,
           },
           shouldThrowOnNotOk: false,
+          retries: UPLOAD_API_REQUEST_RETRIES,
```
AbbanMustafa, line 273: https://github.com/expo/eas-cli/blob/main/packages/worker/src/utils/turtleFetch.ts#L53
so before we explicitly did not retry on POST, are we sure we can now without some form of ensuring idempotency?

sjchmiela: we did not retry on POST _by default).

i think so:
- creating upload sessions
  - in builds appends to the array (limit of 5 but i don't think we're going to hit that)
  - in job runs creates workflow artifacts (unique index on name, but if the request failed then i'd rather fail trying again than no trying again?)
- saving artifacts ensures a given value is set for a given key so it's idempotent

next time can you please try to answer this question yourself and give a conclusive review to unblock?

hSATAC (approved):
Took a deeper look since the retry thread touches the server side too. This looks good to me.

One suggestion and one question:

### Suggestion: `SystemError.metadata` may leak into user-visible build logs

[[findAndUploadBuildArtifacts.ts#L29](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L29)](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L29), [[#L36](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L36)](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/findAndUploadBuildArtifacts.ts#L36), and [[uploadArtifact.ts#L131](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/uploadArtifact.ts#L131)](https://github.com/expo/eas-cli/blob/edf7eb7b0577b86b138ebe570bc56db27a847740/packages/build-tools/src/steps/functions/uploadArtifact.ts#L131) use `logger.error('...', err)`, which inspect-prints enumerable properties and now includes `metadata` (presigned upload URLs and response payloads from the cause chain).

The practical risk seems low since the URLs are PUT-only and short-lived, but it does add credential-shaped noise to user-visible logs. Using `logger.error({ err }, '...')` would route through Bunyan's serializers and avoid logging the metadata.

### Question: does "auto-waive builds" need a www-side companion?

I may be missing something here, but from my reading of the code, the current `SERVER_ERROR` auto-waive logic appears to exist only for job runs ([[TurtleJobRunsRouter.ts#L183-L186](https://github.com/expo/universe/blob/81a0f3a38f2498c56b4360ca9d6ba3fe944dcb08/server/www/src/apiv2/TurtleJobRunsRouter.ts#L183-L186)](https://github.com/expo/universe/blob/81a0f3a38f2498c56b4360ca9d6ba3fe944dcb08/server/www/src/apiv2/TurtleJobRunsRouter.ts#L183-L186)). The orchestrator intentionally never sends `waivedAt` for builds ([[update_www.go#L54](https://github.com/expo/workflow-orchestration/blob/9241d2cef9a8a11b29436253c23605e995a43718/internal/utils/update_www.go#L54)](https://github.com/expo/workflow-orchestration/blob/9241d2cef9a8a11b29436253c23605e995a43718/internal/utils/update_www.go#L54)), and I couldn't find equivalent handling on the build-update path.

If I'm understanding that correctly, builds failing these uploads would still be recorded as `SERVER_ERROR` and remain billable. Is there another piece I'm missing, or should the scope here be clarified to job runs only?

sjchmiela:
Fixed `, err` vs `{ err }` when logging errors, great catch!

Adding waiving system error builds in https://github.com/expo/universe/pull/28145.

hSATAC (approved):

PR #3855, opened 2026-06-12
[worker] Resolve runtime cache settings from job env and fallback on npm cache failures

AbbanMustafa:
# Why

We want to enable npm cache proxy usage per build job, but the worker was loading runtime settings before the job environment was available. That meant job-scoped environment variables  like `EAS_USE_NPM_CACHE=1` https://github.com/expo/universe/pull/27727 could not control cache URL resolution.

We also need cache proxy rollout to be safe: if the proxy is unavailable or misconfigured, builds should fall back to the normal npm registry behavior instead of failing during the post-prebuild install.
  
# How

  - Move worker runtime settings loading into `createBuildContext`, where the job env is available.
  - Resolve runtime cache URLs from job env plus runner cache URL fallbacks.
  - Set `NPM_CONFIG_REGISTRY` only when the job has enabled npm cache usage.
  - Add post-prebuild npm install fallback:
    - enabled only when `EAS_USE_NPM_CACHE=1`
    - detects install failures caused by the configured npm cache registry
    - reports warning-level Sentry events
    - retries the same npm install after removing only `NPM_CONFIG_REGISTRY`, so npm falls back to project/user/default registry settings
  - Add Sentry reporting for non-fatal npm cache registry errors that npm logs while still exiting `0`.

# Test Plan

  - Added unit coverage for job-env runtime cache settings resolution.
  - Added unit coverage for post-prebuild npm cache fallback behavior:
    - retries without the cache registry when the cache install fails
    - reports non-fatal cache registry errors to Sentry when npm logs an error but exits `0`
    - does not retry when `EAS_USE_NPM_CACHE` is not enabled
  - Validated the failure path with a local build by setting `EAS_USE_NPM_CACHE=1` from www. Cache private network could not resolve/connect to the internal cache host, which reproduced the proxy failure path and confirmed the build retried through the normal npm registry behavior, and logged to sentry

https://expoio.sentry.io/issues/7511993527/?environment=development&project=1837720&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d

Screenshot 2026-06-11 at 9 34 16 PM

NOOP when env var is not set

Screenshot 2026-06-11 at 11 24 49 PM


packages/build-tools/src/common/installDependencies.ts
```diff
@@ -69,11 +72,81 @@ export async function installDependenciesAsync({
...
+}: {
+  packageManager: PackageManager;
+  env: Record<string, string | undefined>;
+  cwd: string;
+  logger: Exclude<SpawnOptions['logger'], undefined>;
+  infoCallbackFn?: SpawnOptions['infoCallbackFn'];
+  useFrozenLockfile: boolean;
+}): Promise<void> {
+  const npmCacheUrl = env.NPM_CONFIG_REGISTRY;
+  const npmCacheErrorTracker = createNpmCacheRegistryErrorTracker({ env, npmCacheUrl });
```
AbbanMustafa: this captures cache failures that npm only prints to stdout/stderr, like audit endpoint ENOTFOUND, so we still log them to Sentry when npm exits 0 and skips the catch/fallback path

packages/worker/src/env.ts
```diff
@@ -42,6 +42,7 @@ export function getBuildEnv({
   const cocoapodsCacheUrl = RuntimeSettings.getCocoapodsCacheUrl();
 
   setEnv(env, 'NPM_CACHE_URL', npmCacheUrl);
+  setEnv(env, 'NPM_CONFIG_REGISTRY', npmCacheUrl);
```
sjchmiela, line 45: gosh it took me so long to find docs proving this should work lol

does this replace `npm config set registry`?

packages/worker/src/context.ts
```diff
@@ -32,9 +33,21 @@ export function createBuildContext<TJob extends Job>({
   projectId: string;
   buildId: string;
   buildLogger: bunyan;
-}): BuildContext<TJob> {
-  const env = getBuildEnv({ job, projectId, metadata, buildId });
+}): Promise<BuildContext<TJob>> {
   const childLogger = buildLogger.child({ buildId });
+  await RuntimeSettings.loadAsync({
+    environment: config.env,
+    logger: childLogger,
```
sjchmiela: so actually about this logger -- i don't think we should use the build-facing logger. i don't think we need users to see "Failed to fetch worker runtime settings, using defaults", especially if it's in an "(unnamed build phase)"

i kept this logger from the start so "if we ever were to inspect logs of the service" we would have a record of this failure. maybe it was a bad idea and we should just rely only on sentry and env vars printing as distinction?

packages/worker/src/context.ts
```diff
@@ -32,9 +33,21 @@ export function createBuildContext<TJob extends Job>({
...
+    env: job.builderEnvironment?.env,
+    cacheUrlFallbacks: {
+      npm: config.npmCacheUrl,
+      nodejs: config.nodeJsCacheUrl,
+      maven: config.mavenCacheUrl,
+      cocoapods: config.cocoapodsCacheUrl,
+    },
```
sjchmiela: how about we land https://github.com/expo/eas-cli/pull/3856 first?

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -103,26 +109,50 @@ export namespace RuntimeSettings {
...
+    return getCacheUrl(
+      'cocoapods',
+      'EAS_COCOAPODS_CACHE_URL',
+      'EAS_USE_COCOAPODS_CACHE',
+      cacheUrlFallbacks.cocoapods
+    );
+  }
+}
+
+function getCacheUrl(
```
sjchmiela: let's not do this, i know codex wanted to do it ssoooooo bad. we can inline variables and checks in these four places in one file

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -33,20 +33,26 @@ type CacheUrlFallbacks = {
   cocoapods?: string | null;
 };
 
+type RuntimeEnvironment = Record<string, string | undefined>;
```
sjchmiela: we already have `Env` and `BuildEnv` types that should be able to denote this

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -87,3 +160,119 @@ export function resolvePackagerDir(ctx: BuildContext<Job>): string {
...
+    }: {
+      packageManager: PackageManager;
+      cwd: string;
+      useFrozenLockfile: boolean;
+    }): void {
+      if (!firstErrorLine) {
+        return;
+      }
+      const sentryError = new Error('Non-fatal npm cache registry error during dependency install');
+      sentryError.name = 'NpmCacheRegistryNonFatalError';
```
sjchmiela: wouldn't it be better if we made it a subclass of error?

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -69,11 +72,81 @@ export async function installDependenciesAsync({
...
+        npmCacheUrl,
+        useFrozenLockfile,
+        originalErrorMessage: getErrorMessage(err),
+        status: getErrorField(err, 'status'),
+        signal: getErrorField(err, 'signal'),
+      },
+    });
+
+    const fallbackEnv = { ...env };
+    delete fallbackEnv.NPM_CONFIG_REGISTRY;
```
sjchmiela:
```suggestion
    const { NPM_CONFIG_REGISTRY: _NPM_CONFIG_REGISTRY, ...fallbackEnv } = env;
```
? or maybe not. i think there were days when `delete` had bad rep

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -87,3 +160,119 @@ export function resolvePackagerDir(ctx: BuildContext<Job>): string {
...
+    .join('\n');
+}
+
+function getErrorField(err: unknown, field: 'status' | 'signal'): unknown {
+  return err && typeof err === 'object' ? (err as Record<string, unknown>)[field] : undefined;
+}
+
+function createNpmCacheRegistryInstallError(err: unknown): Error {
+  const error = new Error('Failed to install dependencies using npm cache registry');
+  error.name = 'NpmCacheRegistryInstallError';
```
sjchmiela: subclass

also use `SystemError` / `UserError` when possible

and i think you should be able to pass `err` as `cause`? i wish we would remove `getErrorField` and `getErrorMessage`

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -87,3 +160,119 @@ export function resolvePackagerDir(ctx: BuildContext<Job>): string {
...
+
+function getErrorOutput(err: unknown): string {
+  if (!err || typeof err !== 'object') {
+    return '';
+  }
+  const { stdout, stderr } = err as { stdout?: unknown; stderr?: unknown };
+  return [stdout, stderr]
+    .filter((output): output is string => typeof output === 'string')
+    .join('\n');
+}
```
sjchmiela, line 210: i think we can check if instanceof spawnpromise?

sjchmiela (changes requested):
needs some polish

packages/worker/src/env.ts
AbbanMustafa: Yes! they mention it here  https://docs.npmjs.com/cli/v11/using-npm/config#environment-variables

> Any environment variables that start with npm_config_ will be interpreted as a configuration parameter. For example, putting npm_config_foo=bar in your environment will set the foo configuration parameter to bar.

example
```
> echo "registry=https://from-npmrc.example" > /tmp/.npmrc
> npm config get registry --userconfig=/tmp/.npmrc
https://from-npmrc.example
> export NPM_CONFIG_REGISTRY=https://from-env.example
> npm config get registry --userconfig=/tmp/.npmrc
https://from-env.example
```

sjchmiela: should we remove call to `npm config set registry` then?

packages/build-tools/src/runtimeSettings.ts
```diff
@@ -92,26 +88,34 @@ export namespace RuntimeSettings {
   }
 
   export function getNpmCacheUrl(): string | null {
-    return runtimeSettings.caches?.[process.platform]?.npm
-      ? process.env.EAS_NPM_CACHE_URL || null
-      : null;
+    const envOverride = runtimeEnvironment['EAS_USE_NPM_CACHE'];
+    const enabled =
+      envOverride === '1' ||
+      (envOverride !== '0' && runtimeSettings.caches?.[process.platform]?.npm);
```
sjchmiela: i think if runtime settings is false we should NOT enable a given cache. runtime settings being false is for cache maintenance when we know it's going to not be available so even if a user has `EAS_USE_NPM_CACHE` we should not enable the cache for them

AbbanMustafa: thank you yes, adding back to allow for killswitch

AbbanMustafa:
Validated with updated changes that unset EAS_USE_NPM_CACHE results in a no op

Screenshot 2026-06-12 at 2 07 04 PM

and with env var set, install through cache is attempted and failure is reported to sentry
Screenshot 2026-06-12 at 2 09 03 PM

https://expoio.sentry.io/issues/7545404457/events/latest/?environment=development&project=1837720&query=is%3Aunresolved&referrer=latest-event&statsPeriod=1h

packages/worker/src/env.ts
AbbanMustafa: I think we should! posted validation below

packages/build-tools/src/common/installDependencies.ts
AbbanMustafa: Followed up but this is because spawnpromise is a type not subclass. So the err thrown is a plain Error

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -69,11 +72,80 @@ export async function installDependenciesAsync({
...
+}: {
+  packageManager: PackageManager;
+  env: Record<string, string | undefined>;
+  cwd: string;
+  logger: Exclude<SpawnOptions['logger'], undefined>;
+  infoCallbackFn?: SpawnOptions['infoCallbackFn'];
+  useFrozenLockfile: boolean;
+}): Promise<void> {
+  const npmCacheUrl = env.NPM_CONFIG_REGISTRY;
+  const npmCacheErrorTracker = createNpmCacheRegistryErrorTracker({ env, npmCacheUrl });
```
sjchmiela: let's not introduce this "error tracker" concept, i think we can just inline its logic in this function. we then reduce the logical burden of the code (don't need to see and remember what tracker is), or wonder what reporting non-fatal error means (and that it doesn't actually always report). for `lineTransformer` we can have just an arrow function

sjchmiela (approved):
before switching the killswitch i think we might want to also bring back adhering to `EAS_BUILD_DISABLE_COCOAPODS_CACHE`, `EAS_BUILD_DISABLE_MAVEN_CACHE` and `EAS_BUILD_DISABLE_NPM_CACHE` (https://docs.expo.dev/build-reference/caching/#javascript-dependencies)

PR #3864, opened 2026-06-15
[eas-cli] Support platform-specific update IDs in update:view

keith-kurak:
# Why

User noted that it was difficult to cross reference a platform-specific update ID back to the group ID. It is possible (though not obvious) to put the update ID in this group URL, and it will take you directly to the update: `https://expo.dev/accounts/[account]/projects/[project]/updates/[id]`

Exposing this via the CLI would provide feature parity, but more importantly, open things up to programmatic flows (e.g., you get an update ID from a client bug report and you want to cross reference back to the update group deployment).

# How

Apparently there was already a query available for this, so added a fallback lookup to the existing `eas update:view` command to look for the platform-specific ID.

# Test Plan

image

works with all three ID's, same result:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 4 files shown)


quinlanj (approved):
thanks!

PR #3865, opened 2026-06-16
[eas-cli] share agent templates from llm-configs

davidmokos:
Syncs the generated app `AGENTS.md` and `CLAUDE.md` templates from `expo/llm-configs` during EAS CLI `prepack`, before the package is rebuilt for publishing. Keeps the synced templates bundled in EAS CLI so `eas project:new` still works offline and does not add a runtime `@expo/llm-configs` dependency.

Verified with `yarn workspace eas-cli sync-agent-templates`, byte-for-byte comparison against the upstream raw files, `yarn workspace eas-cli test src/commandUtils/new/__tests__/projectFiles.test.ts --runInBand`, `yarn workspace eas-cli typecheck-for-build`, and `yarn workspace eas-cli build`. Previously also ran a local `eas project:new` smoke test and confirmed generated agent files matched the shared templates.


Kudo:
side question: could we fetch the doc from github.com/expo/llm-configs/blob/main/expo-app/AGENTS.md directly? we could always get latest agents.md without publishing eas-cli

davidmokos:
There are drawbacks - it would make offline creation, GitHub outage, proxy issues, or timeout produce no agent files. IMHO project creation should not depend on GitHub raw being reachable, bundled fallback is the safer design. What do you think?

Kudo:
that's a good point. in this case, how about to sync files in prepublish phase? we still fetch CLAUDE.md/AGENTS.md from llm_configs, and overwrites [existing templates](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commandUtils/new/templates/AGENTS.md).

some advantages for this approach:
- reduce llm_configs deps overhead. the repo may contain other files that are not used by eas-cli.
- whenever we update to llm_configs, next time when eas-cli published, we get the latest one automatically.
- still have offline support.

davidmokos:
I like that approach! Updated the PR

Kudo (approved):

PR #3866, opened 2026-06-16
[eas-cli] Fix update:embedded topic description

jc-expo:
# Why

The `update:embedded` command group had no explicit topic description, so oclif derived the topic's description from one of its subcommands. As a result, the topic listing rendered:

```
TOPICS
  update:embedded  delete an embedded update registered with EAS Update
```

This is confusing — it describes only the `update:embedded:delete` subcommand rather than the group as a whole.

# How

Added an explicit `update:embedded` topic entry in `packages/eas-cli/package.json` with the description `manage embedded updates registered with EAS Update`, matching the style of the sibling `update` topic (`manage individual updates`).

# Test Plan

Run `eas update --help` and `eas update:embedded --help` and confirm the `update:embedded` group now shows `manage embedded updates registered with EAS Update` instead of the delete-subcommand description.


gwdp (approved):
:fire:

PR #3869, opened 2026-06-16
[eas-cli] Pre-warm bsdiff patches against top-K base and embedded updates

jc-expo:
# Why

Linear [ENG-21079](https://linear.app/expo/issue/ENG-21079/metrics-based-prewarmer).

The previous `prewarmDiffingAsync` base-update selection was too simple: it always picked `updateIds[1]` — the second-most-recent update on the branch (filtered by `platform + runtimeVersion`). If the second-newest was a quick hotfix on top of the newest, most devices in the field are likely on the *third* newest (or older), so the pre-warmed bsdiff patch misses the base that real traffic actually requests.

When a device requests a base that hasn't been pre-warmed, the server kicks off the diff on-demand but, until that patch is ready, serves the full **un-diffed** bundle. So a bad base guess means those clients download the entire update instead of a small patch until the diff finishes computing.

# How

Pre-warm against the top-K candidates instead of a single base, approximating the bases that the lazy path organically catches from real device traffic:

- Issue HEAD requests against the **top 5 most recent updates** on the branch (skipping the current one) rather than just the second-most-recent.
- Also pre-warm patches from the **registered embedded bundles** (up to 2) to the new update — the fresh-install scenario — falling back to the sentinel empty-embedded id when a project has no registered embedded bundle.
- Extracted the per-update logic into `prewarmUpdateDiffsAsync`, kept it best-effort (failures are logged via `Log.debug` and swallowed, never blocking a publish), and added debug logging throughout.

# Test Plan

- `yarn test` for the `eas-cli` package passes.
- Manually published an update with `EXPO_DEBUG=1` and confirmed the debug logs show the top-K recent updates and embedded bundles being queued and HEAD-requested with the `a-im: bsdiff` diffing headers.

### DIFF
--- a/packages/eas-cli/src/update/utils.ts
+++ b/packages/eas-cli/src/update/utils.ts
@@ -18,7 +18,8 @@ import {
 } from '../graphql/generated';
 import { AssetQuery } from '../graphql/queries/AssetQuery';
 import { BranchQuery } from '../graphql/queries/BranchQuery';
-import { learnMore } from '../log';
+import { EmbeddedUpdateQuery } from '../graphql/queries/EmbeddedUpdateQuery';
+import Log, { learnMore } from '../log';
 import { RequestedPlatform } from '../platform';
 import { getActorDisplayName } from '../user/User';
 import groupBy from '../utils/expodash/groupBy';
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/update/utils.ts
```diff
@@ -287,75 +289,173 @@ export function isBundleDiffingEnabled(exp: ExpoConfig): boolean {
...
+      return [];
+    }
+
+    // Pre-warm the patch from the bundle embedded in the native binary to the new update.
+    // This is what fresh installs request, so generating it ahead of time avoids an
+    // expensive on-demand diff. Falls back to the empty/default embedded id when the project has no
+    // registered embedded bundle.
+    const embeddedUpdateQuery = await EmbeddedUpdateQuery.viewPaginatedAsync(graphqlClient, {
+      appId,
+      filter: { platform, runtimeVersion: update.runtimeVersion },
```
quinlanj: you'll need to also filter by `channel` to get the set of builds (with embedded updates) that are eligible to receive the published update

quinlanj (changes requested):
looking good so far! 💯

packages/eas-cli/src/update/utils.ts
```diff
@@ -287,75 +291,221 @@ export function isBundleDiffingEnabled(exp: ExpoConfig): boolean {
...
+    {
+      appId,
+      first: PREWARM_CHANNELS_LIMIT,
+    }
+  );
+  return channels.edges
+    .map(edge => edge.node)
+    .filter(channel => getBranchIds(getBranchMapping(channel.branchMapping)).includes(branchId))
+    .map(channel => channel.name);
+}
```
quinlanj, line 326: I forgot there's no graphql api for this yet. Can you open up a separate pr and add a `channelPaginated` edge off the [UpdateBranch](https://github.com/expo/universe/blob/e418d3ae176a78c13bcd190cd8492aa4198fa7a8/server/www/src/graphql/types/UpdateBranch.ts#L19) node to implement relay pagination like this:

```
  const connection = await ChannelEntity.knexLoader(viewerContext, queryContext).loadPageAsync({
    first,
    after,
    where: sql`filter_branch_id(${entityField('branchMapping')}) @>
  ${JSON.stringify(branch.getID())}::JSONB`,
    pagination: {
      strategy: PaginationStrategy.STANDARD,
      orderBy: [{ fieldName: 'createdAt', order: OrderByOrdering.DESCENDING }],
    },
  });
```

Also if you could add a comment saying to only use this internally in the graphql api, that'd be great:

```
  """
  INTERNAL USE ONLY. Channels whose branchMapping references this branch.
  Branch -> channels is not a first-class relationship in EAS Update; do not
  depend on this in public integrations.
  """
```

PR #3872, opened 2026-06-17
[build-tools] upload maestro screenshots

hSATAC:
> Depends on https://github.com/expo/universe/pull/28185 — www must persist the artifact `metadata` this PR sends.

# Why

Surface Maestro per-flow failure screenshots in the workflow Maestro test results UI, shown next to the failing test-case row. Part of ENG-19036.

# How

- **build-tools:** After each Maestro attempt (junit output only), harvest failure screenshots from the debug dirs, then **reduce to what the UI actually surfaces**: a flow that failed on every attempt uploads only its final-attempt screenshot, while a flaky/mixed flow keeps every failed-attempt screenshot (classified from the per-attempt JUnit pass/fail, not screenshot counts). The kept screenshots upload as workflow artifacts carrying `metadata` (`kind`, `flowName`, `attemptIndex`, `capturedAtMs`) that the website matches to the corresponding test-case result. Best-effort and verdict-neutral: errors are logged rather than thrown, and uploads are capped at 30 per job.
- **worker:** Thread an optional `metadata` field through the artifact upload path (`uploadArtifact` → `uploadWorkflowArtifactAsync` → `createUploadSessionAsync`) so it reaches www.

# Test Plan

Unit tests added/updated (`maestroScreenshots.test.ts`, `maestroTests.test.ts`), all passing in CI — covering filename parsing, the mtime/`capturedAt` harvest gating, the pure-vs-flaky reduction (final-only for all-failed flows, every attempt for flaky), the 30-upload cap, upload-before-failure-throw, and per-item error swallowing.

### DIFF
--- a/packages/build-tools/src/context.ts
+++ b/packages/build-tools/src/context.ts
@@ -45,6 +45,7 @@ export type ArtifactToUpload =
       type: GenericArtifactType;
       name: string;
       paths: string[];
+      metadata?: Record<string, unknown>;
     };
 
 export interface BuildContextOptions {
--- a/packages/build-tools/src/steps/functions/maestroScreenshots.ts
+++ b/packages/build-tools/src/steps/functions/maestroScreenshots.ts
@@ -1,0 +1,159 @@
+import { bunyan } from '@expo/logger';
+import { type Dirent } from 'fs';
+import fs from 'fs/promises';
+import path from 'path';
+
+export interface ParsedFailureScreenshot {
+  flowName: string;
+  capturedAtMs: number;
+}
+
+// Maestro failure screenshot: `screenshot-[shard-N-]❌-<epochMillis>-(<flowName>).png`.
+// The flow name may contain parentheses, so anchor on the `-(` after the epoch and the
+// `).png` suffix rather than scanning for parens.
+const FAILURE_SCREENSHOT_PATTERN = /^screenshot-(?:shard-\d+-)?❌-(\d+)-\((.+)\)\.png$/u;
+
+export function parseFailureScreenshotFilename(filename: string): ParsedFailureScreenshot | null {
+  const match = filename.match(FAILURE_SCREENSHOT_PATTERN);
+  if (!match) {
+    return null;
+  }
+  const capturedAtMs = Number(match[1]);
+  if (!Number.isFinite(capturedAtMs)) {
+    return null;
+  }
+  return {
+    flowName: match[2],
+    capturedAtMs,
+  };
+}
+
+export interface HarvestedScreenshot {
+  fileAbsPath: string;
+  displayName: string; // `Failure Screenshot: <flowName> (attempt <n>)`
+  metadata: {
+    kind: 'maestro-test-screenshot';
+    // The Maestro flow name straight from the filename (already '/'->'_' substituted by
+    // Maestro). The website matches this against `attempt.name` (normalized the same way).
+    flowName: string;
+    attemptIndex: number;
+    capturedAtMs: number;
+  };
+}
+
+// Scans every debug dir under testsDirectory whose mtime is recent enough (mtime-based to cover
+// the maestro <2.5.0 bug that splits one invocation across two dirs), then keeps only screenshots
+// whose own capturedAtMs is at/after capturedSinceMs. Never throws: an unreadable dir is logged
+// and skipped so a screenshot harvest can't affect the maestro step outcome.
+export async function harvestFailureScreenshotsAsync(args: {
+  testsDirectory: string;
+  capturedSinceMs: number;
+  attemptIndex: number;
+  logger: bunyan;
+}): Promise<HarvestedScreenshot[]> {
+  // Gate dirs by mtime floored to the second, so a dir created within the same second as the
+  // attempt start isn't stat'd below the baseline. The exact capturedAtMs gate below is what
+  // prevents mis-attributing an earlier attempt's screenshot.
+  const dirSinceMtimeMs = Math.floor(args.capturedSinceMs / 1000) * 1000;
+  let entries: Dirent[];
+  try {
+    entries = await fs.readdir(args.testsDirectory, { withFileTypes: true });
+  } catch (err: any) {
+    args.logger.info({ err }, `Skipping screenshot harvest: cannot read ${args.testsDirectory}.`);
+    return [];
+  }
+
+  const results: HarvestedScreenshot[] = [];
+  for (const entry of entries) {
+    if (!entry.isDirectory()) {
+      continue;
+    }
+    const dirPath = path.join(args.testsDirectory, entry.name);
+    let filenames: string[];
+    try {
+      if ((await fs.stat(dirPath)).mtimeMs < dirSinceMtimeMs) {
+        continue;
+      }
+      filenames = await fs.readdir(dirPath);
+    } catch (err: any) {
+      args.logger.info({ err }, `Skipping unreadable debug dir ${dirPath}.`);
+      continue;
+    }
+    for (const filename of filenames) {
+      const parsed = parseFailureScreenshotFilename(filename);
+      // Re-check capture time per file: a dir's mtime can advance after this attempt began (or
+      // fall within the floored-second dir baseline), so the dir gate alone would re-attribute an
+      // earlier attempt's screenshot to this one.
+      if (parsed === null || parsed.capturedAtMs < args.capturedSinceMs) {
+        continue;
+      }
+      results.push({
+        fileAbsPath: path.join(dirPath, filename),
+        displayName: `Failure Screenshot: ${parsed.flowName} (attempt ${args.attemptIndex + 1})`,
+        metadata: {
+          kind: 'maestro-test-screenshot',
+          flowName: parsed.flowName,
+          attemptIndex: args.attemptIndex,
+          capturedAtMs: parsed.capturedAtMs,
+        },
+      });
+    }
+  }
+  return results;
+}
+
+// Maestro substitutes '/' -> '_' in screenshot filenames (so HarvestedScreenshot.flowName
+// is already in that form) but keeps '/' in the JUnit testcase name. Normalize the JUnit
+// name the same way before comparing — matching the website's own normalization.
+function normalizeFlowName(name: string): string {
+  return name.replace(/\//g, '_');
+}
+
+// A flow is a "pure failure" when it failed and never passed across all attempts. The website
+// surfaces only the final screenshot for such a flow, so keeping its earlier attempts would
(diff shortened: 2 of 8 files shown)


hSATAC:
* **#3878** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3872** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3872?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3872?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/context.ts
```diff
@@ -53,7 +53,11 @@ export interface BuildContextOptions {
   logBuffer: LogBuffer;
   env: Env;
   cacheManager?: CacheManager;
-  uploadArtifact: (spec: { artifact: ArtifactToUpload; logger: bunyan }) => Promise<
+  uploadArtifact: (spec: {
+    artifact: ArtifactToUpload;
+    metadata?: Record<string, unknown>;
```
sjchmiela: should maybe metadata be part of https://github.com/expo/eas-cli/blob/17eb5f0687d15df518918090d023b7e764ac260c/packages/build-tools/src/context.ts#L44-L48?

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -258,6 +270,24 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
...
+        // junit: test-case-result rows (and therefore the summary icons) only exist for junit
+        // runs, so harvesting other formats would just create orphan artifacts the website hides.
+        if (outputFormat === 'junit') {
+          harvested.push(
+            ...(await harvestFailureScreenshotsAsync({
+              testsDirectory,
+              // Floor the dir-mtime gate so a dir created within the same second isn't stat'd
+              // below the baseline; the exact capturedAtMs gate prevents mis-attribution (see
+              // harvestFailureScreenshotsAsync).
+              dirSinceMtimeMs: Math.floor(attemptStartedAtMs / 1000) * 1000,
```
sjchmiela: maybe this can be fully done by `harvestFailureScreenshotsAsync`? like it would receive `capturedSinceMs` and calculate `dirSinceMtimeMs = Math.floor(capturedSinceMs / 1000) * 1000`?

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -330,6 +360,58 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
         }
       }
 
+      // Reduce to the screenshots the website actually surfaces: it shows every failed
+      // attempt for flaky/mixed flows but only the final attempt for flows that failed
+      // every time. Classify from the per-attempt JUnit (the semantic pass/fail source),
```
sjchmiela: maybe it makes sense to display failure screenshot on the flaked tests too? so a user sees what was the flake about?

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -330,6 +360,58 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
...
+      // then drop the undisplayed earlier attempts of all-failed flows so they don't waste
+      // the per-job artifact budget or inflate the "screenshots attached" count.
```
sjchmiela: we can increase the budget if we need to and probably will need to

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -330,6 +360,58 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
...
+          `Found ${harvested.length} failure screenshots; uploading only the first ${MAX_SCREENSHOT_UPLOADS}.`
+        );
+      }
+      // Copy each shot to an ASCII-safe name outside testsDirectory (the originals contain a
+      // non-ASCII marker and testsDirectory is uploaded wholesale as the tarball).
+      const safeScreenshotDir = path.join(os.tmpdir(), 'eas-maestro-screenshots');
+      try {
+        await fs.mkdir(safeScreenshotDir, { recursive: true });
+      } catch (err: any) {
+        logger.warn(`Failed to create the failure-screenshot staging dir: ${err?.message ?? err}`);
```
sjchmiela: should we return here?

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -330,6 +360,58 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
...
+      if (harvested.length > MAX_SCREENSHOT_UPLOADS) {
+        logger.warn(
+          `Found ${harvested.length} failure screenshots; uploading only the first ${MAX_SCREENSHOT_UPLOADS}.`
+        );
+      }
+      // Copy each shot to an ASCII-safe name outside testsDirectory (the originals contain a
+      // non-ASCII marker and testsDirectory is uploaded wholesale as the tarball).
+      const safeScreenshotDir = path.join(os.tmpdir(), 'eas-maestro-screenshots');
+      try {
+        await fs.mkdir(safeScreenshotDir, { recursive: true });
```
sjchmiela: should we `force: true`?

packages/build-tools/src/steps/functions/maestroTests.ts
sjchmiela:
```suggestion
        logger.warn({ err }, `Failed to create the failure-screenshot staging dir.`);
```

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -330,6 +360,58 @@ export function createMaestroTestsBuildFunction(): BuildFunction {
...
+      // still upload. The cap keeps us well under www's 50-artifact-per-job limit.
+      const MAX_SCREENSHOT_UPLOADS = 30;
+      if (harvested.length > MAX_SCREENSHOT_UPLOADS) {
+        logger.warn(
+          `Found ${harvested.length} failure screenshots; uploading only the first ${MAX_SCREENSHOT_UPLOADS}.`
+        );
+      }
+      // Copy each shot to an ASCII-safe name outside testsDirectory (the originals contain a
+      // non-ASCII marker and testsDirectory is uploaded wholesale as the tarball).
+      const safeScreenshotDir = path.join(os.tmpdir(), 'eas-maestro-screenshots');
```
sjchmiela: should we do a mkdtemp so it's not global?

packages/build-tools/src/steps/functions/maestroScreenshots.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export interface ParsedFailureScreenshot {
+  flowName: string;
+  capturedAtMs: number;
+}
+
+// Maestro failure screenshot: `screenshot-[shard-N-]❌-<epochMillis>-(<flowName>).png`.
+// The flow name may contain parentheses, so anchor on the `-(` after the epoch and the
+// `).png` suffix rather than scanning for parens.
+const FAILURE_SCREENSHOT_PATTERN = /^screenshot-(?:shard-\d+-)?❌-(\d+)-\((.+)\)\.png$/u;
```
sjchmiela, line 14: eh, if only junit was pointing to the screenshot file path…

packages/build-tools/src/steps/functions/maestroScreenshots.ts
```diff
@@ -0,0 +1,157 @@
...
+  capturedSinceMs: number;
+  attemptIndex: number;
+  logger: bunyan;
+}): Promise<HarvestedScreenshot[]> {
+  let entries: Dirent[];
+  try {
+    entries = await fs.readdir(args.testsDirectory, { withFileTypes: true });
+  } catch (err: any) {
+    args.logger.info(
+      `Skipping screenshot harvest: cannot read ${args.testsDirectory} (${err?.message ?? err}).`
```
sjchmiela:
```suggestion
      { err }, `Skipping screenshot harvest: cannot read ${args.testsDirectory}.`
```
?

packages/build-tools/src/steps/functions/maestroScreenshots.ts
```diff
@@ -0,0 +1,157 @@
...
+    // The Maestro flow name straight from the filename (already '/'->'_' substituted by
+    // Maestro). The website matches this against `attempt.name` (normalized the same way).
+    flowName: string;
+    attemptIndex: number;
+    capturedAtMs: number;
+  };
+}
+
+// Scans every debug dir under testsDirectory whose mtime is at/after dirSinceMtimeMs (mtime-based
+// to cover the maestro <2.5.0 bug that splits one invocation across two dirs), then keeps only
```
sjchmiela: let's remember to drop any additional complexity as soon as we decide to drop the extra support for <2.5?

or maybe we want to do it already?

sjchmiela (approved):

packages/build-tools/src/steps/functions/maestroTests.ts
hSATAC: Yes we're displaying every screenshots for flaked tests. We only display the final screenshot for the always failed flow.

packages/build-tools/src/context.ts
hSATAC: Absolutely...can't believe I didn't see it. Will fix.

packages/build-tools/src/steps/functions/maestroTests.ts
hSATAC: Hmm I don't think `fs.mkdir()` supports `force:` but will use `mkdtemp` instead.

packages/build-tools/src/steps/functions/maestroScreenshots.ts
hSATAC: Probably something worth adding to the JUnit report...

packages/build-tools/src/steps/functions/maestroScreenshots.ts
hSATAC: I have a Linear ticket for that.

Around 75% of users are on 2.6+ today. If we drop the old path now, ~25% of users would lose these features immediately.

I was thinking something closer to 90%+ adoption before removing it, but happy to discuss.

packages/build-tools/src/steps/functions/maestroTests.ts
hSATAC: Will modify the comments a bit so it's easier to read.

PR #3881, opened 2026-06-18
[build-tools] Fix Argent remote session startup

sjchmiela:
Fixes Argent remote simulator sessions getting stuck immediately after starting the Argent tool-server.

The latest Argent support change switched the startup path to `bunx @swmansion/argent server start ...` and awaited that process. When the server process stays in the foreground, the build step never advances to read the tool-server state file, start the ngrok tunnel, or report remote config back to the API.

This restores detached process launching for the Argent tool-server while keeping the new server/auth flow. If the state file is not created in time, the error now includes captured Argent process output to aid diagnosis.

Failing job run referenced while debugging: https://expo.dev/accounts/sjchmiela/projects/test-production/job-runs/019eda48-6467-789d-8c5e-a7ed5f4ec89d

### DIFF
--- a/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
@@ -18,6 +18,7 @@ import {
   getDeviceRunSessionIdOrThrow,
   getNgrokAuthtokenOrThrow,
   getNgrokTunnelDomainOrThrow,
+  spawnDetached,
   startNgrokTunnelAsync,
   startServeSimWithTunnelAsync,
   uploadRemoteSessionConfigAsync,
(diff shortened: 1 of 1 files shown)


packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
```diff
@@ -88,16 +89,29 @@ export function createStartArgentRemoteSessionBuildFunction(
...
-      const { port: toolServerPort, token: toolServerToken } = await waitForFileAsync({
-        filePath: ARGENT_STATE_FILE,
-        timeoutMs: STARTUP_TIMEOUT_MS,
-        description: 'argent tool-server state',
-        parse: parseArgentToolServerState,
-      });
+      let toolServerPort: number;
+      let toolServerToken: string | undefined;
+      try {
+        const toolServerState = await waitForFileAsync({
```
hSATAC, line 99: Looks good overall. Small nit:

If I'm reading this correctly, waitForFileAsync will always take 60 seconds to fail if server start --detach exits early. Is that right?

I don't think this is a blocker, just wanted to point it out.

hSATAC (approved):

packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
sjchmiela: I think so! Added https://linear.app/expo/issue/6e765588-ab35-4741-9aa5-235aa4d4d3cd

PR #3882, opened 2026-06-18
[eas-cli] Add tiered build-credit usage warnings (approaching/at/over)

sarahlane8:
# Why

The existing pre-build warning fires at a single 85% threshold and gives Free vs paid plan messaging but doesn't distinguish between *approaching* the limit, *reaching* it, and *exceeding* it. As users move through those states, the message they need to see changes — and the current single-tier message is technically inaccurate once they hit 100%.

This PR splits the warning into three tiers, lowers the threshold to 80% to match Expo's email warning cadence, and standardizes the wording across the existing dashboard's "billing period" language.

# How

- **Expanded the existing warning GraphQL query** (`AccountUsageForOverageWarning`) to also return `overageMetrics { value }` and `totalCost`. These already exist on the schema; the query just wasn't asking for them.
- **Added `classifyUsageTier`**: returns `'approaching' | 'at' | 'over' | null` based on plan value, limit, and overage count. `over` wins if any overage is recorded; `at` if `planValue >= limit` with no overage; `approaching` if `>= 80%`.
- **Rewrote `displayOverageWarning`** with tier-specific messaging:
  - **Approaching**: progress bar + percent + "you'll be charged / blocked" warning
  - **At**: "you've reached your limit" + plan-specific consequence
  - **Over (paid)**: overage count + dollar amount accrued
  - **Over (Free)**: defensively renders the `at` message since Free can't accumulate overage (blocked at the limit). Kept as a safety net for data inconsistencies.
- **Standardized wording**: all three tiers now say *"this billing period"* (was "for this month"), matching the existing `eas account:usage` dashboard.
- **Did not change** the call site logic in `runBuildAndSubmit.ts` — the warning still fires once per `eas build` invocation, before the first build, and is skipped for `--local` builds.

# Test Plan

- **26 unit tests** pass, covering every (tier × plan) combination plus the classifier, progress bar, and account-name billing URL.
- **Local preview script** (not committed; lives in gitignored `packages/eas-cli/tmp/`) renders every variant for visual inspection. Sample output:

  ```
  === Approaching · Paid (90%) ===
  You've used 90% of your included build credits this billing period. ███████████████████████████░░░
  Additional usage beyond your limit will be charged at pay-as-you-go rates. See usage in billing.: https://expo.dev/accounts/demo-account/settings/billing

  === At limit · Free (100%) ===
  You've reached your included build credits this billing period.
  New builds are blocked until your billing period resets. Upgrade your plan to continue building.: https://expo.dev/accounts/demo-account/settings/billing

  === Over · Paid (multiple overage builds) ===
  You've used 12 builds beyond your included credits this billing period ($18.00 in overages so far).
  Additional builds continue at pay-as-you-go rates. See usage in billing.: https://expo.dev/accounts/demo-account/settings/billing
  ```

- **Not yet verified end-to-end** against a real account on staging — the GraphQL fields already exist on the schema, but it would be good to confirm the query returns expected data against a real over-limit account before merging.

## Copy review

Public-facing customer warning copy. Would appreciate eyes from PM / customer success / growth before merging.

### DIFF
--- a/packages/eas-cli/src/utils/usage/checkForOverages.ts
+++ b/packages/eas-cli/src/utils/usage/checkForOverages.ts
@@ -4,7 +4,9 @@ import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/creat
 import { AccountQuery } from '../../graphql/queries/AccountQuery';
 import Log, { link } from '../../log';
 
-const THRESHOLD_PERCENT = 85;
+const APPROACHING_THRESHOLD_PERCENT = 80;
+
+export type UsageTier = 'approaching' | 'at' | 'over';
 
 export async function maybeWarnAboutUsageOveragesAsync({
   graphqlClient,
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/utils/usage/checkForOverages.ts
```diff
@@ -4,7 +4,9 @@ import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/creat
 import { AccountQuery } from '../../graphql/queries/AccountQuery';
 import Log, { link } from '../../log';
 
-const THRESHOLD_PERCENT = 85;
+const APPROACHING_THRESHOLD_PERCENT = 80;
```
sarahlane8, line 7: I moved this from 85 to 80 to better match the email warnings, thoughts?

keith-kurak: this makes sense to me

sjchmiela (approved):

PR #3883, opened 2026-06-19
[build-tools] Fix Argent artifact upload size

sjchmiela:
Since https://github.com/software-mansion/argent/pull/347 Argent exposes a `GET /artifacts/` endpoint which returns a `size` for every artifact. This `size` is then being used to create a `www` upload session (which validates and includes the `size` as `Content-Length` of the signed upload URL).

The `size` is being calculated in Argent as `fs.stat()` which works great for single files, but for directories it returns 96 bytes. The actual archiving of a given directory (producing an artifact) is only happening when downloading the artifact with `GET /artifacts/:id`.

This pull request changes the artifacts behavior around Argent from:
- list artifacts
- trust `size`
- pipe `GET /artifacts/:id` directly to signed upload URL

to:
- list artifacts
- download from `GET /artifacts/:id` to temporary directory
- check the `size` of the actually downloaded file
- pipe it to signed upload URL.

## Test

You can see which traces have been created _with_ the fix and which _without_ it.

Zrzut ekranu 2026-06-19 o 13 03 01

### DIFF
--- a/packages/build-tools/src/steps/utils/argentArtifacts.ts
+++ b/packages/build-tools/src/steps/utils/argentArtifacts.ts
@@ -1,4 +1,5 @@
 import { SystemError } from '@expo/eas-build-job';
+import { type bunyan } from '@expo/logger';
 import { createReadStream, createWriteStream } from 'node:fs';
 import { mkdtemp, rm, stat } from 'node:fs/promises';
 import fetch from 'node-fetch';
@@ -19,7 +20,6 @@ const ArgentArtifactSchema = z.object({
   id: z.string(),
   filename: z.string(),
   mimeType: z.string(),
-  size: z.number(),
   isDirectory: z.boolean().optional(),
 });
 const ArgentArtifactsListResponseSchema = z.object({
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -104,39 +112,52 @@ export async function uploadArgentArtifactAsync(
...
+  const temporaryDirectory = await mkdtemp(path.join(os.tmpdir(), 'argent-artifact-'));
+  try {
+    const temporaryArtifactPath = path.join(temporaryDirectory, path.basename(filename));
+    await downloadArgentArtifactToFileAsync({
+      artifact,
+      toolsUrl,
+      toolsAuthToken,
+      destinationPath: temporaryArtifactPath,
+    });
+    const { size } = await stat(temporaryArtifactPath);
```
sswrk, line 134: I think it's worth adding a comment explaining why we do this even though there is non-null `size` field in Argent artifact Zod schema. Or removing it from the zod schema as it's unused.

sswrk (approved):

PR #3885, opened 2026-06-19
Use sjchmiela serve-sim previews without WebRTC

sjchmiela:
## Summary

- switch remote device run sessions to the `serve-sim-sjchmiela` preview helper
- report only the preview URL from serve-sim remote sessions
- avoid forcing WebRTC for tunneled previews for now, keeping the AVCC/H.264 path with MJPEG fallback

## Why

The live WebRTC preview path can get stuck during `/webrtc/offer` while the helper waits for server-side ICE gathering, leaving the browser simulator surface at `connecting`. The AVCC and MJPEG stream endpoints are healthy in that state, so this keeps the new helper while avoiding the broken WebRTC signaling path.

## Test plan

- `yarn oxfmt --write packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts packages/build-tools/src/steps/utils/__tests__/remoteDeviceRunSession.test.ts`
- `yarn --cwd packages/build-tools jest-unit src/steps/utils/__tests__/remoteDeviceRunSession.test.ts`
- `git diff --check main..HEAD`

### DIFF
--- a/packages/build-tools/src/steps/functions/startServeSimRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startServeSimRemoteSession.ts
@@ -28,19 +28,18 @@ export function createStartServeSimRemoteSessionBuildFunction(
 
       await selectXcodeDeveloperDirectoryAsync({ env, logger });
 
-      const { previewUrl, streamUrl } = await startServeSimWithTunnelAsync(ctx, {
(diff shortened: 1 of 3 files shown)


sjchmiela:
we're going to use expo-device-hub instead

PR #3892, opened 2026-06-23
Fix repack AAB output path

Kudo:
# Why

Repacked Android builds should produce an APK output path when the source app path ends in `.aab`.

# How

Removed the unused `output_path` input and generate the repacked output path internally. Generated `.aab` paths are rewritten to `.apk`.

# Test Plan

- `yarn run -T oxfmt packages/build-tools/src/steps/functions/repack.ts packages/build-tools/src/steps/functions/__tests__/repack.test.ts`
- `yarn --cwd packages/build-tools jest-unit repack.test.ts --watchman=false`

### DIFF
--- a/packages/build-tools/src/steps/functions/__tests__/repack.test.ts
+++ b/packages/build-tools/src/steps/functions/__tests__/repack.test.ts
@@ -48,6 +48,47 @@ describe(createRepackBuildFunction, () => {
     expect(repackStep.outputById['output_path'].value).toBe('/path/to/output_app');
   });
 
+  it('should rename explicit aab output path to apk', async () => {
+    const repack = createRepackBuildFunction();
+    const repackStep = repack.createBuildStepFromFunctionCall(
+      createGlobalContextMock({
+        staticContextContent: {
+          job: createTestAndroidJob(),
+        },
+      }),
+      {
+        callInputs: {
+          platform: 'android',
+          source_app_path: '/path/to/source_app.aab',
+          output_path: '/path/to/output_app.aab',
+        },
+      }
+    );
+
+    await repackStep.executeAsync();
+    expect(repackStep.outputById['output_path'].value).toBe('/path/to/output_app.apk');
+  });
+
+  it('should rename generated aab output path to apk', async () => {
+    const repack = createRepackBuildFunction();
+    const repackStep = repack.createBuildStepFromFunctionCall(
+      createGlobalContextMock({
+        staticContextContent: {
+          job: createTestAndroidJob(),
+        },
+      }),
+      {
+        callInputs: {
+          platform: 'android',
+          source_app_path: '/path/to/source_app.aab',
+        },
+      }
+    );
+
+    await repackStep.executeAsync();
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/functions/__tests__/repack.test.ts
```diff
@@ -48,6 +48,47 @@ describe(createRepackBuildFunction, () => {
     expect(repackStep.outputById['output_path'].value).toBe('/path/to/output_app');
   });
 
+  it('should rename explicit aab output path to apk', async () => {
```
sjchmiela: isn't this weird? like what's the purpose of `output_path` input if it can be different in the end haha

maybe we can drop the input altogether?

Kudo: that's kind of intentional to support aab. repack wouldn't be ideal tool to support production aab build, but it's fine to repack from aab source build.
open to idea if there's a better way to support aab. (or even fine that just not to support aab)

sjchmiela: i'm talking only about `output_path` being both an input and an output. i'm all for repacking an aab. if i were to pass in some value to `output_path` as input i would expect the artifact to be where i told it to be

Kudo: i got your point! in the latest commit, it preserves the output_path if explicitly specifying an output_path.

packages/build-tools/src/steps/functions/repack.ts
```diff
@@ -109,9 +109,11 @@ export function createRepackBuildFunction(): BuildFunction {
...
 
       const sourceAppPath = inputs.source_app_path.value as string;
-      const outputPath =
-        (inputs.output_path.value as string) ??
-        path.join(tmpDir, `repacked-${randomUUID()}${path.extname(sourceAppPath)}`);
+      const outputPath = createOutputPath({
+        requestedOutputPath: inputs.output_path.value as string | undefined,
+        sourceAppPath,
+        tmpDir,
+      });
```
sjchmiela, line 110:
```suggestion
      const outputPath = inputs.output_path.value as string | undefined || createOutputPath({
        sourceAppPath,
        tmpDir,
      });
```
it feels clearer to me

sjchmiela (approved):
thank you!

we could also add `Datadog.log` for when we do see `output_path` being used (and not used), and if it isn't, remove it?

Kudo:
> we could also add Datadog.log for when we do see output_path being used (and not used), and if it isn't, remove it?

i think output_path is not used.
 https://github.com/expo/universe/blob/48c43fd54a33fc911756cefe52937f842113d049/server/www/src/data/entities/workflow/job/WorkflowRepackJob.ts#L223-L236

want me to remove it in this pr?

Kudo:
> want me to remove it in this pr?

now i entirely remove the `inputs.output_path` since it's not used

sjchmiela (approved):

PR #3894, opened 2026-06-25
[eas-cli]: warn user on missing primary locale

luke-h1:
# Why
This PR warns the user if an app's metadata they're attempting to update/add is missing the primary locale (and/or if they have another non-primary locale in their config). Providing a non-primary locale while you are missing primary locale can cause some confusion as it will show up blank in the ASC's dashboard

# How

Log a warning to let the user know they might want to fill out the primary locale and/or change the app's primary language

1. Before creating apple tasks, we get the locale `config.getLocales()` and check whether the user has provided an empty object for their primary locale and whether they are using any non-primary locales i.e. 

```
// store.config.json 

en-gb: {}
en-us: { title: "cool app", description: "cool stuff" }
```

# Test Plan

Unit tests - not sure if there is a more automated / interactive way of building the CLI and testing a version so I just went with unit tests for the time being! happy to add more test coverage if there's a better way though

Screenshot 2026-06-25 at 01 03 15

### DIFF
--- a/packages/eas-cli/src/metadata/__tests__/upload.test.ts
+++ b/packages/eas-cli/src/metadata/__tests__/upload.test.ts
@@ -1,9 +1,10 @@
 import { uploadMetadataAsync } from '../upload';
 import { MetadataValidationError } from '../errors';
+import Log from '../../log';
 
 jest.mock('../auth', () => ({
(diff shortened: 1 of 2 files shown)


luke-h1:
/changelog-entry chore warn user on missing primary locale

PR #3895, opened 2026-06-25
[eas-cli] Connect existing PostHog accounts via browser handoff

gwdp:
# Why

Existing PostHog users couldn't connect — `eas integrations:posthog:connect` dead-ended with a "not supported yet" message when the account's email already had a PostHog account.

# How

On the existing-user (`requires_auth`) branch, open the approval URL in the browser, poll for the connection the website callback creates, then continue with the rest of `connect`. Non-interactive runs refuse with a clear message instead of hanging. Pairs with the `universe` stack (the `startPostHogConnection` / `completePostHogConnection` mutations + the `/posthog/callback` website scene).

# Test Plan

- `yarn test connect` (added: browser handoff opens the URL + polls, non-interactive refusal, poll timeout).
- run `connect` with an email that already has a PostHog account, approve in the browser, confirm it links and continues.

carbon (4)

### DIFF
--- a/packages/eas-cli/src/commands/integrations/posthog/connect.ts
+++ b/packages/eas-cli/src/commands/integrations/posthog/connect.ts
@@ -1,6 +1,7 @@
 import { ExpoConfig } from '@expo/config';
 import spawnAsync from '@expo/spawn-async';
 import { Flags } from '@oclif/core';
+import openBrowserAsync from 'better-opn';
 import chalk from 'chalk';
 import dotenv from 'dotenv';
 import * as fs from 'fs-extra';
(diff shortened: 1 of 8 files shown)


packages/eas-cli/src/commands/integrations/posthog/connect.ts
```diff
@@ -276,6 +274,97 @@ export default class IntegrationsPostHogConnect extends EasCommand {
...
+    }
+
+    if (result.__typename === 'PostHogOrganizationConnection') {
+      spinner.succeed(`Created PostHog organization ${chalk.bold(result.posthogOrganizationName)}`);
+      return result;
+    }
+
+    spinner.stop();
+    if (nonInteractive) {
+      throw new Error(
```
mackenco, line 302: slightly odd to me that we have this error at the end of this function if not interactive, but the beginning of the next function if interactive (although not too big of a deal)

mackenco (approved):

packages/eas-cli/src/commands/integrations/posthog/connect.ts
gwdp: Thanks for the review @mackenco kept it there on purpose.`startConnectionAsync` result is the first point we actually know it's an existing PostHog user (vs a brand-new one), so it's the earliest we can tell a browser handoff is needed at all. 

Checking `nonInteractive` any sooner would wrongly bail on valid non-interactive new-user runs.

PR #3897, opened 2026-06-25
Use exec when launching Gradle

sjchmiela:
## Summary

Use `exec` in the shell wrapper that launches Gradle from both `build-tools` Gradle entry points.

This replaces the intermediate shell process with `gradlew`. Supposedly it's better.

Stepping stone to https://github.com/expo/eas-cli/pull/3898.


sjchmiela:
Landed this already https://github.com/expo/eas-cli/pull/3898

PR #3900, opened 2026-06-25
[eas-build-job] Allow undefined env values

sjchmiela:
## Summary

Widens the shared build `Env` type to allow `undefined` values, matching Node/process env semantics where undefined entries can represent absent values.

## Details

- Changes `Env` to `Record<string, string | undefined>` in `@expo/eas-build-job`.
- Allows `ctx.reportError` extras to contain undefined values for diagnostic metadata.
- Updates strict string boundaries to explicitly require values where needed, such as GraphQL IDs and static context API URLs.
- Filters undefined values out of `displayRuntimeInfo` string-only display maps.

## Validation

CI should pass.

### DIFF
--- a/packages/build-tools/src/steps/functions/downloadArtifact.ts
+++ b/packages/build-tools/src/steps/functions/downloadArtifact.ts
@@ -13,6 +13,7 @@ import os from 'node:os';
 import path from 'node:path';
 import stream from 'stream';
 import { promisify } from 'util';
+import nullthrows from 'nullthrows';
 import { z } from 'zod';
 
 import { formatBytes } from '../../utils/artifacts';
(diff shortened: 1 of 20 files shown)


packages/build-tools/src/utils/expoUpdates.ts
```diff
@@ -257,7 +258,7 @@ async function logDiffFingerprints({
           }
         }
       `),
-      { id: ctx.env.EAS_BUILD_ID }
+      { id: nullthrows(ctx.env.EAS_BUILD_ID, 'EAS_BUILD_ID is not set') }
```
sswrk: Is it safe to do nullthrows here? What about local builds?

sjchmiela: Fixed!!

packages/build-tools/src/steps/functions/downloadArtifact.ts
```diff
@@ -1,4 +1,4 @@
-import { UserError } from '@expo/eas-build-job';
+import { SystemError } from '@expo/eas-build-job';
```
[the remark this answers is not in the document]
sjchmiela: I think it's ok

sswrk (approved):

PR #3908, opened 2026-06-27
[eas-cli] migrate off deprecated Build.project and BuildMutation.cancel

tomek-agent:
# Why

The deprecated-GraphQL-field usage dashboard shows eas-cli is the top reader of several `@deprecated` output fields. `Build.project` alone accounts for ~394k resolver hits and `BuildMutation.cancel` for the cancel path. Both have drop-in replacements with identical server behavior, so moving off them removes the largest safely-removable slice of eas-cli's deprecated-field traffic.

# How

- `BuildFragment` selects `app` instead of `project`. On the server `project: Project!` and `app: App!` resolve through the same `getAppAsync()`, so this is a 1:1 swap. Because `app` is always an `App`, the `Project` union branch in `getBuildLogsUrl` collapses to the account/slug path.
- The `build cancel` command uses the `cancelBuild(buildId:)` mutation instead of the deprecated `build(buildId:).cancel`.
- `generated.ts` was regenerated against the committed schema so the diff is limited to these two documents (no unrelated schema refresh).

Intentionally left out:
- `Build.runtimeVersion` / `Build.channel` (the two next-biggest) are **not** safe to move yet: the deprecated string resolvers fall back to build metadata (`runtime?.version ?? metadata.runtimeVersion`) that the `runtime` / `updateChannel` objects don't, so switching would return null for older builds. Unblocking them needs a small server change to add the same fallback.
- `Update.runtimeVersion` is safe but threads through the update-formatting types (`FormatUpdateParameter`, `UpdateToRepublish`, `RolloutUpdate`); it belongs in its own PR.
- `Account.users` has no replacement ("No longer supported").

# Test Plan

Regenerated GraphQL types, built eas-cli, and ran the affected unit suites (build cancel, build delete, workflow logs) — all green. The `project → app` swap is behavior-preserving: both fields share the server resolver and `app` is non-null, so the log-URL, internal-distribution-URL, and run-cache-path consumers receive the same values. CI passes.


tomek-agent:
/changelog-entry chore migrate off deprecated GraphQL fields Build.project (use app) and BuildMutation.cancel (use cancelBuild)

PR #3918, opened 2026-06-30
[eas-cli] Re-upload assets stuck in polling loop when GCF finalize retry window is exhausted

gwdp:
# Why

`eas update` uploads each asset to GCS, and a cloud function then finalizes it in the database. If that function's own retries run out before the GCS event ages out (~40s), the event gets dropped. The asset never finalizes, so the CLI keeps polling until it times out and fails the whole publish. It's usually transient, which makes it extra frustrating to hit.

# How

The fix is small: while polling, if an asset still hasn't finalized after 55s, upload it again. A fresh upload means a fresh GCS event and a fresh retry window for the cloud function. We try up to 3 times.

If the assets still haven't finalized after all 3 attempts, we now fail loudly with the list of stuck assets instead of quietly publishing an update that points at them.

The caller's existing 180s inactivity timeout stays in place on purpose. It's the safety net for what this loop can't see, like a single request that hangs (there are no per-request timeouts yet), so nothing waits forever.

While I was here, the retry work made it worth tidying the upload flow: signing and uploading are decoupled so uploads run from one saturated concurrency pool instead of draining at each 100-asset chunk boundary, and progress is derived from the still-missing set instead of a separate variable that could drift.

# Test Plan

CI passes (unit tests cover chunked signing, re-uploading only the still-missing assets, the exhaustion error, and cancellation).

I also ran `eas update` locally against a real project to check it at scale:

<details>
<summary> Success — 248 new assets (URL signing splits into 3 batches: 100 / 100 / 50)</summary>

```
✔ Exported bundle(s)
✔ Uploaded 2 app bundles
✔ Uploaded 248 assets (reused 38 assets)
✔ Published!
```

</details>

<details>
<summary> Failure — forced finalization to never succeed (re-uploads 3×, then fails loudly)</summary>

```
✔ Exported bundle(s)
✔ Uploaded 2 app bundles
Update not published: 6 asset(s) uploaded but were not processed by the server after 3 attempts:
- assets/stress/img000.png
- assets/stress/img001.png
  …
The upload succeeded, but the server did not finish processing these assets in time.
This is usually temporary. Re-run `eas update` to try again, and contact support if it keeps happening.
    Error: update command failed.
```

</details>


packages/eas-cli/src/project/publish.ts
```diff
@@ -578,6 +579,16 @@ export async function uploadAssetsAsync(
     ); // linear backoff
     missingAssets = await filterOutAssetsThatAlreadyExistAsync(graphqlClient, missingAssets);
     missingAssetStorageKeys = new Set(missingAssets.map(a => a.storageKey));
+
+    if (missingAssets.length > 0 && Date.now() - uploadedAt > 55_000) {
```
quinlanj: add a comment here to explain the significance of 55_000 (why did we choose that)

gwdp: Updated, thank you

quinlanj (approved):
🚀. cc: @wschurman since you are the author for most of the asset logic here

packages/eas-cli/src/project/publish.ts
```diff
@@ -516,80 +518,117 @@ export async function uploadAssetsAsync(
...
-      uniqueAssets.map(asset => ({
-        asset,
-        finished: !missingAssetStorageKeys.has(asset.storageKey),
-      }))
-    );
+  const assetLimitPromise = PublishQuery.getAssetLimitPerUpdateGroupAsync(graphqlClient, projectId);
+
+  // Re-upload assets that don't finalize in time to trigger a fresh finalize event and retry window.
+  let pendingAssets = missingAssets;
+  for (let attempt = 0; attempt < MAX_UPLOAD_ATTEMPTS && pendingAssets.length > 0; attempt++) {
```
wschurman, line 614: What's the interaction between this and the activity timeout mechanism at the callsite?

packages/eas-cli/src/project/publish.ts
```diff
@@ -516,80 +518,117 @@ export async function uploadAssetsAsync(
...
-    missingAssetStorageKeys = new Set(missingAssets.map(a => a.storageKey));
-    await timeoutPromise; // await after filterOutAssetsThatAlreadyExistAsync for easy mocking with jest.runAllTimers
-    timeout += 1;
-    onAssetUploadResultsChanged?.(
-      uniqueAssets.map(asset => ({
-        asset,
-        finished: !missingAssetStorageKeys.has(asset.storageKey),
-      }))
-    );
+  const assetLimitPromise = PublishQuery.getAssetLimitPerUpdateGroupAsync(graphqlClient, projectId);
```
wschurman: nit: move to L630 or even just parallelize in the Promise.all call containing assetsWithStorageKey

```
const [assetLimitPerUpdateGroup, assetsWithStorageKey] = await Promise.all([
    PublishQuery.getAssetLimitPerUpdateGroupAsync(graphqlClient, projectId),
    Promise.all(
    assets.map(async asset => {
      return {
        ...asset,
        storageKey: await getStorageKeyForAssetAsync(asset),
      };
    })
  )]);
```

gwdp: Good call :bow: Parallelized it as you suggested

packages/eas-cli/src/project/publish.ts
gwdp: it's complementary. The new code only handles the finalize-event-aging case (re-upload mints a fresh event, up to 3×). Preferred to keep the 180s watchdog untouched, so it still backstops everything else, like a hung request. Let me know if you have different thoughts!

wschurman (approved):

gwdp:
Thanks @wschurman , was sorting out CI before re-asking review :bow:

gwdp:
Blocked by https://github.com/expo/eas-cli/pull/3938; Will rebase once that is merged

PR #3925, opened 2026-07-01
[build-tools] Apply npm cache registry only to the dependency install

AbbanMustafa:
# Why

#3855 started exporting NPM_CONFIG_REGISTRY into the global build env so installs would resolve packages through our internal npm cache. Because it was a global env var, every npm invocation in a build (workflow job steps, custom steps, and internal npm -g install … calls during environment setup) routed through the cache, but only the managed prebuild install had a fallback. When the killswitch was disabled (expo/universe#28199), this unguarded load overwhelmed the Verdaccio instance and impacted jobs.

# How

- Removed the global NPM_CONFIG_REGISTRY export from worker/src/env.ts (the line added in #3855)
   - EAS_BUILD_NPM_CACHE_URL remains as the signal carrying the resolved cache URL into the build.
- `installDependenciesWithNpmCacheFallbackAsync` now reads the URL from EAS_BUILD_NPM_CACHE_URL (gated on EAS_USE_NPM_CACHE=1) and injects NPM_CONFIG_REGISTRY only into that install's spawn env. On a cache-related failure it retries with the base env (default registry)
- After this change, unrelated npm invocations no longer hit the cache; only the guarded install does, and it still falls back if the cache is unavailable.

# Test Plan

- Unit tests in installDependencies.test.ts cover: registry applied per-spawn on the first attempt, retry without the registry on a cache failure, non-fatal cache-registry error reporting, no retry on unrelated failures, and no cache use when EAS_USE_NPM_CACHE is not enabled.
- worker/src/__unit__/env.test.ts asserts NPM_CONFIG_REGISTRY is no longer present in the global build env while EAS_BUILD_NPM_CACHE_URL still is — the regression guard proving the leak is closed.

### DIFF
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -109,7 +109,6 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
     return;
   }
 
-  const cacheEnv = { ...env, NPM_CONFIG_REGISTRY: npmCacheUrl };
   let firstErrorLine: string | undefined;
   let errorLineCount = 0;
 
@@ -117,7 +116,7 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
     await (
       await installDependenciesAsync({
         packageManager,
-        env: cacheEnv,
+        env: { ...env, NPM_CONFIG_REGISTRY: npmCacheUrl },
         logger,
         infoCallbackFn,
         lineTransformer: (line: string) => {
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -93,7 +93,7 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
   infoCallbackFn?: SpawnOptions['infoCallbackFn'];
   useFrozenLockfile: boolean;
 }): Promise<void> {
-  const npmCacheUrl = env.EAS_USE_NPM_CACHE === '1' ? env.EAS_BUILD_NPM_CACHE_URL : undefined;
+  const npmCacheUrl = env.EAS_BUILD_NPM_CACHE_URL;
 
   if (!npmCacheUrl) {
     await (
@@ -108,6 +108,8 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
     ).spawnPromise;
     return;
   }
+
+  logger.info(`Installing dependencies using the npm cache registry (${npmCacheUrl}).`);
 
   let firstErrorLine: string | undefined;
   let errorLineCount = 0;
--- a/packages/build-tools/src/common/installDependencies.ts
+++ b/packages/build-tools/src/common/installDependencies.ts
@@ -93,7 +93,23 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
   infoCallbackFn?: SpawnOptions['infoCallbackFn'];
   useFrozenLockfile: boolean;
 }): Promise<void> {
-  const npmCacheUrl = env.NPM_CONFIG_REGISTRY;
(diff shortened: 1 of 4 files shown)


packages/worker/src/env.ts
```diff
@@ -43,7 +43,6 @@ export function getBuildEnv({
   const cocoapodsCacheUrl = RuntimeSettings.getCocoapodsCacheUrl();
 
   setEnv(env, 'NPM_CACHE_URL', npmCacheUrl);
```
sjchmiela: should we also remove this?

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -93,19 +93,35 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
...
+        logger,
+        infoCallbackFn,
+        cwd,
+        useFrozenLockfile,
+      })
+    ).spawnPromise;
+    return;
+  }
+
+  const cacheEnv = { ...env, NPM_CONFIG_REGISTRY: npmCacheUrl };
```
sjchmiela: i think we can inline this

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -93,19 +93,35 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
   infoCallbackFn?: SpawnOptions['infoCallbackFn'];
   useFrozenLockfile: boolean;
 }): Promise<void> {
-  const npmCacheUrl = env.NPM_CONFIG_REGISTRY;
+  const npmCacheUrl = env.EAS_USE_NPM_CACHE === '1' ? env.EAS_BUILD_NPM_CACHE_URL : undefined;
+
```
sjchmiela, line 97: can we add a helpful log like https://github.com/expo/eas-cli/blob/cc49343b39add8db67a9b3301ae02423a3823c4c/packages/build-tools/src/ios/pod.ts#L50 for every situation?

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -93,19 +93,35 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
   infoCallbackFn?: SpawnOptions['infoCallbackFn'];
   useFrozenLockfile: boolean;
 }): Promise<void> {
-  const npmCacheUrl = env.NPM_CONFIG_REGISTRY;
+  const npmCacheUrl = env.EAS_USE_NPM_CACHE === '1' ? env.EAS_BUILD_NPM_CACHE_URL : undefined;
```
sjchmiela: i think runtime settings already check for EAS_USE_NPM_CACHE?

packages/build-tools/src/common/installDependencies.ts
```diff
@@ -93,19 +93,35 @@ export async function installDependenciesWithNpmCacheFallbackAsync({
...
         packageManager,
-        env,
+        env: cacheEnv,
         logger,
         infoCallbackFn,
         lineTransformer: (line: string) => {
-          if (isNpmCacheRegistryErrorLine(line, { env, npmCacheUrl })) {
+          if (isNpmCacheRegistryErrorLine(line, npmCacheUrl)) {
             firstErrorLine ??= line;
             errorLineCount += 1;
```
sjchmiela, line 127: now that i think about it maybe it would have been nice to kill this promise if registry error line is detected? maybe too smart

sjchmiela (changes requested):

packages/build-tools/src/common/installDependencies.ts
AbbanMustafa: I will need play around with this a bit more, but yes it could be better to "be smart" and fail eagerly??

sjchmiela (approved):

PR #3927, opened 2026-07-01
[eas-cli] Fix Free plan users being falsely warned they hit their bui…

sarahlane8:
…ld limit

The "over" usage tier fired whenever the summed value of any overageMetrics row was greater than zero, independent of the plan limit or whether a charge was actually incurred. For Free plans that message collapsed into the "you've reached your limit, upgrade to keep building" warning, so free users well below their limit (e.g. 6 builds) were told they were blocked.

Free plans can never be in an "over" state (they are blocked at the limit and never charged pay-as-you-go), and even paid plans now require a real billable overage cost (totalCost > 0) before showing the overage warning.

# Why

# How

# Test Plan

### DIFF
--- a/packages/eas-cli/src/utils/usage/checkForOverages.ts
+++ b/packages/eas-cli/src/utils/usage/checkForOverages.ts
@@ -35,16 +35,18 @@ export async function maybeWarnAboutUsageOveragesAsync({
 
     const overageCount = buildMetrics.overageMetrics.reduce((sum, o) => sum + o.value, 0);
     const overageCostCents = buildMetrics.totalCost;
+    const hasFreePlan = subscription.name === 'Free';
     const tier = classifyUsageTier({
       planValue: planMetric.value,
       limit: planMetric.limit,
       overageCount,
+      overageCostCents,
+      hasFreePlan,
     });
     if (!tier) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


LipLoody (approved):
Sound Good, I've check how `overageMetrics` is computed and it should work fine

packages/eas-cli/src/utils/usage/checkForOverages.ts
```diff
@@ -64,12 +66,22 @@ export function classifyUsageTier({
...
+  // Only paid plans can incur pay-as-you-go overages, and only when there is an
+  // actual billable cost. Free plans are blocked at the limit and are never charged,
+  // so they can never be in an "over" state — a nonzero overage row (e.g. a usage
+  // rollup returned by the API) must not be treated as billable usage. Without this
+  // guard, free users well below their limit were shown a "you've reached your limit,
+  // upgrade to keep building" warning.
+  if (!hasFreePlan && overageCount > 0 && overageCostCents > 0) {
     return 'over';
   }
   if (limit > 0 && planValue >= limit) {
```
LipLoody, line 87: Free limits are per-platform (15 iOS / 15 Android), but the at check only looks at the total (30)

LipLoody (approved):
Can be fix in follow up PR

PR #3928, opened 2026-07-01
[eas-build-job] Add local composite function schema

sswrk:
# Why

First step toward local composite functions for EAS Workflows: reusable step groups referenced via `uses:` in `.eas/workflows/*.yml`.

This PR adds the shared contract in `@expo/eas-build-job` so all consumers agree on composite function config shape.

# How

Introduce the composite function config schema in `@expo/eas-build-job`, the single source of truth for the shape of a `function.yml` file. No runtime behavior changes yet.

Added:
- `CompositeFunctionConfigZ` / `CompositeFunctionConfig`: Zod schema and type for `function.yml` (inputs, outputs, runs.steps)
- `CompositeFunctionCatalog`: map of normalized path → config

# Test Plan

Added unit tests for `CompositeFunctionConfigZ`.

PR #3929, opened 2026-07-01
[steps][2/4] Expand composite function references into build steps

sswrk:
# Why

EAS workflows need reusable local composite functions: a step like `uses: ./.eas/functions/setup` should run the steps defined in `function.yml`, not be treated as a built-in function.

This PR adds the structural layer for that: turning a composite function call into concrete workflow steps with correct ids, scoping, and control flow. Input interpolation, outputs, and loading `function.yml` from disk are handled in separate PRs.

# How

When `StepsConfigParser` sees a local `uses:` path (`./` or `../`), `CompositeFunctionExpander` replaces the call with a flat list of inner steps:

- ID namespacing: inner step `read` under caller `setup` becomes `setup__read`
- Each expanded step carries a `BuildStepCompositeFunctionScope` so `${{ steps.* }}` and `${{ inputs.* }}` resolve against composite-function-local names
- Nested local composite functions are expanded recursively with cycle and depth checks
- Local composite functions are rejected in hooks

# Test Plan

Added expansion unit tests covering nesting, cycles, depth limits, and id collisions.

### DIFF
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -28,7 +28,7 @@ import { evaluateIfCondition as evaluateIfConditionExpression } from './utils/js
 import { BIN_PATH } from './utils/shell/bin';
 import { getShellCommandAndArgs } from './utils/shell/command';
 import { spawnAsync } from './utils/shell/spawn';
-import { interpolateWithInputs, interpolateWithOutputs, parseOutputPath } from './utils/template';
+import { interpolateWithInputs, interpolateWithOutputs } from './utils/template';
 
 export enum BuildStepStatus {
   NEW = 'new',
@@ -518,8 +518,7 @@ export class BuildStep extends BuildStepOutputAccessor {
   }
 
   private getLegacyStepOutputValue(path: string): string {
-    const { stepId, outputId } = parseOutputPath(path);
-    return this.ctx.global.getStepOutputValue(`steps.${stepId}.${outputId}`) ?? '';
+    return this.ctx.global.getStepOutputValue(path) ?? '';
   }
 
   private async collectAndValidateOutputsAsync(outputsDir: string): Promise<void> {
--- a/packages/steps/src/BuildStep.ts
+++ b/packages/steps/src/BuildStep.ts
@@ -136,8 +136,6 @@ export class BuildStep extends BuildStepOutputAccessor {
   public readonly stepEnvOverrides: BuildStepEnv;
   public readonly ifCondition?: string;
   public readonly compositeFunctionScope?: BuildStepCompositeFunctionScope;
-  /** Prefixed expansion steps; top-level composite function outputs stay public. */
-  public readonly isCompositeFunctionInternal: boolean;
   public readonly timeoutMs?: number;
   public readonly __metricsId?: string;
   public readonly __hookId?: HookAnchorId;
@@ -167,7 +165,6 @@ export class BuildStep extends BuildStepOutputAccessor {
       env,
       ifCondition,
       compositeFunctionScope,
-      isCompositeFunctionInternal,
       timeoutMs,
       __metricsId,
       __hookId,
@@ -184,7 +181,6 @@ export class BuildStep extends BuildStepOutputAccessor {
       env?: BuildStepEnv;
       ifCondition?: string;
       compositeFunctionScope?: BuildStepCompositeFunctionScope;
-      isCompositeFunctionInternal?: boolean;
       timeoutMs?: number;
       __metricsId?: string;
       __hookId?: HookAnchorId;
@@ -206,7 +202,6 @@ export class BuildStep extends BuildStepOutputAccessor {
     this.shell = shell ?? '/bin/bash -eo pipefail';
     this.ifCondition = ifCondition;
     this.compositeFunctionScope = compositeFunctionScope;
-    this.isCompositeFunctionInternal = isCompositeFunctionInternal ?? false;
     this.timeoutMs = timeoutMs;
     this.__metricsId = __metricsId;
     this.__hookId = __hookId;
@@ -222,15 +217,24 @@ export class BuildStep extends BuildStepOutputAccessor {
     this.outputsDir = getTemporaryOutputsDirPath(ctx, this.id);
     this.envsDir = getTemporaryEnvsDirPath(ctx, this.id);
 
+    this.registerSelf(ctx);
+  }
+
+  /**
+   * Consulted by registerSelf() in the constructor; overrides may only read state
+   * assigned before that call (outputById, compositeFunctionScope).
+   */
+  public get isCompositeFunctionInternal(): boolean {
+    return this.compositeFunctionScope !== undefined;
+  }
+
+  protected registerSelf(ctx: BuildStepGlobalContext): void {
     ctx.registerStep(this);
   }
 
   public async executeAsync(): Promise<void> {
     try {
-      this.ctx.logger.info(
-        { marker: BuildStepLogMarker.START_STEP },
-        `Executing build step "${this.displayName}"`
-      );
+      this.logStepStart();
       this.status = BuildStepStatus.IN_PROGRESS;
 
       await fs.mkdir(this.outputsDir, { recursive: true });
@@ -279,10 +283,7 @@ export class BuildStep extends BuildStepOutputAccessor {
         await executionPromise;
       }
 
-      this.ctx.logger.info(
-        { marker: BuildStepLogMarker.END_STEP, result: BuildStepStatus.SUCCESS },
-        `Finished build step "${this.displayName}" successfully`
-      );
+      this.logStepSuccess();
       this.status = BuildStepStatus.SUCCESS;
     } catch (err) {
       // Downstream error handling relies on real Errors; wrap non-Error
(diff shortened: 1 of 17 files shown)


sswrk:
* **#3929** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3929?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4013](https://github.com/expo/eas-cli/pull/4013) <a href="https://app.graphite.com/github/pr/expo/eas-cli/4013?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4014](https://github.com/expo/eas-cli/pull/4014) <a href="https://app.graphite.com/github/pr/expo/eas-cli/4014?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3929?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3956** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sswrk:
@claude review

packages/steps/src/StepsConfigParser.ts
```diff
@@ -113,14 +133,30 @@ export class StepsConfigParser extends AbstractConfigParser {
...
       }
-      const maps = { buildFunctionById, buildFunctionGroupById };
       const before = this.constructHookSideEntries(anchorId, 'before', validatedHooks, maps);
-      const anchorStep = this.createBuildStepFromNonGroupStepConfig(stepConfig, buildFunctionById);
+      const createdSteps = this.createBuildStepsFromNonGroupStepConfig(
+        stepConfig,
+        maps,
+        compositeFunctionExpander
+      );
+      if (createdSteps.length !== 1) {
```
hSATAC: The multi-step case is rejected here, but a single-step composite slips through — the anchor comes from the call step's `__hook_id` stamp before expansion, so with `createdSteps.length === 1` we still wrap hooks around the expanded inner step (the `__hookId` stripping doesn't catch it, since the stamp's on the call, not the expanded step).

Only hits if someone hand-writes `__hook_id` on a `uses: ./...` step, so not a real hole — just a bit inconsistent with the guard right above. Maybe reject any anchor on a composite path regardless of count? Or if single-step's intentional, a comment + reworded error would do it.

packages/steps/src/BuildStep.ts
```diff
@@ -438,25 +483,43 @@ export class BuildStep extends BuildStepOutputAccessor {
...
+    return skipLegacyOutputInterpolation
+      ? interpolatedWithInputsAndGlobalContext
+      : interpolateWithOutputs(interpolatedWithInputsAndGlobalContext, path =>
+          this.getLegacyStepOutputValue(path)
+        );
+  }
+
+  private getLegacyStepOutputValue(path: string): string {
+    const { stepId, outputId } = parseOutputPath(path);
+    return this.ctx.global.getStepOutputValue(`steps.${stepId}.${outputId}`) ?? '';
```
hSATAC: Tiny one: `parseOutputPath` splits the path, then this line rebuilds the same `steps.x.y` string and `getStepOutputValue` parses it again — the callback already gets the full path, so `path => this.ctx.global.getStepOutputValue(path) ?? ''` does the same. Helper's still worth it for the two call sites though.

hSATAC (approved):
Nice, this looks solid — expansion, namespacing, cycle/depth, the `if` gating all read well. Two small notes inline, nothing blocking.

packages/steps/src/StepsConfigParser.ts
sswrk: Fixed not rejecting the single-step composite in https://github.com/expo/eas-cli/pull/3929/commits/80a029725b33e7978017f5c1a1b2aa2d116eeb5b

packages/steps/src/BuildStep.ts
sswrk: Ohh, that's a nice simplification. Changed that.

packages/steps/src/utils/localCompositeFunctions.ts
```diff
@@ -1,7 +1,44 @@
...
+      `Local composite function path "${trimmed}" must not contain interpolation ("\${{ ... }}"). The "uses" path for a local composite function must be a static, literal path.`
+    );
+  }
+  if (trimmed.includes('\\')) {
+    throw new BuildConfigError(
+      `Local composite function path "${trimmed}" must not contain backslashes. Use forward slashes as path separators.`
+    );
+  }
+  const normalized = path.posix.normalize(trimmed.replace(/\/+$/, ''));
+  if (normalized === '.' || normalized === '..') {
```
sjchmiela: Well, theoretically I guess someone could make it so? Does it make sense to prevalidate like this?

packages/steps/src/hooks.ts
```diff
@@ -125,6 +126,11 @@ export function constructHookEntriesFromValidatedSteps(
       });
       continue;
     }
+    if (isCompositeFunctionPath(step.uses)) {
+      throw new BuildConfigError(
+        `Local composite function steps ("uses: ${step.uses}") are not supported in hooks.`
```
sjchmiela, line 131: It might be useful to support this in the future

packages/steps/src/StepsConfigParser.ts
```diff
@@ -214,36 +256,52 @@ export class StepsConfigParser extends AbstractConfigParser {
...
-  private createBuildStepFromFunctionStepConfig(
+  private createBuildStepsFromFunctionStepConfig(
     step: FunctionStep,
-    buildFunctionById: BuildFunctionById
-  ): BuildStep {
+    { buildFunctionById }: FunctionMaps,
+    compositeFunctionExpander: CompositeFunctionExpander
+  ): BuildStep[] {
+    if (isCompositeFunctionPath(step.uses)) {
+      return compositeFunctionExpander.expandCompositeFunctionStep(
```
sjchmiela: I wonder if the other way to handle composite actions, i.e. to have a tree of BuildSteps wouldn't be easier, also considering hooks. I don't have a concrete proposal and this pull request is good, just wondering. What I mean by a tree would be sort of like

- buildstep checkout
  - after hooks
    - buildstep custom composite action
      - buildstep 1
      - buildstep 2
- buildstep usenpmrc
- buildstep another composite
  - buildstep x

I think BuildWorkflow being a separate thing and the only thing capable of executing steps requires us to pre-parse everything to a single array which is difficult. Not saying BuildStep should always have a BuildStep children array, but maybe a subclass, like CompositeBuildStep should.

I just wish things were simpler… thank you to both of you for digging here

sjchmiela (approved):

packages/steps/src/StepsConfigParser.ts
sswrk: I support the idea of making the code design reflect the domain as much as possible. But since this is a brownfield codebase, I'm playing it safe, doing additive changes and not interfering with the core. Of course, we should also avoid shipping something that needs refactor from day one, but I think for V1 the balance is fine.

I made one small change towards this direction, that was a low-hanging fruit refactor. See https://github.com/expo/eas-cli/pull/3929/commits/c50f0a3c6ff064e86c533c24edcff5c52b077fce -- this introduces the `CompositeBuildStep` in the parsing phase.

packages/steps/src/hooks.ts
sswrk: Coming soon https://github.com/expo/eas-cli/pull/4062

packages/steps/src/StepsConfigParser.ts
sjchmiela: wow!!

packages/steps/src/utils/localCompositeFunctions.ts
```diff
@@ -1,7 +1,44 @@
+import path from 'path';
+
+import { BuildConfigError } from '../errors';
+
 // Local composite functions referenced via `uses: ./path` or `uses: ../path` in EAS workflows.
 // Not supported in `.eas/build/*.yml` custom build configs.
 
+const JOB_CONTEXT_INTERPOLATION_REGEXP = /\$\{\{(.+?)\}\}/;
+
+function localCompositeFunctionPathIsInterpolated(uses: string): boolean {
```
sjchmiela: nit: `doesLocalCompositeFunctionPathRequireInterpolation` (i think putting the questioning verb at the beginning makes the purpose clearer)

packages/steps/src/CompositeBuildStep.ts
```diff
@@ -0,0 +1,77 @@
...
+      scope: BuildStepCompositeFunctionScope;
+      children: BuildStep[];
+    }
+  ) {
+    super(ctx, {
+      id,
+      displayName,
+      outputs: [],
+      fn: async () => {},
+      ifCondition: '${{ always() }}',
```
sjchmiela, line 32: worth a comment haha

sjchmiela (approved):

packages/steps/src/utils/localCompositeFunctions.ts
sswrk: Renamed.

packages/steps/src/utils/localCompositeFunctions.ts
sswrk: No technical need for this. Relaxed it.

PR #3932, opened 2026-07-01
[eas-cli] Show accurate build overage count in pre-build usage warning

sarahlane8:
The overage count shown to paid users ("N builds beyond your included credits") summed the value of every overageMetrics row, including rows without build metadata (e.g. rollup/summary rows), which could inflate the number. Filter to genuine per-worker build overage rows (those carrying billingResourceClass + platform metadata) before summing.

Fetch that metadata on overageMetrics in the warning query so the rows can be distinguished.

### DIFF
--- a/packages/eas-cli/src/graphql/queries/AccountQuery.ts
+++ b/packages/eas-cli/src/graphql/queries/AccountQuery.ts
@@ -201,6 +201,12 @@ export const AccountQuery = {
                       overageMetrics {
                         id
                         value
+                        metadata {
+                          ... on AccountUsageEASBuildMetadata {
+                            billingResourceClass
+                            platform
+                          }
+                        }
                       }
                       totalCost
                     }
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/graphql/queries/AccountQuery.ts
```diff
@@ -201,6 +201,12 @@ export const AccountQuery = {
                       overageMetrics {
                         id
                         value
+                        metadata {
```
douglowder, line 204: Should this change be done server-side, so that the correct overage count is also available on the web site?

sarahlane8: Checked the website — it isn't affected. It computes build overage from the Stripe line-item quantity (get-current-usage-data.ts:88-90), not by summing overageMetrics, so it doesn't have this bug. This PR is a CLI-only fix.

The row that was inflating the CLI count is the fast-failed waived-builds row (no platform/billingResourceClass), which this PR's filter excludes.

We could move this server-side, but there's a tradeoff: the waived row can't just be removed from overageMetrics, because the website reads it from there for its waived-builds display. So the clean server-side option is to add a dedicated server-computed overage count field that the CLI reads instead — non-breaking, and no website change needed.

What do you think? Double check my thoughts!

douglowder: Thanks for the explanation! Let's go ahead and merge this change in the CLI, and later we can investigate a correct unified server-side solution.

douglowder (approved):

LipLoody (approved):

PR #3938, opened 2026-07-02
[eas-cli] Regenerate GraphQL schema and fix affected types

gwdp:
# Why

The committed GraphQL schema drifted from the server, so the "schema up-to-date" check is red on `main` and every open PR.

# How

Ran `yarn generate-graphql-code` to resync, then fixed the two spots that stopped typechecking against the new schema: narrowed a query's return type to the fields it selects, and handled a now-nullable ASC field.

# Test Plan

CI passes.

### DIFF
--- a/packages/eas-cli/graphql.schema.json
+++ b/packages/eas-cli/graphql.schema.json
@@ -420,6 +420,22 @@
             "deprecationReason": null
           },
           {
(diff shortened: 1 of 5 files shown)


sjchmiela (approved):
Ooo, sorry!!

PR #3939, opened 2026-07-02
[eas-cli] Upload agent-device session artifacts

szdziedzic:
# Why

Agent-device now exposes daemon artifact inventory and download endpoints, and EAS remote sessions should upload those artifacts to DeviceRunSession artifacts the same way Argent sessions do. This lets artifacts produced during an agent-device remote session become available through the existing EAS artifact download UI/API instead of staying only on the VM.

Release constraint: the EAS VM installs `agent-device@latest` unless `package_version` is provided. npm `agent-device@latest` is currently `0.18.2`, which predates the landed artifact endpoint changes in callstack/agent-device#1020 and callstack/agent-device#1027. This PR wires EAS CLI support, but a real run needs an agent-device release that contains those endpoints, or the step must be run with such a package version once available. If the installed daemon does not expose `/artifacts`, EAS logs once and disables artifact uploads for that session.

# How

- Start agent-device daemon with `AGENT_DEVICE_RETAIN_ARTIFACTS=1` so the local remote client and the VM uploader can both download the same artifacts.
- After the remote config is reported and the DeviceRunSession is in progress, poll the local daemon over `http://127.0.0.1:<daemonPort>/artifacts` using the daemon token, independent of the public ngrok tunnel.
- Download each new artifact by ID and upload it through the existing DeviceRunSession artifact upload helper.
- Upload artifacts sequentially per poll cycle so the poller cannot spawn unbounded simultaneous disk/network work. Successfully uploaded IDs are tracked so failed uploads can retry on the next poll.
- Update the git fallback clone URL to `https://github.com/callstack/agent-device.git`.
- Add focused tests for listing, unsupported inventory, uploading, and upload retry behavior.

# Test Plan

- `corepack yarn workspace @expo/build-tools jest-unit packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/argentArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/deviceRunSessionArtifacts.test.ts`
- `corepack yarn oxfmt --check packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts`
- `corepack yarn oxlint --type-aware --import-plugin --tsconfig ./tsconfig.oxlint.json packages/build-tools/src/steps/functions/startAgentDeviceRemoteSession.ts packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts`
- `corepack yarn workspace eas-cli typecheck`
- `corepack yarn workspace @expo/build-tools typecheck`
- `corepack yarn typecheck`

### DIFF
--- a/packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
+++ b/packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
@@ -1,0 +1,191 @@
+import { SystemError } from '@expo/eas-build-job';
+import { type bunyan } from '@expo/logger';
+import { createReadStream, createWriteStream } from 'node:fs';
+import { mkdtemp, rm, stat } from 'node:fs/promises';
+import fetch from 'node-fetch';
+import os from 'node:os';
+import path from 'node:path';
+import { pipeline } from 'node:stream/promises';
+import { z } from 'zod';
+
+import { CustomBuildContext } from '../../customBuildContext';
+import { Sentry } from '../../sentry';
+import { formatBytes } from '../../utils/artifacts';
+import { sleepAsync } from '../../utils/retry';
+import { uploadDeviceRunSessionArtifactAsync } from './deviceRunSessionArtifacts';
+
+const AGENT_DEVICE_ARTIFACT_UPLOAD_POLL_INTERVAL_MS = 5_000;
+
+const AgentDeviceArtifactSchema = z.object({
+  id: z.string(),
+  filename: z.string(),
+});
+const AgentDeviceArtifactsListResponseSchema = z.object({
+  artifacts: z.array(AgentDeviceArtifactSchema),
+});
+
+export type AgentDeviceArtifact = z.infer<typeof AgentDeviceArtifactSchema>;
+
+class AgentDeviceArtifactsUnsupportedError extends SystemError {
+  constructor() {
+    super('agent-device daemon does not expose artifact inventory.');
+  }
+}
+
+export async function pollAgentDeviceArtifactsForUploadAsync(
+  ctx: CustomBuildContext,
+  {
+    deviceRunSessionId,
+    daemonUrl,
+    daemonToken,
+    logger,
+  }: {
+    deviceRunSessionId: string;
+    daemonUrl: string;
+    daemonToken: string;
+    logger: bunyan;
+  }
+): Promise<void> {
+  logger.info('Started polling agent-device daemon for artifacts.');
+  const uploadedArtifactIds = new Set<string>();
+  let listArtifactsErrorCount = 0;
+
+  for (;;) {
+    try {
+      const artifacts = await listAgentDeviceArtifactsAsync({ daemonUrl, daemonToken });
+      listArtifactsErrorCount = 0;
+      for (const artifact of artifacts) {
+        if (uploadedArtifactIds.has(artifact.id)) {
+          continue;
+        }
+        try {
+          await uploadAgentDeviceArtifactAsync(ctx, {
+            deviceRunSessionId,
+            daemonUrl,
+            daemonToken,
+            artifact,
+            logger,
+          });
+          uploadedArtifactIds.add(artifact.id);
+        } catch (err) {
+          const error = err instanceof Error ? err : new Error(String(err));
+          Sentry.capture('Could not upload agent-device remote session artifact', error);
+          logger.warn({ err: error }, 'Could not upload agent-device remote session artifact.');
+        }
+      }
+    } catch (err) {
+      if (err instanceof AgentDeviceArtifactsUnsupportedError) {
+        logger.warn(
+          'agent-device daemon does not expose artifact inventory; remote session artifact uploads are disabled.'
+        );
+        return;
+      }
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
```diff
@@ -0,0 +1,195 @@
...
+    daemonToken: string;
+    artifact: AgentDeviceArtifact;
+    logger: bunyan;
+  }
+): Promise<void> {
+  logger.info(`Downloading artifact ${artifact.filename}.`);
+  const temporaryDirectory = await mkdtemp(path.join(os.tmpdir(), 'agent-device-artifact-'));
+  try {
+    const temporaryArtifactPath = path.join(temporaryDirectory, path.basename(artifact.filename));
+    await downloadAgentDeviceArtifactToFileAsync({
```
sjchmiela, line 142: not sure if worth it, but maybe agent-device _does_ expose a "host path" which we could use and skip the download step. argent did not and also it promises to clean artifacts up with no RETAIN flag which is why i added download there. not that it matters, just noting

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
```diff
@@ -0,0 +1,195 @@
...
+          Sentry.capture('Could not upload agent-device remote session artifact', error);
+          logger.warn({ err: error }, 'Could not upload agent-device remote session artifact.');
+        }
+      }
+    } catch (err) {
+      if (err instanceof AgentDeviceArtifactsUnsupportedError) {
+        logger.warn(
+          'agent-device daemon does not expose artifact inventory; remote session artifact uploads are disabled.'
+        );
+        await new Promise<never>(() => {});
```
sjchmiela: what about… `return;` haha

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
```diff
@@ -0,0 +1,195 @@
...
+  createdAt: z.string(),
+  expiresAt: z.string(),
+});
+const AgentDeviceArtifactsListResponseSchema = z.object({
+  artifacts: z.array(AgentDeviceArtifactSchema),
+});
+
+export type AgentDeviceArtifact = z.infer<typeof AgentDeviceArtifactSchema>;
+
+class AgentDeviceArtifactsUnsupportedError extends Error {
```
sjchmiela: i… guess it's a systemerror in a way? i know it's a throwaway token

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
```diff
@@ -0,0 +1,195 @@
...
+
+const AGENT_DEVICE_ARTIFACT_UPLOAD_POLL_INTERVAL_MS = 5_000;
+
+const AgentDeviceArtifactSchema = z.object({
+  id: z.string(),
+  filename: z.string(),
+  mimeType: z.string(),
+  sizeBytes: z.number(),
+  createdAt: z.string(),
+  expiresAt: z.string(),
```
sjchmiela: i'd keep the required API as small as possible (probably don't need most of it)

sjchmiela (approved):

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
szdziedzic: Yep, much cleaner. I changed the unsupported artifact-inventory path to log once and return, and added a regression test for old daemons that do not expose /artifacts.

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
szdziedzic: I checked the merged agent-device contract. The inventory exposes downloadable artifacts, not a stable host path, and directory artifacts are archived by that endpoint. I kept the HTTP download path so EAS stays coupled to the public daemon contract rather than daemon-local storage details.

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
szdziedzic: Agreed. I changed the unsupported-inventory sentinel to extend SystemError. It still only drives the non-fatal compatibility branch where older agent-device versions do not expose /artifacts.

packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
szdziedzic: Good call. I narrowed the parsed daemon inventory to the fields EAS actually consumes: id and filename. The uploader still gets the authoritative byte size from the downloaded temp file before creating the DeviceRunSession artifact.

PR #3944, opened 2026-07-03
[build-tools] Add eas/posthog_flag_rollout workflow function

gwdp:
# Why

Manage a PostHog feature flag straight from an EAS workflow. Flip it on or ramp its rollout as part of a deploy, instead of doing it by hand in PostHog.

# How

Adds `eas/posthog_flag_rollout`. It uses the personal API key from `integrations:posthog:connect` (`POSTHOG_CLI_API_KEY`/`POSTHOG_CLI_PROJECT_ID`), or the `api_key`/`project_id` inputs. There's no `host` input on purpose, so the personal key only ever goes to `POSTHOG_CLI_HOST`.

The `rollout_percentage` input updates the flag's catch-all release condition and leaves targeted groups alone. Failures fail the step. Set `ignore_error: true` to log a warning and continue instead. Bad inputs, unknown flags and 403 scope errors always fail. Lookup and update aren't atomic, so a parallel dashboard edit can be lost.

# Test Plan

CI passes. Enabled a real flag from a local run with this workflow. It needs a personal key with feature-flag scopes. The last job fails the lookup on purpose and just warns.

<details>
<summary>Test workflow</summary>

```yaml
name: PostHog flag rollout test

jobs:
  enable_flag:
    steps:
      - uses: eas/posthog_flag_rollout
        with:
          flag: eas-workflow-test-flag
          active: true

  gradual_rollout:
    steps:
      - uses: eas/posthog_flag_rollout
        with:
          flag: eas-workflow-test-flag
          rollout_percentage: 50

  set_payload:
    steps:
      - uses: eas/posthog_flag_rollout
        with:
          flag: eas-workflow-test-flag
          active: true
          payload:
            welcome_message: hello from a workflow

  ignored_failure:
    steps:
      - uses: eas/posthog_flag_rollout
        with:
          flag: eas-workflow-test-flag
          active: true
          api_key: phx_invalid_local_test
          ignore_error: true
```
(Validate posthog side, feature flag is updated with enw values)

Screenshot 2026-07-03 at 10 21 39 AM
Screenshot 2026-07-03 at 11 49 00 AM

</details>

# Examples

These are simple workflows just to exercise the step. More fun and useful ones, mixed with EAS Update and deploys, are coming in the docs.

<details>
<summary>Ship a feature behind a flag</summary>

Publish the update, then enable the flag at 10% once the code is out.

```yaml
name: Ship a feature behind a PostHog flag

jobs:
  update:
    type: update
    params:
      branch: main
[...]

PR #3934, opened 2026-07-02
[build-tools] Add eas/posthog_capture_event workflow function

gwdp:
# Why

First PostHog workflow step: send an event from a workflow (e.g. mark a deploy).

# How

add `eas/posthog_capture_event` function; Uses the env vars `integrations:posthog:connect` already sets, or `api_key`/`host` inputs.

No `distinct_id` = anonymous event, so CI runs don't create person profiles. Failures fail the step; set `ignore_error: true` to log a warning and continue instead (same as `upload_artifact`).

# Test Plan

CI passes. Ran this workflow against a local — all three jobs green, both events visible in the project's PostHog activity (anonymous one under the `eas-workflow` person), and the ignored failure logs a warning without failing the step:

```yaml
name: PostHog capture test

jobs:
  anonymous_event:
    steps:
      - uses: eas/posthog_capture_event
        with:
          event: eas_workflow_test
          properties:
            source: coin-flip-workflow-test

  identified_event:
    steps:
      - uses: eas/posthog_capture_event
        with:
          event: eas_workflow_test_identified
          distinct_id: gwdp-local-test
          properties:
            source: coin-flip-workflow-test
            variant: identified

  ignored_failure:
    steps:
      - uses: eas/posthog_capture_event
        with:
          event: eas_workflow_test_ignored
          host: https://example.com
          ignore_error: true
```

### Workflow succeeds 
Screenshot 2026-07-02 at 4 51 42 PM

### Data appears on my posthog organization
Screenshot 2026-07-02 at 4 52 57 PM

PR #3945, opened 2026-07-03
[build-tools] Add eas/posthog_wait_for_metric workflow function

gwdp:
# Why

Gate an EAS workflow on a PostHog metric. Hold a rollout until errors stay low (or any metric clears) before it continues, instead of watching dashboards by hand. imo this is the coolest 🔥 

# How

Adds `eas/posthog_wait_for_metric`. It polls a [HogQL](https://posthog.com/docs/sql) `query` until `operator`/`threshold` is met, every `interval_seconds` (default 30) for up to `timeout_seconds` (default 600). Auth is the same personal key as the flag step, and the key needs the query read scope. (good docs coming)

It also sets a `value` output with the query result that met the threshold, so a later step can use `${{ steps.<step id>.outputs.value }}`.

There's no `ignore_error` here on purpose. It's a gate, so timing out fails the step. Network errors, 429 and 5xx retry until the deadline. Auth and query errors fail right away.

# Test Plan

CI passes. Test workflow below. The second and third jobs fail on purpose (timeout and bad query shape). Not live-run yet, since the key from connect lacks the query scope. Will run it before undrafting.

<details>
<summary>Test workflow</summary>

```yaml
name: PostHog wait-for-metric test

jobs:
  wait_cleared:
    steps:
      - uses: eas/posthog_wait_for_metric
        with:
          query: select count() from events
          operator: gte
          threshold: 0

  wait_timeout:
    steps:
      - uses: eas/posthog_wait_for_metric
        with:
          query: select count() from events
          operator: lt
          threshold: 0
          interval_seconds: 5
          timeout_seconds: 20

  wait_bad_query:
    steps:
      - uses: eas/posthog_wait_for_metric
        with:
          query: select event from events limit 1
          operator: gte
          threshold: 0
          interval_seconds: 5
          timeout_seconds: 15
```

Screenshot 2026-07-03 at 2 32 52 PM
Screenshot 2026-07-03 at 2 32 57 PM

</details>

# Examples

These are simple workflows just to exercise the step. More fun and useful ones, mixed with EAS Update and deploys, are coming in the docs.

<details>
<summary>Send an event and wait for it</summary>

Two jobs with no `needs`, so they run at the same time. One sends the event, the other waits for it to land. This needs real concurrency. It won't clear on a local runner set to concurrency 1, since the jobs run one after another there.

```yaml
name: Send an event and await it in parallel

jobs:
  send_ping:
    steps:
      - uses: eas/posthog_capture_event
        with:
          event: workflow_ping
          properties:
            source: parallel-demo

  await_ping:
    steps:
      - uses: eas/posthog_wait_for_metric
        with:
          query: select count() from events where event = 'workflow_ping' and timestamp > now() - interval 5 minute
          operator: gte
          threshold: 1
          interval_seconds: 10
          timeout_seconds: 300
```

</details>

<details>
<summary>Use the metric value in a later step</summary>

The gate exposes the reading it settled on as an output. A later step reads it with `${{ steps.gate.outputs.value }}`.

```yaml
name: Gate on a metric, then use its value

jobs:
  gated_announce:
    steps:
      - id: gate
        uses: eas/posthog_wait_for_metric
        with:
          query: select count() from events where event = '$exception' and timestamp > now() - interval 1 hour
          operator: lte
          threshold: 5
      - uses: eas/posthog_capture_event
        with:
          event: rollout_gate_cleared
          properties:
            error_count: ${{ steps.gate.outputs.value }}
```

</details>


gwdp:
* **#3949** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3948** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3947** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3947?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3945** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3945?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3944** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3944?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3934** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/utils/posthog.ts
```diff
@@ -1,8 +1,65 @@
...
+  }
+
+  if (response.status === 403) {
+    throw new Error(
+      'Running the PostHog query was forbidden (403). The personal API key likely lacks the "query:read" scope.'
+    );
+  }
+  if (response.status >= 500 || response.status === 429) {
+    logger.warn(`Running the PostHog query failed with status ${response.status}; retrying.`);
+    return undefined;
```
sjchmiela: should we maybe throw a typed error here so the caller knows it should wait?

packages/build-tools/src/steps/functions/waitForPostHogMetric.ts
```diff
@@ -0,0 +1,184 @@
...
+        required: false,
+      }),
+    ],
+    outputProviders: [BuildStepOutput.createProvider({ id: 'value', required: false })],
+    fn: async (stepCtx, { inputs, outputs, env }) => {
+      const { logger } = stepCtx;
+
+      const operator = inputs.operator.value as string;
+      if (!(operator in OPERATORS)) {
+        throw new Error(
```
sjchmiela: `UserError`

packages/build-tools/src/steps/functions/waitForPostHogMetric.ts
```diff
@@ -0,0 +1,184 @@
...
+          `Invalid "operator" "${operator}". Must be one of: ${Object.keys(OPERATORS).join(', ')}.`
+        );
+      }
+      const timeoutSeconds = inputs.timeout_seconds.value as number;
+      const intervalSeconds = inputs.interval_seconds.value as number;
+      if (timeoutSeconds <= 0 || intervalSeconds <= 0) {
+        throw new Error('"timeout_seconds" and "interval_seconds" must be greater than 0.');
```
sjchmiela: `UserError`

sjchmiela (approved):
i guess it would be even cooler if we could do it not-from-a-VM (`type: posthog-wait-for-metric`) and poll from a background job

packages/build-tools/src/steps/functions/waitForPostHogMetric.ts
gwdp: Updated :bow: Hopefully that sticks in my head now :sweat:

packages/build-tools/src/steps/functions/waitForPostHogMetric.ts
gwdp: Updated, ty

packages/build-tools/src/steps/utils/posthog.ts
gwdp: great call, that's cleaner but also dont retry non retryable errors, which I was doing before. Added `PosthogRetryableError` — the client throws it on network/5xx/429/bad-json and the poll loops catch it to keep waiting, while a 403/other are now fatal

gwdp:
> i guess it would be even cooler if we could do it not-from-a-VM (`type: posthog-wait-for-metric`) and poll from a background job

Tysm for the idea, it would be even cooler, tracking it under ENG-23097 as a follow-up

PR #3946, opened 2026-07-03
[build-tools] Wait for remote sessions to stop cleanly

sjchmiela:
## Summary

- replace the remote device session infinite waits with polling for the device run session to become `STOPPED`
- let agent-device, Argent, and serve-sim session steps return normally after the stop signal so the worker can report success and the launcher can close the VM
- make Argent artifact polling stoppable, drain one final artifact list on shutdown, and wait briefly for pending uploads

## Why

Stopping a device run session currently cancels the backing turtle job run, which tears down the VM immediately. That skips graceful shutdown work such as finishing recordings and uploading session artifacts. This change makes the worker observe the session stop state and finish cleanup before exiting.

## Validation

- `yarn oxfmt <modified files>`
- `yarn --cwd packages/build-tools test -- remoteDeviceRunSession`
- `yarn --cwd packages/build-tools test -- argentArtifacts`
- `yarn --cwd packages/build-tools typecheck`
- `git diff --check`

## Note

This relies on the backend/API behavior changing so `ensureDeviceRunSessionStopped` marks the device run session `STOPPED` without canceling the turtle job run, and allows artifact uploads while the stopped session is still cleaning up.

### DIFF
--- a/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
@@ -55,7 +55,7 @@ export function createStartArgentRemoteSessionBuildFunction(
         allowedValueTypeName: BuildStepInputValueTypeName.STRING,
       }),
     ],
-    fn: async ({ logger, global }, { inputs, env }) => {
+    fn: async ({ logger, global }, { inputs, env, signal }) => {
       // Fail fast before any expensive setup if the injected env
       // vars are missing: DEVICE_RUN_SESSION_ID (to report the remote config
       // back to the API server), EAS_SIMULATOR_NGROK_TUNNEL_DOMAIN (base domain
@@ -123,12 +123,15 @@ export function createStartArgentRemoteSessionBuildFunction(
       }
       logger.info(`Argent tool-server is listening on port ${toolServerPort}.`);
       const artifactPollAbortController = new AbortController();
+      const artifactPollSignal = signal
+        ? AbortSignal.any([signal, artifactPollAbortController.signal])
+        : artifactPollAbortController.signal;
       const artifactPollingPromise = pollArgentArtifactsForUploadAsync(ctx, {
         deviceRunSessionId,
         toolsUrl: `http://127.0.0.1:${toolServerPort}`,
         toolsAuthToken: toolServerToken,
         logger,
-        signal: artifactPollAbortController.signal,
+        signal: artifactPollSignal,
       });
 
       try {
@@ -170,6 +173,7 @@ export function createStartArgentRemoteSessionBuildFunction(
           ctx,
           deviceRunSessionId,
           logger,
+          signal,
         });
       } finally {
         artifactPollAbortController.abort();
--- a/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
@@ -24,6 +24,7 @@ import {
   startNgrokTunnelAsync,
   startServeSimWithTunnelAsync,
   uploadRemoteSessionConfigAsync,
+  waitForDeviceRunSessionStoppedAsync,
   waitForFileAsync,
 } from '../utils/remoteDeviceRunSession';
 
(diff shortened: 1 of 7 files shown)


packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -77,9 +86,29 @@ export async function pollArgentArtifactsForUploadAsync(
...
-    await sleepAsync(ARGENT_ARTIFACT_UPLOAD_POLL_INTERVAL_MS);
+  };
+
+  while (!signal.aborted) {
+    await listAndQueueArtifactUploadsAsync();
+    await sleepUntilAbortedAsync(ARGENT_ARTIFACT_UPLOAD_POLL_INTERVAL_MS, signal);
   }
+
+  logger.info('Argent artifact polling stopped; draining remaining artifacts.');
+  await listAndQueueArtifactUploadsAsync();
```
szdziedzic, line 120: The drain awaits `listArgentArtifactsAsync`, whose fetch has no timeout and no signal (node-fetch v2 defaults to no timeout), and the 30s `ARGENT_ARTIFACT_UPLOAD_CLEANUP_TIMEOUT_MS` only bounds `waitForPendingUploadsAsync` below. If the tool-server accepts the connection but never responds — plausible while it's dying at session end — this await never settles, the `finally` in `startArgentRemoteSession.ts` awaits `artifactPollingPromise` forever, and the job hangs indefinitely. Same for the in-loop list call on line 101, which also delays abort reaction until the in-flight request settles. Suggest a per-request timeout on the list/download fetches (e.g. AbortSignal.timeout(10_000)), or run the entire drain inside the same 30s race.

packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
```diff
@@ -120,51 +122,65 @@ export function createStartArgentRemoteSessionBuildFunction(
         );
       }
       logger.info(`Argent tool-server is listening on port ${toolServerPort}.`);
-      void pollArgentArtifactsForUploadAsync(ctx, {
+      const artifactPollAbortController = new AbortController();
```
szdziedzic, line 125: this is what Fable found: 
```
The steps framework already passes an AbortSignal into every step fn (BuildStep.ts:418-431; maestroTests.ts:150 uses it). If a step timeout ever fires, BuildStep abandons the fn promise and aborts its signal — this private controller is never aborted, so the artifact poller and the status poll keep hitting the network every 5s until process exit. Suggest destructuring signal from the fn args and chaining: AbortSignal.any([signal, artifactPollAbortController.signal]), ideally threading it into waitForDeviceRunSessionStoppedAsync too.
```

packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -172,3 +201,68 @@ async function downloadArgentArtifactToFileAsync({
...
+  try {
+    await Promise.race([
+      Promise.allSettled([...pendingUploads]),
+      new Promise<never>((_resolve, reject) => {
+        timeout = setTimeout(() => {
+          reject(
+            new SystemError(
+              `Timed out after ${Math.round(timeoutMs / 1000)}s waiting for Argent artifact uploads.`,
+              {
+                trackingCode: 'SIMULATOR_ARTIFACT_SLOW_UPLOAD',
```
szdziedzic, line 272: This SystemError/trackingCode can never reach the classification pipeline: its only consumer is the finally in startArgentRemoteSession.ts, which catches and downgrades to a warn (deliberately, per the review-fixes commit), and the Sentry capture doesn't carry the code either. If success-with-warn is the intended behavior, drop the SystemError scaffolding (log-and-return) or attach the code to the Sentry event — as written it promises a classification that can't happen.

packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -172,3 +201,68 @@ async function downloadArgentArtifactToFileAsync({
   }
   await pipeline(response.body, createWriteStream(destinationPath));
 }
+
+async function sleepUntilAbortedAsync(timeoutMs: number, signal: AbortSignal): Promise<void> {
```
szdziedzic, line 227: Another catch by Fable: 
```
This is a drop-in for try { await setTimeout(timeoutMs, undefined, { signal }); } catch {} from node:timers/promises (already imported in AndroidEmulatorUtils.ts, IosSimulatorUtils.ts, uploadToAsc.ts) — verified equivalent on already-aborted, mid-sleep abort, and listener cleanup. Deletes ~30 lines plus the dedicated "removes the abort listener" test, which guards behavior Node core owns.
```

szdziedzic (commented):
some interesting findings

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: Addressed. `listArgentArtifactsAsync` and artifact downloads now use `ARGENT_ARTIFACT_FETCH_TIMEOUT_MS` via an abort signal, and the in-loop list call also receives the poller/step signal so abort can interrupt an in-flight request. The final drain still does one last pass, but each fetch is bounded before pending uploads enter the 30s cleanup wait.

packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
sjchmiela: Addressed. `startArgentRemoteSession` now destructures the step `signal`, combines it with the local artifact-poller controller using `AbortSignal.any(...)`, passes that combined signal to `pollArgentArtifactsForUploadAsync`, and threads the step signal into `waitForDeviceRunSessionStoppedAsync`.

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: Leaving this as-is intentionally for now. `startArgentRemoteSession` catches cleanup failures so a user-requested stop can still finish successfully, but keeping the explicit `SystemError` and tracking code preserves structured context for this helper and for Sentry/logging if we later decide to surface cleanup timeout classification differently.

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: Addressed. Both abortable sleeps now use `setTimeout` from `node:timers/promises` with `{ signal }`, so listener cleanup is owned by Node core and the custom listener-management code/test are gone.

packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -159,6 +201,10 @@ async function downloadArgentArtifactToFileAsync({
 }): Promise<void> {
   const response = await fetch(new URL(`/artifacts/${artifact.id}`, toolsUrl).toString(), {
     headers: toolsAuthToken ? { Authorization: `Bearer ${toolsAuthToken}` } : {},
+    signal: createTimeoutSignal({
```
szdziedzic: `AbortSignal.timeout` starts ticking at fetch time and node-fetch v2 aborts
the body mid-stream when it fires - so this 10s cap bounds the *entire
download*, not just the connection. An artifact that takes >10s to stream (a
large session recording, or the tool-server producing the file slowly) now
fails mid-download, and combined with the seen-before-success skip above it's
permanently lost with only a warn.

The 10s cap really only needs to protect the list call that gates the drain;
downloads were previously bounded by the 30s pending-uploads race, which is
arguably the right bound for them. Suggest a separate, larger timeout for
downloads (or an idle/headers-only timeout rather than whole-stream).

packages/build-tools/src/steps/utils/argentArtifacts.ts
```diff
@@ -35,38 +37,51 @@ export async function pollArgentArtifactsForUploadAsync(
...
   const seenArtifactIds = new Set<string>();
+  const pendingUploads = new Set<Promise<void>>();
   let listArtifactsErrorCount = 0;
 
-  for (;;) {
+  const queueArtifactUpload = (artifact: ArgentArtifact): void => {
+    if (seenArtifactIds.has(artifact.id)) {
+      return;
+    }
+    seenArtifactIds.add(artifact.id);
```
szdziedzic, line 58: The id is marked seen before the upload succeeds and never unmarked in the
`.catch`, so an artifact whose upload fails transiently mid-session is skipped
by the final drain - permanently lost, even though the drain exists precisely
to not lose artifacts (the test only covers late-*appearing* artifacts, not
failed ones). The new 10s fetch timeout makes transient download failures more
likely, so this path got more reachable.

Note the naive fix (delete the id in the catch) has trade-offs: a permanently
failing artifact would be re-downloaded and Sentry-captured every 5s, and
`uploadDeviceRunSessionArtifactAsync` isn't idempotent (each attempt runs
`createArtifactUploadSession`). A bounded retry count, or clearing failed ids
only before the final drain, is the safer shape. Fine as a follow-up.

packages/build-tools/src/steps/utils/argentArtifacts.ts
szdziedzic, line 93: Now that the in-loop list fetch is tied to the poller/step signal, an abort
landing mid-fetch rejects with an AbortError and ends up here — filing a
Sentry event and a warn for a perfectly normal shutdown. An early
`if (signal?.aborted) { return []; }` guard in the catch avoids the noise.

Related nit: there are now two private copies of `sleepUntilAbortedAsync`
(here and in remoteDeviceRunSession.ts) - could live in a shared util.

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: Un-bounded download here by removing the timeout signal from downloading artifacts.

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: I think we can start with what we have — don't need to be super safe around downloading/uploading aritfacts. We'll keep track of failures thanks to Sentry so if things fail randomly we'll know about it and add retries.

packages/build-tools/src/steps/utils/argentArtifacts.ts
sjchmiela: Added the `signal?.aborted` check

szdziedzic (approved):
fix tests

PR #3950, opened 2026-07-03
[build-tools] Add iOS Simulator recorder

sjchmiela:
# Why

Add a small Swift helper for recording iOS Simulator screen output from EAS build workers. The helper produces either segmented fMP4 output with manifest metadata for playlist construction or one continuous MP4 when segmentation is disabled.

# How

- Added a `record-sim` Swift package under `@expo/build-tools` resources.
- Added `yarn workspace @expo/build-tools build:record-sim` for local builds into `packages/build-tools/bin/record-sim`.
- Updated iOS worker packaging to build `record-sim` from source and copy only the executable into the worker archive.
- Ignored the generated local `bin/record-sim` binary.

# Test Plan

- `yarn workspace @expo/build-tools build:record-sim`
- `packages/build-tools/bin/record-sim --help`
- `bash -n packages/worker/package.sh`
- `./node_modules/.bin/oxfmt --check packages/build-tools/.gitignore packages/build-tools/package.json packages/worker/package.sh packages/build-tools/src/index.ts`

### DIFF
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -23,6 +23,7 @@
     "watch": "tsc --watch --preserveWatchOutput",
     "prebuild": "yarn gql",
     "build": "tsc",
+    "build:record-sim": "mkdir -p bin && swift build -c release --package-path resources/record-sim --build-path resources/record-sim/.build && cp resources/record-sim/.build/release/record-sim bin/record-sim && chmod +x bin/record-sim",
     "typecheck": "tsc",
     "prepack": "rimraf dist \"*.tsbuildinfo\" && yarn gql && tsc -p tsconfig.build.json",
     "jest-unit": "jest --config jest/unit-config.ts",
--- a/packages/build-tools/package.json
+++ b/packages/build-tools/package.json
@@ -23,7 +23,7 @@
     "watch": "tsc --watch --preserveWatchOutput",
     "prebuild": "yarn gql",
     "build": "tsc",
-    "build:record-sim": "mkdir -p bin && swift build -c release --package-path resources/record-sim --build-path resources/record-sim/.build && cp resources/record-sim/.build/release/record-sim bin/record-sim && chmod +x bin/record-sim",
+    "build:record-sim": "mkdir -p bin && record_sim_bin_path=$(swift build -c release --package-path resources/record-sim --build-path resources/record-sim/.build --show-bin-path) && swift build -c release --package-path resources/record-sim --build-path resources/record-sim/.build && cp \"$record_sim_bin_path/record-sim\" bin/record-sim && chmod +x bin/record-sim",
     "typecheck": "tsc",
     "prepack": "rimraf dist \"*.tsbuildinfo\" && yarn gql && tsc -p tsconfig.build.json",
     "jest-unit": "jest --config jest/unit-config.ts",
--- a/packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
+++ b/packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
@@ -2,8 +2,10 @@ import Foundation
 import ObjectiveC
 
 enum XcodeDeveloperDirectory {
-    static func current() -> String {
-        if let developerDir = ProcessInfo.processInfo.environment["DEVELOPER_DIR"], !developerDir.isEmpty {
+    static let current: String = {
+        if let developerDir = ProcessInfo.processInfo.environment["DEVELOPER_DIR"],
+            !developerDir.isEmpty
+        {
             return developerDir
         }
 
@@ -21,38 +23,62 @@ enum XcodeDeveloperDirectory {
         process.waitUntilExit()
         return String(data: pipe.fileHandleForReading.readDataToEndOfFile(), encoding: .utf8)?
             .trimmingCharacters(in: .whitespacesAndNewlines) ?? fallback
-    }
+    }()
 }
 
 enum PrivateSimulatorFrameworks {
-    static func load() {
-        let developerDir = XcodeDeveloperDirectory.current()
+    static func load() throws {
+        try loadResult.get()
+    }
+
+    private static let loadResult: Result<Void, Error> = {
+        let developerDir = XcodeDeveloperDirectory.current
         let candidates = [
             "/Library/Developer/PrivateFrameworks/CoreSimulator.framework/CoreSimulator",
             "\(developerDir)/Library/PrivateFrameworks/CoreSimulator.framework/CoreSimulator",
             "\(developerDir)/../SharedFrameworks/SimulatorKit.framework/SimulatorKit",
             "\(developerDir)/Library/PrivateFrameworks/SimulatorKit.framework/SimulatorKit",
         ]
+        var failures: [String] = []
         for path in candidates {
-            _ = dlopen(path, RTLD_NOW)
+            if dlopen(path, RTLD_NOW) == nil {
+                let message = dlerror().map { String(cString: $0) } ?? "unknown error"
+                failures.append("\(path): \(message)")
+            }
         }
-    }
+        guard NSClassFromString("SimServiceContext") != nil else {
+            let tried = candidates.joined(separator: ", ")
+            let errors = failures.isEmpty ? "no dlopen errors" : failures.joined(separator: "; ")
+            return .failure(
+                RecorderError.make(
+                    8,
+                    "Failed to load CoreSimulator/SimulatorKit (tried: \(tried); errors: \(errors))"
+                )
+            )
+        }
+        return .success(())
+    }()
 }
 
 enum SimulatorDeviceLookup {
-    static func find(udid: String) -> NSObject? {
+    private static let context: NSObject? = {
         guard let contextClass = NSClassFromString("SimServiceContext") as? NSObject.Type else {
             return nil
         }
         let sharedSelector = NSSelectorFromString("sharedServiceContextForDeveloperDir:error:")
-        guard let context = contextClass.perform(sharedSelector, with: XcodeDeveloperDirectory.current(), with: nil)?
+        return contextClass.perform(
+            sharedSelector, with: XcodeDeveloperDirectory.current, with: nil)?
             .takeUnretainedValue() as? NSObject
-        else {
+    }()
+
+    static func find(udid: String) -> NSObject? {
+        guard let context else {
             return nil
         }
         let deviceSetSelector = NSSelectorFromString("defaultDeviceSetWithError:")
-        guard let deviceSet = context.perform(deviceSetSelector, with: nil)?
-            .takeUnretainedValue() as? NSObject
+        guard
+            let deviceSet = context.perform(deviceSetSelector, with: nil)?
+                .takeUnretainedValue() as? NSObject
         else {
             return nil
         }
--- a/packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
+++ b/packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
@@ -1,0 +1,66 @@
+import Foundation
+import ObjectiveC
+
+enum XcodeDeveloperDirectory {
+    static func current() -> String {
+        let fallback = "/Applications/Xcode.app/Contents/Developer"
+        let pipe = Pipe()
+        let process = Process()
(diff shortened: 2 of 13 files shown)


packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
```diff
@@ -0,0 +1,615 @@
...
+        }
+
+        guard let destinationAddress = CVPixelBufferGetBaseAddress(destination) else {
+            throw RecorderError.make(42, "Pixel buffer has no base address")
+        }
+        let sourceAddress = IOSurfaceGetBaseAddress(surface)
+        let sourceStride = IOSurfaceGetBytesPerRow(surface)
+        let destinationStride = CVPixelBufferGetBytesPerRow(destination)
+        let copyBytes = min(width * 4, sourceStride, destinationStride)
+        for row in 0..<height {
```
szdziedzic: Heap buffer overflow when the framebuffer height grows mid-recording. The pixel-buffer
pool is sized once from the *first* frame (`startWriter`, sourcePixelBufferAttributes),
but this loop runs over the *live* snapshot's `height`, and `copyBytes` only clamps
per-row width -not row count. Surface replacement with different dimensions is an
anticipated event in this design (`onSurfaceChange` exists, `surfaceSnapshot()` re-picks
the largest descriptor), so e.g. a landscape→portrait rotation makes `memcpy` write rows
past the destination allocation → heap corruption/crash mid-build.

Fix: clamp the loop to `min(height, CVPixelBufferGetHeight(destination))` (and same idea
for width). Longer-term, a dimension change probably needs a writer restart anyway -
frames scaled into the old geometry are wrong output even when they don't crash.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
```diff
@@ -0,0 +1,615 @@
...
+
+        let capturedAt = monotonicClock.elapsedTime()
+        let capturedAtWallClock = Date()
+        do {
+            let pixelBuffer = try copySurfaceToOwnedPixelBuffer(
+                surface,
+                width: surfaceWidth,
+                height: surfaceHeight
+            )
+            lastSeed = seed
```
szdziedzic: A frame dropped by the writer is never re-captured, so the recording can show stale
content indefinitely. `lastSeed` is committed here, *before* the async append — but
`appendOnWriterQueue` can silently drop that frame (`guard input.isReadyForMoreMediaData
else { return }`, no error, no rollback). After the drop, every recovery path is blocked:
the seed-skip refuses to re-capture the same seed, `probeFramebufferHealth` only fires on
`observedSeed != lastSeed` (they're equal), and boundary/tail frames re-append
`lastAppendedPixelBuffer` - the *previous* pixels. One busy-encoder moment on the last
screen change before a quiet period (or before stop) and the file permanently shows the
old content. With `expectsMediaDataInRealTime` + `maxPendingFrames: 8`, not-ready windows
are realistic under load.

Fix: update `lastSeed` only after a successful append (or reset it when a frame is
dropped) so the next probe/callback re-captures.

packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
```diff
@@ -0,0 +1,70 @@
...
+import ObjectiveC
+
+enum XcodeDeveloperDirectory {
+    static func current() -> String {
+        if let developerDir = ProcessInfo.processInfo.environment["DEVELOPER_DIR"], !developerDir.isEmpty {
+            return developerDir
+        }
+
+        let fallback = "/Applications/Xcode.app/Contents/Developer"
+        let pipe = Pipe()
```
szdziedzic, line 13: This spawns `/usr/bin/xcode-select -p` (Process + waitUntilExit) on every call, and it's
called once per second for the lifetime of the recording: the 1s boundary timer →
`probeSimulatorState()` → `SimulatorDeviceLookup.stateString()` → `find()` → `current()`.
`DEVELOPER_DIR` is unset on workers, so the subprocess path is the normal case - and the
timer runs on `callbackQueue`, the same serial queue that receives frame callbacks, so
every probe stalls capture for the subprocess duration and churns one process/second for
hours-long builds. `find()` also re-enumerates the entire device set each time.

Fix: cache the developer dir in a `static let` (it can't change mid-run) and reuse the
`SimServiceContext` between probes.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
```diff
@@ -0,0 +1,615 @@
...
+                let boundaryPTS = CMTime(seconds: boundary, preferredTimescale: 1_000_000_000)
+                return try appendHeldFrame(at: boundaryPTS)
+            } catch {
+                firstError = error
+                signalFirstFrameReadyIfNeeded()
+                return false
+            }
+        }
+        if didAppendBoundaryFrame {
+            nextBoundaryElapsed = boundary + configuration.segmentDuration
```
szdziedzic: Cross-queue data race on `nextBoundaryElapsed`: read/written on `callbackQueue` here and
at the guard above, and read/written on `writerQueue` in `appendPendingBoundaryFrames`.
Concrete interleaving: while this timer path holds a stale `boundary`, the writer-queue
loop advances `nextBoundaryElapsed` past several boundaries; this line then resets it
*backwards*. The PTS guard in `appendHeldFrame` makes the duplicate appends no-ops, so it
mostly self-heals, but segment cadence can drift — and it's a formal data race (TSan-red,
UB on a shared Double).

Fix: keep all `nextBoundaryElapsed` access on `writerQueue` — this assignment can move
inside the `writerQueue.sync` block that computes `didAppendBoundaryFrame`.

packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
```diff
@@ -0,0 +1,70 @@
...
+    static func load() {
+        let developerDir = XcodeDeveloperDirectory.current()
+        let candidates = [
+            "/Library/Developer/PrivateFrameworks/CoreSimulator.framework/CoreSimulator",
+            "\(developerDir)/Library/PrivateFrameworks/CoreSimulator.framework/CoreSimulator",
+            "\(developerDir)/../SharedFrameworks/SimulatorKit.framework/SimulatorKit",
+            "\(developerDir)/Library/PrivateFrameworks/SimulatorKit.framework/SimulatorKit",
+        ]
+        for path in candidates {
+            _ = dlopen(path, RTLD_NOW)
```
szdziedzic: All dlopen results are discarded. If CoreSimulator/SimulatorKit fail to load (missing or
relocated Xcode — SimulatorKit already moved once in Xcode 27, wrong `xcode-select` path
on a worker image), the failure cascades as `NSClassFromString → nil → find() → nil` and
surfaces as error 1 “Simulator <udid> not found” — sending the operator hunting for a
nonexistent simulator problem instead of a broken Xcode install.

Fix: after the dlopen loop, `guard NSClassFromString("SimServiceContext") != nil` and
throw a dedicated “failed to load CoreSimulator/SimulatorKit (tried: <paths>)” error,
optionally including `dlerror()`.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
```diff
@@ -0,0 +1,615 @@
...
+        }
+
+        var output: CVPixelBuffer?
+        let status = CVPixelBufferPoolCreatePixelBuffer(kCFAllocatorDefault, pool, &output)
+        guard status == kCVReturnSuccess, let destination = output else {
+            throw RecorderError.make(41, "Failed to allocate pixel buffer from pool: \(status)")
+        }
+
+        CVPixelBufferLockBaseAddress(destination, [])
+        IOSurfaceLock(surface, .readOnly, nil)
```
szdziedzic: The `kern_return_t` from `IOSurfaceLock` is unchecked — if the lock fails (e.g. simulator
torn down mid-copy), the copy proceeds on an unlocked surface and can emit torn/garbage
pixels. Notably the destination base address *is* checked a few lines up, but the source
path isn't. Checking the return and bailing on failure is a one-liner.

packages/build-tools/package.json
szdziedzic, line 16: `files` includes `bin/`, but `prepack` neither builds nor prunes `bin/record-sim`, and the
binary is gitignored. npm-packlist normally excludes untracked gitignored files even
inside a `files` directory, and nothing consumes the binary from npm today (the worker
builds it from source in `package.sh`) — so this is a future-trap rather than a live bug:
whether a publisher's stale local arm64 binary sneaks into the tarball depends on npm
ignore-precedence and machine state. Worth making deliberate: either build it in
`prepack` or exclude it explicitly (e.g. `!bin/record-sim` in files / .npmignore).

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. The copy path now clamps both width and height to the destination pixel buffer dimensions before memcpy, so a resized framebuffer cannot write past the pool-sized buffer.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. `lastSeed` now tracks the last successfully appended frame, not the last copied/queued frame, so a writer drop can be retried by the next callback/probe instead of permanently suppressing that seed.

packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. `xcode-select -p` is cached in a static value, and `SimulatorDeviceLookup` now reuses a shared `SimServiceContext` instead of recreating it for each state probe.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. The boundary timer now performs all `nextBoundaryElapsed` reads/writes inside the `writerQueue.sync` block, matching the ownership used by pending boundary appends.

packages/build-tools/resources/record-sim/Sources/RecordSim/PrivateSimulatorServices.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. Framework loading now returns a dedicated error when `SimServiceContext` is unavailable after trying the candidate CoreSimulator/SimulatorKit paths, including the attempted paths and dlopen errors.

packages/build-tools/resources/record-sim/Sources/RecordSim/SimulatorRecorder.swift
sjchmiela: Fixed in a5e5333984884264a557bbe4f197028909c72e07. The copy path now checks `IOSurfaceLock` before reading, and also checks the destination pixel-buffer lock before writing.

packages/build-tools/package.json
sjchmiela: Leaving this one unresolved for now. This thread is outdated and calls out a future packaging trap rather than a live worker packaging bug; the worker path still builds `record-sim` from source and `bin/record-sim` remains gitignored. We should address this separately if we decide the build-tools npm package should deliberately include or exclude that binary.

sjchmiela: We expect the current setup to work on EAS: the worker packaging path builds `record-sim` from source for the iOS VM tarball, and `@expo/build-tools` knowingly does not ship `bin/record-sim` in the npm package. If private API/Xcode drift turns this into a real compatibility problem, we will change how `record-sim` gets onto the VM, most likely by building it while the worker is running on that image instead of treating it as a prebuilt package artifact.

szdziedzic (approved):

PR #3952, opened 2026-07-03
[worker] warn on unknown build environment

gwdp:
# Why

A workflow job with an environment that doesn't exist (e.g. `staging`, vs `development`/`preview`/`production`) gets no warning. It silently resolves to zero environment variables and fails later in a confusing, unrelated-looking way (e.g. the app config failing to read).

# How

Warn in the "Spin up build environment" phase when the job's environment isn't `development`/`preview`/`production`.

# Test Plan

Ran this job against a local worker:

```yaml
jobs:
  verify_fingerprint_job_outputs:
    name: Verify fingerprint job outputs
    type: custom
    environment: staging
    runs_on: linux-medium
    steps:
      - run: echo "=== test job ==="
```

Screenshot 2026-07-03 at 4 18 14 PM
Screenshot 2026-07-03 at 4 21 17 PM

### DIFF
--- a/packages/worker/src/build.ts
+++ b/packages/worker/src/build.ts
@@ -19,6 +19,26 @@ import { Analytics, Event, logProjectDependenciesAsync } from './external/analyt
 import { prepareRuntimeEnvironment } from './runtimeEnvironment';
 import { cleanUpWorkingdir } from './workingdir';
 
+const KNOWN_BUILD_ENVIRONMENTS = ['development', 'preview', 'production'];
+
+type EnvironmentWarningContext = {
+  metadata?: { environment?: string };
+  logger: Pick<bunyan, 'warn'>;
+  markBuildPhaseHasWarnings: () => void;
+};
+
+export function warnOnUnknownEnvironment(ctx: EnvironmentWarningContext): void {
+  const environment = ctx.metadata?.environment;
+  if (environment && !KNOWN_BUILD_ENVIRONMENTS.includes(environment)) {
+    ctx.markBuildPhaseHasWarnings();
+    ctx.logger.warn(
+      `Unknown environment "${environment}". Expected one of: ${KNOWN_BUILD_ENVIRONMENTS.join(
+        ', '
+      )}. No environment variables were added for it.`
+    );
+  }
+}
+
 export async function build({
   ctx,
   buildId,
@@ -40,6 +60,7 @@ export async function build({
           { job: omit(ctx.job, 'secrets', 'projectArchive') },
           'Builder is ready, starting build'
         );
+        warnOnUnknownEnvironment(ctx);
       },
       { doNotMarkStart: true }
     );
--- a/packages/worker/src/__unit__/build.test.ts
+++ b/packages/worker/src/__unit__/build.test.ts
@@ -1,0 +1,75 @@
+import { BuildContext, Builders } from '@expo/build-tools';
+import { Platform } from '@expo/eas-build-job';
+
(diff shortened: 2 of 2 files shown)


gwdp:
* **#3952** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/worker/src/build.ts
```diff
@@ -40,6 +54,7 @@ export async function build({
           { job: omit(ctx.job, 'secrets', 'projectArchive') },
           'Builder is ready, starting build'
         );
+        warnOnUnknownEnvironment(ctx);
```
gwdp, line 63: started with a new build phase but thought it was overkill and moved here. Let me know if you have any different thoughts

packages/worker/src/build.ts
```diff
@@ -19,6 +19,26 @@ import { Analytics, Event, logProjectDependenciesAsync } from './external/analyt
...
+
+type EnvironmentWarningContext = {
+  metadata?: { environment?: string };
+  logger: Pick<bunyan, 'warn'>;
+  markBuildPhaseHasWarnings: () => void;
+};
+
+export function warnOnUnknownEnvironment(ctx: EnvironmentWarningContext): void {
+  const environment = ctx.metadata?.environment;
+  if (environment && !KNOWN_BUILD_ENVIRONMENTS.includes(environment)) {
```
sjchmiela, line 32: We support custom environments, so specifying a not-known environment is not an issue per se as long as the user has environment variables in that environment?

Initially I thought we might color the environment pill yellow on workflow run page if the environment has no variables, but that would color it yellow if a user does not use EAS Environment Variables at all which is bad?

Maybe we do this ^ but only if environment is not one of the known environments? Or sth like that

gwdp: Tysm for the review! I missed the custom environments UI, and the www enum only listing dev/preview/prod made me assume they weren't a thing. Should've asked claude:rolling_on_the_floor_laughing:

Agreed that false warnings would be bad here. I think the cleaner fix is to have `www` set a flag (that's where the resolved vars actually live from what I see) saying the environment matched no variables, and have the worker just react to it instead of hardcoding the known-environments list. wdyt?

sjchmiela: I think it depends on where do we want the flag to live!

We could have it be a column / flag (property in some new JSONB column) on a Workflow Job. Then, based on the flag supplied at creation time we would later show the warning in the UI or not. This requires adding a new field.

We could have it be something calculated dynamically, like we open workflow run page, website gets the list of environments used in the workflow run, website asks www if those environments currently have variables assigned to them. Bad thing about it is that this may drift as we don't have a notion of "environment variables at a given point in time".

We could have it be, as you suggest, a flag on the job, but then we can't show the warning on non-VM jobs!

I _think_ the first is the most correct one.

gwdp: At first I was okay with not showing it for non-VM jobs, but reasoning through, the warning I'm showing was never the perfect place imho. Further on, maybe a new `warnings`on WorkflowJob would allow us to have those warnings issues on www and more flexibility on how to show? (Maybe in the annotations?). Will dig further but sharing in case you have a different take or any input

PR #3956, opened 2026-07-06
[steps][1/4] Add `isCompositeFunctionPath` and reject local composite functions in custom builds

sswrk:
# Why

Local composite functions use `uses: ./...` paths. Custom build configs (`.eas/build/*.yml`) must reject these early with a clear error.

# How

- Adds `isCompositeFunctionPath` in `localCompositeFunctions.ts`
- Rejects local composite function paths in `validateAllFunctionsExist` for custom builds

# Test Plan

Added unit tests.

### DIFF
--- a/packages/steps/src/BuildConfig.ts
+++ b/packages/steps/src/BuildConfig.ts
@@ -8,6 +8,7 @@ import { BuildRuntimePlatform } from './BuildRuntimePlatform';
 import { BuildStepEnv } from './BuildStepEnv';
 import { BuildStepInputValueType, BuildStepInputValueTypeName } from './BuildStepInput';
 import { BuildConfigError, BuildWorkflowError } from './errors';
+import { isLocalCompositeFunctionPath } from './utils/localCompositeFunctions';
 import { BUILD_STEP_OR_BUILD_GLOBAL_CONTEXT_REFERENCE_REGEX } from './utils/template';
 
 export type BuildFunctions = Record<string, BuildFunctionConfig>;
@@ -438,6 +439,18 @@ export function validateAllFunctionsExist(
     }
   }
   const calledFunctionsOrFunctionGroup = Array.from(calledFunctionsOrFunctionGroupsSet);
+  const compositeFunctionPaths = calledFunctionsOrFunctionGroup.filter(
+    isLocalCompositeFunctionPath
+  );
+  if (compositeFunctionPaths.length > 0) {
+    throw new BuildConfigError(
+      `Local composite functions (${compositeFunctionPaths
+        .map(compositeFunctionPath => `"${compositeFunctionPath}"`)
+        .join(
+          ', '
+        )}) are not supported in ".eas/build/*.yml" custom builds. Local composite functions can only be used in EAS workflows (".eas/workflows/*.yml").`
+    );
+  }
   const externalFunctionIdsSet = new Set(externalFunctionIds);
   const externalFunctionGroupsIdsSet = new Set(externalFunctionGroupsIds);
   const nonExistentFunctionsOrFunctionGroups = calledFunctionsOrFunctionGroup.filter(
--- a/packages/steps/src/__tests__/BuildConfig-test.ts
+++ b/packages/steps/src/__tests__/BuildConfig-test.ts
@@ -1042,6 +1042,40 @@ describe(validateAllFunctionsExist, () => {
       validateAllFunctionsExist(buildConfig, { externalFunctionIds: [] });
     }).toThrowError(/Calling non-existent functions: "eas\/build"/);
   });
+  test('rejects local composite function references with a dedicated error', () => {
+    const buildConfig: BuildConfig = {
+      build: {
+        steps: ['./.eas/functions/my-function'],
+      },
+    };
+
+    expect(() => {
+      validateAllFunctionsExist(buildConfig, { externalFunctionIds: [] });
+    }).toThrowError(
+      /Local composite functions \("\.\/\.eas\/functions\/my-function"\) are not supported in "\.eas\/build\/\*\.yml" custom builds\. Local composite functions can only be used in EAS workflows \("\.eas\/workflows\/\*\.yml"\)\./
+    );
+  });
+  test('rejects object-form local composite function references with a dedicated error', () => {
+    const buildConfig: BuildConfig = {
+      build: {
+        steps: [
+          {
+            './.eas/functions/my-function': {
+              inputs: {
+                foo: 'bar',
+              },
+            },
+          },
+        ],
+      },
+    };
+
+    expect(() => {
+      validateAllFunctionsExist(buildConfig, { externalFunctionIds: [] });
+    }).toThrow(
+      /Local composite functions \("\.\/\.eas\/functions\/my-function"\) are not supported in "\.eas\/build\/\*\.yml" custom builds\. Local composite functions can only be used in EAS workflows \("\.eas\/workflows\/\*\.yml"\)\./
+    );
+  });
   test('non-existent namespaced functions with skipNamespacedFunctionsOrFunctionGroupsCheck = false', () => {
     const buildConfig: BuildConfig = {
       build: {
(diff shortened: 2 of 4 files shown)


sswrk:
* **#3929** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3929?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4013](https://github.com/expo/eas-cli/pull/4013) <a href="https://app.graphite.com/github/pr/expo/eas-cli/4013?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4014](https://github.com/expo/eas-cli/pull/4014) <a href="https://app.graphite.com/github/pr/expo/eas-cli/4014?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3956** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3956?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3956?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const JOB_CONTEXT_INTERPOLATION_REGEXP = /\$\{\{(.+?)\}\}/;
+
+function actionReferenceIsInterpolated(uses: string): boolean {
+  return JOB_CONTEXT_INTERPOLATION_REGEXP.test(uses);
+}
+
+export function parseActionReference(uses: string): string | null {
+  const trimmed = uses.trim();
+  if (!trimmed.startsWith('./')) {
```
sjchmiela: oh! all these years i somehow thought those are relative from a workflow file!

this being said, could we support `../shared-functions/...` for monorepos?

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+export function isActionReference(uses: string): boolean {
+  return parseActionReference(uses) !== null;
+}
+
+/** Loads only actions transitively referenced by `rootSteps`. Unreferenced files are ignored. */
+export async function buildActionCatalogFromStepsAsync({
+  rootSteps,
+  loadAction,
+}: BuildActionCatalogFromStepsOptions): Promise<ActionCatalog> {
+  const catalog: ActionCatalog = {};
```
sjchmiela: @hSATAC found some specific behavior about checking stuff in objects and he's using `Object.create(null)` in the hooks pull request, not sure if we want to do it here too

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// Local composite actions referenced via `uses: ./path/to/action` in EAS workflows.
+// Not supported in `.eas/build/*.yml` custom build configs.
+
+type ActionLoader = (ref: string) => Promise<ActionConfig>;
+
+type BuildActionCatalogFromStepsOptions = {
+  rootSteps: readonly unknown[];
+  loadAction: ActionLoader;
+};
```
sjchmiela: can we inline this?

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
+import { ActionCatalog, ActionConfig } from '@expo/eas-build-job';
+import { promises as fs } from 'fs';
+import path from 'path';
+
+// Local composite actions referenced via `uses: ./path/to/action` in EAS workflows.
+// Not supported in `.eas/build/*.yml` custom build configs.
+
+type ActionLoader = (ref: string) => Promise<ActionConfig>;
```
sjchmiela: and this too?

sjchmiela: do we need to make it generic (could there be a static `loadAction` function)?

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+    // Action path missing: no symlink to resolve, so path.resolve() containment is enough.
+  }
+
+  const relative = path.relative(realProjectRoot, realActionPath);
+  return relative !== '' && !relative.startsWith('..') && !path.isAbsolute(relative);
+}
+
+function collectActionReferencesFromSteps(
+  steps: readonly unknown[],
+  into: Set<string> = new Set()
```
sjchmiela: can we merge sets in callsite?

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export function parseActionReference(uses: string): string | null {
+  const trimmed = uses.trim();
+  if (!trimmed.startsWith('./')) {
+    return null;
+  }
+  // If a uses path is interpolated, we can't check if it's valid until the workflow actually runs.
+  if (actionReferenceIsInterpolated(trimmed)) {
+    return null;
+  }
```
sjchmiela: vibe-comment -- it feels to me like `parseActionReference` sometimes (if non-interpolated) returning path and sometimes not (if interpolated) is weird.

also… maybe we can disallow using interpolation in `uses` for now

also - would it be easier to reference actions/functions by their path, not "some ref"?

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,113 @@
...
+): Promise<boolean> {
+  const resolvedProjectRoot = path.resolve(projectRoot);
+  const resolvedActionPath = path.resolve(actionPath);
+
+  let realProjectRoot = resolvedProjectRoot;
+  let realActionPath = resolvedActionPath;
+
+  try {
+    // Follow symlinks to reject paths that escape the project (e.g. via `.eas/actions/evil`).
+    realActionPath = await fs.realpath(resolvedActionPath);
```
sjchmiela: what attack would this allow (honest question)? maybe we can allow it

sjchmiela (commented):
preliminary review -- i would also maybe recommend moving the here-unused functions to the pull request which starts using them cause it's harder to understand the context like this

packages/steps/src/utils/localActions.ts
sswrk: I assume this is about https://github.com/expo/eas-cli/pull/3957#discussion_r3569827569

We should be fine with the current approach, we're validating that the keys are local paths which won't collide with object prototype properties.

packages/steps/src/utils/localActions.ts
sswrk: About making a static function: the passed loader function will involve filesystem access (which is the responsibility of `build-tools` in the stack), so this approach is about dependency inversion.

packages/steps/src/utils/localActions.ts
sswrk: Removed the project directory boundary guard. This was more of a "least privilege principle" thing, but as communicated offline, workflows ensure proper isolation, so this isn't really needed in this layer.

sswrk:
> i would also maybe recommend moving the here-unused functions to the pull request which starts using them cause it's harder to understand the context like this

Yeah, sorry about this. This shaped up like this because this code is like a "hot spot" that multiple things will depend on (including CLI and build tools). I'll see if I can reorganize the PR stack to be fully linear and introduce these functions at the first usage.

EDIT: Done. Trimmed down this PR, moved the yet-unused functions to be introduced higher in the stack.

packages/steps/src/utils/localActions.ts
sjchmiela: bleh

packages/steps/src/utils/localActions.ts
```diff
@@ -0,0 +1,7 @@
+// Local composite actions referenced via `uses: ./path` or `uses: ../path` in EAS workflows.
+// Not supported in `.eas/build/*.yml` custom build configs.
+
+export function isActionPath(uses: string): boolean {
```
sjchmiela: nit: add "local" adjective?

sjchmiela (approved):

sjchmiela (approved):

PR #3957, opened 2026-07-06
[build-tools] unify workflows hook

hSATAC:
# Why

Workflow hooks (`before_install_node_modules`, `before_submit`, …) are currently spliced into the generated `steps[]` on the server, with every job type wiring them by hand, and they silently don't work in managed native builds. We're unifying this so the server sends `job.hooks` on the payload and the worker runs them itself — one implementation, one set of semantics (ENG-22614).

This PR is the worker half and ships dormant: hooks only run when `job.hooks` is present, which the server doesn't send yet, so there is no behavior change until the server side switches over.

# How

- `@expo/eas-build-job` gains a `HOOK_ANCHORS` registry (the single source of anchor names) and an internal `__hook_id` field on `CommonStepZ`. Anchor binding is self-declared by its carrier: a function declares its anchor with an internal `__hookId` on its `BuildFunction` definition (the `__metricsId` pattern), and server-generated anchor steps carry the `__hook_id` stamp. A stamp wins over the invoked function's declaration, and stamp values are loose strings on purpose: an older worker treats anchor names newer than itself as inert metadata.
- Hooks execute in the engine rather than being spliced into `steps[]` at parse time. `StepsConfigParser` takes a required `hooks` option and constructs hook steps per anchor occurrence into a `hooksByAnchorStep` map; this also works inside `uses: eas/build` expansions, via the `__hookId` stamped onto constructed steps. `BuildWorkflow.executeAsync` then runs them around each anchor: the anchor's gate is evaluated once, before-hooks run first, the anchor runs in a `try`, after-hooks in the `finally`. So `before_x` runs iff the anchor runs, and `after_x` runs whenever the anchor executed, pass or fail.
- A hook entry is the unit the user authored: a function-group hook step expands into one entry whose `if:` is evaluated once for the whole group. Groups are allowed in both hook sides.
- Version-skew safety: hook keys the worker doesn't recognize are fully inert (skipped before validation, surfaced as build-log warnings), so a newer server can never fail an older worker. Hook steps under known keys are validated like job steps and fail with a `BuildConfigError` naming the key.
- The hook primitives (`constructHookEntriesAsync`, `executeHookStepsAsync`, `validateHookStepsAsync`) are exported from `@expo/steps` for the managed-native hook runner that follows this PR.
- Observability: each executed hook side reports one `eas.workflow.hook` Datadog event (tags: anchor, timing, result, plus the anchor's own outcome on after events), emitted at execution through the engine's provider callback. A volume drop is the silent-drop signal for the rollout; unmatched hook keys stay build-log warnings, so a user whose hook didn't run finds out. (The metric name is allowlisted on the server-side turtle-metrics endpoint, which lands with the switchover.)
- One engine bugfix that applies immediately, hooks or not: a step throwing a falsy value (e.g. `Promise.reject(undefined)`) used to be swallowed; it now fails the job with a descriptive wrapped error.

# Test Plan

Unit tests across the three packages: hook construction and per-occurrence attachment (including anchors inside a real `uses: eas/build` expansion), the run-iff-anchor-runs gate, after-on-failure, the `always()` matrix, group-entry `if:` semantics, multi-occurrence wraps, skew cases (unknown keys and unregistered stamps stay inert), dormancy (no `job.hooks` → identical behavior), the falsy-throw fix, and metric emission. Registry consistency tests pin exactly which functions declare anchors. CI passes.


packages/eas-build-job/src/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * `functionId` binds an anchor to an EAS build function: any step invoking that
+ * function (directly, or inside a function-group expansion) is an occurrence of
+ * the anchor. Anchors without a functionId are bound via an internal stamp
+ * field (`__hook_id`) on EAS-generated shell steps.
+ */
+export const HOOK_ANCHORS = {
+  install_node_modules: {
+    description: 'The step that installs node modules.',
+    functionId: 'eas/install_node_modules',
```
sjchmiela: uh it's still weird to me we mix `__hook_id` with `functionId` here…

packages/build-tools/src/utils/hookMetrics.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Emits one `eas.workflow.hook` distribution per inserted hook step (steps
+ * world; the native runner will emit the same metric with `world: 'native'`).
+ * The turtle-job-runs proxy forwards these tags as-is; the turtle-builds proxy
+ * (custom builds) also adds the standard build tags (image, sdk_version,
+ * build_mode, …), the same as every build metric.
+ */
+export function reportWorkflowHookMetrics(insertedHooks: readonly InsertedHook[]): void {
+  for (const { anchor, timing, kind } of insertedHooks) {
+    Datadog.distribution('eas.workflow.hook', 1, {
```
sjchmiela: do we want this to be datadog **logs**?

packages/steps/src/BuildStep.ts
```diff
@@ -135,6 +141,13 @@ export class BuildStep extends BuildStepOutputAccessor {
   public readonly ifCondition?: string;
   public readonly timeoutMs?: number;
   public readonly __metricsId?: string;
+  // The function this step was created from, when it came from a `uses:` call
+  // or a function-group expansion. TS-only reference — never serialized.
+  public readonly sourceFunction?: BuildFunction;
+  // When set, this step is an after-hook of the referenced step: it runs iff
+  // that step executed (pass or fail), and is skipped iff that step skipped.
+  // TS-only reference — never serialized.
+  public readonly runAfterStep?: BuildStep;
```
sjchmiela: i wonder… what if we made `BuildWorkflow` understand workflows instead? it feels to me like this might be a bit easier than having to tie steps together with `runAfterStep`? (i also am not sure why there's no `runBeforeStep` — what if a hookable step has an `if` that evaluates to false?)

```ts
  public async executeAsync(): Promise<void> {
    let maybeError: Error | null = null;
    for (const step of this.buildSteps) {
      let shouldExecuteStep = false;

      try {
        shouldExecuteStep = step.shouldExecuteStep();
      } catch (err: any) {
        step.ctx.logger.error({ err });
        step.ctx.logger.error(
          `Runner failed to evaluate if it should execute step "${step.displayName}", using step's if condition "${step.ifCondition}". This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
        );
        maybeError = maybeError ?? err;
        this.ctx.markAsFailed();
      }

      if (shouldExecuteStep) {


// get step.getHookId()
// run this.hooks[`before_${hook_id}`]


        const startTime = performance.now();
        let stepResult: StepMetricResult;
        try {
          await step.executeAsync();
          stepResult = 'success';
        } catch (err: any) {
          stepResult = 'failed';
          maybeError = maybeError ?? err;
          this.ctx.markAsFailed();
        } finally {
          this.collectStepMetrics(step, stepResult!, performance.now() - startTime);

// evaluate this.hooks[`after_${hook_id}`]
        }
      } else {
        step.skip();
      }
    }

    if (maybeError) {
      throw maybeError;
    }
  }
```

wdyt

sjchmiela (commented):
thank you for moving so far with this!

packages/steps/src/BuildStep.ts
hSATAC: Yeah I like this approach. It's easier to understand and maintain especially when we started to deal with the native build hooks. I'll refactor toward this.

packages/build-tools/src/utils/hookMetrics.ts
hSATAC: I'd probably lean toward keeping this as a `Datadog.distribution`, mainly because it's more convenient for the dashboards/graphs we want to build.

That said, I don't feel strongly about it—if there's a compelling reason to use `Datadog.log` instead, I'm happy to switch.

packages/eas-build-job/src/hooks.ts
hSATAC: Taking a step back and thinking about it again, I think I agree with you. I was probably a bit too focused on centralizing everything and overlooked some flexibility.

With the current design, a single anchor (`__hook_id`) can be associated with one step and multiple TypeScript helpers. But if that's the case, why not extend that idea so an anchor can be associated with multiple steps as well as multiple helpers?

I can already think of cases where this would be useful—for example, if we have two similar but slightly different steps that should both map to the same anchor.

I'll try reworking the design in that direction.

packages/build-tools/src/customBuildContext.ts
```diff
@@ -63,8 +69,12 @@ export class CustomBuildContext<TJob extends Job = Job> implements ExternalBuild
   private _env: Env;
   private readonly expoApiV2BaseUrl?: string;
   private readonly pendingMetricUploads: Promise<void>[] = [];
+  // The `world` tag on hook metrics is a property of the context owner: the
+  // generic/custom steps runners pass 'steps'; the native hook runner (not
+  // yet landed) will pass 'native'. Never hard-coded at the emission site.
+  private readonly world: 'steps' | 'native';
```
sjchmiela: what purpose does it serve? (i wish we didn't add this haha)

packages/steps/src/AbstractConfigParser.ts
```diff
@@ -30,16 +31,21 @@ export abstract class AbstractConfigParser {
...
+      hooksByAnchorStep,
+    });
     await new BuildWorkflowValidator(workflow).validateAsync();
     return workflow;
   }
 
   protected abstract parseConfigToBuildStepsAndBuildFunctionByIdMappingAsync(): Promise<{
     buildSteps: BuildStep[];
     buildFunctionById: BuildFunctionById;
+    hooksByAnchorStep?: ReadonlyMap<BuildStep, AnchorHooks>;
```
sjchmiela:
```suggestion
    hooksByAnchorStep: ReadonlyMap<BuildStep, AnchorHooks>;
```
?

packages/steps/src/BuildFunctionGroup.ts
```diff
@@ -65,7 +65,10 @@ export class BuildFunctionGroup {
 export function createBuildFunctionGroupByIdMapping(
   buildFunctionGroups: BuildFunctionGroup[]
 ): BuildFunctionGroupById {
-  const buildFunctionGroupById: BuildFunctionGroupById = {};
+  // Null prototype: a lookup of an Object prototype property name ("toString",
+  // "constructor", ...) must return undefined, not an inherited function that
+  // then gets called as a group.
+  const buildFunctionGroupById: BuildFunctionGroupById = Object.create(null);
```
sjchmiela: isn't it kinda weird we need to do this? we're already validating the anchor IDs? or maybe i'm missing some edge case where a user adds `before_constructor` hook and we'll try to call it?

packages/steps/src/BuildStep.ts
```diff
@@ -135,6 +137,9 @@ export class BuildStep extends BuildStepOutputAccessor {
   public readonly ifCondition?: string;
   public readonly timeoutMs?: number;
   public readonly __metricsId?: string;
+  // The function this step was created from, when it came from a `uses:` call
+  // or a function-group expansion. TS-only reference — never serialized.
+  public readonly sourceFunction?: BuildFunction;
```
sjchmiela: can we instead store `__hookId?: string`? i think that's going to be neater?

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+ * its result), and the authored step's `if:` rides on the entry: group
+ * expansion drops an outer `if:`, so the executor evaluates `ifCondition`
+ * once before executing the entry's steps. Plain `run:` and single-function
+ * entries keep their step-level `if:` instead (`ifCondition` stays unset).
+ * Entry conditions evaluate against the global context only — a group call's
+ * `with:` inputs are not visible to them.
+ */
+export interface HookEntry {
+  steps: BuildStep[];
+  kind: 'run' | 'uses';
```
sjchmiela: it feels somewhat weird we keep `kind` here since what does hook entry care what kind is it?

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+  ifCondition?: string;
```
sjchmiela, line 35: do we need to "copy" if condition from the step to entry?

packages/steps/src/BuildWorkflow.ts
```diff
@@ -1,68 +1,154 @@
...
+      hooksByAnchorStep,
+    }: {
+      buildSteps: BuildStep[];
+      buildFunctions: BuildFunctionById;
+      hooksByAnchorStep?: ReadonlyMap<BuildStep, AnchorHooks>;
+    }
   ) {
     this.buildSteps = buildSteps;
     this.buildFunctions = buildFunctions;
+    this.hooksByAnchorStep = hooksByAnchorStep ?? new Map();
```
sjchmiela, line 34: one _more_ solution crossed my mind: add `before: BuildStep[]` and `after: BuildStep[]` to… `BuildStep`!

then the executor can do like
```
fn executeStep(step) {
  if step.shouldExecuteStep() {
    execute before steps
    execute step
    execute after steps
  }
}  
```

maybe too weird

packages/steps/src/BuildWorkflow.ts
```diff
@@ -1,68 +1,154 @@
...
+    const recordFailure = (err: unknown): void => {
+      if (!hasFailed) {
+        hasFailed = true;
+        firstError = err;
+      }
+    };
+
     for (const step of this.buildSteps) {
-      let shouldExecuteStep = false;
+      const hooks = this.hooksByAnchorStep.get(step);
```
sjchmiela, line 73:
```suggestion
      const hooks = this.hooksByAnchorStep.get(step) ?? { before: [], after: [] };
```
maybe slightly neater since there's less ifs?

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+ * Sets every hook step's display status to skipped without evaluating
+ * conditions (display parity with the anchor's own skip).
+ */
+export function skipHookEntries(entries: readonly HookEntry[]): void {
+  for (const entry of entries) {
+    for (const step of entry.steps) {
+      step.skip();
+    }
+  }
+}
```
sjchmiela: i'd personally inline it -- it's not too much and it doesn't introduce a function that maybe only calls `.skip()` or maybe does something else -- who knows

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+    shell: step.shell,
+    command: step.run,
+    env: step.env,
+    ifCondition: step.if,
+    __metricsId: step.__metrics_id,
+  });
+}
+
+/** Package-internal: every `uses:` in `steps` must name a known function or group. */
+export function validateAllStepFunctionsExist(
```
sjchmiela, line 193: i don't think `hooks.ts` is the right place for this

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+  entries: HookEntry[],
+  options: {
+    anchor: HookAnchorId;
+    timing: 'before' | 'after';
+    /**
+     * The global failure state captured BEFORE the anchor's gate decision.
+     * Defaults to the state at call time. Only failures occurring after this
+     * baseline make no-`if:` before-entries skip.
+     */
+    baselineFailure?: boolean;
```
sjchmiela: i think it's always `hasAnyPreviousStepFailed`? (i haven't see a situation when it's not it)? maybe we can drop it?

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+            `step "${step.displayName}"`,
+            step.ifCondition
+          );
+          recordFailure(err);
+          entryFailed = true;
+        }
+      } else {
+        // After-side no-if steps run unconditionally — past sibling failures
+        // too. Before-side: an entry whose explicit condition evaluated true
+        // behaves like a single step whose if: passed — its no-if children
```
sjchmiela: what children?

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+        ctx.baseLogger.debug({ err }, 'Reporting the workflow hook metric failed');
+      }
+    }
+  }
+
+  return { failedLocally, firstError };
+}
+
+// The no-input variant of a step-level `if:` (a group hook entry has no step
+// to hang the authored condition on).
```
sjchmiela: not sure i follow. why not use the same interpolation context we pass to step's if?

packages/steps/src/hooks.ts
sjchmiela: like maybe we can make entries just be steps? maybe a stupid comment

packages/steps/src/hooks.ts
```diff
@@ -0,0 +1,421 @@
...
+ *   explicit `failedLocally` state, never a global-flag delta (the flag may
+ *   already be true when the anchor passed via `always()`). Failure state
+ *   predating the anchor's gate decision (`baselineFailure`) is ignored,
+ *   as required by the "runs iff the anchor runs" invariant.
+ * - `after`: runs unconditionally — past the anchor's own failure AND past an
+ *   earlier after-entry's failure.
+ * A user `if:` is always evaluated against the real global context, so
+ * `failure()` / `success()` keep their global meaning on both sides.
+ */
+export async function executeHookStepsAsync(
```
sjchmiela: i'd put it in the executor (buildworkflow.ts), but maybe you have a different idea

packages/steps/src/StepsConfigParser.ts
```diff
@@ -41,184 +58,237 @@ export class StepsConfigParser extends AbstractConfigParser {
...
+      const maybeFunctionGroup = isStepFunctionStep(stepConfig)
+        ? buildFunctionGroupById[stepConfig.uses]
+        : undefined;
+      if (maybeFunctionGroup !== undefined) {
+        // The unregistered-stamp check runs FIRST: a future-anchor stamp from
+        // a newer server must render the step inert on this worker (no anchor,
+        // NO error) — skew safety outranks the group fence below.
+        if (stepConfig.__hook_id !== undefined && isHookAnchorId(stepConfig.__hook_id)) {
+          throw new BuildConfigError(
+            `A function group call ("${stepConfig.uses}") cannot carry the hook anchor stamp "${stepConfig.__hook_id}": the group call expands into its steps, so there is no single step to attach hooks to.`
```
sjchmiela: why can't we attach before to the first step and after to last? may be a follow up

sjchmiela: or maybe unnecessary altogether, idk

sjchmiela (approved):

packages/build-tools/src/customBuildContext.ts
hSATAC: This is just for metrics and I don't think it's that important. Let's remove it.

packages/steps/src/BuildFunctionGroup.ts
hSATAC: You're right. Nice catch. We don't actually need this.

packages/steps/src/hooks.ts
hSATAC: I thought it's nice to have this in metrics. Sure let's remove it.

packages/steps/src/hooks.ts
hSATAC: We still need the entry for group calls: the group's `if:` has to be evaluated once for the whole group. If each expanded step evaluated it on its own, something like `if: ${{ !failure() }}` would flip to false mid-group once an inner step fails.

packages/steps/src/BuildWorkflow.ts
hSATAC: Hmm I have mixed feeling about this. Let me put this on hold and fix other places first and maybe it will be clearer when we revisit the code again.

packages/steps/src/BuildWorkflow.ts
hSATAC: Not sure this passes the typecheck but let me try to reduce the ifs

packages/steps/src/StepsConfigParser.ts
hSATAC: A group's first or last step doesn't always run even when the group itself runs — the first one can be skipped by its own if:, and the last one never runs if the group fails halfway. So hooks attached to those steps would sometimes just not fire. And for hooking a moment inside a group, declaring the anchor on the inner function already works.

PR #3973, opened 2026-07-06
[eas-cli] Accept navigation metric aliases as positional args to observe:metrics

douglowder:
# Why

When running `observe:metrics` in non-interactive mode, navigation metric names could not be passed into the command; only app startup metrics were accepted.

# How

- Included the navigation metrics in the validation check
- Renamed the short names (keys) for the navigation metrics to avoid confusion with app startup metric short names

# Test Plan

- Tested locally
- Unit tests added and updated

### DIFF
--- a/packages/eas-cli/src/commands/observe/metrics.ts
+++ b/packages/eas-cli/src/commands/observe/metrics.ts
@@ -19,12 +19,7 @@ import {
   ObserveTimeRangeFlags,
   ObserveUpdateIdFlag,
 } from '../../observe/flags';
-import {
-  METRIC_ALIASES,
-  METRIC_SHORT_NAMES,
-  NAVIGATION_METRIC_ALIASES,
-  resolveMetricName,
-} from '../../observe/metricNames';
+import { METRIC_ALIASES, METRIC_SHORT_NAMES, resolveMetricName } from '../../observe/metricNames';
 import { buildObserveEventsJson, buildObserveEventsTable } from '../../observe/formatEvents';
 import { appObservePlatformFromFlag, appPlatformsFromFlag } from '../../observe/platforms';
 import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
@@ -41,7 +36,7 @@ export default class ObserveMetrics extends EasCommand {
     metric: Args.string({
       description: 'Metric to query (e.g. tti, cold_launch, nav_tti)',
       required: false,
-      options: [...Object.keys(METRIC_ALIASES), ...Object.keys(NAVIGATION_METRIC_ALIASES)],
+      options: Object.keys(METRIC_ALIASES),
     }),
   };
 
@@ -95,7 +90,7 @@ export default class ObserveMetrics extends EasCommand {
     } else if (flags['non-interactive']) {
       throw new EasCommandError(
         'A metric argument is required in non-interactive mode. Available metrics: ' +
-          [...Object.keys(METRIC_ALIASES), ...Object.keys(NAVIGATION_METRIC_ALIASES)].join(', ')
+          Object.keys(METRIC_ALIASES).join(', ')
       );
     } else {
       const choices = Object.entries(METRIC_SHORT_NAMES).map(([fullName, displayName]) => ({
--- a/packages/eas-cli/src/commands/observe/metrics.ts
+++ b/packages/eas-cli/src/commands/observe/metrics.ts
@@ -19,7 +19,12 @@ import {
   ObserveTimeRangeFlags,
   ObserveUpdateIdFlag,
 } from '../../observe/flags';
-import { METRIC_ALIASES, METRIC_SHORT_NAMES, resolveMetricName } from '../../observe/metricNames';
+import {
+  METRIC_ALIASES,
+  METRIC_SHORT_NAMES,
+  NAVIGATION_METRIC_ALIASES,
+  resolveMetricName,
+} from '../../observe/metricNames';
 import { buildObserveEventsJson, buildObserveEventsTable } from '../../observe/formatEvents';
 import { appObservePlatformFromFlag, appPlatformsFromFlag } from '../../observe/platforms';
 import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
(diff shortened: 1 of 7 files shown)


douglowder:
* **#3973** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3973?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/observe/metricNames.ts
```diff
@@ -10,8 +10,8 @@ export const METRIC_ALIASES: Record<string, string> = {
 };
 
 export const NAVIGATION_METRIC_ALIASES: Record<string, string> = {
-  cold_ttr: 'expo.navigation.cold_ttr',
-  warm_ttr: 'expo.navigation.warm_ttr',
+  nav_cold_ttr: 'expo.navigation.cold_ttr',
+  nav_warm_ttr: 'expo.navigation.warm_ttr',
   nav_tti: 'expo.navigation.tti',
 };
```
Ubax, line 17: Can we combine them into single `METRIC_ALIASES`?

Something like:

```ts
export const METRIC_ALIASES: Record<string, string> = {
  ...NAVIGATION_METRIC_ALIASES,
  tti: 'expo.app_startup.tti',
  ttr: 'expo.app_startup.ttr',
  cold_launch: 'expo.app_startup.cold_launch_time',
  warm_launch: 'expo.app_startup.warm_launch_time',
  bundle_load: 'expo.app_startup.bundle_load_time',
  update_download: 'expo.updates.download_time',
}
```

Then you could remove 

```ts
if (NAVIGATION_METRIC_ALIASES[input]) {
  return NAVIGATION_METRIC_ALIASES[input];
}
```

and 

```ts
KNOWN_FULL_NAVIGATION_NAMES.has(input) ||
```

packages/eas-cli/src/observe/__tests__/metricNames.test.ts
```diff
@@ -43,12 +43,29 @@ describe(resolveMetricName, () => {
...
+
+  it('passes through navigation full metric names unchanged', () => {
+    expect(resolveMetricName('expo.navigation.cold_ttr')).toBe('expo.navigation.cold_ttr');
+    expect(resolveMetricName('expo.navigation.warm_ttr')).toBe('expo.navigation.warm_ttr');
+    expect(resolveMetricName('expo.navigation.tti')).toBe('expo.navigation.tti');
+  });
+
+  it('lists both app-startup and navigation aliases in the error message on unknown alias', () => {
+    expect(() => resolveMetricName('bogus')).toThrow(/nav_tti/);
+    expect(() => resolveMetricName('bogus')).toThrow(/cold_launch/);
```
Ubax: Can you modify this tests to be a bit more accurate?

Maybe even something like:

```ts
expect(() => resolveMetricName('bogus')).toMatchSnapshot();
```

This would make it easy to update, yet verify that all the metrics are there

packages/eas-cli/src/commands/observe/metrics.ts
```diff
@@ -34,9 +39,9 @@ export default class ObserveMetrics extends EasCommand {
 
   static override args = {
     metric: Args.string({
-      description: 'Metric to query (e.g. tti, cold_launch)',
+      description: 'Metric to query (e.g. tti, cold_launch, nav_tti)',
       required: false,
-      options: Object.keys(METRIC_ALIASES),
+      options: [...Object.keys(METRIC_ALIASES), ...Object.keys(NAVIGATION_METRIC_ALIASES)],
```
Ubax: This could be reverted if the metrics are spread in `METRIC_ALIASES`

packages/eas-cli/src/observe/metricNames.ts
```diff
@@ -10,8 +10,8 @@ export const METRIC_ALIASES: Record<string, string> = {
...
 
 export const NAVIGATION_METRIC_ALIASES: Record<string, string> = {
-  cold_ttr: 'expo.navigation.cold_ttr',
-  warm_ttr: 'expo.navigation.warm_ttr',
+  nav_cold_ttr: 'expo.navigation.cold_ttr',
+  nav_warm_ttr: 'expo.navigation.warm_ttr',
```
Ubax: Should we display a warning if someone uses the old aliases? I'm not sure if we advertised this as stable or unstable (we can just remove the alias without a breaking change)

packages/eas-cli/src/observe/metricNames.ts
douglowder: Yes indeed. We still need to have the navigation aliases separately listed for observe:routes, but we can certainly improve this change.

packages/eas-cli/src/observe/metricNames.ts
douglowder: I don't think this is necessary since we are still in beta

packages/eas-cli/src/observe/metricNames.ts
```diff
@@ -1,18 +1,19 @@
...
   warm_launch: 'expo.app_startup.warm_launch_time',
   bundle_load: 'expo.app_startup.bundle_load_time',
   update_download: 'expo.updates.download_time',
-};
-
-export const NAVIGATION_METRIC_ALIASES: Record<string, string> = {
-  cold_ttr: 'expo.navigation.cold_ttr',
-  warm_ttr: 'expo.navigation.warm_ttr',
-  nav_tti: 'expo.navigation.tti',
+  ...NAVIGATION_METRIC_ALIASES,
```
Ubax, line 10: Can we do it the other way around?

```suggestion
  ...NAVIGATION_METRIC_ALIASES,
  tti: 'expo.app_startup.tti',
  ttr: 'expo.app_startup.ttr',
  cold_launch: 'expo.app_startup.cold_launch_time',
  warm_launch: 'expo.app_startup.warm_launch_time',
  bundle_load: 'expo.app_startup.bundle_load_time',
  update_download: 'expo.updates.download_time',
```

This way it will be clear, that metrics specified here will override any from `NAVIGATION_METRIC_ALIASES`. Otherwise when you look at this variable it may seem that they will be there, but may not be and one need to check `NAVIGATION_METRIC_ALIASES` as well

Ubax (approved):
One nit. Otherwise looks great 🎉

PR #3976, opened 2026-07-07
Bump minimatch dependency

szdziedzic:
# Why

Resolve https://github.com/expo/eas-cli/security/dependabot/282 by moving every `minimatch` path to a patched release line while keeping Yarn hardened mode happy.

# How

Bumped the direct eas-cli `minimatch` dependency from 5.1.2 to 5.1.9.

Bumped root `lerna` to 9.0.7 and refreshed the lockfile so its transitive `minimatch` dependency resolves to 3.1.4.

Vendored `eslint-plugin-graphql@4.0.0` with its legacy rule names intact, but updated its `graphql-config` dependency to a patched major version because upstream no longer publishes dependency updates.

Updated GraphQL schema/generated files after the CI GraphQL verification job surfaced live schema changes.

Added a changelog entry for the dependency bump.

# Test Plan

- YARN_ENABLE_HARDENED_MODE=1 corepack yarn install --immutable
- corepack yarn verify-graphql-code
- corepack yarn run -T lerna run build --scope eas-cli --include-dependencies
- corepack yarn typecheck
- corepack yarn lint
- corepack yarn fmt
- corepack yarn fmt:check
- corepack yarn lint-changelog
- git diff --check
- corepack yarn why minimatch


szdziedzic:
Closing this approach; it ended up being too heavy for the dependency bump.

PR #3980, opened 2026-07-07
[build-tools] Wire agent-device artifact kinds

sjchmiela:
## Summary

- Preserve `artifactType` from agent-device artifact inventory responses.
- Send that value to the device run session artifact upload-session mutation as `kind`.
- Make the local device run session artifact upload helper require an explicit `kind: string | undefined` so untyped uploads are intentional.
- Regenerate EAS CLI GraphQL schema/types with the repository codegen script and update the one fixture affected by regenerated schema drift.

## Why

agent-device now emits semantic artifact types, and universe now stores them on device run session artifacts. This wires the build-tools upload path between those two changes so users can see producer-owned artifact kinds instead of relying on filenames.

## Validation

- `yarn workspace eas-cli generate-graphql-code`
- `yarn oxfmt packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts packages/build-tools/src/steps/utils/deviceRunSessionArtifacts.ts packages/build-tools/src/steps/utils/argentArtifacts.ts packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/deviceRunSessionArtifacts.test.ts packages/eas-cli/src/graphql/generated.ts packages/eas-cli/graphql.schema.json`
- `yarn oxfmt --check packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts packages/build-tools/src/steps/utils/deviceRunSessionArtifacts.ts packages/build-tools/src/steps/utils/argentArtifacts.ts packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/deviceRunSessionArtifacts.test.ts packages/eas-cli/src/graphql/generated.ts packages/eas-cli/graphql.schema.json`
- `yarn workspace @expo/build-tools jest-unit packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/deviceRunSessionArtifacts.test.ts packages/build-tools/src/steps/utils/__tests__/argentArtifacts.test.ts`
- `yarn workspace eas-cli typecheck`
- `git diff --check`

### DIFF
--- a/packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
+++ b/packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
@@ -18,6 +18,7 @@ const AGENT_DEVICE_ARTIFACT_UPLOAD_POLL_INTERVAL_MS = 5_000;
 
 const AgentDeviceArtifactSchema = z.object({
   id: z.string(),
+  artifactType: z.string().nullish(),
   filename: z.string(),
 });
 const AgentDeviceArtifactsListResponseSchema = z.object({
@@ -152,6 +153,7 @@ export async function uploadAgentDeviceArtifactAsync(
       artifactId: artifact.id,
       name: `${artifact.filename} (${artifact.id})`,
       filename: artifact.filename,
+      kind: artifact.artifactType ?? undefined,
       size,
       stream: createReadStream(temporaryArtifactPath),
     });
--- a/packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts
+++ b/packages/build-tools/src/steps/utils/__tests__/agentDeviceArtifacts.test.ts
@@ -49,9 +49,19 @@ describe(listAgentDeviceArtifactsAsync, () => {
           artifacts: [
             {
               id: 'artifact-id',
+              artifactType: 'agent-device-test-report',
               filename: 'report.json',
               mimeType: 'application/json',
               sizeBytes: 123,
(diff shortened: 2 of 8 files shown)


packages/build-tools/src/steps/utils/agentDeviceArtifacts.ts
```diff
@@ -18,6 +18,7 @@ const AGENT_DEVICE_ARTIFACT_UPLOAD_POLL_INTERVAL_MS = 5_000;
 
 const AgentDeviceArtifactSchema = z.object({
   id: z.string(),
+  artifactType: z.string().optional(),
```
szdziedzic: `z.string().optional()` accepts a *missing* `artifactType` but rejects an explicit `null` — and since validation runs on the whole `/artifacts` response, a single `"artifactType": null` would fail `safeParse` for the entire inventory: `listAgentDeviceArtifactsAsync` throws on every poll and no artifacts get uploaded for that session (only warn logs, retrying forever).

If the daemon's serializer can ever emit `null` for an absent type (or we want to be robust across daemon versions), this would be null-proof:

​```ts
artifactType: z.string().nullish(),
​```

plus `kind: artifact.artifactType ?? undefined` at the upload call site. If the daemon provably omits the key instead of sending `null`, feel free to ignore.

szdziedzic (approved):
LGTM - one small finding from claude

PR #3982, opened 2026-07-07
[eas-cli] Improve build request error details

szdziedzic:
# Why

Build request failures can currently hide useful underlying details when the GraphQL client returns an empty `graphQLErrors` array. In the reported case, `EXPO_DEBUG=1` only surfaced `[]`, while the user-facing error stayed generic despite the HTTP 400 response carrying more context.

# How

Updated `handleBuildRequestError` to preserve existing friendly handling for known server-side EAS error codes, while enriching the generic build request failure path with network error messages, HTTP response status, and request IDs when available. Example output for the reported case (HTTP 400 with empty `graphQLErrors`):

```
Build request failed. Make sure you are using the latest eas-cli version. If the problem persists, report the issue.
Network error: Request failed: 400 (Bad Request)
Response status: 400 Bad Request
Request ID: 8f14e45f-ceea-467f-a1d2-91b50fabc001
```

Also changed debug logging to emit structured build request details, including GraphQL errors, network error fields, response status/url, and selected response headers. Response headers are read through an allowlist (`expo-request-id`, `x-request-id`, `content-type`) and the response body is never serialized, so no sensitive data can end up in the output.

Follow-up refinements:

- The response request ID is deduplicated against the GraphQL extension request IDs by comparing the IDs themselves, instead of substring matching on formatted lines.
- Debug output always includes the top-level error message, instead of a conditional that guessed when it was redundant.
- Collapsed the single-use formatting helpers, renamed the object-building ones from `format*` to `collect*`, and `formatBuildRequestErrorDetails` now receives the already-parsed `graphQLErrors` array instead of re-deriving it.

# Test Plan

- `corepack yarn test src/build` — 102 passed, including new tests for the `x-request-id` fallback, the request-ID dedupe, and the top-level error message fallback
- `corepack yarn workspace eas-cli typecheck`
- `yarn fmt` / `yarn lint`
- Smoke test against the compiled output with `EXPO_DEBUG=1`, simulating an HTTP 400 with empty `graphQLErrors` (the reported case), a pure network failure (`ENOTFOUND`), and a GraphQL error + response sharing the same request ID — all produced the expected user-facing message and structured debug output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 - [build-tools] Skip embedded bundle upload for development client builds instead of warning that the bundle is missing. ([#3940](https://github.com/expo/eas-cli/pull/3940) by [@gwdp](https://github.com/gwdp))
 - [eas-cli] Retry uploading assets that don't finish processing during `eas update`, instead of failing the update. ([#3918](https://github.com/expo/eas-cli/pull/3918) by [@gwdp](https://github.com/gwdp))
+- [eas-cli] Show network and HTTP response details for EAS Build request failures when GraphQL errors are empty. ([#3982](https://github.com/expo/eas-cli/pull/3982) by [@szdziedzic](https://github.com/szdziedzic))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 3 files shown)


szdziedzic:
/changelog-entry bug-fix Improve EAS Build request failures to include network and HTTP response details when GraphQL errors are empty.

tsapeta (approved):

douglowder (approved):

PR #3987, opened 2026-07-07
[eas-cli] Support drill-down from event lists in observe:session

douglowder:
# Why

When session ID is omitted and the `observe:session` command is running interactively, walk the user through picking an event whose session they want to inspect. Non-interactive without a session ID still errors as before.

The goal is to support a flow similar to that already available on our website, where a user can select an event from a displayed list, and navigate to the session timeline that contains that event.

# How

Add new flags `--event-name` and `--sort-order`. If these are not provided, user will be prompted first to select from a list of event names (including custom events, if applicable), then select a sort order. The command also now takes the same time period flags as the other commands (`--days`, `--start`, `--end`).

Once the user has made the above selections, the user can choose from the list of events shown (up to 25), and the session for that event will be shown.

# Test Plan

- Tested manually with an SDK 57 project
- New unit tests added

### DIFF
--- a/packages/eas-cli/src/commands/observe/session.ts
+++ b/packages/eas-cli/src/commands/observe/session.ts
@@ -1,15 +1,32 @@
-import { Args } from '@oclif/core';
+import { Args, Flags } from '@oclif/core';
 
 import EasCommand from '../../commandUtils/EasCommand';
+import { EasCommandError } from '../../commandUtils/errors';
 import { EasNonInteractiveAndJsonFlags } from '../../commandUtils/flags';
+import { ExpoGraphqlClient } from '../../commandUtils/context/contextUtils/createGraphqlClient';
 import Log from '../../log';
-import { fetchObserveSessionEventsAsync } from '../../observe/fetchSessions';
-import { ObserveProjectIdFlag } from '../../observe/flags';
+import { ObserveQuery } from '../../graphql/queries/ObserveQuery';
+import { EventsOrderPreset } from '../../observe/fetchEvents';
+import {
+  fetchObserveSessionEventsAsync,
+  fetchSessionLogCandidatesAsync,
+  fetchSessionMetricCandidatesAsync,
+} from '../../observe/fetchSessions';
+import { ObserveProjectIdFlag, ObserveTimeRangeFlags } from '../../observe/flags';
 import {
   buildObserveSessionEventsJson,
   buildObserveSessionEventsTable,
+  formatLogCandidateTitle,
+  formatMetricCandidateTitle,
 } from '../../observe/formatSessions';
+import {
+  METRIC_SHORT_NAMES,
+  isKnownMetricName,
+  resolveMetricName,
+} from '../../observe/metricNames';
 import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
+import { resolveTimeRange } from '../../observe/startAndEndTime';
+import { ExpoChoice, selectAsync } from '../../prompts';
 import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 
 // Fixed at 100 — the maximum page size accepted by the underlying events and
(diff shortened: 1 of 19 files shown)


douglowder:
* **#3987** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3987?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3987?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eas-cli/src/observe/fetchSessions.ts
```diff
@@ -146,3 +146,82 @@ export async function fetchObserveSessionEventsAsync(
...
+  options: FetchSessionLogCandidatesOptions
+): Promise<SessionLogCandidate[]> {
+  const { events } = await fetchObserveCustomEventsAsync(graphqlClient, appId, {
+    eventName: options.eventName,
+    limit: options.limit,
+    startTime: options.startTime,
+    endTime: options.endTime,
+  });
+  const ascending = options.orderAscending;
+  const sorted = [...events].sort((a, b) => {
```
kadikraman: I don't think we should sort this client side - if the server doesn't have an orderby we should add it if we need it and then sort on the server

packages/eas-cli/src/commands/observe/session.ts
```diff
@@ -85,3 +145,145 @@ export default class ObserveSession extends EasCommand {
...
+  graphqlClient: ExpoGraphqlClient;
+  projectId: string;
+  eventNameFlag: string | undefined;
+  sort: string | undefined;
+  timeRangeFlags: { days: number | undefined; start: string | undefined; end: string | undefined };
+}): Promise<string> {
+  const { startTime, endTime } = resolveTimeRange(timeRangeFlags);
+
+  const eventNameChoice: EventNameChoice = eventNameFlag
+    ? { name: eventNameFlag, isMetric: isKnownMetricName(eventNameFlag) }
```
kadikraman, line 174: We're likely going to have custom _metrics_ down the line so you could track timings of custom values here in which case this would break since it assumes built-in metrics or nothing.

kadikraman: Maybe we can be explicit here and ask for a metric or an event name?

packages/eas-cli/src/commands/observe/session.ts
```diff
@@ -57,17 +89,45 @@ export default class ObserveSession extends EasCommand {
...
+        flags.days !== undefined ||
+        flags.start !== undefined ||
+        flags.end !== undefined;
+      if (pickerFlagsProvided) {
+        throw new EasCommandError(
+          'The picker flags (--event-name, --sort, --days, --start, --end) describe how to find a session and cannot be combined with a session ID argument.'
+        );
+      }
+      sessionId = args.sessionId;
+    } else if (flags['non-interactive']) {
```
kadikraman: I believe the existing convention is to treat `--json` as non-interactive as well vs here we'd be asking for prompts.

douglowder: This was a great catch. We should be using resolveNonInteractiveAndJsonFlags() function from flags.ts, and it appears that several observe commands need to be adjusted to use the correct function.

packages/eas-cli/src/commands/observe/session.ts
douglowder: I think the right way to handle this is to extend the isKnownMetricName method to include custom metrics, once they are supported.

packages/eas-cli/src/commands/observe/session.ts
douglowder: Fixed across all observe commands

packages/eas-cli/src/observe/fetchSessions.ts
douglowder: Fixed -- server already has orderBy so only client change was needed

kadikraman (approved):

PR #3988, opened 2026-07-08
[build-tools] Fix iOS Simulator `apsd` disable for iOS 26.5

sjchmiela:
## Summary

Fixes the iOS Simulator `apsd` shutdown helper to use the launchd domain requested by iOS 26.5:

- disables and boots out `user/foreground/com.apple.apsd` first
- falls back to `system/com.apple.apsd` for compatibility
- verifies whether the service is still `state = running`
- treats `launchctl print` returning service-not-found as a successful stopped state

## Root cause

The previous implementation used `system/com.apple.apsd`. On iOS 26.5, `launchctl` warns to switch to `user/foreground/com.apple.apsd`, and workflow logs showed `bootout system/com.apple.apsd` exiting non-zero. That made a successful or partially successful disable look like a failure and could leave `apsd` running.

## Validation

- `yarn test packages/build-tools/src/utils/__tests__/IosSimulatorUtils.test.ts`
- `yarn typecheck` in the original checkout
- `git diff --check`

I also verified locally on a disposable iOS 26.5 simulator clone that `disable` + `bootout` with `user/foreground/com.apple.apsd` stops the service, and subsequent `launchctl print` returns service-not-found.

### DIFF
--- a/packages/build-tools/src/utils/IosSimulatorUtils.ts
+++ b/packages/build-tools/src/utils/IosSimulatorUtils.ts
@@ -226,16 +226,32 @@ export namespace IosSimulatorUtils {
     udid: IosSimulatorUuid;
     env: NodeJS.ProcessEnv;
   }): Promise<void> {
-    await spawn(
-      'xcrun',
-      ['simctl', 'spawn', udid, 'launchctl', 'disable', 'system/com.apple.apsd'],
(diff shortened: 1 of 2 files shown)


gwdp (approved):
thank you for patching this! 🔥

PR #3994, opened 2026-07-09
[build-tools] Wire `record-sim` into EAS functions

sjchmiela:
## Summary

- add separate `eas/start_ios_simulator_recordings` and `eas/finish_ios_simulator_recordings` build functions
- poll booted iOS Simulators and record each session with `record-sim`
- retry failed recorder processes up to three times per simulator boot
- upload completed MP4 files as screen-recording device run session artifacts with recording metadata
- extend device run session artifact uploads to accept metadata

## Why

Device run sessions need simulator screen recordings that can start independently from remote sessions and be finalized and uploaded when the build workflow finishes.

## Impact

Workflows can explicitly start simulator recording polling and later finish and upload recordings. Missing `record-sim` binaries or device run session IDs disable recording or upload without failing the workflow.

## Validation

- `yarn workspace @expo/build-tools jest-unit steps/utils/__tests__/deviceRunSessionArtifacts.test.ts --runInBand`
- `yarn oxfmt --check` on all changed files
- `git diff --check origin/main...HEAD`

### DIFF
--- a/packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
+++ b/packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
@@ -75,124 +75,124 @@ type IosSimulatorRecordingSession = {
 
 let activeIosSimulatorRecordingSession: IosSimulatorRecordingSession | null = null;
 
-export async function startIosSimulatorRecordingsAsync({
-  env,
-  logger,
-}: {
-  env: Env;
-  logger: bunyan;
-}): Promise<void> {
-  if (activeIosSimulatorRecordingSession) {
-    logger.info('iOS Simulator screen recording polling is already running.');
-    return;
-  }
+export namespace IosSimulatorRecordingUtils {
+  export async function startAsync({ env, logger }: { env: Env; logger: bunyan }): Promise<void> {
+    if (activeIosSimulatorRecordingSession) {
+      logger.info('iOS Simulator screen recording polling is already running.');
+      return;
+    }
 
-  const recordSimCommand = await resolveRecordSimCommandAsync({ env });
-  if (!recordSimCommand) {
-    logger.warn(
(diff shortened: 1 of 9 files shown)


packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  session.logger.info(`Starting screen recording for ${displayName}.`);
+  const recordingSpawn = spawn(
+    session.recordSimCommand,
+    ['--udid', udid, '--output', outputDirectory, '--segment-duration', '0'],
+    {
+      env: session.env,
+      stdio: ['ignore', 'pipe', 'pipe'],
+    }
+  );
```
szdziedzic, line 234: I checked it with 5.6 Sol and I see a potential problem: `record-sim` defaults to 30 Mbps (which is about 3.75 MB/s), while device-run artifacts are limited to 2 GiB. At the target bitrate, a recording can exceed that limit after roughly 9.5 minutes, even though sessions can run for 45 minutes to 2 hours. The upload failure is caught and only logged, so longer sessions may silently end up without a recording.

should we segment these recordings?

sjchmiela: I think we use variable bitrate, MP4 is pretty good at compressing no movement to static screens. Let's see how things go and add segmentation later?

szdziedzic (approved):
LGTM overall

discussed offline: add metadata field we will use to know which of the recordings is system recording in WWW

sjchmiela:
```
{"udid":"7F21F24E-4BA4-4C21-9071-5AD6089D6708","firstFrameAt":"2026-07-10T13:15:35.009Z","__eas_screen_recording":"1"}
```

Zrzut ekranu 2026-07-10 o 15 21 33

PR #3990, opened 2026-07-09
[build-tools] pass metrics CORS origin to serve-sim

gwdp:
Ref ENG-23622

# Why

So the expo.dev dashboard can read serve-sim's live `/metrics` stream cross-origin.

# How

Reads `EAS_SIMULATOR_METRICS_CORS_ORIGIN` from the worker env and passes it to serve-sim as `--metrics-cors-origin` when launching the tunnel.

# Test Plan

CI passes. Unit tests cover the env getter and the arg builder. Also verified end-to-end locally that the origin reaches serve-sim's `--metrics-cors-origin`.

PR #3995, opened 2026-07-10
[build-tools] collect serve-sim session metrics with a start/collect step pair

gwdp:
Ref [ENG-23622](https://linear.app/expo/issue/ENG-23622/eas-simulator-cpumemory-inspector-in-the-ui)

# Why

So a simulator session's CPU/memory shows up in the dashboard — live while it runs, and from an artifact after it ends.

# How

* `eas/start_serve_sim_metrics` starts a background poller that discovers each running serve-sim from its local registry (`${tmpdir}/serve-sim/server-<udid>.json`) and streams its `/metrics` to a per-device NDJSON file, reconnecting if a stream drops.
* `eas/collect_serve_sim_metrics` (run with `if: always()`, universe side) stops the poller and uploads the files as artifacts.
* Every iOS session type runs serve-sim (serve-sim, argent, agent-device), so all three are covered. Best-effort throughout.

# Test Plan

CI passes. Unit tests cover registry discovery (malformed/missing files), the SSE → NDJSON stream, the start→collect lifecycle including reconnect after a failed attempt, and the upload (skips missing/empty, swallows failures).  <br>Also ran the full flow locally via `eas simulator start --type serve-sim` against a local worker: the poller streamed `/metrics` and the NDJSON uploaded as an artifact.

https://github.com/user-attachments/assets/369d97e6-bb62-48fd-9ce5-070f495ee047

### DIFF
--- a/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
+++ b/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
@@ -1,6 +1,7 @@
 import { BuildFunction, BuildRuntimePlatform } from '@expo/steps';
 
 import { type CustomBuildContext } from '../../customBuildContext';
+import { Sentry } from '../../sentry';
 import { getDeviceRunSessionIdOrThrow } from '../utils/remoteDeviceRunSession';
 import { uploadServeSimMetricsFileAsync } from '../utils/serveSimMetricsArtifacts';
 import { ServeSimMetricsRecorder } from '../utils/serveSimMetricsRecorder';
@@ -20,11 +21,19 @@ export function createCollectServeSimMetricsBuildFunction(ctx: CustomBuildContex
       }
       try {
         const deviceRunSessionId = getDeviceRunSessionIdOrThrow(env);
-        for (const { udid, filePath } of collected) {
-          await uploadServeSimMetricsFileAsync(ctx, { deviceRunSessionId, udid, filePath, logger });
+        for (const { udid, filePath, meta } of collected) {
+          await uploadServeSimMetricsFileAsync(ctx, {
+            deviceRunSessionId,
+            udid,
+            filePath,
+            meta,
+            logger,
+          });
         }
       } catch (err) {
-        logger.warn({ err }, 'Could not upload serve-sim metrics.');
+        const error = err instanceof Error ? err : new Error(String(err));
+        Sentry.capture('Could not upload serve-sim metrics', error);
+        logger.warn({ err: error }, 'Could not upload serve-sim metrics.');
       }
     },
   });
--- a/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
+++ b/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
@@ -1,0 +1,31 @@
+import { BuildFunction, BuildRuntimePlatform } from '@expo/steps';
+
+import { type CustomBuildContext } from '../../customBuildContext';
+import { getDeviceRunSessionIdOrThrow } from '../utils/remoteDeviceRunSession';
+import { uploadServeSimMetricsFileAsync } from '../utils/serveSimMetricsArtifacts';
+import { ServeSimMetricsRecorder } from '../utils/serveSimMetricsRecorder';
+
+export function createCollectServeSimMetricsBuildFunction(ctx: CustomBuildContext): BuildFunction {
+  return new BuildFunction({
+    namespace: 'eas',
+    id: 'collect_serve_sim_metrics',
+    name: 'Collect serve-sim metrics',
+    __metricsId: 'eas/collect_serve_sim_metrics',
+    supportedRuntimePlatforms: [BuildRuntimePlatform.DARWIN],
+    fn: async ({ logger }, { env }) => {
+      const collected = await ServeSimMetricsRecorder.finishAsync({ logger });
+      if (collected.length === 0) {
+        logger.info('No serve-sim metrics collected; skipping upload.');
+        return;
+      }
+      try {
+        const deviceRunSessionId = getDeviceRunSessionIdOrThrow(env);
+        for (const { udid, filePath } of collected) {
+          await uploadServeSimMetricsFileAsync(ctx, { deviceRunSessionId, udid, filePath, logger });
+        }
+      } catch (err) {
+        logger.warn({ err }, 'Could not upload serve-sim metrics.');
+      }
+    },
+  });
+}
--- a/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
+++ b/packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
@@ -21,12 +21,12 @@ export function createCollectServeSimMetricsBuildFunction(ctx: CustomBuildContex
       }
       try {
         const deviceRunSessionId = getDeviceRunSessionIdOrThrow(env);
-        for (const { udid, filePath, meta } of collected) {
+        for (const { udid, filePath, metadata } of collected) {
           await uploadServeSimMetricsFileAsync(ctx, {
             deviceRunSessionId,
             udid,
             filePath,
-            meta,
+            metadata,
             logger,
           });
         }
--- a/packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
+++ b/packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
@@ -15,11 +15,13 @@ export async function uploadServeSimMetricsFileAsync(
     deviceRunSessionId,
     udid,
     filePath,
+    meta,
     logger,
   }: {
     deviceRunSessionId: string;
     udid: string;
     filePath: string;
+    meta?: Record<string, unknown>;
     logger: bunyan;
   }
 ): Promise<void> {
@@ -27,11 +29,11 @@ export async function uploadServeSimMetricsFileAsync(
   try {
     ({ size } = await stat(filePath));
   } catch {
-    logger.info(`No serve-sim metrics captured for ${udid}; skipping upload.`);
+    logger.info(`No serve-sim metrics file was written for ${udid}; skipping upload.`);
     return;
   }
   if (size === 0) {
-    logger.info(`No serve-sim metrics captured for ${udid}; skipping upload.`);
+    logger.info(`serve-sim metrics file for ${udid} is empty; skipping upload.`);
     return;
   }
   try {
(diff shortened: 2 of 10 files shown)


gwdp:
* **#3995** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/3995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3990** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3990?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
```diff
@@ -0,0 +1,31 @@
...
+        logger.info('No serve-sim metrics collected; skipping upload.');
+        return;
+      }
+      try {
+        const deviceRunSessionId = getDeviceRunSessionIdOrThrow(env);
+        for (const { udid, filePath } of collected) {
+          await uploadServeSimMetricsFileAsync(ctx, { deviceRunSessionId, udid, filePath, logger });
+        }
+      } catch (err) {
+        logger.warn({ err }, 'Could not upload serve-sim metrics.');
```
szdziedzic: Should we report it to Sentry?

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
```diff
@@ -0,0 +1,54 @@
...
+    return;
+  }
+  try {
+    logger.info(`Uploading serve-sim metrics for ${udid} (${formatBytes(size)}).`);
+    await uploadDeviceRunSessionArtifactAsync(ctx, {
+      deviceRunSessionId,
+      artifactId: `metrics-${udid}`,
+      name: METRICS_ARTIFACT_FILENAME,
+      filename: METRICS_ARTIFACT_FILENAME,
+      kind: undefined,
```
szdziedzic: How will www/the dashboard identify this artifact among a session's artifacts? Screen recordings pass `kind: 'screen-recording'` plus `__eas_type` metadata precisely for this; here the only signals are `filename === 'metrics.ndjson'` and `metadata.udid` - and an Argent user artifact could legitimately also be named `metrics.ndjson`. If the dashboard keys on filename that seems fragile; `kind: 'performance-metrics'` and `__eas_type` metadata field would match the screen-recording precedent.

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (size === 0) {
+    logger.info(`No serve-sim metrics captured for ${udid}; skipping upload.`);
+    return;
+  }
+  try {
+    logger.info(`Uploading serve-sim metrics for ${udid} (${formatBytes(size)}).`);
+    await uploadDeviceRunSessionArtifactAsync(ctx, {
+      deviceRunSessionId,
+      artifactId: `metrics-${udid}`,
+      name: METRICS_ARTIFACT_FILENAME,
```
szdziedzic: Every other DRS artifact type builds a descriptive and unique display name (recordings include device name + short UDID), but here `name` is the literal `metrics.ndjson` for every device - a two-device session produces two indistinguishable entries in any artifact list.

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
```diff
@@ -0,0 +1,246 @@
...
+      buffer += decoder.decode(chunk as Buffer, { stream: true });
+      let newline: number;
+      while ((newline = buffer.indexOf('\n')) !== -1) {
+        const line = buffer.slice(0, newline);
+        buffer = buffer.slice(newline + 1);
+        if (line === '') {
+          event = 'message';
+        } else if (line.startsWith('event:')) {
+          event = line.slice('event:'.length).trim();
+        } else if (line.startsWith('data:') && event !== 'meta') {
```
szdziedzic: Skipping the `meta` frame for NDJSON homogeneity is reasonable, but it carries `hostCores`, `sampleIntervalMs`, and `schemaVersion`, which a consumer of the artifact may want (e.g. to normalize `cpuPct`). Since it's already parsed out here, consider surfacing the last-seen meta so collect can stash it in the artifact `metadata` alongside `udid`.

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
```diff
@@ -0,0 +1,246 @@
...
+
+async function pollServeSimMetricsAsync(session: ServeSimMetricsSession): Promise<void> {
+  const { abortController, logger } = session;
+  while (!abortController.signal.aborted) {
+    for (const server of await readServeSimServersAsync(session.stateDir)) {
+      if (session.activeStreams.has(server.udid)) {
+        continue;
+      }
+      const failures = session.attempts.get(server.udid) ?? 0;
+      if (failures >= MAX_CONSECUTIVE_STREAM_FAILURES_PER_DEVICE) {
```
szdziedzic, line 117: Parity gap with `IosSimulatorRecordingUtils`, which clears failure counts when a device is no longer booted so a rebooted sim gets retried: here `attempts` persists for the whole session, so a serve-sim that restarts for the same UDID after 10 consecutive failures never gets collected again (its registry file disappearing/reappearing doesn't reset the count). Probably fine for best-effort telemetry - flagging in case you want to clear counts for UDIDs absent from the registry, like the recordings poller does.

packages/build-tools/src/steps/functions/collectServeSimMetrics.ts
gwdp: Adding although [https://app.graphite.com/github/pr/expo/eas-cli/3995/[build-tools]-collect-serve-sim-session-metrics-with-a-start%2Fcollect-step-pair#lines-packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts-R51-R51](https://app.graphite.com/github/pr/expo/eas-cli/3995/%5Bbuild-tools%5D-collect-serve-sim-session-metrics-with-a-start%2Fcollect-step-pair#lines-packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts-R51-R51) is more likely to be hit; Thanks for calling this out

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
gwdp: The rows were distinct under the hood (unique id + `udid`), but they all read as `metrics.ndjson`. Tweaked, so name is now `Performance metrics (<short-udid>)` , so a multi-sim session is distinguishable at a glance too.

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
gwdp: You're right that filename alone was fragile! Switched to `kind: 'performance-metrics'` + `__eas_type` (same as screen recordings) so the `type` is explicit and can't clash with a user file that happens to be named `metrics.ndjson `:pray:

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
gwdp: Thank you for the flag, I did that in my ssh project but forgot here, indded I want to handle it :laughing:. The poller now drops the failure count for any UDID that's left the registry, so a restarted serve-sim on the same UDID gets retried instead of staying stuck after 10 failures

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
gwdp: Loved the idea, now collect stashes `hostCores`/`sampleIntervalMs`/`schemaVersion` in the artifact `metadata` alongside `udid`

szdziedzic (approved):

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (size === 0) {
+    logger.info(`serve-sim metrics file for ${udid} is empty; skipping upload.`);
+    return;
+  }
+  try {
+    logger.info(`Uploading serve-sim metrics for ${udid} (${formatBytes(size)}).`);
+    await uploadDeviceRunSessionArtifactAsync(ctx, {
+      deviceRunSessionId,
+      artifactId: `metrics-${udid}`,
+      name: `Performance metrics (${udid.slice(0, 8)})`,
```
sjchmiela, line 44: would be nice to add device name here / to metadata

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
```diff
@@ -0,0 +1,64 @@
...
+    deviceRunSessionId,
+    udid,
+    filePath,
+    meta,
+    logger,
+  }: {
+    deviceRunSessionId: string;
+    udid: string;
+    filePath: string;
+    meta?: Record<string, unknown>;
```
sjchmiela:
```suggestion
    metadata?: Record<string, unknown>;
```

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
```diff
@@ -0,0 +1,271 @@
...
+type ServeSimServer = { udid: string; url: string };
+
+type ServeSimMetricsSession = {
+  logger: bunyan;
+  stateDir: string;
+  pollIntervalMs: number;
+  outputDirectory: string;
+  files: Map<string, string>;
+  attempts: Map<string, number>;
+  meta: Map<string, Record<string, unknown>>;
```
sjchmiela:
```suggestion
  metadata: Map<string, Record<string, unknown>>;
```
?

sjchmiela (approved):

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
gwdp: I'm sorry, I must have missed the notification for this 🙇‍♂️ Applying suggestions (might need some test changes so applying outside of gh)

packages/build-tools/src/steps/utils/serveSimMetricsRecorder.ts
gwdp: Updated, ty!

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
gwdp: https://github.com/expo/eas-cli/pull/3995/changes/ff025b4b11b288f095bf9f9a203574286192ee26

packages/build-tools/src/steps/utils/serveSimMetricsArtifacts.ts
gwdp: Doing a follow-up on this, great idea!

PR #3999, opened 2026-07-10
[build-tools] Collect agent-device session activity

szdziedzic:
## Why

`agent-device` writes privacy-shaped usage events to per-session `events.ndjson` files. DeviceRunSession consumers need those records independently of the Bunyan runtime-log stream.

## How

- Incrementally tail every `~/.agent-device/sessions/*/events.ndjson` file and preserve partial UTF-8 characters and partial lines across polls.
- Validate the stable source envelope with Zod while accepting new source versions, event kinds, optional fields, and extra fields.
- Normalize records into a producer-neutral DeviceRunSession event format that future Argent and serve-sim collectors can also emit.
- Use `RemoteLoggerStream` to overwrite the marked GCS artifact with a complete snapshot every five seconds.
- Rely on the GCS upload session to enforce the 100 MiB limit with `x-goog-content-length-range`.
- Namespace event IDs by source directory and keep them monotonic across truncation.
- Report bounded setup, parse, collection, and cleanup failures to Sentry without including event payloads.

Collection remains best-effort, so an event reporting failure cannot fail the simulator session. Worker runtime logs are unchanged.

## Test Plan

- `corepack yarn workspace @expo/build-tools typecheck`
- `corepack yarn workspace @expo/build-tools jest-unit --runInBand src/steps/utils/__tests__/deviceRunSessionEvents.test.ts`
- Coverage includes split UTF-8 characters, partial lines, truncation, multiple source directories, upstream schema changes, parse summaries, continuous collection failures, setup failures, and GCS upload headers.

## Rollout

#4027 has merged. Deploy this collector after expo/universe#28937 is available. Website and CLI consumers can deploy after this begins producing marked artifacts.

### DIFF
--- a/packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
+++ b/packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
@@ -197,7 +197,19 @@ export async function startAgentDeviceEventCollectionAsync({
         }
       }
     }
-  })();
+  })()
+    .catch(err => {
+      const error = err instanceof Error ? err : new Error(String(err));
+      logger.warn({ err: error }, 'Agent-device event collection poller failed.');
+      Sentry.capture('Agent-device event collection poller failed', error, {
+        level: 'warning',
+        tags: { phase: 'agent-device-event-collection', operation: 'poll' },
+        extras: { deviceRunSessionId },
+      });
+    })
+    .catch(() => {
+      // A diagnostics failure must not recreate an unhandled rejection in this best-effort poller.
+    });
 
   return {
     stopAsync: async () => {
--- a/packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
+++ b/packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
@@ -1,0 +1,401 @@
+import { SystemError } from '@expo/eas-build-job';
+import { type bunyan } from '@expo/logger';
+import { graphql } from 'gql.tada';
+import fs from 'node:fs';
+import path from 'node:path';
+import { StringDecoder } from 'node:string_decoder';
+import { setTimeout as setTimeoutAsync } from 'node:timers/promises';
+import { z } from 'zod';
+
+import { type CustomBuildContext } from '../../customBuildContext';
+import RemoteLoggerStream from '../../logging/RemoteLoggerStream';
+import { Sentry } from '../../sentry';
+
+const POLL_INTERVAL_MS = 1_000;
+const UPLOAD_INTERVAL_MS = 5_000;
+
+const AGENT_DEVICE_EVENT_KIND_TO_TYPE: Record<string, string> = {
+  'request.started': 'operation.started',
+  'request.finished': 'operation.completed',
+  'action.recorded': 'interaction.recorded',
+};
+
+const AgentDeviceEventSchema = z
+  .object({
+    version: z.number(),
+    ts: z.string(),
+    session: z.string(),
+    kind: z.string(),
+    requestId: z.string().optional().catch(undefined),
+    command: z.string().optional().catch(undefined),
+    status: z.string().optional().catch(undefined),
+    summary: z.string().optional().catch(undefined),
+    details: z.record(z.string(), z.unknown()).optional().catch(undefined),
+  })
+  .passthrough();
+
+const CREATE_DEVICE_RUN_SESSION_EVENT_LOG_UPLOAD_SESSION_MUTATION = graphql(`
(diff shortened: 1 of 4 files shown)


packages/build-tools/src/index.ts
```diff
@@ -1,8 +1,10 @@
 import * as Builders from './builders';
-import GCSLoggerStream from './gcs/LoggerStream';
 import GCS from './gcs/client';
+import RemoteLoggerStream from './logging/RemoteLoggerStream';
 
-export { Builders, GCS, GCSLoggerStream };
+export { Builders, GCS, RemoteLoggerStream };
+/** @deprecated Use RemoteLoggerStream. */
+export { RemoteLoggerStream as GCSLoggerStream };
```
sjchmiela: i don't think we need to keep this compatible

packages/build-tools/src/storage/uploadWithSignedUrl.ts
```diff
@@ -0,0 +1,65 @@
...
+import { URL } from 'url';
+
+import { RetryOptions, retryOnUploadFailure } from './retry';
+
+export interface SignedUrl {
+  url: string;
+  headers: Record<string, string>;
+}
+
+export interface UploadWithSignedUrlParams {
```
sjchmiela: nit: i prefer to do these as `type` cause i don't expect these should accept unknown properties

packages/build-tools/src/gcs/client.ts
```diff
@@ -1,59 +1,13 @@
...
-    if (!resp.ok) {
-      let body: string | undefined;
-      try {
-        body = await resp.text();
-      } catch {}
-      throw new Error(
-        `Failed to upload file: status: ${resp.status} status text: ${resp.statusText}, body: ${body}`
-      );
-    }
+  export type SignedUrl = RemoteSignedUrl;
```
sjchmiela: let's remove gcs altogehter?

sjchmiela (changes requested):
could you also ask clanker to split it into 2 prs:
1. remoteloggerstream
2. agent-device activity?

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
```diff
@@ -0,0 +1,401 @@
...
+        }
+      }
+    }
+  }
+`);
+
+type AgentDeviceEvent = z.infer<typeof AgentDeviceEventSchema>;
+
+export type DeviceRunSessionEvent = {
+  schemaVersion: 1;
```
sjchmiela: are we going to compress the logs? if not i suggest we use shorter prop names (v) and remove obsolete properties (`deviceRunSessionId`, `producerVersion`?, `name`, `occurredAt` -> `ts`)

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
```diff
@@ -0,0 +1,401 @@
...
+  data?: Record<string, unknown>;
+};
+
+type EventFileState = {
+  offset: number;
+  nextSequenceNumber: number;
+  nextLineNumber: number;
+  pending: string;
+  decoder: StringDecoder;
+};
```
sjchmiela, line 72: https://github.com/ribeiromiranda/jsonl-stream?

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
```diff
@@ -0,0 +1,401 @@
...
+        reportEventLogFailure(error, 'cleanup');
+      }
+    },
+  };
+}
+
+async function createEventLogUploadSessionAsync(
+  ctx: CustomBuildContext,
+  deviceRunSessionId: string
+): Promise<{ url: string; headers: Record<string, string> }> {
```
sjchmiela: isn't this signedurl?

sjchmiela (approved):

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
szdziedzic: I used shorter prop names

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
szdziedzic: I looked into this. `jsonl-stream` handles chunk decoding, but we would still need our file discovery, offsets, truncation, multi-file state, and event ID logic. It also terminates the stream on a malformed record, whereas here we intentionally report and skip bad records while continuing collection. Given that, I think the small local `StringDecoder` + line buffer is a better fit and avoids adding a dependency.

We can revisit it if we want to in the future.

PR #4003, opened 2026-07-11
[eas-cli] Add `eas browse` command

jonsamp:
# Why

`eas open` opens the project dashboard but has no way to jump straight to a specific project subpage (builds, submissions, updates, workflows, hosting, etc.). This adds a  `eas browse` command that accepts an optional page argument so users can go directly where they want from the terminal. This is similar to the `gh browse` command.

I want this so it's easier to stay in the command line and to zoom to exactly where i need. I also want agents to be able to use this command to share where on our website they can find things in a more human-friendly format.

Linear: ENG-24214

# How

- Add `eas browse [PAGE]`. `PAGE` is a whitelisted argument mapped to its dashboard path via `PROJECT_PAGES`; both the EAS command name and the website wording resolve to the same page (e.g. `build`/`builds`, `submit`/`submissions`, `workflow`/`workflows`/`cicd`). With no argument it opens the project dashboard.
- Add a `--no-browser` / `-n` flag to print the URL instead of opening a browser.
- Extract URL building into `getProjectPageUrl(accountName, projectName, page?)` in `build/utils/url.ts`; `getProjectDashboardUrl` now delegates to it. Account and project names are `encodeURIComponent`-escaped, consistent with the other builders in that file.
- Convert `eas open` into a thin alias that delegates to `Browse.run([])`. It's the same as calling `eas browse`.

# Test Plan

- Run `eas browse [name of subpage]` and also test `--no-browser|-n`. Examples: `eas browse build`, `eas browse cicd`.
- Run `eas browse` with no arguments and make sure you see the project overview dashboard.
- Run tests


douglowder (approved):
Looks good! I pushed a commit to fix the CHANGELOG entry, and tested locally

PR #4005, opened 2026-07-12
[eas-cli] Add `eas chat` command

jonsamp:
# Why

[ENG-24455](https://linear.app/expo/issue/ENG-24455/eas-chat-command): bring the dashboard AI chat assistant to the terminal so developers (and other agents) can ask questions about their Expo account and EAS projects without leaving the CLI.

# How

* New top-level command `eas chat "message"` (`src/commands/chat.ts`).
* **Streaming:** parses the AI SDK UI message stream (SSE) and writes the assistant's text to stdout incrementally, surfacing tool calls while loading.
* **Interactive multi-turn:** after each answer it prompts `Reply` (default) so you can follow up; the full history is resent each turn.
* **Flags:** `-p/--project` (focus on a project, validated via `AppQuery.byFullNameAsync`, resolves the owner account and frames the question), `-a/--account` (scope), `--non-interactive`, and `--json` (agent-friendly single-turn structured output including the assistant's `response` and the `toolCalls` it used).
* The command is hidden for now since this API feature is not generally available.

# Test Plan

Run `eas chat` and you should see our AI output:

https://github.com/user-attachments/assets/3c396a19-fd75-4960-8ed2-13f10ee9d774


packages/eas-cli/src/chat/chatClient.ts
```diff
@@ -0,0 +1,449 @@
...
+    const part: Extract<ChatMessagePart, { type: 'text' }> = { type: 'text', text: '' };
+    if (id) {
+      textMessagePartsById.set(id, part);
+    }
+    currentTextMessagePart = part;
+    assistantMessageParts.push(part);
+    return part;
+  };
+
+  const writeAssistantLine = (rendered: string, withNewline: boolean): void => {
```
douglowder, line 214: Terminal control-char / ANSI injection from streamed output (renderMarkdown.ts:21-51, chatClient.ts:214-229).
Assistant text is written straight to process.stdout after renderMarkdownLine, which applies chalk styling but does not strip C0/C1 control characters or raw ANSI escape sequences. The model's answers can embed tool results — build logs, branch/channel names, env-var names, account-member names — which are user/project-controlled and can contain escape sequences. A crafted branch name or log line echoed verbatim by the model would be emitted raw to the terminal (cursor manipulation, screen clear, OSC-8 hyperlinks, etc.).
- Suggestion: strip control chars (everything except \n/\t) from each line before rendering/writing — e.g. line.replace(/[\u0000-\u0008\u000b-\u001f\u007f-\u009f]/g, '') at the top of renderMarkdownLine, and add a test that an escape sequence in a text-delta is neutralized. Moderate severity (needs the model to echo attacker content), and it's hidden, so a fast-follow is fine — but worth closing before wider rollout.

douglowder (changes requested):

douglowder:
@jonsamp I looked through this and tested it locally. Looks very good overall. Claude scan caught a possible security issue with unfiltered text being sent to the terminal output, see inline.

packages/eas-cli/src/chat/chatClient.ts
jonsamp: Fixed by stripping C0/C1 control characters, while preserving tabs and newlines, at the start of renderMarkdownLine before Markdown parsing or Chalk styling. I also added a streaming regression test that injects ESC, C1 CSI, and BEL characters through a text-delta and verifies they are not emitted to stdout.

douglowder (approved):

PR #4009, opened 2026-07-13
Upgrade Lodash and GraphQL Code Generator

sjchmiela:
# Why

Supersedes #3968.

Dependabot alert 354 requires `lodash` and `lodash-es` 4.18.1. The original PR upgraded the worker's direct dependency and enforced the patched transitive version with a root resolution. That resolution is no longer needed: this PR upgrades the EAS-owned declarations and the GraphQL Code Generator packages that brought in the remaining restrictive `lodash ~4.17.0` dependency.

# How

- Upgrade direct Lodash declarations in build-tools, local-build-plugin, template-file, and worker to 4.18.1.
- Upgrade the scripts project's `lodash-es` declaration and its independent lockfile to 4.18.1.
- Upgrade the complete GraphQL Code Generator family used by EAS CLI:
  - `@graphql-codegen/cli` 5.0.0 → 6.3.1
  - `@graphql-codegen/introspection` 4.0.0 → 5.0.2
  - `@graphql-codegen/typescript` 4.0.1 → 5.0.10
  - `@graphql-codegen/typescript-operations` 4.0.1 → 5.1.0
- Regenerate the checked-in GraphQL TypeScript output.
- Keep the root free of `resolutions` and regenerate both affected lockfiles.

These are the newest aligned Codegen releases before `typescript-operations` v6. That v6 upgrade changes generated operation type semantics (including nullable response-field presence) and causes broad fixture/type failures across EAS CLI, so it should be handled as a separate GraphQL contract migration rather than bundled into this security dependency update. The selected packages use `@graphql-codegen/plugin-helpers` 6.3.x, which no longer depends on Lodash.

The root dependency graph now reports only `lodash@4.18.1`, selected through ordinary declared dependency ranges. The separate scripts project reports only `lodash-es@4.18.1`. There is no resolution or override.

# Test Plan

- `yarn install`
- `yarn why lodash` — only `lodash@4.18.1`; no Codegen consumer
- `cd scripts && yarn install && yarn why lodash-es` — only `lodash-es@4.18.1`
- `yarn build` — all 13 projects passed
- `yarn workspace eas-cli verify-graphql-code` — generated output compiles against the current live schema
- `cd scripts && yarn typecheck`
- `yarn lint-changelog`
- template-file tests — 7 passed
- worker tests — 51 passed
- `yarn fmt`


szdziedzic (changes requested):
scripts/package.json still pins lodash-es@4.17.21, which is affected by the [same advisory](https://github.com/advisories/GHSA-r5fr-rjxr-66jc)

szdziedzic (approved):

PR #4018, opened 2026-07-15
[build-tools] Remove serve-sim stream URL output

sjchmiela:
I found it sometimes isn't printed.

---

## Summary

- Stop waiting for `serve-sim` to emit a `Stream:` URL before reporting startup success.
- Upload only `previewUrl` in the serve-sim remote config payload.
- Remove `streamUrl` from the simulator session query selection and generated query result type.
- Simplify the serve-sim URL parser to match only the tunnel URL.

## Why

The stream URL is no longer part of the expected serve-sim output path. Keeping it in the startup contract made the build step wait for and upload a value that is not needed by the CLI display flow.

## Test plan

- `yarn --cwd packages/build-tools typecheck`
- `yarn --cwd packages/eas-cli typecheck`
- `yarn --cwd packages/build-tools jest-unit src/steps/utils/__tests__/remoteDeviceRunSession.test.ts`
- `yarn --cwd packages/eas-cli test src/commands/simulator/__tests__/get.test.ts src/commands/simulator/__tests__/start.test.ts --runInBand`
- `git diff --check`

### DIFF
--- a/packages/build-tools/src/steps/functions/startServeSimRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startServeSimRemoteSession.ts
@@ -29,19 +29,18 @@ export function createStartServeSimRemoteSessionBuildFunction(
 
       await selectXcodeDeveloperDirectoryAsync({ env, logger });
 
-      const { previewUrl, streamUrl } = await startServeSimWithTunnelAsync(ctx, {
+      const { previewUrl } = await startServeSimWithTunnelAsync(ctx, {
         baseDomain: ngrokTunnelDomain,
         env,
         logger,
         timeoutMs: STARTUP_TIMEOUT_MS,
       });
       logger.info(`Preview URL: ${previewUrl}`);
-      logger.info(`Stream URL: ${streamUrl}`);
 
       await uploadRemoteSessionConfigAsync({
         ctx,
         deviceRunSessionId,
-        remoteConfig: { previewUrl, streamUrl },
+        remoteConfig: { previewUrl },
         logger,
       });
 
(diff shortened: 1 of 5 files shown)


packages/build-tools/src/steps/functions/startServeSimRemoteSession.ts
```diff
@@ -29,19 +29,18 @@ export function createStartServeSimRemoteSessionBuildFunction(
...
         timeoutMs: STARTUP_TIMEOUT_MS,
       });
       logger.info(`Preview URL: ${previewUrl}`);
-      logger.info(`Stream URL: ${streamUrl}`);
 
       await uploadRemoteSessionConfigAsync({
         ctx,
         deviceRunSessionId,
-        remoteConfig: { previewUrl, streamUrl },
+        remoteConfig: { previewUrl },
```
szdziedzic, line 43: This is currently incompatible with the API contract. On `expo/universe` main, the `SERVE_SIM` remote-config schema still requires both `previewUrl` and `streamUrl`, and `startDeviceRunSession` validates the payload against that schema before persisting it. Sending only `{ previewUrl }` will therefore reject the mutation and prevent the session from starting.

Additionally, `streamUrl` remains a non-null GraphQL field, and the website plus older EAS CLI versions still query it. Could we keep sending it until a backward-compatible server/client migration is deployed, or link the coordinated backend change if one already exists?

- [Remote-config schema](https://github.com/expo/universe/blob/main/server/www/src/data/entities/device-run-session/DeviceRunSessionEntity.ts)
- [Mutation validation](https://github.com/expo/universe/blob/main/server/www/src/graphql/mutations/DeviceRunSessionMutation.ts)

Maybe we should just remove "pure" serve-sim session as we always launch it with argent/agent-device anyway

szdziedzic (changes requested):

sjchmiela:
Updated after the coordinated backend change landed: https://github.com/expo/universe/pull/29064

This branch no longer queries or uses `streamUrl`, and the checked-in GraphQL schema/types now match the new API contract: `streamUrl` remains available only as a deprecated nullable field for older clients. The serve-sim startup payload now sends only `previewUrl`, which `www` accepts after that migration.

szdziedzic (approved):

PR #4020, opened 2026-07-15
[build-tools] Use unique simulator recording artifact names

sjchmiela:
## Summary

- include a shortened simulator UDID and first-frame timestamp in each screen recording artifact name
- format the timestamp as a readable, stable UTC value
- add `__eas_type: "screen-recording"` metadata for typed artifact detection
- cover recordings from a rebooted simulator and simultaneous same-model simulators

Artifact names now use this format:

```
iPhone 16 screen recording (01234567-…, started at Jul 10, 2026, 10:00:00.000 UTC)
```

## Why

A simulator can reboot during one device run session, producing multiple recordings with the same device name and UDID. A session can also contain two same-model simulators whose recordings start at precisely the same time. Device name alone, or device name plus timestamp alone, therefore does not reliably distinguish every recording.

The first eight UDID characters distinguish same-model simulators without putting the full UUID in the user-facing name. The millisecond first-frame timestamp distinguishes subsequent recordings after a reboot. The full UDID and canonical ISO timestamp remain in artifact metadata for exact identity, parsing, and ordering.

The new `__eas_type` marker provides a descriptive discriminator for screen recordings. The existing `__eas_screen_recording: "1"` marker remains temporarily so currently deployed consumers continue recognizing these artifacts during migration.

The artifact metadata also provides `udid`, `deviceName`, `runtimeDisplayName`, `firstFrameAt`, and encoded `width` and `height`.

## Validation

- focused screen recording upload tests covering rebooted and simultaneous same-model simulators
- build-tools TypeScript typecheck

### DIFF
--- a/packages/build-tools/src/steps/functions/__tests__/uploadDeviceRunSessionScreenRecordings.test.ts
+++ b/packages/build-tools/src/steps/functions/__tests__/uploadDeviceRunSessionScreenRecordings.test.ts
@@ -70,6 +70,7 @@ describe(createUploadDeviceRunSessionScreenRecordingsBuildFunction, () => {
         expect.objectContaining({
           name: 'iPhone 16 screen recording (01234567-…, started at Jul 10, 2026, 10:00:00.000 UTC)',
           metadata: {
+            __eas_type: 'screen-recording',
             __eas_screen_recording: '1',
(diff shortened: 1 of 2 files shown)


szdziedzic (approved):
__eas_type for the win

PR #4021, opened 2026-07-15
[eas-cli] remove deprecated GraphQL field usage

tchayen:
## Why

We can remove deprecated fields that are not used by the CLI anymore.

## How

- `BuildFragment`: `project` -> `app`, `channel` -> `updateChannel { id name }`, `runtimeVersion` -> `runtime { id version }` (this is a breaking change noted in changelog).
- `build:cancel` uses `cancelBuild`, `AccountFragment` replaces the full `users` list with `viewerUserPermission { role }`, `Account.appleTeams` moves to `appleTeamsPaginated`, dropped deprecated `selectedImage`.
- The submissions use `Build.appIdentifier` instead of `androidConfig.applicationIdentifier`.

## Test plan

CI passes. Tested against staging with test account.


packages/eas-cli/src/update/utils.ts
```diff
@@ -34,16 +34,9 @@ export type FormatUpdateParameter = Pick<Update, 'id' | 'createdAt' | 'message'>
     | null;
 };
 
-export type UpdateJsonInfo = { branch: string } & Pick<
+export type UpdateJsonInfo = { branch: string; runtimeVersion: string } & Pick<
```
szdziedzic, line 37: This leaves the update JSON contract inconsistent. `getUpdateJsonInfosForUpdates` maps the new `runtime` object back to the legacy flat `runtimeVersion`, so commands such as `eas update --json`, `update:view`, and `update:edit` retain the old shape. Meanwhile, `update:republish --json` prints the raw `UpdateFragment` and emits nested `runtime`.

https://github.com/expo/eas-cli/blob/a1f3ec784ada8e6010f6a96b449dd3e0d4f2f6ab/packages/eas-cli/src/update/republish.ts#L230

This seems to be breaking change.

We should probably map it to `runtimeVersion` as well for republish.

szdziedzic (changes requested):
It's 99% correct, although I found 1 problem:

szdziedzic:
cc: @wschurman or @quinlanj if you want to take a look at update specific things

packages/eas-cli/src/update/utils.ts
tchayen: It is a breaking change, yes. For now I reversed it to remap the field and dropped changelog mention of it. WDYT: do we want to minimize breaking changes (or even avoid at all cost) or are we ok to have a cut off point now and release a major?

szdziedzic (approved):

packages/eas-cli/src/submit/utils/formatSubmission.ts
```diff
@@ -47,10 +47,6 @@ export function formatGraphQLSubmission(submission: SubmissionWithSubmittedBuild
     },
     ...(submission.platform === AppPlatform.Android
       ? [
-          {
-            label: 'Application Identifier',
-            value: sanitize(submission.androidConfig?.applicationIdentifier),
-          },
```
szdziedzic, line 53: I think we should use `submission.submittedBuild?.appIdentifier` to display app id here?

szdziedzic (approved):

PR #4023, opened 2026-07-16
[eas-build-job] internal refactor general hook

hSATAC:
# Why

Part of unifying how workflow hooks work. This is the `@expo/eas-build-job` + `@expo/steps` side, preparing a simpler, looser hook API for consumers to build on.

# How

- `eas-build-job`: add `PublicStepZ`, the user-authored step shape (`StepZ` without the internal `__metrics_id` / `__hook_id` stamp fields), so consumers don't re-derive it.
- `eas-build-job`: `HOOK_ANCHORS` becomes a plain list of anchor names (was a record keyed by anchor with an unused `description`); add `HOOK_KEYS`, the full set of `before_` / `after_` keys.
- `steps`: drop `warnAboutUnmatchedHookKeys`. A hook key whose anchor never occurs in a job is already inert, so the warning was just noise.

# Test Plan

CI passes. Added a `PublicStepZ` test asserting the internal stamp fields are stripped.

### DIFF
--- a/packages/eas-build-job/src/__tests__/hooks.test.ts
+++ b/packages/eas-build-job/src/__tests__/hooks.test.ts
@@ -2,19 +2,13 @@ import { HOOK_ANCHORS, parseHookKey } from '../hooks';
 
 describe('HOOK_ANCHORS', () => {
   it('contains the v1 anchors', () => {
-    expect(Object.keys(HOOK_ANCHORS).sort()).toEqual([
+    expect([...HOOK_ANCHORS].sort()).toEqual([
       'checkout',
       'install_node_modules',
       'maestro_cloud',
       'maestro_tests',
       'submit',
     ]);
-  });
-
-  it('every entry has a non-empty description', () => {
-    for (const entry of Object.values(HOOK_ANCHORS)) {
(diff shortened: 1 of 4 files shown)


packages/eas-build-job/src/step.ts
```diff
@@ -158,6 +158,11 @@ export type ShellStep = z.infer<typeof ShellStepZ>;
 
 export const StepZ = z.union([ShellStepZ, FunctionStepZ]);
 
+// Internal stamp fields EAS sets; users must not forge them.
+const INTERNAL_STEP_FIELDS = { __metrics_id: true, __hook_id: true } as const;
+
+export const PublicStepZ = z.union(StepZ.options.map(member => member.omit(INTERNAL_STEP_FIELDS)));
+
```
sjchmiela: to me this scrutiny doesn't seem useful? do you think it's useful?

sjchmiela (approved):
thank you!

packages/eas-build-job/src/step.ts
hSATAC: Yeah, that's mostly just for my own peace of mind 😄 I agree it's probably not worth it. Let's remove it.

PR #4025, opened 2026-07-16
[build-tools] Distinguish simulator recording artifact names

sjchmiela:
# Why

A simulator can reboot during a device run session and produce multiple screen recordings for the same device. Those recordings previously shared the same artifact name, which violates the API's per-session artifact-name uniqueness constraint and prevents later recordings from uploading.

[Sentry issue](https://expoio.sentry.io/issues/7604712142/?project=130569&query=is%3Aunresolved%20requestId%3A8e5b9779-0903-4e1c-96d0-e7d698047e31&referrer=issue-stream)

# How

Include the recording's existing first-frame ISO timestamp in its artifact name. Each recording keeps the same device metadata and stable UUID-backed filename, while recordings created after simulator reboots no longer collide on the display name.

# Test Plan

- `yarn jest-unit src/steps/functions/__tests__/uploadDeviceRunSessionScreenRecordings.test.ts --runInBand`
- `yarn typecheck` in `packages/build-tools`
- `yarn lint` (0 errors; existing unrelated warnings only)
- `yarn fmt:check`


sjchmiela:
Superseded by https://github.com/expo/eas-cli/pull/4020

PR #4029, opened 2026-07-16
[eas-build-job] Add SSH build phase

gwdp:
# Why

First PR in the worker SSH stack (ENG-24547). An SSH session needs its own build phase so it shows up as a distinct step in the build logs and on the website instead of being folded into an unrelated phase.

# How

Adds an `SSH_SESSION` entry to the `BuildPhase` enum, with its log display name and website id. No runtime behavior yet; later PRs in the stack use the phase.

# Test Plan

Enum addition with no runtime behavior. CI passes.

PR #4030, opened 2026-07-16
[build-tools] Add worker-side SSH session helper

gwdp:
# Why

The worker needs a piece that starts an SSH session on the build VM and keeps the worker alive while someone is connected.

# How

New helper resolves the upterm client vendored in the worker package, dials the relay, reports the connection details back so the CLI can attach, and holds on an idle timer. Redialing is capped at a few attempts per drop so a flapping relay can't retry forever. Connection secrets are scrubbed from logs.

# Test Plan

Unit tests cover the connection parsing, secret redaction, process-liveness detection, the capped redial, and the idle hold. CI passes.


gwdp:
* **#4032** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4031** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4031?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4030** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,222 @@
...
+  '..',
+  '..',
+  'bin',
+  `upterm-${resolveUptermArch(process.arch)}`
+);
+const UPTERM_KEEPALIVE_SLEEP_SECONDS = 6 * 60 * 60; // 6 hours
+const CONNECTION_POLL_INTERVAL_MS = 500;
+const CONNECTION_STARTUP_TIMEOUT_MS = 60_000;
+
+export const DEFAULT_SSH_RELAY_SERVER_URL = 'wss://uptermd.upterm.dev';
```
sjchmiela: i think we might want to decide we don't want to use the public upterm host ever for "security"

sjchmiela (changes requested):
how aout using `upterm session current --output json`?

gwdp:
> upterm session current --output json

​Great call, looking into it; I think it needs the admin socket, which came only later with me trying to get client count, so yes, if that works ,less weight to lift here and will go with it, tysm!

packages/build-tools/src/utils/upterm.ts
gwdp: Chopping it out for now and bringing up discussing

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+  return arch === 'arm64' ? 'arm64' : 'amd64';
+}
+
+const UPTERM_BIN_PATH = path.join(
+  __dirname,
+  '..',
+  '..',
+  'bin',
+  `upterm-${resolveUptermArch(process.arch)}`
+);
```
sjchmiela: https://github.com/expo/eas-cli/pull/4031#pullrequestreview-4773244189

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+  isAlive: () => boolean;
+  redialAsync: () => Promise<SshConnectionConfig>;
+  stopAsync: () => Promise<void>;
+};
+
+type UptermSessionJson = {
+  sessionId?: string;
+  host?: string;
+  clientCount?: number;
+};
```
sjchmiela: should this be parsed with zod?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+
+export function redactConnectionSecrets(text: string): string {
+  let redacted = text;
+  // Every session token upterm emitted (userinfo before `@`), so each exact token can be scrubbed.
+  for (const [, token] of text.matchAll(/upterm proxy wss:\/\/([^@\s]+)@/g)) {
+    redacted = redacted.split(token).join('<redacted>');
+  }
+  // Catch-all: redact URL userinfo (scheme://<userinfo>@) for any other credential in the text.
+  return redacted.replace(/([a-z][a-z0-9+.-]*:\/\/)[^@\s/]+@/gi, '$1<redacted>@');
+}
```
sjchmiela, line 94: does it make sense to redact connection secrets for errors (errors mean the session is no longer active?)

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+
+export function isUptermProcessAlive(
+  child:
+    | { exitCode: number | null; signalCode: NodeJS.Signals | null; killed: boolean }
+    | null
+    | undefined
+): boolean {
+  return child != null && child.exitCode === null && child.signalCode === null && !child.killed;
+}
```
sjchmiela: should this be part of `processes.ts`? and then the return property could be the SpawnPromise i think

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+
+function killUptermProcessGroup(child: { pid?: number; kill: () => void } | undefined): void {
+  if (child?.pid == null) {
+    return;
+  }
+  try {
+    process.kill(-child.pid, 'SIGTERM');
```
sjchmiela: why negate?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+    );
+  }
+  return UPTERM_BIN_PATH;
+}
+
+async function findAdminSocketPathAsync(uptermSocketDir: string): Promise<string | null> {
+  const entries = await fs.readdir(uptermSocketDir).catch(() => [] as string[]);
+  const socketName = entries.find(entry => entry.endsWith('.sock'));
+  return socketName ? path.join(uptermSocketDir, socketName) : null;
+}
```
sjchmiela, line 145: why do we need admin socket and not just the default?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  const stateDir = await fs.mkdtemp(path.join(os.tmpdir(), 'eas-ssh-'));
+  const hostKeyPath = path.join(stateDir, 'id_host');
+  const forceCommandPath = path.join(stateDir, 'join.sh');
+  const uptermSocketDir = path.join(stateDir, 'upterm');
+
+  await spawn('ssh-keygen', ['-t', 'ed25519', '-N', '', '-f', hostKeyPath, '-q'], {
+    logger: ctx.logger,
+  });
+  await fs.writeFile(forceCommandPath, '#!/usr/bin/env bash\nexec bash -l\n', { mode: 0o755 });
```
sjchmiela, line 206: worht a comment?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  let currentProcess: ReturnType<typeof spawn> | null = null;
```
sjchmiela: i think we use SpawnPromise as a type in a few places

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,318 @@
...
+        relayServerUrl,
+        '--accept',
+        '--skip-host-key-check',
+        '-i',
+        hostKeyPath,
+        '--force-command',
+        forceCommandPath,
+        '--',
+        'bash',
+        '-lc',
```
sjchmiela, line 266: what's the difference between force command and this bash?

sjchmiela (changes requested):

packages/build-tools/src/utils/upterm.ts
gwdp: Moved towards GCS backed binary, tysm (you mentioned this earlier in the review process, so sorry for missing it)

packages/build-tools/src/utils/upterm.ts
gwdp: updated, indeed a good call

packages/build-tools/src/utils/upterm.ts
gwdp: It does actually, spawn errors still could carry stdout/stderr with the token, and logs outlive the session. Cheap insurance if you are not strong

packages/build-tools/src/utils/upterm.ts
gwdp: nice one, moved to processes.ts, `isChildProcessAlive`

packages/build-tools/src/utils/upterm.ts
gwdp: I was caught on this one (when I didn't have it and learned some), `kill(-pid)` means process group. We spawn detached, so without the negate the sleep child can survive redial. Kept + commented.

packages/build-tools/src/utils/upterm.ts
gwdp: Added. ty!

packages/build-tools/src/utils/upterm.ts
gwdp: My bad, ty!

packages/build-tools/src/utils/upterm.ts
gwdp: Without it, session current hits upterm’s default socket, which can still be the old one after a rediall; that was blowing up client-count reads. I can dig up why that would happen on upterm but it sounded the easy way to guarantee the current connect admin socket without any risks; Let me know if any thoughts

packages/build-tools/src/utils/upterm.ts
gwdp: --force-command is what user get (login shell). The sleep after `--` is just the host-side process so upterm host stays up with nobody connected

gwdp: Added some comments

packages/build-tools/src/utils/processes.ts
```diff
@@ -1,5 +1,18 @@
 import spawn from '@expo/turtle-spawn';
 
+/**
+ * Whether a spawned child is still running. Prefer this over checking exitCode alone —
+ * signal-terminated children can have exitCode null with a non-null signalCode.
+ */
+export function isChildProcessAlive(
+  child:
+    | { exitCode: number | null; signalCode: NodeJS.Signals | null; killed: boolean }
```
sjchmiela: shouldn't this be `ChildProcess` from nodejs types?

packages/build-tools/src/utils/processes.ts
```diff
@@ -1,5 +1,18 @@
...
+    | null
+    | undefined
```
sjchmiela: it's a bit weird we're allowing passing in null and undefined isn't it?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+    ...(typeof spawnErr.stdout === 'string'
+      ? { stdout: redactConnectionSecrets(spawnErr.stdout) }
+      : {}),
+    ...(typeof spawnErr.stderr === 'string'
+      ? { stderr: redactConnectionSecrets(spawnErr.stderr) }
+      : {}),
+  };
+}
+
+function killUptermProcessGroup(child: { pid?: number; kill: () => void } | undefined): void {
```
sjchmiela: -> `processes.ts`?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+
+export type TurtleSshTarget =
+  | { turtleJobRunId: string; turtleBuildId?: never }
+  | { turtleJobRunId?: never; turtleBuildId: string };
+
+export type JobWithOptionalSsh = { ssh?: SshSettings };
+
+export function isWorkflowSshEnabled(job: JobWithOptionalSsh): boolean {
+  return job.ssh != null;
+}
```
sjchmiela: why doesn't this use `Generic.Job`?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+
+export function getWorkflowJobIdOrThrow(env: Env): string {
+  const workflowJobId = env.__WORKFLOW_JOB_ID;
+  if (!workflowJobId) {
+    throw new SystemError(
+      '__WORKFLOW_JOB_ID is not set. It should be present in the job environment when ssh is enabled.'
+    );
+  }
+  return workflowJobId;
+}
```
sjchmiela: this is not used (and rightly so -- nothing in ssh should rely on a job being / not being a owrklfow job)

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+    );
+  }
+  return idleTimeoutSeconds;
+}
+
+export function getSshRelayServerUrl(job: JobWithOptionalSsh): string {
+  const relayServerUrl = job.ssh?.relayServerUrl;
+  if (!relayServerUrl) {
+    throw new SystemError(
+      'job.ssh.relayServerUrl is required when ssh is enabled on the job. The worker dials our SSH relay; the public upterm host is not allowed.'
```
sjchmiela: make this a user-readable error maybe?

bonus points for adding `trackingCode`

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+  hasPlatform,
+}: {
+  buildId: string;
+  hasPlatform: boolean;
+}): TurtleSshTarget {
+  if (hasPlatform) {
+    return { turtleBuildId: buildId };
+  }
+  return { turtleJobRunId: buildId };
+}
```
sjchmiela: if i were to vote i would either inline this wherever we do it, or pass in actual `platform` so we don't artificially split logic?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+    parts.push(hours === 1 ? '1 hour' : `${hours} hours`);
+  }
+  if (minutes > 0) {
+    parts.push(minutes === 1 ? '1 minute' : `${minutes} minutes`);
+  }
+  if (seconds > 0 || parts.length === 0) {
+    parts.push(seconds === 1 ? '1 second' : `${seconds} seconds`);
+  }
+  return parts.join(' ');
+}
```
sjchmiela: let's make it a reusable function like https://github.com/expo/eas-cli/blob/9edb10ca55521cc3b0d79ab6bc75159fbfce2eda/packages/eas-cli/src/utils/timer.ts#L33 and move out of this file

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+      connectionConfig: {
+        ...connectionConfig,
+        type: 'UPTERM_V1',
+      },
+      sessionSettings: { idleTimeoutSeconds },
+    })
+    .toPromise();
+  if (result.error || !result.data) {
+    throw new SystemError(
+      `Failed to create or update the SSH session: ${result.error?.message ?? 'no data returned'}`
```
sjchmiela: `, { cause: result.error }`?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+    );
+  }
+  const session = result.data.turtleSshSession.createOrUpdateTurtleSshSession as {
+    sessionSettings: Pick<SshSettings, 'idleTimeoutSeconds'>;
+  };
```
sjchmiela: hmmm i think the response should come out typed.

https://github.com/expo/eas-cli/blob/42a487e6951ebe7314be12c2ef44768dcb1f7f5c/packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts#L363

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+    json = JSON.parse(raw);
+  } catch {
+    return null;
+  }
+  const parsed = UptermSessionJsonZ.safeParse(json);
+  if (!parsed.success) {
+    return null;
+  }
+  return connectionConfigFromUptermSession(parsed.data);
+}
```
sjchmiela: looks like not used

packages/build-tools/src/utils/upterm.ts
sjchmiela:
```suggestion
function killProcessGroup(child: ChildProcess): void {
```
? and move to `processes.ts`?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+
+async function tryUptermOnPathAsync(env: Env): Promise<string | null> {
+  try {
+    // Use the job env (same PATH the host spawn gets), not process.env.
+    await spawn('upterm', ['version'], { stdio: 'pipe', env });
+    return 'upterm';
+  } catch {
+    return null;
+  }
+}
```
sjchmiela: inline

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+ * platform binary from gs://turtle-v2/upterm (same pattern as xclogparser).
+ */
+export async function resolveUptermPathAsync(env: Env): Promise<string> {
+  const onPath = await tryUptermOnPathAsync(env);
+  if (onPath) {
+    return onPath;
+  }
+
+  const objectName = resolveUptermGcsObjectName();
+  const cacheDir = path.join(os.tmpdir(), 'eas-upterm');
```
sjchmiela: i think we usually use mkdtemp

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+    // download below
+  }
+
+  await fs.mkdir(cacheDir, { recursive: true });
+  const url = `${UPTERM_GCS_BASE_URL}/${objectName}`;
+  try {
+    await downloadFile(url, cachePath, { retry: 3, timeout: UPTERM_DOWNLOAD_TIMEOUT_MS });
+    await fs.chmod(cachePath, 0o755);
+  } catch (err) {
+    await fs.rm(cachePath, { force: true }).catch(() => {});
```
sjchmiela: i keep going back and forth on whether it makes sense to remove or not

packages/build-tools/src/utils/upterm.ts
sjchmiela: good, let's add a comment

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+    // Missing socket is unknown, not proof of zero clients — a brief gap during
+    // redial must not look like "idle" to the post-job teardown path.
+    return null;
+  }
+  const session = await readCurrentSessionJsonAsync(uptermPath, adminSocketPath);
+  if (!session || typeof session.clientCount !== 'number') {
+    return null;
+  }
+  return session.clientCount;
+}
```
sjchmiela: i would move to the handle function returned from the host

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+
+async function getConnectedClientCountAsync(
+  uptermPath: string,
+  uptermSocketDir: string
+): Promise<number | null> {
+  const adminSocketPath = await findAdminSocketPathAsync(uptermSocketDir);
+  if (!adminSocketPath) {
+    // Missing socket is unknown, not proof of zero clients — a brief gap during
+    // redial must not look like "idle" to the post-job teardown path.
+    return null;
```
sjchmiela: should we throw error in this case and catch it somewhere higher up, where we could use `retry` or sth?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,371 @@
...
+        hostKeyPath,
+        '--force-command',
+        forceCommandPath,
+        '--',
+        'bash',
+        '-lc',
+        `sleep ${UPTERM_KEEPALIVE_SLEEP_SECONDS}`,
+      ],
+      {
+        env: { ...ctx.env, XDG_RUNTIME_DIR: stateDir },
```
sjchmiela, line 272: XDG_RUNTIME_DIR with a comment

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+    idleTimeoutSeconds,
+  };
+}
+
+/**
+ * Keeps the tunnel alive for as long as the session should stay reachable, and resolves once it
+ * should be torn down. The idle timeout only starts counting once the job has finished, so the
+ * tunnel is reachable for the whole job even when `idleTimeoutSeconds` is 0.
+ */
+export async function superviseSshSessionAsync({
```
sjchmiela, line 180: shouldn't this receive the `handle` as arugment?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+        );
+      } else if (connectedClientCount < previousClientCount) {
+        logger.info(
+          connectedClientCount === 0
+            ? 'The SSH client disconnected.'
+            : `An SSH client disconnected (${connectedClientCount} still connected).`
+        );
+      }
+      previousClientCount = connectedClientCount;
+    }
```
sjchmiela: this feels slightly misleading cause ti may happen that a user disconnects and another connects and we won't print anything because the count is still 1

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+        unknownSince = Date.now();
+      }
+    }
+
+    if (jobHasFinished && connectedClientCount === 0) {
+      if (idleTimeoutSeconds === 0) {
+        logger.info('The job finished and no SSH client is connected. Closing the session.');
+        return;
+      }
+      if (idleSince !== null && Date.now() - idleSince >= idleTimeoutMs) {
```
sjchmiela, line 241: we use idleTimeoutSeconds in one place and idleTimeoutMs in another, this feels wierd

also not sure if it makes sense to split these two?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,331 @@
...
+      unknownSince = null;
+    } else if (connectedClientCount === 0) {
+      unknownSince = null;
+      if (!jobHasFinished) {
+        idleSince = null;
+      } else if (idleSince === null) {
+        idleSince = Date.now();
+      }
+    } else {
+      // null = unknown: do not treat as idle (could still have a live client), but bound the wait.
```
sjchmiela: does this happen often? it's weird we have to handle it like this… maybe we could never return `null` from `getConnectedClientCount` and instead make it retry (inside) getting the count? idk

sjchmiela (changes requested):

packages/build-tools/src/utils/upterm.ts
gwdp: added, ty

packages/build-tools/src/utils/processes.ts
gwdp: Missed this one, ty

packages/build-tools/src/utils/processes.ts
gwdp: tightened it 🙇‍♂️

packages/build-tools/src/utils/upterm.ts
gwdp: Updated (killProcessGroup)

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: missed that after introducing the type, updated

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: removed it and kept `isSshEnabled` check only

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: all in for the bonus points

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: inlined, no hard preference on my side

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: pulled out, ty!

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: it comes back typed now through the gql mutation, stack with the type dependency not merged wasn't cool, I made a little mess, sorry

packages/build-tools/src/utils/upterm.ts
gwdp: cleaned up, ty for catching this

packages/build-tools/src/utils/upterm.ts
gwdp: moved, ty

packages/build-tools/src/utils/upterm.ts
gwdp: good call! moved to getConnectedClientCountAsync

packages/build-tools/src/utils/upterm.ts
gwdp: this one took me a bit but indeed cleaner, ty for flagging it

packages/build-tools/src/utils/upterm.ts
gwdp: Added

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: I prefer the external interface to be flat/simple, but fair point for consistency

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: collapsed to a single idle clock, one `idleSince` and one `idleTimeoutMs`

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: Great idea, that's what it does now; `getConnectedClientCountAsync` retries inside and throws if it still can't read, so no null and the supervisor doesn't have to reason about unknown counts

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: Good point,, but the limitation is that upterm `session current --output json` only hands back a count, it does len(connected_clients) server-side and drops the list, so from the worker a 1-for-1 swap looks identical to a steady connection. 

To log real join/leave events I think I can read upterm's connected_clients (they carry a public-key fingerprint) off its gRPC admin api instead of the CLI. I might defer now as a known limitation and go for a follow-up if that works for you?

packages/build-tools/src/utils/upterm.ts
gwdp: You're right, switched to mkdtemp and dropped the fixed cache dir

packages/build-tools/src/utils/upterm.ts
gwdp: with the move to mkdtemp there's no shared cache to worry about anymore

gwdp:
Applied changes but want to do a little more testing before requesting review.

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,245 @@
...
+      }
+    }
+  }
+`);
+
+export type TurtleSshTarget =
+  | { turtleJobRunId: string; turtleBuildId?: never }
+  | { turtleJobRunId?: never; turtleBuildId: string };
+
+export function isSshEnabled(job: Pick<Generic.Job, 'ssh'>): boolean {
```
sjchmiela:
```suggestion
export function isSshEnabled(job: Pick<Job, 'ssh'>): boolean {
```
?

https://github.com/expo/eas-cli/blob/789cf62bd6d7b0b4173ad3e790426ac6807b9731/packages/eas-build-job/src/job.ts#L9

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,245 @@
...
+  return job.ssh != null;
+}
+
+export function getSshIdleTimeoutSeconds(job: Pick<Generic.Job, 'ssh'>): number {
+  const idleTimeoutSeconds = job.ssh?.idleTimeoutSeconds;
+  if (idleTimeoutSeconds == null) {
+    throw new SystemError('SSH is enabled but the job is missing an idle timeout.', {
+      trackingCode: 'SSH_IDLE_TIMEOUT_MISSING',
+    });
+  }
```
sjchmiela: this relies on `www` providing a default value, right? should we default to 0?

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,245 @@
...
+  if (
+    !Number.isFinite(idleTimeoutSeconds) ||
+    !Number.isInteger(idleTimeoutSeconds) ||
+    idleTimeoutSeconds < 0 ||
+    idleTimeoutSeconds > MAX_SSH_IDLE_TIMEOUT_SECONDS
+  ) {
+    throw new SystemError(
+      `SSH idle timeout must be an integer between 0 and ${MAX_SSH_IDLE_TIMEOUT_SECONDS} seconds.`,
```
sjchmiela: nit: include what it was?

packages/build-tools/src/utils/turtleSshSession.ts
sjchmiela: same comment to all other places

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,245 @@
...
+  }: {
+    target: TurtleSshTarget;
+    connectionConfig: SshConnectionConfig & { reconnecting?: boolean };
+    idleTimeoutSeconds: number;
+  }
+): Promise<{ idleTimeoutSeconds: number }> {
+  const result = await ctx.graphqlClient
+    .mutation(CREATE_OR_UPDATE_TURTLE_SSH_SESSION_MUTATION, {
+      turtleJobRunId: target.turtleJobRunId ?? null,
+      turtleBuildId: target.turtleBuildId ?? null,
```
sjchmiela: just sharing -- what i recently did for logs was https://github.com/expo/universe/blob/1ae9b878798169e2de9e05af590e549061aa2f51/server/www/src/graphql/mutations/RealtimeLogsMutation.ts#L26-L34

i think it works nicely

packages/build-tools/src/utils/turtleSshSession.ts
```diff
@@ -0,0 +1,245 @@
...
+    throw err;
+  }
+
+  const ensureConnectedAsync = async (): Promise<void> => {
+    if (host.isAlive()) {
+      return;
+    }
+    for (let attempt = 1; attempt <= MAX_SSH_REDIALS; attempt++) {
+      try {
+        if (!host.isAlive()) {
```
sjchmiela, line 142: do we need this check?

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,310 @@
...
+  }
+  return { type: 'upterm-v1', host, secret: parsed.sessionId };
+}
+
+export function redactConnectionSecrets(text: string): string {
+  let redacted = text.replace(CONTROL_CHARACTERS, '');
+  for (const [, token] of redacted.matchAll(/upterm proxy wss?:\/\/([^@\s]+)@/g)) {
+    redacted = redacted.split(token).join('<redacted>');
+  }
+  return redacted.replace(/([a-z][a-z0-9+.-]*:\/\/)[^@\s/]+@/gi, '$1<redacted>@');
```
sjchmiela, line 93: this would benefit slightly from an explanation (what is the output, what are we catching, replacing)

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,310 @@
...
+export async function resolveUptermPathAsync(env: Env): Promise<string> {
+  try {
+    await spawn('upterm', ['version'], { stdio: 'pipe', env });
+    return 'upterm';
+  } catch {}
+
+  const objectName = resolveUptermGcsObjectName();
+  const downloadDir = await fs.mkdtemp(path.join(os.tmpdir(), 'eas-upterm-'));
+  const uptermPath = path.join(downloadDir, objectName);
+  const url = `${UPTERM_GCS_BASE_URL}/${objectName}`;
```
sjchmiela, line 124: we could also use cocoapods cache to proxy. maybe as a follow up

packages/build-tools/src/utils/upterm.ts
```diff
@@ -0,0 +1,310 @@
...
+  adminSocketPath: string
+): Promise<UptermSessionJson | null> {
+  try {
+    const result = await spawn(
+      uptermPath,
+      ['session', 'current', '--admin-socket', adminSocketPath, '--output', 'json'],
+      { stdio: 'pipe' }
+    );
+    const parsed = UptermSessionJsonZ.safeParse(JSON.parse(result.stdout));
+    return parsed.success ? parsed.data : null;
```
sjchmiela, line 158: should we capture an error to sentry if this fails?

sjchmiela (approved):

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: https://github.com/expo/eas-cli/pull/4030#discussion_r3747883751 I went checking to make sure I wasnt dropping the ball. Updated.

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: yeah it does, and www already defaults it to 0 itself, so throwing here was guarding a contract that already holds, updated, ty!

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: Added, ty

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: Hm... that's a good one, I think I do prefer your way 🤔 (baking a change)

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: it looks redundant but it earns its keep on retries. On attempt 1 it's always true, on later attempts the host was already redialed fine, and only the www call failed, so without the check we'd redial a healthy host every time round. Left it.

packages/build-tools/src/utils/upterm.ts
gwdp: added, ty

packages/build-tools/src/utils/upterm.ts
gwdp: tracking under https://linear.app/expo/issue/ENG-26047/serve-the-upterm-download-through-the-cache-proxy

packages/build-tools/src/utils/upterm.ts
gwdp: agreed there's a gap, but moved it up a level. This read has two callers and both retry on purpose, so a single failure is normal while the socket comes up or during a redial, and capturing here would fire on every poll. The real hole was `superviseSshSessionAsync`, which caught both terminal failures, logged a warn and quietly closed the session. Captures there now, ty for the nudge

packages/build-tools/src/utils/turtleSshSession.ts
gwdp: https://github.com/expo/universe/pull/30238 🙈

PR #4031, opened 2026-07-16
[worker] Wire SSH session into build

gwdp:
## Why

Wiring SSH into the worker build lifecycle: when a job has SSH enabled, open the session in `SSH_SESSION` and keep the VM up until supervision says it is safe to tear down.

## How

- Start the session early in `build()`, pass a `hasJobFinished` signal into the supervisor, and `await` teardown in `finally`
- Leave `SSH_SESSION` open in the log UI for the life of the tunnel (`doNotMarkEnd`, emit END on teardown)
- Opening the session is best-effort: failure warns and the job continues without SSH
- Upterm comes from PATH (image bake) or a runtime download from `gs://turtle-v2/upterm` via build-tools; this PR does not vendor binaries into the worker package

## Test Plan

Worker unit tests for the SSH phase. CI

### DIFF
--- a/packages/worker/package.sh
+++ b/packages/worker/package.sh
@@ -131,6 +131,38 @@ if [[ "$PLATFORM" == "ios" ]]; then
   popd >/dev/null 2>&1
 fi
 
+# Bake the upterm SSH client into the worker package (build-tools/bin/upterm-<arch>) for both
+# arches so jobs use the vendored binary instead of downloading it at runtime; the worker resolves
+# the right one via process.arch, so a mixed-arch worker pool is fine.
+upterm_version="0.24.0"
+case "$PLATFORM" in
+  ios) upterm_os="darwin" ;;
+  android) upterm_os="linux" ;;
+  *) echo "Unsupported platform for the upterm client: $PLATFORM" >&2; exit 1 ;;
+esac
+upterm_bin_dir="$target_root_dir/packages/build-tools/bin"
+mkdir -p "$upterm_bin_dir"
+for upterm_arch in amd64 arm64; do
+  case "${upterm_os}_${upterm_arch}" in
+    darwin_amd64) upterm_sha256="cc65a3c73e993ba22ef0fffd028dfb363529d71968380bf8f2a9804d53fe7f0a" ;;
+    darwin_arm64) upterm_sha256="aafa1330bb452abd308b78ed2cead7c8bdcb1aa2c486468fbd47dd4b567a8b30" ;;
+    linux_amd64) upterm_sha256="a6324d76c6d962236c22e4a27a2c4ffd17aef11dfa3da33d14dfcff4f4de60b3" ;;
+    linux_arm64) upterm_sha256="9015aabeeb837ef4c503db56a54a90139b5965a9f16f162b098b1c5acf8d1584" ;;
+  esac
+  upterm_tarball="$tmp_dir/upterm_${upterm_os}_${upterm_arch}.tar.gz"
+  curl -fsSL -o "$upterm_tarball" \
+    "https://github.com/owenthereal/upterm/releases/download/v${upterm_version}/upterm_${upterm_os}_${upterm_arch}.tar.gz"
+  if command -v sha256sum >/dev/null 2>&1; then
+    echo "${upterm_sha256}  ${upterm_tarball}" | sha256sum -c -
+  else
+    echo "${upterm_sha256}  ${upterm_tarball}" | shasum -a 256 -c -
+  fi
+  tar -xzf "$upterm_tarball" -C "$upterm_bin_dir" upterm
+  mv "$upterm_bin_dir/upterm" "$upterm_bin_dir/upterm-${upterm_arch}"
+  chmod +x "$upterm_bin_dir/upterm-${upterm_arch}"
+  rm -f "$upterm_tarball"
+done
+
 # Create backward-compatible symlink for orchestrator
 # The orchestrator expects ./src/services/worker/dist/main.js
 # but the new structure has ./packages/worker/dist/main.js
--- a/packages/worker/package.sh
+++ b/packages/worker/package.sh
@@ -131,38 +131,6 @@ if [[ "$PLATFORM" == "ios" ]]; then
   popd >/dev/null 2>&1
 fi
 
-# Bake the upterm SSH client into the worker package (build-tools/bin/upterm-<arch>) for both
-# arches so jobs use the vendored binary instead of downloading it at runtime; the worker resolves
-# the right one via process.arch, so a mixed-arch worker pool is fine.
-upterm_version="0.24.0"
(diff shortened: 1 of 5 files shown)


gwdp:
* **#4032** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4031** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4031?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4031?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4030** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4030?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/worker/package.sh
```diff
@@ -131,6 +131,38 @@ if [[ "$PLATFORM" == "ios" ]]; then
...
+  *) echo "Unsupported platform for the upterm client: $PLATFORM" >&2; exit 1 ;;
+esac
+upterm_bin_dir="$target_root_dir/packages/build-tools/bin"
+mkdir -p "$upterm_bin_dir"
+for upterm_arch in amd64 arm64; do
+  case "${upterm_os}_${upterm_arch}" in
+    darwin_amd64) upterm_sha256="cc65a3c73e993ba22ef0fffd028dfb363529d71968380bf8f2a9804d53fe7f0a" ;;
+    darwin_arm64) upterm_sha256="aafa1330bb452abd308b78ed2cead7c8bdcb1aa2c486468fbd47dd4b567a8b30" ;;
+    linux_amd64) upterm_sha256="a6324d76c6d962236c22e4a27a2c4ffd17aef11dfa3da33d14dfcff4f4de60b3" ;;
+    linux_arm64) upterm_sha256="9015aabeeb837ef4c503db56a54a90139b5965a9f16f162b098b1c5acf8d1584" ;;
```
sjchmiela: i think we only need to worry about darwin_arm64 and linux_amd64?

packages/worker/package.sh
```diff
@@ -131,6 +131,38 @@ if [[ "$PLATFORM" == "ios" ]]; then
...
+  esac
+  upterm_tarball="$tmp_dir/upterm_${upterm_os}_${upterm_arch}.tar.gz"
+  curl -fsSL -o "$upterm_tarball" \
+    "https://github.com/owenthereal/upterm/releases/download/v${upterm_version}/upterm_${upterm_os}_${upterm_arch}.tar.gz"
```
sjchmiela: let's put the binaries in https://console.cloud.google.com/storage/browser/turtle-v2/upterm?authuser=0&project=exponentjs&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)) and download them from there instead? we do the same with xclogparser (https://console.cloud.google.com/storage/browser/turtle-v2/xclogparser?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&authuser=0&project=exponentjs)

packages/worker/package.sh
```diff
@@ -131,6 +131,38 @@ if [[ "$PLATFORM" == "ios" ]]; then
   popd >/dev/null 2>&1
 fi
 
+# Bake the upterm SSH client into the worker package (build-tools/bin/upterm-<arch>) for both
+# arches so jobs use the vendored binary instead of downloading it at runtime; the worker resolves
+# the right one via process.arch, so a mixed-arch worker pool is fine.
```
sjchmiela: i…'m not sure it makes sense to bake upterm binaries into the worker?

- it makes worker larger which makes download and unpack slower (for all jobs, not only ones with ssh enabled)
- a single download in runtime shouldn't be a problem

instead i think we should be baking upterm into the images (and downloading if it's missing). looks like we already bake it in on linux https://github.com/expo/turtle-v2/blob/7681b8c2c9c325dce6fd9e32ad1e1bd7b4b19e8c/infra/vm-images/ansible/roles/common-android-worker/tasks/upterm.yml#L9-L11

sjchmiela (changes requested):
ugh sorry i had this review from 4 days ago not sent

packages/worker/package.sh
gwdp: Ty for the reasoning, makes total sense 🙇‍♂️ Moving towards the recommended approach!

packages/worker/package.sh
gwdp: Cleaned up, easy to add (linux arm if ever needed 👀 )

packages/worker/package.sh
gwdp: Added there (someone already experimented with it a while ago 😆 )

packages/worker/src/sshSession.ts
```diff
@@ -0,0 +1,99 @@
...
+  buildId: string;
+  logger: bunyan;
+  hasJobFinished: () => boolean;
+}): Promise<{ done?: Promise<void> }> {
+  let done: Promise<void> | undefined;
+
+  await ctx.runBuildPhase(
+    BuildPhase.SSH_SESSION,
+    async () => {
+      const phaseStartedAt = Date.now();
```
sjchmiela: use `performance.now` for durations

packages/worker/src/sshSession.ts
```diff
@@ -0,0 +1,99 @@
...
+          idleTimeoutSeconds === 0
+            ? 'It stays open for the whole job and closes once the job finishes and no client is connected.'
+            : `It stays open for the whole job, then closes after ${TurtleSshSession.formatSshIdleTimeoutForLog(idleTimeoutSeconds)} with no client connected. The worker stays up until then.`
+        );
+
+        const sshLogger = logger.child({ phase: BuildPhase.SSH_SESSION });
+        done = (async () => {
+          let result = BuildPhaseResult.SUCCESS;
+          try {
+            await TurtleSshSession.superviseSshSessionAsync({
```
sjchmiela, line 45: maybe the supervisesshsessionasync should be inlined here?

sjchmiela: or at least in this file

sjchmiela (approved):
this is good, i think the code layout due to the stack nature of the feature has grown a bit too wide, but it looks great

packages/worker/src/sshSession.ts
gwdp: Leaning toward keeping it as its own function on purpose. The full supervise loop has its own unit tests but also would bloat this file. Let me know if you are soft on this 🙇‍♂️

packages/worker/src/sshSession.ts
gwdp: Updated

PR #4032, opened 2026-07-16
[eas-cli] Add workflow:ssh command

gwdp:
# Why

Lets a developer open a shell on the worker running a workflow job:

```
eas workflow:ssh <workflow-job-id>
```

# How

Resolves the job's SSH session over the EAS API, waits for the worker to come online (spinner, up to ~5 min), then opens a direct `ssh` to the relay against a generated config.

- Tokens after the id run as a remote command instead of opening an interactive shell.
- `--show-connect` prints the connect commands instead of opening a session — the direct one plus an `upterm proxy` WebSocket fallback for networks that block it.
- Not-enabled, ended, still-pending, and unknown-id cases print a clear message and exit non-zero instead of a raw error.

# Test Plan

Unit tests cover argv parsing (`--show-connect` position, passthrough command) and connect-status resolution. Manually opened a session against a live local run and ran a remote command. CI passes.

### DIFF
--- a/packages/eas-cli/src/commands/workflow/ssh.ts
+++ b/packages/eas-cli/src/commands/workflow/ssh.ts
@@ -133,7 +133,7 @@ export default class WorkflowSsh extends EasCommand {
       loggedIn: { graphqlClient },
     } = await this.getContextAsync(WorkflowSsh, { nonInteractive: true });
 
-    const connectInfo = await WorkflowJobSshQuery.connectInfoForWorkflowJobAsync(
+    const connectInfo = await WorkflowJobSshQuery.connectInfoForResourceIdAsync(
       graphqlClient,
       resourceId
     );
@@ -234,7 +234,7 @@ async function waitForSessionToOpenAsync(
   const deadline = Date.now() + SESSION_OPEN_TIMEOUT_MS;
   try {
     while (Date.now() < deadline) {
-      const connectInfo = await WorkflowJobSshQuery.connectInfoForWorkflowJobAsync(
+      const connectInfo = await WorkflowJobSshQuery.connectInfoForResourceIdAsync(
         graphqlClient,
         workflowJobId
       );
--- a/packages/eas-cli/src/commands/workflow/ssh.ts
+++ b/packages/eas-cli/src/commands/workflow/ssh.ts
@@ -224,7 +224,6 @@ async function waitForSessionToOpenAsync(
   const deadline = Date.now() + SESSION_OPEN_TIMEOUT_MS;
   try {
     while (Date.now() < deadline) {
-      await sleepAsync(SESSION_OPEN_POLL_INTERVAL_MS);
       const connectInfo = await WorkflowJobSshQuery.connectInfoForWorkflowJobAsync(
         graphqlClient,
         workflowJobId
@@ -238,6 +237,7 @@ async function waitForSessionToOpenAsync(
         spinner.succeed('The ssh session is ready.');
         return connectInfo?.session?.connectionConfig ?? null;
       }
+      await sleepAsync(SESSION_OPEN_POLL_INTERVAL_MS);
     }
     spinner.fail(
       'Timed out waiting for the ssh session to open. The worker may still be starting up; try again in a moment.'
--- a/packages/eas-cli/src/commands/workflow/ssh.ts
+++ b/packages/eas-cli/src/commands/workflow/ssh.ts
@@ -15,15 +15,17 @@ import Log from '../../log';
 import { ora } from '../../ora';
 import { sleepAsync } from '../../utils/promise';
 
-export const CONNECTION_HOST_REGEX = /^[A-Za-z0-9.-]+(?::\d+)?$/;
-export const CONNECTION_SECRET_REGEX = /^[A-Za-z0-9._~:/+=-]+$/;
-
-export function splitConnectionHost(connectionHost: string): { host: string; port?: number } {
-  const match = connectionHost.match(/^(.+):(\d+)$/);
-  if (!match) {
-    return { host: connectionHost };
+export function splitConnectionHost(connectionHost: string): { hostname: string; port?: number } {
+  try {
+    const { hostname, port: stringPort } = new URL(`ssh://${connectionHost}`);
+    const port = stringPort === '' ? undefined : Number(stringPort);
+    return { hostname, port };
+  } catch (err) {
+    throw new Error(
+      'Unexpected connection host reported for this ssh session. Update eas-cli and try again, or contact support if it persists.',
+      { cause: err }
+    );
   }
-  return { host: match[1], port: Number(match[2]) };
 }
 
 const SSH_INSECURE_OPTS = '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null';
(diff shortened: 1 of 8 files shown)


gwdp:
* **#4032** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4032?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4032?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4031** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4031?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4030** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4030?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (session?.connectionConfig?.reconnecting) {
+    return 'pending';
+  }
+  if (!sshRequested) {
+    return 'not-enabled';
+  }
+  return jobCompleted ? 'ended' : 'pending';
+}
+
+export function parseSshArgv(rawArgv: readonly string[]): {
```
sjchmiela, line 83: oh no does oclif really not support this in a nicer way?

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+  const deadline = Date.now() + SESSION_OPEN_TIMEOUT_MS;
+  try {
+    while (Date.now() < deadline) {
+      const connectInfo = await WorkflowJobSshQuery.connectInfoForWorkflowJobAsync(
+        graphqlClient,
+        workflowJobId
+      );
+      const status = resolveSshConnectStatus(connectInfo);
+      if (status === 'unknown' || status === 'not-enabled' || status === 'ended') {
+        spinner.fail('The ssh session ended before it opened.');
```
sjchmiela: this message is not accurate for all statuses, especially for `not-enabled`

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+): Promise<NonNullable<WorkflowJobSshSession['connectionConfig']> | null> {
+  const spinner = ora('Waiting for the worker to open the ssh session').start();
+  const deadline = Date.now() + SESSION_OPEN_TIMEOUT_MS;
+  try {
+    while (Date.now() < deadline) {
+      const connectInfo = await WorkflowJobSshQuery.connectInfoForWorkflowJobAsync(
+        graphqlClient,
+        workflowJobId
+      );
+      const status = resolveSshConnectStatus(connectInfo);
```
sjchmiela, line 232: maybe a `switch` over `status` would be clearer?

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+        `SSH was not enabled for "${resourceId}". Start the run with \`eas workflow:run --ssh\` to enable it.`
+      );
+      process.exitCode = 1;
+      return;
+    }
+    if (status === 'ended') {
+      Log.error('This ssh session has ended.');
+      process.exitCode = 1;
+      return;
+    }
```
sjchmiela: this logic is duplicated in waitForSessionToOpenAsync

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (!CONNECTION_HOST_REGEX.test(connectionHost)) {
+      throw new Error(
+        'Unexpected connection host reported for this ssh session. Update eas-cli and try again, or contact support if it persists.'
+      );
+    }
+    if (!CONNECTION_SECRET_REGEX.test(secret)) {
+      throw new Error(
+        'Unexpected connection token reported for this ssh session. Update eas-cli and try again, or contact support if it persists.'
+      );
+    }
```
sjchmiela: why are we validating what www receives? shouldn't www validate inputs instead?

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+  WorkflowJobSshSession,
+} from '../../graphql/queries/WorkflowJobSshQuery';
+import Log from '../../log';
+import { ora } from '../../ora';
+import { sleepAsync } from '../../utils/promise';
+
+export const CONNECTION_HOST_REGEX = /^[A-Za-z0-9.-]+(?::\d+)?$/;
+export const CONNECTION_SECRET_REGEX = /^[A-Za-z0-9._~:/+=-]+$/;
+
+export function splitConnectionHost(connectionHost: string): { host: string; port?: number } {
```
sjchmiela: would `new URL` give us similar results?

packages/eas-cli/src/commands/workflow/ssh.ts
```diff
@@ -0,0 +1,250 @@
...
+      return;
+    }
+
+    const configDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'eas-workflow-ssh-'));
+    try {
+      const configPath = path.join(configDir, 'config');
+      await fs.promises.writeFile(
+        configPath,
+        [
+          'Host eas-workflow-ssh',
```
sjchmiela: let's include part of the resource id so terminal tabs get named uniquely?

packages/eas-cli/src/graphql/queries/WorkflowJobSshQuery.ts
```diff
@@ -0,0 +1,141 @@
...
+
+function toConnectInfo(
+  job: WorkflowJobSshPollQuery['workflowJobs']['byId']
+): WorkflowJobSshConnectInfo {
+  const hasTurtleTarget = job.turtleJobRun != null || job.turtleBuild != null;
+  return {
+    sshRequested:
+      job.type !== WorkflowJobType.GetBuild &&
+      hasTurtleTarget &&
+      job.workflowRun.sshSettings != null,
```
sjchmiela: eh

sjchmiela (approved):

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: Updated, ty for catching. The wait loop now uses the same per status message as the first check

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: yup, good call 🙇‍♂️

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: deduped, ty

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: ❤️ updated

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: It does support it just with some tradeoffs either way. Tried `'--': false` which lets us accept `workflow:ssh jr-1 ls -la` like real ssh but then we lose flag validation; So a typo like `--show-conect jr-1` silently becomes the job id instead of failing; The other option is the standard --, which means typing like `eas workflow:ssh jr-1 -- ls -la`

Keeping the small parser so we get the ssh-like form and still error on bad flags. Happy to switch to either if you are strong on this.

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: This was caught on my little sec review; it's guarding injection rather than input validation per say

packages/eas-cli/src/graphql/queries/WorkflowJobSshQuery.ts
gwdp: fair, chopped `GET_BUILD` out

packages/eas-cli/src/commands/workflow/ssh.ts
gwdp: We don't have a protocol at this string; a new URL would need me to fake a protocol. Keeping it if that works

AHGIJMKLKKZNPJKQR: I think using ``new URL(`ssh://${connectionHost}`)`` here (or `URL.parse`) might be better, because there are some weird edge cases like `host.name:`, which should return `{ host: "host.name" }`, but returns `{ host: "host.name:" }`.

PR #4035, opened 2026-07-17
[build-tools] Add `ref` input to `eas/checkout` function

sswrk:
# Why

Currently, `eas/checkout` has no manual `ref` customization, it's determined basing on the ref it's triggered. This is a limitation to teams that don't find this sufficient.

# How

Added an optional `ref` input to the `eas/checkout` function. When set, the step moves the cloned sources as before and then runs a new `fetchAndCheckoutRefAsync` git util: a shallow `git fetch origin <ref>` followed by a checkout of `FETCH_HEAD` (recreating the tag or branch name locally, since the initial clone may have already created it).

The input is only valid when project sources come from a git repository.

# Test Plan

Added unit tests.


packages/build-tools/src/common/git.ts
```diff
@@ -55,6 +57,43 @@ export async function shallowCloneRepositoryAsync({
...
+  ref,
+  repositoryDirectory,
+}: {
+  ref: string;
+  repositoryDirectory: string;
+}): Promise<void> {
+  if (!(await fs.pathExists(path.join(repositoryDirectory, '.git')))) {
+    throw new UserError(
+      'EAS_CHECKOUT_NOT_A_GIT_REPOSITORY',
+      `Cannot check out ref "${ref}": ${repositoryDirectory} is not a git repository.`
```
sjchmiela:
```suggestion
      `Cannot check out ref "${ref}": ${repositoryDirectory} is not a Git repository.`
```

packages/build-tools/src/common/git.ts
```diff
@@ -55,6 +57,43 @@ export async function shallowCloneRepositoryAsync({
...
+    );
+  }
+
+  const { name, type } = getStrippedBranchOrTagName(ref);
+  try {
+    await spawn('git', ['fetch', 'origin', '--depth', '1', '--no-tags', name], {
```
sjchmiela: supposedly if a tag and a branch share a name, a tag is going to get picked by git and we should ideally not strip `refs/...` if possible

packages/build-tools/src/common/git.ts
```diff
@@ -55,6 +57,43 @@ export async function shallowCloneRepositoryAsync({
...
+      cwd: repositoryDirectory,
+    });
+    if (type === 'tag') {
+      await spawn('git', ['checkout', 'FETCH_HEAD'], { cwd: repositoryDirectory });
+      // --force because the initial clone may have already created this tag.
+      await spawn('git', ['tag', '--force', name], { cwd: repositoryDirectory });
+    } else {
+      // -B because a branch with this name may exist from the initial clone.
+      await spawn('git', ['checkout', '-B', name, 'FETCH_HEAD'], { cwd: repositoryDirectory });
```
sjchmiela, line 90: supposedly a commit hash ref would be better as detached head

sjchmiela (approved):

PR #4041, opened 2026-07-19
[eas-cli] Add env:set and deprecate env:create and env:update

jonsamp:
# Why

We want to combine `eas env:create` and `eas env:update` into `eas env:set` for a nicer CLI command.

Linear: ENG-15351

# How

- Add `eas env:set` to create or update project and account environment variables.
- Adds `--name` and `--value` to `:get`, `:update`, and `:delete`, to make them consistent. `--variable-name` and other existing flags remain untouched so there aren't breaking changes.
- Preserve existing environment memberships and file variable types during updates.
- Add JSON output for automation.
- Hide `eas env:create` and `eas env:update` from help while keeping them available with deprecation warnings.

# Test Plan

Try these examples.

### Create, then update the same variable

```sh
eas env:set development --name EAS_ENV_SET_REVIEW --value first --visibility plaintext --scope project --non-interactive
eas env:set development --name EAS_ENV_SET_REVIEW --value second --visibility plaintext --scope project --non-interactive
eas env:get development --variable-name EAS_ENV_SET_REVIEW --scope project --non-interactive
```

The first command should report that it created the variable, the second should report that it updated it, and `env:get` should print `EAS_ENV_SET_REVIEW=second`.

### Verify JSON output

```sh
eas env:set preview --name EAS_ENV_SET_JSON_REVIEW --value json-value --visibility plaintext --scope project --json
```

Standard output should contain only valid JSON for the created variable, including its name, value, and `preview` environment.

### DIFF
--- a/packages/eas-cli/src/commands/env/set.ts
+++ b/packages/eas-cli/src/commands/env/set.ts
@@ -9,11 +9,13 @@ import {
   EASEnvironmentVariableScopeFlag,
   EASEnvironmentVariableScopeFlagValue,
   EASMultiEnvironmentFlag,
-  EASNonInteractiveFlag,
   EASVariableVisibilityFlag,
+  EasNonInteractiveAndJsonFlags,
+  resolveNonInteractiveAndJsonFlags,
 } from '../../commandUtils/flags';
 import {
   EnvironmentSecretType,
+  EnvironmentVariableFragment,
   EnvironmentVariableScope,
   EnvironmentVariableVisibility,
 } from '../../graphql/generated';
@@ -24,6 +26,7 @@ import {
   getDisplayNameForProjectIdAsync,
   getOwnerAccountForProjectIdAsync,
 } from '../../project/projectUtils';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
 import {
   parseVisibility,
   promptVariableEnvironmentAsync,
@@ -41,6 +44,7 @@ interface RawSetFlags {
   scope: EASEnvironmentVariableScopeFlagValue;
   environment?: string[];
   'non-interactive': boolean;
+  json: boolean;
 }
 
 interface SetFlags {
@@ -51,6 +55,14 @@ interface SetFlags {
   scope: EnvironmentVariableScope;
   environment?: string[];
   'non-interactive': boolean;
+  json: boolean;
+}
+
+interface ResolvedVariableDetails {
+  value: string;
+  visibility: EnvironmentVariableVisibility;
+  type: EnvironmentSecretType | undefined;
+  fileName: string | undefined;
 }
 
 export default class EnvSet extends EasCommand {
@@ -70,7 +82,7 @@ export default class EnvSet extends EasCommand {
       description: 'Name of the variable',
     }),
     value: Flags.string({
-      description: 'Text value or the variable',
+      description: 'Text value of the variable',
     }),
     type: Flags.option({
       description: 'The type of variable',
@@ -79,7 +91,7 @@ export default class EnvSet extends EasCommand {
     ...EASVariableVisibilityFlag,
     ...EASEnvironmentVariableScopeFlag,
     ...EASMultiEnvironmentFlag,
-    ...EASNonInteractiveFlag,
+    ...EasNonInteractiveAndJsonFlags,
   };
 
   static override contextDefinition = {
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/commands/env/set.ts
```diff
@@ -0,0 +1,387 @@
...
+      newType = EnvironmentSecretType.String;
+    }
+
+    let newVisibility = visibility ? parseVisibility(visibility) : undefined;
+
+    if (!type && !value && !nonInteractive) {
+      newType = await promptVariableTypeAsync(nonInteractive, existingVariable?.type);
+    }
+
+    if (!newVisibility) {
```
douglowder: This is going to fail if in non-interactive mode and visibility is not defined. It should also be guarded by `!nonInteractive`)

packages/eas-cli/src/commands/env/set.ts
```diff
@@ -0,0 +1,387 @@
...
+
+    if (!value) {
+      value = await promptVariableValueAsync({
+        nonInteractive,
+        hidden: newVisibility !== EnvironmentVariableVisibility.Public,
+        filePath: (newType ?? existingVariable?.type) === EnvironmentSecretType.FileBase64,
+      });
+    }
+
+    newVisibility = newVisibility ?? EnvironmentVariableVisibility.Public;
```
douglowder: You should do `(newVisibility ?? existingVariable?.visibility)` here

douglowder (changes requested):
See inline.

douglowder (commented):
## Overview
Adds `eas env:set` as a create-or-update (upsert) for project/account environment variables, and hides `env:create`/`env:update` behind deprecation warnings. `set.ts` looks up an existing variable by name + overlapping-environment, then routes to `updateAsync` or `createFor{App,SharedVariable}Async`, merging environments and preserving file type. JSON output added. Deprecations are clean (`hidden = true` + `Log.warn`), and CHANGELOG/README are updated per convention. Overall the structure is solid and readable.

## Correctness

**🔴 Blocking — non-interactive update requires `--visibility` and otherwise resets visibility (`set.ts:334`, `set.ts:353`).**
`resolveVariableDetailsAsync` calls `promptVariableVisibilityAsync(nonInteractive, existing?.visibility)` unconditionally when `--visibility` is omitted, and `promptVariableVisibilityAsync` **throws in non-interactive mode**. So:
```sh
eas env:set development --name FOO --value newval --scope project --non-interactive
```
on an existing `FOO` errors with *"The `--visibility` flag must be set…"*, even though the goal is just to change the value.

This is a **regression** vs `env:update`, which guards all prompting behind `if (!nonInteractive)` (`update.ts:276`) and leaves `visibility` undefined so the mutation **preserves** the existing value. It also contradicts this PR's stated goal ("Preserve existing environment memberships and file variable types") — visibility is the one attribute *not* preserved. Even interactively, `set.ts:353`'s `?? Public` means an omitted visibility resolves to a concrete value rather than "keep existing."

**Suggested fix:** when `existingVariable` is present and `--visibility` isn't provided, default to `existingVariable.visibility` (skip the prompt/throw) — mirroring how type is preserved at `set.ts:342`. The same reasoning applies to prompting for `type`/`value` on update, but visibility is the one that hard-fails.

**🟡 `findExistingVariableAsync` can silently create a duplicate (`set.ts:288`).**
It only treats a variable as "existing" when its environments overlap the requested ones. `eas env:set production --name FOO` when `FOO` exists only in `development` → no overlap → falls to the **create** path. Depending on the API's name-uniqueness rules that either errors confusingly or creates a second same-named variable. Worth confirming this matches intended behavior (and ideally covering it with a test).

**🟡 Merge-only environments.** `environmentsToSet` (`set.ts:143`) only unions environments — `set` can add an environment to a variable but never remove one. Probably intended for upsert semantics, but worth a note in the help text so users don't expect `set` to scope down.

## Test coverage
Good breadth (create/update × project/account, environment preservation, positional env arg, file-type preservation, invalid file path, JSON output, aliases). Two gaps, both tied to the bug above:
- **No test updates an existing variable without `--visibility`** — every update test passes `--visibility secret`, so the failing/reset path is invisible. Add a non-interactive update-without-`--visibility` test asserting the existing visibility is preserved (it will currently fail, catching the bug).
- No test for the non-overlapping-environment same-name case (the `findExistingVariableAsync` create-vs-update decision).

## Minor / style
- `sanitizeFlags` (`set.ts:378`) spreads `RawSetFlags` into `SetFlags`, carrying `json`/`non-interactive` that are then re-derived via `resolveNonInteractiveAndJsonFlags` and never read off the sanitized object — mild redundancy.
- The Project vs Shared branches (`set.ts:162-236`) duplicate the create/update-then-report block; could be factored to reduce ~40 lines, though it's readable as-is.
- `findExistingVariableAsync` filters `scope === Project` on the project branch but has no matching filter on the shared branch (`set.ts:300`) — harmless if `sharedAsync` only returns shared vars, but the asymmetry invites confusion.

## PR description accuracy
The body says it "Adds `--name` and `--value` to `:get`, `:update`, and `:delete`." The diff doesn't touch `get.ts`/`delete.ts`, and `update.ts` only receives the deprecation — only `set.ts` gains `--name`/`--value` (with `variable-name`/`variable-value` aliases). Either those changes are missing or the description should be corrected.

## Verdict
Nice consolidation and a clean deprecation path. Worth resolving the **non-interactive visibility regression** (fix + a test) before merge; the duplicate-create edge and the description mismatch are worth addressing but non-blocking.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

douglowder (approved):

PR #4042, opened 2026-07-19
[eas-cli] Handle new server-side errors thrown when observe features are blocked

douglowder:
# Why

New server code for observe features may throw errors when a feature is not available for an account.

# How

Handle these errors correctly, and surface the correct server-side error message.

# Test Plan

- Testing against accounts in staging
- New unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [eas-cli] Handle new server-side errors thrown when observe features are blocked. ([#4042](https://github.com/expo/eas-cli/pull/4042) by [@douglowder](https://github.com/douglowder))
 - [eas-cli] Add `eas project:icon:set` to upload a project icon from the command line. ([#4068](https://github.com/expo/eas-cli/pull/4068) by [@brentvatne](https://github.com/brentvatne))
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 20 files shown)


douglowder:
* **#4042** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4042?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4042?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

brentvatne:
One gap in the plan-gate handling: `fetchTotalEventCountAsync` still swallows the error and returns `0`.

This PR rethrows `isObservePlanGateError` in the per-platform fetch helpers, but `fetchTotalEventCountAsync` (`packages/eas-cli/src/observe/fetchEvents.ts:94`) has its own per-platform catch that isn't updated:

```ts
try {
  const versions = await ObserveQuery.appVersionsAsync(graphqlClient, { ... });
  return versions.reduce(...);
} catch {
  return 0;   // <- swallows plan-gate errors too
}
```

So if the server returns the plan-gate error from the `appVersions` (total-count) query while the events query succeeds, the `Promise.all` wrapper never sees it and the command renders a misleading `0 total events` instead of the upgrade message.

Suggested fix: rethrow the gate error here as well, matching the other helpers:

```ts
} catch (error) {
  if (isObservePlanGateError(error)) {
    throw error;
  }
  return 0;
}
```

_(Flagged by the experimental AI reviewer; sharing since it looks like a real inconsistency with the rest of the PR.)_

kadikraman (approved):
Checked against an account without a plan, and seeing the errors as expected ✔️

douglowder:
> One gap in the plan-gate handling: fetchTotalEventCountAsync still swallows the error and returns 0.

This finding was valid, and is now fixed.

PR #4043, opened 2026-07-20
[eas-cli] Prompt for platform in simulator:start instead of erroring

szdziedzic:
# Why

Running `eas simulator:start` without `--platform` failed with `Missing required flag platform`. Since the command already has an interactive session-wait flow, it's friendlier to prompt the user to pick a platform rather than erroring out.

# How

- Made the `--platform` flag optional in `eas simulator:start`.
- Added a `resolvePlatformAsync` helper that:
  - Maps the flag value to `AppPlatform` when `--platform` is provided.
  - Shows a `Select platform` dropdown (Android / iOS) when the flag is omitted in interactive mode, matching the existing pattern used by `eas build:run` and `src/platform.ts`.
  - Throws a clear error (`The --platform flag must be set when running in non-interactive mode.`) when the flag is omitted under `--non-interactive`/`--json`, since a prompt can't be shown without a TTY.

# Test Plan

Added unit tests in `start.test.ts`:
- Prompts to select the platform when `--platform` is omitted, and the selected platform flows into the session mutation.
- Throws (without prompting) when `--platform` is omitted in non-interactive mode.

All simulator `start` tests pass; `yarn typecheck` and `yarn lint` are clean.

### DIFF
--- a/packages/eas-cli/src/commands/simulator/start.ts
+++ b/packages/eas-cli/src/commands/simulator/start.ts
@@ -42,6 +42,12 @@ const OUT_CONFIG_TYPE_VALUES = {
   Env: 'env',
   Dotenv: 'dotenv',
 } as const;
+const PLATFORM_FLAG_VALUES = ['android', 'ios'] as const;
+type PlatformFlagValue = (typeof PLATFORM_FLAG_VALUES)[number];
+const APP_PLATFORM_BY_FLAG_VALUE: Record<PlatformFlagValue, AppPlatform> = {
+  android: AppPlatform.Android,
+  ios: AppPlatform.Ios,
+};
 
 export default class SimulatorStart extends EasCommand {
   static override hidden = true;
@@ -52,7 +58,7 @@ export default class SimulatorStart extends EasCommand {
     platform: Flags.option({
       char: 'p',
       description: 'Device platform',
-      options: ['android', 'ios'] as const,
+      options: PLATFORM_FLAG_VALUES,
     })(),
     type: Flags.option({
       description: 'Type of simulator session to create',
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -120,7 +120,7 @@ export default class SimulatorStart extends EasCommand {
       Log.newLine();
     }
 
-    const platform = flags.platform === 'android' ? AppPlatform.Android : AppPlatform.Ios;
+    const platform = await resolvePlatformAsync(flags.platform, nonInteractive);
```
gwdp: personal preference, not blocker; Could we prompt before we possibly show `Overwriting previous simulator session`?

gwdp (approved):

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -243,6 +243,30 @@ export default class SimulatorStart extends EasCommand {
   }
 }
 
+async function resolvePlatformAsync(
+  platform: 'android' | 'ios' | undefined,
+  nonInteractive: boolean
+): Promise<AppPlatform> {
+  if (platform) {
+    return platform === 'android' ? AppPlatform.Android : AppPlatform.Ios;
```
sjchmiela: Do we want `--platform szymon` to become an iOS session?

sjchmiela (approved):
➕ to Gabe's comment

packages/eas-cli/src/commands/simulator/start.ts
szdziedzic: Good call - moved the platform prompt above the overwrite warning. Thanks!

packages/eas-cli/src/commands/simulator/start.ts
szdziedzic: Oclif's `options: ['android', 'ios']` already rejects anything else at parse time, so `szymon` never reaches this code. But you're right that the ternary read like it silently mapped unknowns to iOS, so I replaced it with an explicit exhaustive `Record<PlatformFlagValue, AppPlatform>` map, and the flag `options` now reuse the same tuple so the two can't drift apart.

PR #4044, opened 2026-07-20
[build-tools] run workflow hooks around managed native build phases

hSATAC:
# Why

Managed native builds (`type: build` without custom `steps:`) accept `before_*` / `after_*` hooks in the payload but silently drop them today. This runs them, starting with the `install_node_modules` anchor, so native builds match the steps world.

# How

Hooks execute as real steps through the engine's public hook API, wrapped around the native install phase: before-hooks, the phase, then after-hooks, with the phase error taking precedence over an after-hook failure. They are parsed once during setup, after profile-env resolution and before the install phase. Only `install_node_modules` is wrapped.

# Test Plan

- CI passes.
- Tested locally.


packages/build-tools/src/common/setup.ts
```diff
@@ -81,32 +99,48 @@ export async function setupAsync<TJob extends BuildJob>(ctx: BuildContext<TJob>)
...
+    ctx,
+    hooks: jobHooksRef.current,
+    buildPhase: BuildPhase.INSTALL_DEPENDENCIES,
+    anchor: 'install_node_modules',
+    fn: async () => {
+      // Read package.json fresh here rather than reusing the READ_PACKAGE_JSON
+      // capture: a before_install_node_modules hook may have patched it, and the
+      // steps-world install already reads it fresh. Silent read — READ_PACKAGE_JSON
+      // stays the only logging site.
+      const freshPackageJson = readPackageJson(ctx.getReactNativeProjectDirectory());
```
sjchmiela, line 121: can we do it as a separate change?

packages/eas-build-job/src/errors.ts
```diff
@@ -8,6 +8,7 @@ export enum ErrorCode {
   UNKNOWN_FASTLANE_RESIGN_ERROR = 'EAS_BUILD_UNKNOWN_FASTLANE_RESIGN_ERROR',
   UNKNOWN_GRADLE_ERROR = 'EAS_BUILD_UNKNOWN_GRADLE_ERROR',
   BUILD_TIMEOUT_ERROR = 'EAS_BUILD_TIMEOUT_ERROR',
+  HOOKS_ERROR = 'EAS_BUILD_HOOKS_ERROR',
```
sjchmiela, line 11: fwiw i think this enum isn't used too widely and consistently

packages/build-tools/src/common/jobHooks.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  const customBuildContext = new CustomBuildContext(ctx);
+  const globalContext = new BuildStepGlobalContext(customBuildContext, false);
+  // Hook steps run in the checked-out project directory, not the pre-checkout
+  // target directory the global context defaults to.
+  globalContext.markAsCheckedOut(ctx.logger);
+
+  const externalFunctions = getEasFunctions(customBuildContext);
+  const externalFunctionGroups = getEasFunctionGroups(customBuildContext);
```
sjchmiela, line 56: this will require coordination with @sswrk soon since he's working on custom actions/functions

packages/build-tools/src/builders/ios.ts
```diff
@@ -49,7 +49,22 @@ export default async function iosBuilder(ctx: BuildContext<Ios.Job>): Promise<Ar
 }
 
 async function buildAsync(ctx: BuildContext<Ios.Job>): Promise<void> {
-  await setupAsync(ctx);
+  const jobHooksRef: JobHooksRef = { current: null };
```
sjchmiela, line 52: a worthwhile improvement probably would have been to merge the different step contexts into… less of different subclasses so we can pass one context everywhere and don't need this ref

packages/build-tools/src/builders/ios.ts
```diff
@@ -49,7 +49,22 @@ export default async function iosBuilder(ctx: BuildContext<Ios.Job>): Promise<Ar
...
+  try {
+    await buildInnerAsync(ctx, jobHooksRef);
+  } finally {
+    // Native builds have no other drain path for the hook context's queued
+    // step-metric uploads. Must wrap the WHOLE body — iOS's existing inner
+    // try/finally only starts after setupAsync returns.
+    await jobHooksRef.current?.customBuildContext.drainPendingMetricUploads();
```
sjchmiela, line 59: in launcher there was this notion of `ExitHandler`, feels like we're missing something like it (or, again, a single context)

sjchmiela (approved):

packages/build-tools/src/common/setup.ts
hSATAC: I saw this as part of the hooks change. Since `before_install_node_modules` can modify `package.json`, reusing the value from `READ_PACKAGE_JSON` could leave `shouldUseFrozenLockfile` and the doctor check with stale data. On its own this re-read doesn't change behavior, so keeping it in the same PR felt more natural. Was there a reason you preferred splitting it out?

packages/build-tools/src/builders/ios.ts
hSATAC: Agreed on consolidating all the contexts but that's probably is a standalone refactor task on it's own.

packages/build-tools/src/common/jobHooks.ts
hSATAC: Let me merge this and check what @sswrk is working on.

sswrk: This is fine, as for custom functions V1 I'm not adding support for calling the functions within hooks (and no support for attaching hooks to custom functions too). And if we wanted to add support for custom functions in hooks in the future, I think it would be fine too.

PR #4045, opened 2026-07-20
[eas-cli] add simulator:availability command

gwdp:
# Why

Tooling and agents need a way to check whether EAS Simulator is available for an account before invoking it, so they can check up front and degrade gracefully instead of failing partway through (some flows build on EAS first, so failing only at `simulator:start` means waiting through a full build to hit the wall).

# How

Adds a hidden `eas simulator:availability` command that reads the current project owner account's EAS Simulator availability via the existing `accountFeatureGates` field and prints whether it's enabled, with `--json`. No schema change.

# Test Plan

Unit tests cover the enabled/disabled JSON output and the graceful not-enabled message.

### DIFF
--- a/packages/eas-cli/src/graphql/queries/DeviceRunSessionAvailabilityQuery.ts
+++ b/packages/eas-cli/src/graphql/queries/DeviceRunSessionAvailabilityQuery.ts
@@ -15,20 +15,20 @@ export const DeviceRunSessionAvailabilityQuery = {
       graphqlClient
         .query<SimulatorAvailabilityQuery, SimulatorAvailabilityQueryVariables>(
           gql`
-            query SimulatorAvailabilityQuery($appId: String!) {
+            query SimulatorAvailabilityQuery($appId: String!, $filter: [String!]) {
               app {
                 byId(appId: $appId) {
                   id
                   ownerAccount {
                     id
                     name
-                    accountFeatureGates(filter: ["device-run-sessions"])
+                    accountFeatureGates(filter: $filter)
                   }
                 }
               }
             }
           `,
-          { appId },
+          { appId, filter: [DEVICE_RUN_SESSIONS_GATE] },
           { additionalTypenames: ['Account'] }
         )
         .toPromise()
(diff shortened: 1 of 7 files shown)


szdziedzic (commented):
Do we need separate command for this? Can we just catch feature gate error on `eas simulator:start` and wrap it with nicer error message we produce here? I bet agent can figure out that EAS Sim is not enabled based on this info

What do you think?

gwdp:
> Do we need separate command for this? Can we just catch feature gate error on `eas simulator:start` and wrap it with nicer error message we produce here? I bet agent can figure out that EAS Sim is not enabled based on this info
> 
> What do you think?

The reason for the separate command is failing fast — a couple of skill modes have you do an EAS build (~10-20 min) before `simulator:start` is ever reached, so catching it only at start means waiting through a whole build just to hit the wall at the end. Let me know how do you feel about it

packages/eas-cli/src/graphql/queries/DeviceRunSessionAvailabilityQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+        .query<SimulatorAvailabilityQuery, SimulatorAvailabilityQueryVariables>(
+          gql`
+            query SimulatorAvailabilityQuery($appId: String!) {
+              app {
+                byId(appId: $appId) {
+                  id
+                  ownerAccount {
+                    id
+                    name
+                    accountFeatureGates(filter: ["device-run-sessions"])
```
sjchmiela: nit: can we reuse DEVICE_RUN_SESSIONS_GATE here?

sjchmiela (approved):

packages/eas-cli/src/graphql/queries/DeviceRunSessionAvailabilityQuery.ts
gwdp: 100%, my bad

szdziedzic (approved):

PR #4047, opened 2026-07-20
[worker] set EAS_JOB_TYPE for builds and job runs

gwdp:
# Why

`eas update` run inside an EAS Build 500s because the publish sends the build id in a job-run-only field. The CLI needs a way to tell a build apart from a job run at publish time.

# How

Set `EAS_JOB_TYPE` (`build` | `jobRun`) in the build environment, derived from the dispatch message's existing job type.

Groundwork for #4048.

# Test Plan

Unit tests cover both job types. CI passes.

### DIFF
--- a/packages/local-build-plugin/src/build.ts
+++ b/packages/local-build-plugin/src/build.ts
@@ -32,6 +32,7 @@ export async function buildAsync(job: BuildJob, metadata: Metadata): Promise<voi
       EAS_BUILD: '1',
       EAS_BUILD_RUNNER: 'local-build-plugin',
       EAS_BUILD_PLATFORM: job.platform,
+      EAS_JOB_TYPE: 'build',
       EAS_BUILD_WORKINGDIR: path.join(workingdir, 'build'),
       EAS_BUILD_PROFILE: metadata.buildProfile,
       EAS_BUILD_GIT_COMMIT_HASH: metadata.gitCommitHash,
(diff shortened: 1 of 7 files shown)


gwdp:
* **#4047** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4047?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4047?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

gwdp:
@claude review always

sjchmiela (changes requested):
How about we used EAS_BUILD_PLATFORM? We already use platform to differentiate between builds and job runs in a few places.

gwdp:
> How about we used EAS_BUILD_PLATFORM? We already use platform to differentiate between builds and job runs in a few places.

Closing in favor of #4048 only 🙏

Took your suggestion — reusing the existing `EAS_BUILD_PLATFORM` instead of adding `EAS_JOB_TYPE`. That drops the worker change entirely, so the fix is now just the one CLI PR.

Not a huge fan of the implicit signal (you have to know job runs don't carry a platform to get why we're checking it), but reusing what's already there wins. Thanks for the nudge!

PR #4054, opened 2026-07-21
[eas-cli] Prompt for account when linking unconfigured project

williamgrosset:
# Why

When a project has no `extra.eas.projectId` and no `owner` in its app config, project-scoped commands (`eas build`, `eas submit`, `eas update`, etc.) silently linked or created the EAS project under the user's personal account. For users in multiple accounts, this often put projects under the wrong account.

# How

- Replaced the silent personal-account fallback with the same account selection `eas init` uses:
  - `owner` set or single account: no prompt, behavior unchanged
  - multiple accounts (interactive): prompt "Which account should own this project?"
  - non-interactive or robot user: same errors as before
- Extracted `getAccountChoices` and the ViewOnly-permission filter into a shared `src/project/accountSelection.ts` module, and sorted account choices by creation date, newest first (added `createdAt` to the `CurrentUser` query). This also applies to the `eas init` prompt.

# Test Plan

Unit tests added: `accountSelection-test.ts` (choice ordering and personal/team/organization annotations) and `getProjectIdAsync-test.ts` (prompt, `owner` field, single-account, non-interactive, and robot paths).

In a fresh app with no `extra.eas.projectId` and no `owner` in `app.json`, prompts for account selection:

https://github.com/user-attachments/assets/1dddb0f1-641d-4796-9dd9-fb76f60c87d8

With `owner` set in `app.json`, no prompt and uses that account:

https://github.com/user-attachments/assets/d2311795-413c-42e6-b2d3-d5061e5565c4

Also verified:
- `eas init` shows the same newest-first ordering in its account prompt (shared `getAccountChoices`)
- `--non-interactive` with multiple accounts and no `owner` still fails with the "run `eas init`" error


brentvatne:
## 🤖 AI code review

**Decision:** Approve

All specialist reviewers (correctness, security, and consistency) returned no findings. The PR is clean with no critical or warning issues identified.

No findings.

---
_This review is advisory — it never blocks a merge and never auto-approves._

packages/eas-cli/src/project/accountSelection.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function getAccountChoices(
+  actor: Actor,
+  namesWithSufficientPermissions: Set<string>
+): Choice[] {
+  // sorted by account creation date from newest to oldest
+  const sortedAccounts = [...actor.accounts].sort(
+    (a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+  );
```
brentvatne: - i really don't like this approach, in part because the sorting criteria is not a known value to the user, so it looks arbitrary. when you join a new org, it will just appear in some random position in the list. it is better to sort it by data that is visible to the user
- i also don't like it because it is inconsistent with how we sort accounts elsewhere
- on the website account selector we do: personal account, team accounts, organization accounts. within team accounts and org accounts, they're sorted alphabetically. https://github.com/expo/universe/blob/b3e7591d40b70dc79a776fe21a9ce2eb35c2ec0c/server/website/ui/components/SidebarNavigation/components/AccountSelectorContent.tsx#L57-L93. it'd be great to update to using this same convention here if we aren't doing it already. (edit: oh i see we were doing that before and this code changes that. let's revert this change).

brentvatne (changes requested):

packages/eas-cli/src/project/accountSelection.ts
williamgrosset: Good points. I added the sorting in the [last commit](https://github.com/expo/eas-cli/pull/4054/changes/973a1eb60654b96fc92d89284619b1beacc9eb10). The idea was that first-selection bias was affecting people's choice of selecting the personal account instead of the org they recently created. I'm good with reverting back to previous behavior and discussing another approach if it's a consistent problem.

williamgrosset:
@brentvatne Account selector now uses the original sorting behavior:

https://github.com/user-attachments/assets/39c1831d-100a-4d7c-a3e9-96fb0dc974d0

brentvatne (approved):

PR #4057, opened 2026-07-22
[eas-cli] Add --account flag to eas init

williamgrosset:
# Why

There's no way to choose the owning account when configuring an EAS project non-interactively. `eas init --non-interactive` fails unless `owner` is pre-set in the app config, and `--force` silently falls back to a default account. The related error messages also didn't explain how to fix this.

# How

- Added `--account <name>` to `eas init` to name the owning account. Unattended bootstrap: `eas init --account <name> --non-interactive`
- Since `--account` fully specifies the project (`@account/slug`), no confirmation (`--force`) is required to create or link it non-interactively.  `--force` is still needed to overwrite conflicting existing state
- Errors if the account is unknown, inaccessible, or conflicts with the `owner` field
- Rewrote the non-interactive "project not configured" errors (init, build/update/etc., robot tokens) to show the exact command and the accounts you can create projects in, newest-first
- Robot tokens can now bootstrap projects non-interactively

# Test Plan

Unit tests added: `--account` paths in `init.test.ts` (create without prompting, non-interactive create/link without `--force`, missing `--account`, unknown account, owner conflict) and list ordering in `User-test.ts`.

`eas init --account <org> --non-interactive` creates the project under the org:

https://github.com/user-attachments/assets/5f71b0b1-d085-4439-89b7-55566f468be6

`eas build --non-interactive` on an unlinked project prints the new guidance:

https://github.com/user-attachments/assets/38a2f8c8-94e1-40cd-96f5-d52d21607511

### DIFF
--- a/packages/eas-cli/README.md
+++ b/packages/eas-cli/README.md
@@ -1627,7 +1627,7 @@ USAGE
   $ eas init [--account <value> | --id <value>] [--force] [--non-interactive]
 
 FLAGS
-  --account=<value>  Name of the account that should own the project
+  --account=<value>  Name of the account that will own the project
   --force            Whether to create a new project/link an existing project without additional prompts or overwrite
                      any existing project ID when running with --id flag
   --id=<value>       ID of the EAS project to link
@@ -2263,7 +2263,7 @@ USAGE
   $ eas project:init [--account <value> | --id <value>] [--force] [--non-interactive]
 
 FLAGS
-  --account=<value>  Name of the account that should own the project
+  --account=<value>  Name of the account that will own the project
   --force            Whether to create a new project/link an existing project without additional prompts or overwrite
                      any existing project ID when running with --id flag
   --id=<value>       ID of the EAS project to link
--- a/packages/eas-cli/README.md
+++ b/packages/eas-cli/README.md
@@ -1624,9 +1624,10 @@ create or link an EAS project
 
 ```
 USAGE
-  $ eas init [--id <value>] [--force] [--non-interactive]
+  $ eas init [--account <value> | --id <value>] [--force] [--non-interactive]
 
 FLAGS
+  --account=<value>  Name of the account that should own the project
   --force            Whether to create a new project/link an existing project without additional prompts or overwrite
                      any existing project ID when running with --id flag
   --id=<value>       ID of the EAS project to link
@@ -2259,9 +2260,10 @@ create or link an EAS project
 
 ```
 USAGE
-  $ eas project:init [--id <value>] [--force] [--non-interactive]
+  $ eas project:init [--account <value> | --id <value>] [--force] [--non-interactive]
 
 FLAGS
+  --account=<value>  Name of the account that should own the project
   --force            Whether to create a new project/link an existing project without additional prompts or overwrite
                      any existing project ID when running with --id flag
   --id=<value>       ID of the EAS project to link
--- a/packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
+++ b/packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
@@ -39,7 +39,7 @@ export async function fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsyn
 
   if (options.nonInteractive) {
     throw new Error(
-      `EAS project not configured. To configure it non-interactively, choose the account that should own the project and run:\n\n  eas init --account <name> --force --non-interactive\n\nAccounts you can create projects in: ${getCreatableAccountNamesNewestFirst(
+      `EAS project not configured. To configure it non-interactively, choose the account that should own the project and run:\n\n  eas init --account <name> --non-interactive\n\nAccounts you can create projects in: ${getCreatableAccountNamesNewestFirst(
         actor
       ).join(
         ', '
--- a/packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
+++ b/packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
@@ -8,7 +8,7 @@ import { AppQuery } from '../graphql/queries/AppQuery';
 import { link } from '../log';
 import { ora } from '../ora';
 import { confirmAsync } from '../prompts';
-import { Actor } from '../user/User';
+import { Actor, getCreatableAccountNamesNewestFirst } from '../user/User';
 
 /**
  * 1. Looks for an existing project on EAS servers. If found, ask the user whether this is the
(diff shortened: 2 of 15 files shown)


williamgrosset:
~This PR overlaps with #4054. Both add `createdAt` to the `CurrentUser` accounts query for newest-first account ordering (picker there, error messages here). Whichever merges second gets a small rebase to drop the duplicate.~ edit: no longer the case

packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
```diff
@@ -30,18 +30,22 @@ export async function fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsyn
...
   const allAccounts = actor.accounts;
   const accountNamesWhereUserHasSufficientPublishPermissions = new Set(
     allAccounts
       .filter(a => a.users.find(it => it.actor.id === actor.id)?.role !== Role.ViewOnly)
       .map(it => it.name)
   );
+
+  if (options.nonInteractive) {
+    throw new Error(
+      `EAS project not configured. To configure it non-interactively, choose the account that should own the project and run:\n\n  eas init --account <name> --non-interactive\n\nAccounts you can create projects in: ${getCreatableAccountNamesNewestFirst(
```
brentvatne:
```suggestion
      `EAS project not configured. To configure it non-interactively, choose the account that should own the project and run:\n\n  eas init --account <name> --non-interactive\n\nAccounts you have permissions to create projects in: ${getCreatableAccountNamesNewestFirst(
```

packages/eas-cli/src/project/fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsync.ts
```diff
@@ -30,18 +30,22 @@ export async function fetchOrCreateProjectIDForWriteToConfigWithConfirmationAsyn
...
+        actor
+      ).join(
+        ', '
+      )}\n\nAlternatively, set the "owner" field in your app config, or run "eas init" for interactive setup.`
```
brentvatne:
```suggestion
      )}\n\nAlternatively, set the "owner" field in your app config, or run "eas init" to configure it interactively.
```

packages/eas-cli/src/project/projectInitialization.ts
```diff
@@ -214,21 +218,36 @@ export async function initializeWithoutExplicitIDAsync(
       .map(it => it.name)
   );
 
-  // if no owner field, ask the user which account they want to use to create/link the project
-  let accountName = exp.owner;
+  if (accountNameArgument) {
+    if (exp.owner && exp.owner !== accountNameArgument) {
+      throw new Error(
+        `The account specified with --account (${accountNameArgument}) does not match the "owner" field in your app config (${exp.owner}). Pass a matching --account or update the "owner" field.`
```
brentvatne:
```suggestion
        `The account specified with --account (${accountNameArgument}) does not match the "owner" field in your app config (${exp.owner}). Provide a matching --account or update the "owner" field.`
```

packages/eas-cli/src/project/projectInitialization.ts
```diff
@@ -214,21 +218,36 @@ export async function initializeWithoutExplicitIDAsync(
...
+      );
+    }
+    if (!allAccounts.some(a => a.name === accountNameArgument)) {
+      throw new Error(
+        `You don't have access to an account named "${accountNameArgument}". Accounts you can create projects in: ${getCreatableAccountNamesNewestFirst(
```
brentvatne:
```suggestion
        `You are not able to create projects in the "${accountNameArgument}" account. Accounts you have permissions to create projects in: ${getCreatableAccountNamesNewestFirst(
```

brentvatne: is this the correct wording? i assume that we're (ahem) accounting for the fact that you could be a viewer in an org and therefore not be able to create accounts, even if you are able to otherwise sign in

packages/eas-cli/src/project/projectInitialization.ts
```diff
@@ -214,21 +218,36 @@ export async function initializeWithoutExplicitIDAsync(
...
       accountName = allAccounts[0].name;
     } else if (nonInteractive) {
       if (!force) {
         throw new Error(
-          `There are multiple accounts that you have access to: ${allAccounts
-            .map(a => a.name)
-            .join(
-              ', '
-            )}. Explicitly set the owner property in your app config or run this command with the --force flag to proceed with a default account: ${
+          `You have access to multiple accounts. Choose the account that should own this project with the --account flag:\n\n  eas init --account <name> --non-interactive\n\nAccounts you can create projects in: ${getCreatableAccountNamesNewestFirst(
```
brentvatne:
```suggestion
          `You have access to multiple accounts. Choose the account that should own this project with the --account flag:\n\n  eas init --account <name> --non-interactive\n\nAccounts you have permissions to create projects in: ${getCreatableAccountNamesNewestFirst(
```

packages/eas-cli/README.md
```diff
@@ -2237,9 +2238,10 @@ create or link an EAS project
 
 ```
 USAGE
-  $ eas project:init [--id <value>] [--force] [--non-interactive]
+  $ eas project:init [--account <value> | --id <value>] [--force] [--non-interactive]
 
 FLAGS
+  --account=<value>  Name of the account that should own the project
```
brentvatne:
```suggestion
  --account=<value>  Name of the account that will own the project
```

brentvatne (changes requested):

brentvatne:
`--account` looks like it's silently ignored when the project is already linked. In `initializeWithoutExplicitIDAsync` we early-return on `existingProjectId` before the flag is ever read:

```ts
const existingProjectId = exp.extra?.eas?.projectId;
if (existingProjectId) {
  Log.succeed(`Project already linked (ID: ${existingProjectId})`);
  return existingProjectId; // returns before any --account handling
}
```

So `eas init --account X` on an already-linked project is a no-op even when `X` differs from the current owner. I think we should:

- **no-op** when `--account` matches the current owner (fine as-is), and
- **error** when it differs and `--force` isn't passed.

The owner is a cheap lookup — `AppQuery.byIdAsync` already returns `ownerAccount.name` — so, placed *before* the early-return:

```ts
if (existingProjectId && accountNameArgument) {
  const { ownerAccount } = await AppQuery.byIdAsync(graphqlClient, existingProjectId);
  if (ownerAccount.name !== accountNameArgument && !force) {
    throw new Error(
      `This project is already linked to @${ownerAccount.name} (ID: ${existingProjectId}). ` +
      `Pass --force to re-link it to @${accountNameArgument}, or remove the projectId from your app config.`
    );
  }
  // matches (or --force) → falls through to the existing "already linked" no-op / re-link
}
```

One thing to confirm so the error's advice is honest: that `--force` actually re-links an already-linked project to the new account. The current early-return doesn't check `force`, so if that path isn't wired up, "pass `--force`" would be a dead end.

packages/eas-cli/src/project/projectInitialization.ts
williamgrosset: yes, we only list accounts that the user can create projects in. ref: https://github.com/expo/eas-cli/blob/5f20be698097f3d403d7f6381ba98ae362e96616/packages/eas-cli/src/user/User.ts#L5-L14

williamgrosset: ah, this check wasn't aligned: `!allAccounts.some(a => a.name === accountNameArgument`). updated this to: `!accountNamesWhereUserHasSufficientPermissionsToCreateApp.has(accountNameArgument)`

brentvatne (approved):

PR #4060, opened 2026-07-23
[eas-cli] Add --client-id filter to observe:metrics|events commands

keith-kurak:
# Why

Some users want to read all Observe events/metrics by user (EAS client ID) via the CLI, in order to approximate these views:
- `https://expo.dev/accounts/[account]/projects/[project]/observe/events?t=14d&environment=production&easClientId=[id]` (all events by user)
- `https://expo.dev/accounts/[account]/projects/[project]/observe/metrics?metric=expo.app_startup.cold_launch_time&t=14d&environment=production&easClientId=[id]` (all metrics by user)

However `eas obsever:metrics` don't exactly show the same data, as the CLI needs to filter by a particular event/metric, while the web pages show all events/metrics

# How

- Added `--client-id` filter on `observe:events` and `observe:metrics`
- Added  `--all-metrics` flag on `eas observe:metrics`, which will not require you to filter to a single metric. This is consistent with on `eas observe:events --all-events`.
- `eas observe:metrics` now show an additional `metric` field, so you can tell which metric you're looking at when there's more than one type (this field is now on all outputs from these commands for consistency)
- `eas observe:metrics` without a metric or `--all-metrics` still brings up a metrics chooser.

# Test Plan

## Metrics

image

image

#### No metric or flag (same behavior as before)
image

## Events

image


packages/eas-cli/src/commands/observe/metrics.ts
```diff
@@ -88,20 +95,34 @@ export default class ObserveMetrics extends EasCommand {
...
-          Object.keys(METRIC_ALIASES).join(', ')
+        '--all-metrics cannot be combined with a metric argument. Pass a metric to filter by it, or pass --all-metrics to return samples across all metrics.'
       );
+    }
+
+    // --json alone should not suppress the metric picker: only an explicit
+    // --non-interactive flag or a non-interactive terminal (e.g. piped output,
+    // CI) does. In those cases we can't prompt, so we default to all metrics
+    // rather than erroring out.
+    const canPromptForMetric = !flags['non-interactive'] && !!process.stdin.isTTY;
```
kadikraman, line 109: This is incorrect - the [existing documented convention](https://github.com/expo/eas-cli/blob/main/packages/eas-cli/src/commandUtils/flags.ts#L11) is that `--json` implies `--non-interactive`

keith-kurak: I wondered ... like, that sounded familiar, but then I spot-checked a handful of commands and they all did the opposite. Should I change the behavior of `eas observe:metrics --json` to follow non-interactive requirements (it wasn't before) in this PR or leave it the same?

douglowder: Yes please, there is a helper function `resolveNonInteractiveAndJsonFlags()` in `packages/eas-cli/src/commandUtils/flags.ts`.

PR #4061, opened 2026-07-23
[steps] Unify the default gating of steps with no `if:` across steps, composite scopes, and hooks

sswrk:
# Why

For a step with no `if:`, whether the step should run was decided in 3 separate places: `BuildStep.shouldExecuteStep`, the composite function scope (same rule), and the hook executor (its own rule). That bypass means composite scopes never see hook defaults, which is a blocker for adding support for composite functions in hooks.

# How

A single gate now decides whether any step runs, taking a single input: a bool telling whether a step with no `if:` should run. Composite functions called from hooks therefore inherit the hook's rule. Steps with an explicit `if:` are unaffected.

Behavior-preserving refactor.

# Test Plan

- Existing steps/hook tests still pass.
- Added unit coverage for the run-by-default value on `shouldExecuteStep` /
  `isActive`, and hook sibling skip/continue cases for the before/after defaults.

PR #4015, opened 2026-07-14
[steps][4/4] Resolve composite function outputs

sswrk:
# Why

This PR stack adds local composite functions for EAS workflows. A `function.yml` declares `outputs:` with templated values, and the caller should read them as `${{ steps.setup.outputs.version }}`, where `setup` is the id of the calling step.

After the previous PRs, data flows into a composite function (inputs, env) but nothing flows back out. This PR closes that loop.

# How

For a composite function that declares `outputs:`, `CompositeFunctionExpander` appends a synthetic function step after the inner steps, reusing the caller's step id.

- The step runs with `if: ${{ always() }}`
- Output values are interpolated against the composite-function-local step namespace

# Test Plan

Added outputs unit tests.

PR #4014, opened 2026-07-14
[steps][3/4] Interpolate composite function inputs

sswrk:
# Why

This PR stack adds support for local composite functions in workflows. After structural expansion, inner steps still could not see `${{ inputs.* }}`, and caller `env` did not yet reach them at runtime.

This PR adds the data layer: input declaration, validation and interpolation, and runtime resolution of the caller context (env, templated working directories).

# How

- `CompositeFunctionExpander` builds declared composite function inputs as real `BuildStepInput` instances
- `BuildStepCompositeFunctionScope` exposes `${{ inputs.* }}` as lazy getters
- Caller `env` and working directory templating resolve through the composite function scope

# Test Plan

Added inputs and scoping unit tests.

PR #4013, opened 2026-07-13
[steps][3/5] Add action interpolation helpers

sswrk:
# Why

Third PR in the `@expo/steps` local actions stack: small shared helpers for interpolating action-scoped values. Part of ENG-22387.

# How

Adds internal `actionInterpolation.ts`:

- `resolveInterpolatedTarget(target, context, legacyResolver)`: run `${{ }}` via the job interpolator, then route legacy `${ steps.* }` through a caller-provided resolver (so action scopes can rewrite short ids)
- `stringifyInterpolatedResult` / `stringifyOptionalInterpolatedResult`
- `containsUnresolvedTemplateReference` for deferred template checks

Reuses existing `interpolateWithOutputs` / `BUILD_STEP_OUTPUT_EXPRESSION_REGEXP` from `template.ts` instead of a parallel `templateSpans` / `actionInputs` layer. Utilities remain internal (not exported from `index.ts`).

# Test Plan

`cd packages/steps && yarn test` — `actionInterpolation` unit tests.

PR #4062, opened 2026-07-23
[steps] Support local composite functions in hook steps

sswrk:
# Why

Hook steps currently reject `uses: ./...` local composite functions. Users should be able to reuse the same composites in hooks as in the main workflow.

# How

A composite function used in a hook is expanded into its inner steps at parse time, the same way it is in the main workflow, and the whole expansion becomes a single hook entry. The step's `if:` gates the expansion as a whole.

Hook construction now takes a catalog of composite functions. Without it, referencing a composite fails with a missing-function error instead of the previous "not supported in hooks" error.

Wiring reading the composite functions files will be added in follow-up PRs in the stack.

# Test Plan

Added unit tests.

### DIFF
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -165,8 +165,9 @@ export class BuildWorkflow {
  * Default-run rules (a step or entry with no `if:`):
  * - `before`: runs unless a failure occurred within THIS hook sequence;
  *   failures predating the call are ignored — "runs iff the anchor runs".
- * - `after`: runs unconditionally — past the anchor's own failure AND past an
- *   earlier after-entry's failure.
+ * - `after`: each entry starts fresh, past the anchor's own failure AND past
+ *   an earlier after-entry's failure, but a failure within the entry skips
+ *   its later no-`if:` steps.
  * Passed as `runByDefault` so composite scopes share the same missing-`if:` rule.
  * A user `if:` is always evaluated against the real global context, so
  * `failure()` / `success()` keep their global meaning on both sides.
@@ -224,10 +225,11 @@ export async function executeHookStepsAsync(
 
     let entryFailed = false;
     for (const step of entry.steps) {
-      // before skips on in-sequence failure; an entry with a passed if: only skips
-      // on within-entry failure; after always runs.
-      const runByDefault =
-        options.timing === 'after' || (entryHasExplicitCondition ? !entryFailed : !failedLocally);
+      // after and an entry with a passed if: skip on within-entry failure;
+      // before without an entry if: skips on any in-sequence failure.
+      const scopedFailure =
+        options.timing === 'after' || entryHasExplicitCondition ? entryFailed : failedLocally;
+      const runByDefault = !scopedFailure;
       let shouldExecuteStep = false;
       try {
         shouldExecuteStep = step.shouldExecuteStep({ runByDefault });
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -5,6 +5,7 @@ import { BuildFunctionById } from './BuildFunction';
 import { BuildRuntimePlatform } from './BuildRuntimePlatform';
 import { BuildStep } from './BuildStep';
 import { BuildStepGlobalContext } from './BuildStepContext';
+import { CompositeBuildStep } from './CompositeBuildStep';
 import { StepMetricResult } from './StepMetrics';
 import { AnchorHooks, HookEntry } from './hooks';
 import { evaluateIfCondition } from './utils/jsepEval';
@@ -192,7 +193,7 @@ export async function executeHookStepsAsync(
     ctx.markAsFailed();
   };
 
-  let anyStepExecuted = false;
+  let anyAuthoredStepExecuted = false;
   for (const entry of entries) {
     // Truthiness, not presence: an empty `if:` means "no condition", the same
     // as BuildStep.shouldExecuteStep treats it.
@@ -247,7 +248,11 @@ export async function executeHookStepsAsync(
         step.skip();
         continue;
       }
-      anyStepExecuted = true;
+      // The synthetic composite outputs node runs under always(), so it alone
+      // must not make a fully-skipped hook side report a metric.
+      if (!(step instanceof CompositeBuildStep)) {
+        anyAuthoredStepExecuted = true;
+      }
       const startTime = performance.now();
       let stepResult: StepMetricResult = 'success';
       try {
@@ -262,7 +267,7 @@ export async function executeHookStepsAsync(
     }
   }
 
-  if (anyStepExecuted) {
+  if (anyAuthoredStepExecuted) {
     ctx.reportWorkflowHookMetric({
       anchor: options.anchor,
       timing: options.timing,
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -274,8 +274,9 @@ export async function executeHookStepsAsync(
   return { failedLocally, firstError };
 }
 
-// Shared wording for unevaluable `if:` gates. `ifCondition` is omitted when the
-// failure is a composite call-site `if:` evaluated via the step's scope.
+// Shared wording for unevaluable `if:` gates. Callers pass the step's own
+// `if:`, but the throw may come from a composite call-site `if:` evaluated via
+// the step's scope. In that case the quoted condition is not the one that failed.
 function logConditionEvaluationError(
   logger: bunyan,
   err: unknown,
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -272,8 +272,8 @@ export async function executeHookStepsAsync(
   return { failedLocally, firstError };
 }
 
-// The one wording for an `if:` that could not be evaluated — anchor steps,
-// hook steps, and hook group entries all log through here.
+// Shared wording for unevaluable `if:` gates. `ifCondition` is omitted when the
+// failure is a composite call-site `if:` evaluated via the step's scope.
 function logConditionEvaluationError(
   logger: bunyan,
   err: unknown,
@@ -282,6 +282,8 @@ function logConditionEvaluationError(
 ): void {
   logger.error({ err });
   logger.error(
-    `Runner failed to evaluate if it should execute ${subject}, using its if condition "${ifCondition}". This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
+    `Runner failed to evaluate if it should execute ${subject}${
+      ifCondition ? `, using its if condition "${ifCondition}"` : ''
+    }. This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
   );
 }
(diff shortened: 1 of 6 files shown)


sswrk:
* **#4098** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4098?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4097** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4097?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4096** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4095** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4095?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4064** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4064?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4063** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4063?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4062** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4061** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4061?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3931** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3930** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3930?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4015** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4014** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4014?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3929** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3929?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4013](https://github.com/expo/eas-cli/pull/4013) <a href="https://app.graphite.com/github/pr/expo/eas-cli/4013?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3956** <a href="https://app.graphite.com/github/pr/expo/eas-cli/3956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/steps/src/CompositeFunctionExpander.ts
```diff
@@ -45,6 +45,10 @@ export type FunctionMaps = {
   buildFunctionGroupById: BuildFunctionGroupById;
 };
 
+export type FunctionMapsWithExpander = FunctionMaps & {
```
hSATAC: nit, not blocking.

`CompositeFunctionExpander` already holds its own `functionMaps`, so this type carries the maps twice and nothing stops a caller from assembling a bag whose expander was built from a different pair of maps. Passing just the expander, or exposing the maps off it, would make that impossible to get wrong. Happy to leave it if you prefer keeping the destructuring at the call sites as it is.

packages/steps/src/hooks.ts
```diff
@@ -30,6 +31,11 @@ import { createBuildStepOutputsFromDefinition, getShellStepDisplayName } from '.
  * (2) An entry whose explicit `if:` passed behaves like a single step whose
  * `if:` passed: its no-`if:` steps run past earlier entries' failures, while
  * within-entry failures still skip later siblings.
+ *
+ * Composite calls are also one entry, but their call-site `if:` lives on the
+ * expansion scope (not `ifCondition`), matching main-workflow composites. A
+ * passing call `if:` therefore does not grant the group-entry `!entryFailed`
+ * shield; no-`if:` children follow the plain hook default.
```
hSATAC: nit, not blocking, and I think this predates the PR rather than being caused by it, so it is fine as a separate PR.

The paragraph is accurate, but the consequence is sharper than it reads. When an earlier entry has already failed and the call `if:` passes, the no-`if:` children skip as described, while the synthetic outputs node runs anyway, since it is constructed with `ifCondition: '${{ always() }}'`. So the call reports SUCCESS and publishes its declared outputs resolved against skipped children, which come out as empty strings. A later step reading `${{ steps.<call-id>.outputs.x }}` silently gets `''`.

I checked and the same thing happens in the main workflow, so this PR does not introduce it. Raising it here because hooks are where people write `if: always()`, so it will start showing up.

packages/steps/src/BuildWorkflow.ts
```diff
@@ -272,8 +272,8 @@ export async function executeHookStepsAsync(
   return { failedLocally, firstError };
 }
 
-// The one wording for an `if:` that could not be evaluated — anchor steps,
-// hook steps, and hook group entries all log through here.
+// Shared wording for unevaluable `if:` gates. `ifCondition` is omitted when the
+// failure is a composite call-site `if:` evaluated via the step's scope.
```
hSATAC: nit, not blocking, and fine to split into its own PR.

The omission is conditional on the child having an `if:`, not on where the failure came from. When a composite call-site `if:` throws and the child does have its own `if:`, this quotes the child's condition as the culprit even though that condition is valid and did not fail. And because `isActive` memoizes with `??=`, a throw is never cached, so every remaining child re-evaluates the same broken expression and logs the same pair of lines. A 10-step composite produces 10 copies naming 10 different steps, none of which is the one the user actually wrote the condition on.

The same misattribution already exists in the main-workflow loop further up, so it is not new. Flagging it here because this PR is the one touching the message.

packages/steps/src/hooks.ts
```diff
@@ -127,9 +134,16 @@ export function constructHookEntriesFromValidatedSteps(
       continue;
     }
     if (isLocalCompositeFunctionPath(step.uses)) {
-      throw new BuildConfigError(
-        `Local composite function steps ("uses: ${step.uses}") are not supported in hooks.`
-      );
+      entries.push({
+        steps: compositeFunctionExpander
+          .expandCompositeFunctionStep(
```
hSATAC, line 142: nit, not blocking, pre-existing rather than yours, and fine as a separate PR.

`expandCompositeFunctionStep` forwards `name`, `with`, `if` and `env` from the call step and rejects `working_directory` with a good message, but it never reads `step.shell`, so `shell:` on a composite call is silently dropped. Same class of thing as `working_directory`, just unguarded. Worth either rejecting it the same way or forwarding it to the inner steps.

packages/steps/src/hooks.ts
```diff
@@ -127,9 +134,16 @@ export function constructHookEntriesFromValidatedSteps(
...
+            step,
+            parseLocalCompositeFunctionPath(step.uses),
+            BuildStep.getNewId(step.id)
+          )
+          .getFlattenedSteps(),
```
hSATAC, line 147: nit, not blocking, and arguably my area rather than yours. Also fine as a separate PR.

`getFlattenedSteps()` appends the synthetic outputs node whenever the composite declares `outputs:`, and that node runs under `always()`. Since `anyStepExecuted` is the only gate on `reportWorkflowHookMetric`, a hook side where every authored child was skipped still reports one `eas.workflow.hook` event with `result: 'success'` when the composite declares outputs, and reports nothing when it does not. Same hook, same outcome, different signal depending on whether an `outputs:` block happens to be there. I would rather the metric did not depend on that, but it is not urgent.

hSATAC (commented):
A few non-blocking nits. Most of them are about composite behaviour that predates this PR rather than anything it introduces, so they are all fine to split out into a separate PR if you would rather keep this one focused.

hSATAC (approved):
One thing I noticed while reviewing this: I don’t think this PR is doing anything wrong, but it does make an existing behavior difference more likely to surface for users.

Today, when a single hook entry expands into multiple steps, the later steps are gated differently depending on which side the hook runs on:

* In a `before_*` hook, if an earlier step fails, later steps without an `if:` are skipped.
* In an `after_*` hook, later steps still run.

This PR preserves that existing behavior. The difference is that, until now, multi-step hook entries could only come from function groups that we own. We don’t have many function groups, and users are also unlikely to use them inside hooks, so this difference probably hasn’t been very visible in practice.

With composite functions, users can now define and reuse the same multi-step unit themselves:

```yaml
# .eas/functions/publish/function.yml
runs:
  steps:
    - run: ./login.sh
    - run: ./push.sh
```

When used in `steps:`, `./push.sh` is skipped if `./login.sh` fails. When the same composite is used in `after_build`, `./push.sh` still runs. I verified this behavior on the branch.

I think this could be surprising because the author of a composite may not know where it will eventually be used, so the behavior isn’t something they can fully account for inside the composite itself.

I don’t think this needs to block the PR, but I wanted to flag it so we can decide whether we want to document the difference, align the default behavior for `after_*` hooks, or keep it as-is.

One related detail if we document it: `success()` and `failure()` inside a composite reflect the overall job status, not the result of the previous step. That means `if: ${{ success() }}` is not a reliable fail-fast guard inside an `after_*` hook, since the hook may already be running after the job has failed.

To express “only run if the previous step succeeded,” users would need to pass that state explicitly, for example:

```yaml
- id: login
  run: |
    ./login.sh
    set-output ok true
  outputs: [ok]

- run: ./push.sh
  if: ${{ steps.login.outputs.ok == 'true' }}
```

The same behavior applies to function groups, so whatever we decide here could cover both.

sjchmiela (approved):

sswrk:
@hSATAC Thank you for noting that, and yes, this was done like that for the sake of consistency. But taking a step back, as a user I'd expect composite functions/function groups to fast-fail within them even in the `after_*` hook (and then let the next steps in the hook run).

Corrected the behavior for both composite functions and function groups.

packages/steps/src/hooks.ts
sswrk: I think we're fine because https://github.com/expo/eas-cli/blob/main/packages/eas-build-job/src/step.ts#L97

We can follow-up with further improving the handling of this like in `working_directory`.

packages/steps/src/CompositeFunctionExpander.ts
sswrk: Cleaned this up

packages/steps/src/hooks.ts
sswrk: Valid point. Deferring it to a follow-up PR as this seems like a bigger change.

[ENG-25537](https://linear.app/expo/issue/ENG-25537)

packages/steps/src/BuildWorkflow.ts
sswrk: Yeah, not specific to this PR but a valid point. Changed the message a bit and deferring the fix to a follow-up PR.

[ENG-25535](https://linear.app/expo/issue/ENG-25535)

packages/steps/src/hooks.ts
sswrk: Applied a fix: https://github.com/expo/eas-cli/pull/4062/commits/72786a9c62d524e4cfb042ff60ff94e70dcebee9

hSATAC (approved):

PR #3930, opened 2026-07-01
[build-tools] Load local composite function catalog at build runtime

sswrk:
# Why

This PR stack adds local composite functions for EAS workflows (`uses: ./.eas/functions/...`). Earlier PRs define the schema, expand calls into steps, and wire inputs/outputs. The build runtime still needs to load those `function.yml` files from the project and pass them into `StepsConfigParser`.

This PR is the I/O layer: discover referenced local composite functions on disk, validate them, and hand the catalog to the parser for custom builds and generic jobs.

# How

- Adds catalog helpers in `@expo/steps` (`buildCompositeFunctionCatalogFromStepsAsync`, `resolveLocalCompositeFunctionPath`)
- Adds `packages/build-tools/src/steps/compositeFunctions.ts` to resolve paths relative to the EAS project root, read `function.yml` / `function.yaml`, and validate with `CompositeFunctionConfigZ`
- Wires the catalog into `runCustomBuildAsync` and `runGenericJobAsync`

# Test Plan

Added unit tests for catalog loading and custom-build parse integration.

PR #4063, opened 2026-07-23
[build-tools][steps] Load composite function catalogs for hook steps

sswrk:
# Why

Local composite functions can be referenced from hook steps (`uses: ./...`), but build-tools only builds the composite catalog from job `steps`.

# How

- Steps-based jobs: eager catalog loag from job `steps`, pass a lazy `loadCompositeFunction` loader so hook composites load only when their anchor runs
- Native jobs: load only from wrapped anchors
- Warn and skip for registered hook keys whose anchors never appear
 
# Test Plan

Added unit tests.

### DIFF
--- a/packages/build-tools/src/common/jobHooks.ts
+++ b/packages/build-tools/src/common/jobHooks.ts
@@ -1,6 +1,5 @@
 import {
   BuildJob,
-  CompositeFunctionCatalog,
   ErrorCode,
   HookAnchorId,
   HookKey,
@@ -91,30 +90,6 @@ export async function parseJobHooksAsync<TJob extends BuildJob>(
     }
   }
 
-  // Only load composites for anchors this build actually wraps.
-  // Unwrapped registered keys warn-and-skip above.
-  const catalogRootSteps = wrappedAnchors.flatMap(anchor =>
-    (['before', 'after'] as const).flatMap(side => {
-      const steps = hooks[`${side}_${anchor}`];
-      return Array.isArray(steps) ? steps : [];
-    })
-  );
-  let compositeFunctionCatalog: CompositeFunctionCatalog;
-  try {
-    compositeFunctionCatalog = await buildCompositeFunctionCatalogAsync(
-      ctx.getReactNativeProjectDirectory(),
-      { steps: catalogRootSteps, logger: ctx.logger }
-    );
-  } catch (err) {
-    throw new UserError(
-      ErrorCode.HOOKS_ERROR,
-      `Failed to load a local composite function referenced from the job's hooks: ${
-        err instanceof Error ? err.message : String(err)
-      }`,
-      { cause: err }
-    );
-  }
-
   // Construct entries for the wrapped anchors in EXECUTION order (per anchor,
   // before_ then after_) so generated step ids and output references follow the
   // order steps actually run. All entries share one globalContext, so env and
@@ -130,6 +105,11 @@ export async function parseJobHooksAsync<TJob extends BuildJob>(
       }
       let entries: HookEntry[];
       try {
+        // Per key so a bad `uses:` path is attributed to that hook key.
+        const compositeFunctionCatalog = await buildCompositeFunctionCatalogAsync(
+          ctx.getReactNativeProjectDirectory(),
+          { steps, logger: ctx.logger }
+        );
         entries = await constructHookEntriesAsync(globalContext, steps, {
           externalFunctions,
           externalFunctionGroups,
--- a/packages/build-tools/src/common/jobHooks.ts
+++ b/packages/build-tools/src/common/jobHooks.ts
@@ -1,5 +1,6 @@
 import {
   BuildJob,
+  CompositeFunctionCatalog,
   ErrorCode,
   HookAnchorId,
   HookKey,
@@ -12,12 +13,13 @@ import {
   BuildStepGlobalContext,
   HookEntry,
   constructHookEntriesAsync,
+  extendCompositeFunctionCatalogFromStepsAsync,
   validateHookStepsAsync,
 } from '@expo/steps';
 
 import { BuildContext } from '../context';
 import { CustomBuildContext } from '../customBuildContext';
-import { buildCompositeFunctionCatalogAsync } from '../steps/compositeFunctions';
+import { createCompositeFunctionLoader } from '../steps/compositeFunctions';
 import { getEasFunctionGroups } from '../steps/easFunctionGroups';
 import { getEasFunctions } from '../steps/easFunctions';
 
@@ -96,6 +98,11 @@ export async function parseJobHooksAsync<TJob extends BuildJob>(
   // outputs accumulate across keys.
   const hookEntriesByKey: Partial<Record<HookKey, HookEntry[]>> = {};
   const orderedSteps: BuildStep[] = [];
+  const compositeFunctionCatalog: CompositeFunctionCatalog = {};
+  const loadCompositeFunction = createCompositeFunctionLoader(
+    ctx.getReactNativeProjectDirectory(),
+    ctx.logger
+  );
   for (const anchor of wrappedAnchors) {
     for (const side of ['before', 'after'] as const) {
       const key: HookKey = `${side}_${anchor}`;
@@ -105,11 +112,12 @@ export async function parseJobHooksAsync<TJob extends BuildJob>(
       }
       let entries: HookEntry[];
       try {
-        // Per key so a bad `uses:` path is attributed to that hook key.
-        const compositeFunctionCatalog = await buildCompositeFunctionCatalogAsync(
-          ctx.getReactNativeProjectDirectory(),
-          { steps, logger: ctx.logger }
-        );
+        // Extended per key so a bad `uses:` path is attributed to that hook key.
+        await extendCompositeFunctionCatalogFromStepsAsync({
+          catalog: compositeFunctionCatalog,
+          rootSteps: steps,
+          loadCompositeFunction,
+        });
         entries = await constructHookEntriesAsync(globalContext, steps, {
           externalFunctions,
           externalFunctionGroups,
--- a/packages/build-tools/src/common/jobHooks.ts
+++ b/packages/build-tools/src/common/jobHooks.ts
@@ -1,5 +1,6 @@
 import {
   BuildJob,
+  CompositeFunctionCatalog,
   ErrorCode,
   HookAnchorId,
   HookKey,
@@ -17,6 +18,7 @@ import {
 
 import { BuildContext } from '../context';
 import { CustomBuildContext } from '../customBuildContext';
+import { buildCompositeFunctionCatalogAsync } from '../steps/compositeFunctions';
 import { getEasFunctionGroups } from '../steps/easFunctionGroups';
 import { getEasFunctions } from '../steps/easFunctions';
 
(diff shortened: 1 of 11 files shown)


sswrk:
Also moving the 2 open discussions from the previous built-tools PR (https://github.com/expo/eas-cli/pull/3930) here:

- https://github.com/expo/eas-cli/pull/3930#discussion_r3628786646
- https://github.com/expo/eas-cli/pull/3930#pullrequestreview-4768059121

packages/build-tools/src/steps/compositeFunctions.ts
```diff
@@ -76,11 +78,20 @@ async function loadLocalCompositeFunctionConfigAsync(
...
       loadLocalCompositeFunctionConfigAsync(projectRoot, compositeFunctionPath, { logger }),
   });
 }
+
+function hookCatalogRootSteps(hooks: Hooks | undefined): Step[] {
+  if (!hooks) {
+    return [];
+  }
+  return Object.entries(hooks).flatMap(([key, steps]) =>
+    parseHookKey(key) !== null && Array.isArray(steps) ? steps : []
```
hSATAC: I think this may cause an issue with `defaults.hooks`.

We collect composites from every registered hook key here, even if a steps-based job doesn’t have that anchor. This is different from the native path in `jobHooks.ts`, which only loads composites for wrapped anchors.

For example:

```yaml
defaults:
  hooks:
    before_install_node_modules:
      - uses: ./.eas/functions/notify

jobs:
  mobile_build:
    type: build
    params:
      project_root_directory: apps/mobile
  lint:
    type: custom
    steps:
      - run: yarn lint
```

www merges the default hooks into both jobs. If `function.yml` only exists under `apps/mobile/.eas/functions/notify`, `mobile_build` resolves it correctly. But lint would also try to resolve it from the repo root and fail in `PARSE_CUSTOM_WORKFLOW_CONFIG`, even though it has no `install_node_modules` anchor and would never run this hook.

Previously this was ignored because `validateAllStepFunctionsExist` skips local composites. Maybe validating all hooks early is what we want, but I wasn’t sure if this behavior change was intended. We also lose the hook key after flattening the roots, so the error can’t tell the user which hook referenced the composite.

packages/build-tools/src/steps/compositeFunctions.ts
```diff
@@ -76,11 +78,20 @@ async function loadLocalCompositeFunctionConfigAsync(
 
 export async function buildCompositeFunctionCatalogAsync(
   projectRoot: string,
-  { steps, logger }: { steps: readonly Step[]; logger?: bunyan }
+  { steps, hooks, logger }: { steps: readonly Step[]; hooks?: Hooks; logger?: bunyan }
```
hSATAC: Minor and non-blocking: would it be better to keep `hooks: Hooks | undefined` required here, like `StepsConfigParser` does? This prevents a call site from silently forgetting to pass the job hooks.

The three call sites are correct now, but removing `hooks` from `custom.ts` or `generic.ts` still type-checks and passes the tests. `jobHooks.ts` can pass `hooks: undefined` explicitly, which also makes the intentional omission clearer.

hSATAC (commented):
Overall this looks good to me, and the native/steps split makes sense. I left a couple of non-blocking inline comments on things I wasn’t sure were intentional.

One separate follow-up: `eas workflow:validate` only checks composites in `job.steps`, so references from `hooks` or `defaults.hooks` could pass validation and then fail on the worker. Not blocking this PR, but worth tracking.

packages/build-tools/src/steps/compositeFunctions.ts
sswrk: I didn't think of it, thank you so much for pointing this out. The eager loading approach comes from how I designed this before considering hooks, but it falls short in this case.

I changed it a bit, so that in the catalog build phase we don't load composite functions referenced in hooks and instead the parser in `steps` lazily loads them when it find's the hooks anchor.

This should now match the native path.

packages/build-tools/src/steps/compositeFunctions.ts
sswrk: Outdated as of https://github.com/expo/eas-cli/pull/4063/changes/8fa56f7b5698744aa3df6808a9e8f9376d747649

sswrk:
> One separate follow-up: eas workflow:validate only checks composites in job.steps, so references from hooks or defaults.hooks could pass validation and then fail on the worker. Not blocking this PR, but worth tracking.

Thanks for thinking about everything. This is coming in the next PR, I'm also looking into reducing the duplication between build-tools and eas-cli.

packages/build-tools/src/common/jobHooks.ts
```diff
@@ -89,6 +91,30 @@ export async function parseJobHooksAsync<TJob extends BuildJob>(
...
+    );
+  } catch (err) {
+    throw new UserError(
+      ErrorCode.HOOKS_ERROR,
+      `Failed to load a local composite function referenced from the job's hooks: ${
+        err instanceof Error ? err.message : String(err)
+      }`,
+      { cause: err }
+    );
+  }
```
hSATAC: This changes hook error reporting on the native path, and I think it’s a regression from `main`.

On `main`, a bad composite under a wrapped key is handled by `constructHookEntriesAsync`, so the error includes the hook key:

```
Failed to parse hooks.before_install_node_modules: Local composite function "./.eas/functions/missing" does not exist. Expected a "function.yml" (or "function.yaml") file at ...
```

Here, the catalog build gets there first after `catalogRootSteps` has flattened the wrapped anchors, so that context is lost:

```
Failed to load a local composite function referenced from the job's hooks: Local composite function "./.eas/functions/missing" was referenced by a step but no such composite function exists. ...
```

`working_directory` makes this more noticeable because it isn’t a load failure at all. `collectLocalCompositeFunctionPathsFromSteps` throws while collecting paths, but the error still gets the `Failed to load` prefix:

* `main`: `Failed to parse hooks.before_install_node_modules: "working_directory" is not supported ...`
* here: `Failed to load a local composite function referenced from the job's hooks: "working_directory" is not supported ...`

The steps path already avoids this by rethrowing `BuildConfigError` unchanged. Would it be better to build the catalog per key inside the `for (const anchor of wrappedAnchors)` loop below? That would keep the key in scope and let `BuildConfigError` fall through to the existing `Failed to parse hooks.${key}` wrapper.

The new test only checks the inner `no such composite function exists`, so it passes with either behavior.

packages/steps/src/StepsConfigParser.ts
```diff
@@ -175,6 +189,15 @@ export class StepsConfigParser extends AbstractConfigParser {
       }
     }
 
+    for (const hookKey of Object.keys(validatedHooks)) {
```
hSATAC: Non-blocking: this warning can’t be reached for one of the cases it covers.

`validateAllStepFunctionsExist` runs over every registered hook key, even when its anchor isn’t present. Local composites are skipped and handled by the lazy loader, but a plain `uses:` still throws before reaching this warning.

For example:

```yaml
defaults:
  hooks:
    before_submit:
      - uses: eas/some_newer_function

jobs:
  mobile_build:
    type: build
  lint:
    type: custom
    steps:
      - run: yarn lint
```

`lint` has no `submit` anchor, so I’d expect only the warning. Instead it fails with `Calling non-existent functions: "eas/some_newer_function"`.

`jobHooks.ts` deliberately doesn’t validate function existence for unwrapped keys, so native and steps jobs behave differently here. Would it make sense to move this check into `constructHookSideEntriesAsync`, where it would only run for anchors that actually appear? `constructHookEntriesAsync` already validates per key that way.

packages/steps/src/StepsConfigParser.ts
```diff
@@ -175,6 +189,15 @@ export class StepsConfigParser extends AbstractConfigParser {
...
+      const parsed = parseHookKey(hookKey);
+      if (parsed !== null && !seenAnchorIds.has(parsed.anchorId)) {
```
hSATAC: Minor and non-blocking: `parsed` can’t be null here. `validateHooks()` only keeps keys for which `parseHookKey()` returned non-null, so this is narrowing rather than a real check.

Would it be better for `validateHooks()` to return the anchor IDs it already parsed, or type its result as `Partial<Record<HookKey, Step[]>>`? Either would avoid parsing the keys again here.

packages/steps/src/StepsConfigParser.ts
```diff
@@ -175,6 +189,15 @@ export class StepsConfigParser extends AbstractConfigParser {
...
+        this.ctx.baseLogger.warn(
+          `Ignoring "hooks.${hookKey}": this build does not run the "${parsed.anchorId}" step.`
```
hSATAC: Minor and non-blocking: `jobHooks.ts` warns for both an unknown hook key and a registered key whose anchor isn’t wrapped. This adds the second warning with matching wording.

Should `validateHooks()` also warn about the unknown keys it skips? `parseHookKey(hookKey) === null` currently just continues, so a typo like `before_install_node_module` is silently dropped for a steps job but reported for a native one.

hSATAC (commented):
The lazy loader addresses what I raised last time, and the `rejectingLoader()` test covers it. Removing `hooks` from `buildCompositeFunctionCatalogAsync` also resolves my other comment.

I left one blocking comment in `jobHooks.ts` about the hook error messages regressing from `main`. The other three are non-blocking.

packages/steps/src/StepsConfigParser.ts
sswrk: Added the warning https://github.com/expo/eas-cli/pull/4063/commits/0593495f3d361a83e88e9a45eea33588299b0c53

packages/steps/src/StepsConfigParser.ts
sswrk: Cleaned this up https://github.com/expo/eas-cli/pull/4063/changes/083c0ba6193f3f406a568e5da9c725635d84de3e

packages/steps/src/StepsConfigParser.ts
sswrk: Makes sense, suggestion applied: https://github.com/expo/eas-cli/pull/4063/changes/7e123896fa7b5f51d2b7a90162bd593e32b6e600

packages/build-tools/src/common/jobHooks.ts
sswrk: Fair point about the regression. Applied a suggestion and added a test to assert the expected fomrat of the error message: https://github.com/expo/eas-cli/pull/4063/commits/dbb877b6e532e3a934c14c0cd355c4c45e50510d

sswrk:
Re. AI review:

> Hook composite errors lose the hook key

Fair, goes hand in hand with https://github.com/expo/eas-cli/pull/4063#discussion_r3691362936. Addressed: https://github.com/expo/eas-cli/pull/4063/changes/060c136f8223f6c20c477efcfe838f47e8f816d1

> Unknown hook key can spoof terminal output

Leaving this out for now, I think it's not really specific to this PR

hSATAC (approved):

packages/build-tools/src/steps/compositeFunctions.ts
hSATAC: Correction to my example: `params.project_root_directory` doesn't exist. The project root is resolved per run (from the project source or GitHub `baseDirectory`), not per job. So the two-root scenario I described can't happen: with a single root, a local composite either resolves for every job or for none.

The lazy loading change still makes sense, for a simpler reason. `www` merges `defaults.hooks` into every job, while the native path only loads composites for wrapped anchors. With eager loading, a steps job that never reaches the anchor could fail on a broken hook composite it would never run, while a native job in the same run would not. Lazy loading keeps the two paths consistent, and jobs that do reach the anchor still fail.

No change needed here — just correcting the reasoning so it doesn't affect later decisions. This also applies to the `project_root_directory` skip in #4064; I left a comment there.

packages/steps/src/StepsConfigParser.ts
```diff
@@ -252,16 +280,49 @@ export class StepsConfigParser extends AbstractConfigParser {
...
+          );
+        }
+        throw new BuildConfigError(
+          `Failed to load a local composite function referenced from "hooks.${side}_${anchorId}": ${
+            err instanceof Error ? err.message : String(err)
+          }`
+        );
+      }
+    }
     return constructHookEntriesFromValidatedSteps(this.ctx, hookSteps, compositeFunctionExpander);
```
hSATAC: Non-blocking: hook composite expansion errors still lose the hook key on this path. This return is outside both try blocks above, so errors thrown during expansion — such as a missing required input, duplicate IDs, or a composite calling a nonexistent function — come out bare:

```
Input parameter "message" for step "./.eas/functions/notify" is required but it was not set.
```

The native path wraps the whole construction, so the same error starts with `Failed to parse hooks.before_install_node_modules:`. Load errors on this path have included the key since `060c136f`; expansion errors are the remaining case.

Not a big issue since this matches plain job-step composite errors and the message still includes the composite path. But would it make sense to wrap this return with the same `Invalid steps in "hooks.<side>_<anchorId>"` pattern used above, so both paths attribute these errors consistently?

hSATAC (commented):
Left one non-blocking inline comment.

packages/steps/src/StepsConfigParser.ts
sswrk: Improved this 👌

PR #4064, opened 2026-07-23
[eas-cli][build-tools][steps] Validate local composite functions referenced from workflow hooks; extract the shared loader into `steps`

sswrk:
# Why

Workflow validation already checks local composite functions referenced from job steps, but hooks can also use `uses: ./....`. `eas workflow:validate` don't catch problems with composite functions in hooks.

This PR also includes an extraction of duplicated (eas-cli and build-tools) composite function loader to `@expo/steps`.

# How

Validation of workflows in `eas-cli` now reads each job's `hooks` steps too, not just `job.steps`.

Removed a duplicate loader from `eas-cli` and `build-tools`, moved it to `steps`.

# Test Plan

Added unit tests.


sswrk:
Also, moving an open discussion from the previous eas-cli PR in the stack: https://github.com/expo/eas-cli/pull/4064

sswrk:
Re. AI review:

> Validation skips jobs with a custom project root

That's a deliberate skip. Validating it would need injecting the vcs client. Will change it if human reviewers advise doing so

> Invalid function config uses the wrong error type

Not specific to this PR, worth considering as a follow up

> Path text can inject terminal control characters

not specific to this PR, a general pattern

> Validation loads composites from ignored hooks 

It does hold up, but it was deliberate: we can only do static verification here, so that's the only way we can check those functions within `workflow:validate`. Will change it if human reviewers advise doing so.

packages/eas-cli/src/commandUtils/workflow/compositeFunctions.ts
```diff
@@ -1,95 +1,51 @@
...
 
-const COMPOSITE_FUNCTION_FILE_EXTENSIONS = ['yml', 'yaml'] as const;
-
 export async function validateWorkflowLocalCompositeFunctionsAsync(
   parsedYaml: any,
   projectDir: string
 ): Promise<void> {
-  await buildCompositeFunctionCatalogFromStepsAsync({
+  await buildLocalCompositeFunctionCatalogAsync(projectDir, {
     rootSteps: stepsFromWorkflow(parsedYaml),
```
hSATAC, line 10: This still misses `defaults.hooks`. `stepsFromWorkflow()` collects each job's `steps` and `hooks`, but `www` merges `defaults.hooks` into jobs at runtime. A composite referenced only there passes `eas workflow:validate` and then fails on the worker:

```yaml
defaults:
  hooks:
    before_install_node_modules:
      - uses: ./.eas/functions/setup

jobs:
  build:
    type: build
    params:
      platform: android
```

I think the fix is to include `hookStepsFromJob(parsedYaml?.defaults)` in the root steps. A job-level hook key fully overrides the matching default key, so this may validate a broken default that every job overrides and never runs. That seems preferable to skipping `defaults.hooks` entirely.

I'd consider this blocking for this PR, since defaults are the main remaining gap in the hook composite validation it adds.

packages/eas-cli/src/commandUtils/workflow/compositeFunctions.ts
```diff
@@ -1,95 +1,51 @@
...
-  absolutePath: string
-): Promise<CompositeFunctionConfig | null> {
-  let rawContents: string;
-  try {
-    rawContents = await fs.readFile(absolutePath, 'utf-8');
-  } catch (err) {
-    if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') {
-      return null;
+  return Object.entries(jobs).flatMap(([jobKey, job]: [string, any]) => {
+    if (hasCustomProjectRootDirectory(job)) {
```
hSATAC: Non-blocking: I'd remove this skip. `project_root_directory` came from my incorrect example in #4063. It isn't a supported job param and nothing reads it at runtime; the composite root is resolved per run from the project source or GitHub `baseDirectory`, not per job.

Custom-job params allow arbitrary keys, so this currently turns an unknown key into a silent validation bypass: adding `params: { project_root_directory: apps/tv }` makes a broken local composite pass validation even though the worker ignores that key and later fails.

Validating every job against `projectDir` is the same assumption the existing job-step validation makes. A GitHub run can use a different `baseDirectory`, but that is run-level state and isn't visible in the YAML, so this skip can't detect it anyway.

I'd replace the skip and its tests with the monorepo scenario that was removed in this rewrite:

```yaml
# repo/.eas/functions/notify/function.yml             <- decoy at the repo root
# repo/apps/mobile/.eas/functions/notify/function.yml <- the real one
# repo/apps/mobile/.eas/workflows/build.yml

jobs:
  build:
    type: build
    params:
      platform: android
    hooks:
      before_install_node_modules:
        - uses: ./.eas/functions/notify
```

This should pass with the function under `projectDir`, continue allowing an explicit path such as `../../shared/...`, and fail when the function exists only at the repository root. That confirms resolution starts from `projectDir` and never falls back upward.

The PR body note about `project_root_directory` can go as well.

hSATAC (commented):
Left two comments.

packages/eas-cli/src/commandUtils/workflow/compositeFunctions.ts
sswrk: Removed the `project_root_directory` skip.

packages/eas-cli/src/commandUtils/workflow/compositeFunctions.ts
sswrk: Applied the suggested fix (including `hookStepsFromJob(parsedYaml?.defaults)` in the root steps).

hSATAC (approved):
Looking good. 👍

PR #4066, opened 2026-07-23
[steps] Safely index build step inputs and outputs

sjchmiela:
# Why

Build step input and output collections were indexed with ordinary JavaScript objects. An identifier such as `__proto__` would interact with the object's prototype instead of behaving like an ordinary own property. Call-input presence checks also treated inherited property names as supplied inputs.

# How

- Create internally owned input/output dictionaries with a null prototype.
- Keep the existing object-shaped public interfaces, including runtime-loaded custom functions.
- Use `Object.hasOwn` when reading externally supplied call inputs.
- Apply the same safe dictionary representation to condition evaluation and command interpolation.
- Keep conversion only at explicit serialization boundaries.

# Test Plan

Added regressions for `__proto__` identifiers in:

- input and output indexes
- direct build functions
- build function groups
- command interpolation
- `if` conditions
- runtime-loaded custom functions

Validated locally with the steps and build-tools tests and typechecks, formatting, and lint.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,3 +4,7 @@ When working with this repository follow instructions from CLAUDE.md.
 
 - [./CLAUDE.md](./CLAUDE.md)
 - Before committing, run the formatter on modified files.
+- Use null-prototype records for object-shaped dictionaries keyed by user-controlled identifiers.
+  Keep these dictionaries as records throughout runtime code; only convert them at explicit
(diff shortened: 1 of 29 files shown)


sswrk (approved):
😮 

Should we do something in this vein for inputs too?

sjchmiela:
changed it a bit so re-requesting review

sjchmiela:
thank you for thinking of inputs when reviewing!!

sswrk (approved):

PR #4067, opened 2026-07-23
[build-tools] Collect Argent tool-server session events during remote sessions

szdziedzic:
# Why

Agent-device remote simulator sessions already tail agent-device's event log and upload a normalized session-event timeline to the API server (`startAgentDeviceEventCollectionAsync`). Argent gained an equivalent structured event log ([software-mansion/argent](https://github.com/software-mansion/argent), `@swmansion/argent` >= 0.16.0), so Argent remote sessions can now feed the same timeline. This wires that up.

# How

Argent's tool-server writes bunyan JSONL records to its `ARGENT_EVENT_LOG` path (default `~/.argent/tool-server-events.jsonl`) once the `tool-server-event-log` flag is enabled. Each record is tagged with a `type` (`tool.invoked`, `tool.completed`, `tool.failed`, `service.*`, `tool_server.*`, ...).

- **Generic engine.** Extracted the file-tailing (byte-offset + partial-line/multi-byte reassembly + truncation-safe sequence numbering), upload-session creation, poll loop, and failure reporting out of `deviceRunSessionEvents.ts` into a producer-parameterized `startDeviceRunSessionEventCollectionAsync`. A `DeviceRunSessionEventSource` adapter only has to say where its files live and how to normalize one line. Diagnostic messages interpolate the producer id verbatim (no recasing), and phase tags are unchanged for agent-device.
- **Adapters.** Moved the agent-device adapter into `agentDeviceEvents.ts` and added `argentEvents.ts` — mirroring the existing `deviceRunSessionArtifacts.ts` (shared) + `agentDeviceArtifacts.ts`/`argentArtifacts.ts` (per-producer) split. The argent normalizer maps `tool.invoked` → `operation.started` and `tool.completed`/`tool.failed` → `operation.completed` (with `outcome`), correlates them via `toolInvocationId` → `operationId`, and passes other record types through unchanged.
- **Version floor & unconditional enable.** Bumped `MIN_ARGENT_REMOTE_SESSION_VERSION` from `0.12.0` to `0.16.0` — the first version that exposes the event log flag. With that floor guaranteed, `startArgentRemoteSession.ts` enables `tool-server-event-log` **unconditionally** (like the existing `artifacts-list-endpoint` flag). This makes event collection an explicit remote-session contract rather than silently producing no events.
- **Pinned path (review).** The event log path is pinned once and handed to both the tool-server (via `ARGENT_EVENT_LOG` in its spawn env) and the collector, so an ambient `ARGENT_EVENT_LOG` or a change to Argent's default can never make the two disagree.
- **Discovery errors (review).** File discovery treats only `ENOENT` as "no event file yet"; permission/I/O errors propagate to the collector's existing Sentry reporting instead of being silently swallowed.
- **Summary comes from Argent (review).** `summary` is Argent's `msg` verbatim rather than something we synthesize — Argent's `EventLogRecord` requires `msg` and builds it from each tool's `interaction.startedMsg`/`completedMsg`/`failedMsg` hooks, so it is already the human-readable description (parameterized by the actual args/result). `msg` is required in our schema too, so a record without one is a reported contract break rather than invented text.

Stale history is not a concern: Argent truncates the event file when the tool-server starts, and we tail from offset 0 with truncation handling.

Argent was checked against `origin/main` (latest, up to tag `v0.16.1`) — the event log landed in `v0.16.0` (`feat(tool-server): add tool-server event log`, #367). I did not `git pull` the local `~/expo/argent` clone because it had unrelated uncommitted work; I inspected `origin/main` directly instead.

# What we upload, and what the session UI should render

We deliberately upload **every** well-formed record and leave the choice of what to display to the consumer, so a failed session keeps the events that explain it. Filtering here would bake into `@expo/build-tools` (release cadence, and the data is gone for sessions in between); filtering in the UI is reversible with full history intact.

The only records we drop are blank lines and ones failing the schema (missing `time`/`type`/`msg`) — and those are reported to Sentry as `invalid-event`, not silently skipped.

`type` is enough to split the stream; every event also carries the raw Argent type as `data.sourceType`.

| Bucket | Normalized `type` | From Argent `type` |
| --- | --- | --- |
| **Timeline** — what the agent did to the app | `operation.started`, `operation.completed` (+ `outcome: success \| failure`, `durationMs`, `operationId`) | `tool.invoked`, `tool.completed`, `tool.failed` |
| **Diagnostic** — why a session misbehaved | `service.state_change`, `service.error`, `service.registered`, `tool.registered`, `tool_server.started`, `tool_server.stopping`, `tool_server.bind_failed`, `tool_server.start_failed`, `tool_server.deferred_to_existing` | same (passed through unchanged) |

⚠️ **The UI filter is required, not cosmetic.** Argent registers ~38 tools at startup, so an unfiltered timeline opens with ~40 `Tool <id> was registered.` lines before anything the user did. Render `operation.*` in the main timeline and keep the rest for a debug/expandable view.

Two things to know about the timeline text: a tool that has not defined `interaction` hooks yet falls back to Argent's generic strings (`Tool <id> was invoked.`, `Tool <id> completed in 12.34 ms.`), which improves on its own as Argent fills them in; and the diagnostic messages include internal details such as `127.0.0.1:<port>`.

# Test Plan

- `yarn jest-unit` in `packages/build-tools` — full suite passes (898 tests), including:
  - `argentEvents.test.ts` (new): tool lifecycle → operation vocabulary, failure-signal preservation, non-tool passthrough, missing required field (`type`/`msg`) reporting, and a non-`ENOENT` discovery error surfacing to Sentry.
  - `startArgentRemoteSession-orchestration.test.ts` (new): drives the build function's `fn` and verifies the event-log flag is enabled (before launch), the tool-server env and collector share one pinned path, and collection starts before / stops after the session-stop wait.
  - `agentDeviceEvents.test.ts` (moved): agent-device behavior unchanged.
  - `startArgentRemoteSession.test.ts`: version-floor checks + `stopArgentEventCollectionSafelyAsync`.
- `yarn typecheck`, `yarn lint`, `yarn fmt:check` all pass. Rebased onto latest `main`.

### DIFF
--- a/packages/build-tools/src/steps/utils/argentEvents.ts
+++ b/packages/build-tools/src/steps/utils/argentEvents.ts
@@ -1,0 +1,178 @@
+import { type bunyan } from '@expo/logger';
+import fs from 'node:fs';
+import path from 'node:path';
+import { z } from 'zod';
+
+import { type CustomBuildContext } from '../../customBuildContext';
+import {
+  type DeviceRunSessionEvent,
+  type DeviceRunSessionEventParseFailure,
+  startDeviceRunSessionEventCollectionAsync,
+} from './deviceRunSessionEvents';
+
+const ARGENT_PRODUCER = 'argent';
+// Argent's tool-server writes its event log to `~/.argent/tool-server-events.jsonl`
+// (its default $ARGENT_EVENT_LOG path) once the `tool-server-event-log` flag is on.
+export const ARGENT_EVENT_LOG_FILENAME = 'tool-server-events.jsonl';
+
+// Argent records are bunyan log lines tagged with a `type`. Map the tool-lifecycle
+// types onto the shared operation vocabulary; anything else (service.*,
+// tool_server.*, future types) passes through unchanged, mirroring agent-device.
+const ARGENT_EVENT_TYPE_TO_TYPE: Record<string, string> = {
+  'tool.invoked': 'operation.started',
+  'tool.completed': 'operation.completed',
+  'tool.failed': 'operation.completed',
+};
+
+// Bunyan bookkeeping plus the fields we promote to the top level of a
+// DeviceRunSessionEvent — excluded from the free-form `data` bag so it carries
+// only the record's own context (toolId, serviceId, failureSignal, ...).
+const NON_DATA_KEYS = new Set([
+  'name',
+  'hostname',
+  'pid',
+  'v',
+  'time',
+  'msg',
+  'level',
+  'type',
+  'toolInvocationId',
+  'durationMs',
+]);
+
+const ArgentEventSchema = z
+  .object({
+    time: z.string(),
+    type: z.string(),
+    msg: z.string().optional().catch(undefined),
+    level: z.number().optional().catch(undefined),
+    toolId: z.string().optional().catch(undefined),
+    toolInvocationId: z.string().optional().catch(undefined),
+    durationMs: z.number().optional().catch(undefined),
+  })
+  .passthrough();
+
+type ArgentEvent = z.infer<typeof ArgentEventSchema>;
+
+type ParsedArgentEvent = {
+  event?: ArgentEvent;
+  failure?: DeviceRunSessionEventParseFailure;
+};
+
+export async function startArgentEventCollectionAsync({
+  ctx,
+  deviceRunSessionId,
+  stateDir,
+  logger,
+  pollIntervalMs,
+}: {
+  ctx: CustomBuildContext;
+  deviceRunSessionId: string;
+  stateDir: string;
+  logger: bunyan;
+  pollIntervalMs?: number;
+}): Promise<{ stopAsync: () => Promise<void> }> {
+  return startDeviceRunSessionEventCollectionAsync({
+    ctx,
+    deviceRunSessionId,
+    logger,
+    pollIntervalMs,
+    source: {
+      producer: ARGENT_PRODUCER,
+      findEventFilesAsync: () => findArgentEventFilesAsync(stateDir),
+      sourceKeyForFile: eventFile => path.basename(eventFile, path.extname(eventFile)),
+      parseLine: ({ line, sourceKey, sequenceNumber, deviceRunSessionId }) => {
+        const { event, failure } = parseArgentEvent(line);
+        if (!event) {
+          return { failure };
+        }
+        return {
+          event: normalizeArgentEvent({ event, sequenceNumber, sourceKey, deviceRunSessionId }),
+        };
+      },
+    },
+  });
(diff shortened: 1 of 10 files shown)


packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
```diff
@@ -321,80 +322,25 @@ async function collectEventFileAsync({
...
-    ...(typeof durationMs === 'number' ? { durationMs } : {}),
-    summary,
-    data: {
-      ...event.details,
-      session: event.session,
-      sourceVersion: event.version,
-      ...(event.status ? { sourceStatus: event.status } : {}),
-    },
-  };
+function capitalize(value: string): string {
```
sjchmiela: i don't think we should capitalize stuff. it comes with problems. i'd rather maintain a mapping from producer input to producer label or use producer input directly.

(throwback to how long it took to realize why "iOS Build" becomes an "IOS Build" in `BoxHeader` in `website`)

packages/build-tools/src/steps/utils/argentEvents.ts
```diff
@@ -0,0 +1,184 @@
...
+    type,
+    ...(event.toolInvocationId ? { operationId: event.toolInvocationId } : {}),
+    ...(outcome ? { outcome } : {}),
+    ...(typeof event.durationMs === 'number' ? { durationMs: event.durationMs } : {}),
+    summary,
+    data,
+  };
+}
+
+function summarizeArgentEvent(event: ArgentEvent): string {
```
sjchmiela: this is what argent event has `msg` for

https://github.com/software-mansion/argent/pull/367/changes#diff-13131ae11198b09c0243d08f472d4eb34c1452f27a2bbe60db1c3d13f7b19397R151

sjchmiela (approved):
i think the two comments are worth fixing

packages/build-tools/src/steps/utils/argentEvents.ts
szdziedzic: You are right — `EventLogRecord` types `msg` as required and every emit site sets it, so the fallback was reinventing what Argent already gives us. Removed `summarizeArgentEvent` and use `msg` verbatim as the summary; `msg` is now required in the schema too, so a record without one is reported as a contract break instead of getting invented text. (agent-device keeps its fallback because its `summary` really is optional upstream.)

packages/build-tools/src/steps/utils/deviceRunSessionEvents.ts
szdziedzic: I went with using producer input directly

PR #4068, opened 2026-07-24
[eas-cli] Add project:icon:set command

brentvatne:
# Why

Setting a project's icon currently requires going through the website, which is tedious. The GraphQL API already supports icon uploads via app-scoped upload sessions, so this exposes it in the CLI as `eas project:icon:set PATH`.

# How

- Added `UploadSessionMutation.createAppScopedUploadSessionAsync` wrapping the `uploadSession.createAppScopedUploadSession` mutation with type `PROFILE_IMAGE_UPLOAD`, mirroring the existing account-scoped wrapper.
- Added `uploadAppScopedFileAtPathToGCSAsync` to `uploads.ts` to PUT the file to the returned signed URL.
- Added `AppQuery.byIdProfileImageUrlAsync`, a network-only query for `app.byId.profileImageUrl`.
- New command `project:icon:set`: validates the image (exists, PNG/JPEG, ≤ 10 MB), uploads it, then polls `profileImageUrl` (2s interval, 90s timeout) until it changes — the icon is processed asynchronously server-side (a cloud function resizes it and assigns it to the app), same as the website's icon form, which also polls. On success it links to the project overview page; on timeout it notes the icon may still appear shortly.

# Test Plan

- 6 new unit tests in `src/commands/project/icon/__tests__/set.test.ts` covering the upload+poll flow, first-time icon set, validation failures (missing file, unsupported format, oversize), and poll timeout.
- Tried it for real on https://expo.dev/accounts/brent-org/projects/euxy — the icon uploaded and appeared on the project page.
- `yarn typecheck`, `yarn lint`, and `yarn fmt:check` pass; README regenerated via `oclif readme`.

### DIFF
--- a/packages/eas-cli/src/commands/project/icon/set.ts
+++ b/packages/eas-cli/src/commands/project/icon/set.ts
@@ -1,0 +1,130 @@
+import { Args } from '@oclif/core';
+import chalk from 'chalk';
+import fs from 'fs-extra';
+import path from 'path';
+
+import { getProjectDashboardUrl } from '../../../build/utils/url';
+import EasCommand from '../../../commandUtils/EasCommand';
+import { ExpoGraphqlClient } from '../../../commandUtils/context/contextUtils/createGraphqlClient';
+import { EASNonInteractiveFlag } from '../../../commandUtils/flags';
+import { AppUploadSessionType } from '../../../graphql/generated';
+import { AppQuery } from '../../../graphql/queries/AppQuery';
+import Log, { link } from '../../../log';
+import { ora } from '../../../ora';
+import { uploadAppScopedFileAtPathToGCSAsync } from '../../../uploads';
+import { sleepAsync } from '../../../utils/promise';
+
+const ALLOWED_EXTENSIONS = ['.png', '.jpg', '.jpeg'];
+const MAX_IMAGE_SIZE_BYTES = 10 * 1024 * 1024; // matches the upload session's GCS limit
+const POLL_INTERVAL_MS = 2000;
+const POLL_TIMEOUT_MS = 90_000;
+
+export default class ProjectIconSet extends EasCommand {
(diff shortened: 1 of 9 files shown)


packages/eas-cli/src/commands/project/icon/set.ts
```diff
@@ -0,0 +1,130 @@
...
+const POLL_INTERVAL_MS = 2000;
+const POLL_TIMEOUT_MS = 90_000;
+
+export default class ProjectIconSet extends EasCommand {
+  static override description = 'set the project icon displayed on the EAS dashboard';
+
+  static override args = {
+    path: Args.string({
+      required: true,
+      description: 'Path to the icon image (PNG or JPEG, at most 10 MB, ideally square)',
```
tchayen: IIRC API will reject anything not square

tchayen (approved):
Everything looks good here (just unsure about that one error message and worth double checking the 10MB limit)

packages/eas-cli/src/commands/project/icon/set.ts
brentvatne: I checked in universe to be sure — there's no squareness validation anywhere in the pipeline:

- The website's client-side [`validateImageAsync`](https://github.com/expo/universe/blob/main/server/website/common/profileImageUtils.ts) checks extension, MIME type, size, and max dimensions (3000×3000), but not aspect ratio.
- The signed upload URL only enforces a 10 MB cap (`x-goog-content-length-range`).
- The [`profile-image-upload` cloud function](https://github.com/expo/universe/blob/main/functions/profile-image-upload/src/index.ts) does `sharp().resize(1024, 1024).png()` — the default `cover` fit, so non-square images are center-cropped to a square rather than rejected, and the finalize endpoint does no dimension checks.

That said, "ideally square" was underselling the cropping behavior, so I updated the description in 5b501be5 to say non-square images are center-cropped.

PR #4070, opened 2026-07-24
[eas-cli] Use npx in agent-device simulator instructions

szdziedzic:
# Why

EAS Simulator currently suggests running `agent-device` as if it were installed globally. Using `npx` makes the suggested command work with an on-demand local package invocation.

# How

Update the agent-device session instructions to print:

```sh
eas simulator:exec npx agent-device <command>
```

Update the matching simulator start command test expectation.

# Test Plan

- `corepack yarn build`
- `corepack yarn test src/commands/simulator/__tests__/start.test.ts --runInBand` (10 tests passed)
- `oxfmt --check` on both modified files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


sjchmiela (approved):
`npx -y`? or not

PR #4073, opened 2026-07-24
[eas-cli] Add account:avatar:set command

brentvatne:
# Why

Follow-up to #4068, which added `eas project:icon:set`. Setting an account avatar (which for a personal account is your user avatar) has the same problem: it currently requires going through the website. The GraphQL API already supports avatar uploads via account-scoped upload sessions, so this exposes it in the CLI as `eas account:avatar:set PATH [ACCOUNT_NAME]`.

There is no user-scoped upload session — a user's picture is their personal account's profile image (the website's personal settings page uploads to `user.primaryAccount.id`), so a single account-scoped command covers both users and organizations.

# How

- New command `account:avatar:set`: validates the image (exists, PNG/JPEG, ≤ 10 MB), resolves the target account the same way `account:usage` does (explicit `ACCOUNT_NAME` argument → the only account → interactive prompt; errors in `--non-interactive` mode without a name), uploads via the existing account-scoped upload session with type `PROFILE_IMAGE_UPLOAD`, then polls `account.byId.profileImageUrl` (2s interval, 90s timeout) until it changes — avatars are processed asynchronously server-side (resized to 1024×1024, non-square images center-cropped), same as project icons. On success it links to the account settings page.
- Extracted the image validation and polling logic shared with `project:icon:set` into `src/utils/profileImages.ts` (`validateProfileImageAsync`, `pollForProfileImageChangeAsync`) and refactored the project command to use it.
- Added `AccountQuery.byIdProfileImageUrlAsync`, a network-only query used for the before/after polling.
- Made `uploadAccountScopedFileAtPathToGCSAsync`'s progress handler optional (avatars are small enough not to need one).

# Test Plan

- 6 new unit tests in `src/commands/account/avatar/__tests__/set.test.ts` covering account resolution (explicit argument, single account, interactive prompt), the non-interactive error, the no-access error, and missing-file validation. The 6 existing `project:icon:set` tests still pass against the shared helpers.
- Tried it for real on https://expo.dev/accounts/brent-org/settings with a random non-square (600×400) image from the web — the avatar uploaded, was center-cropped, and appeared in the account settings.
- `yarn typecheck`, `yarn lint`, and `yarn fmt:check` pass; README regenerated via `oclif readme`.


brentvatne:
@copilot resolve the merge conflicts in this pull request

PR #4084, opened 2026-07-26
[eas-cli] Add eas testflight:feedback and eas testflight:crashes

brentvatne:
# Why

The Expo MCP server has `testflight_feedback` and `testflight_crashes`, but there was no way to get
either from the CLI. Nothing was blocking it: `@expo/apple-utils` (already a dependency) exposes
`App.getBetaFeedbackScreenshotSubmissionsAsync`, `App.getBetaFeedbackCrashSubmissionsAsync`, and
`BetaCrashLog.getCrashLogAsync` — the same App Store Connect endpoints the MCP calls. This wires
them up.

It also makes the EAS workflow `app_store_connect.beta_feedback` trigger actionable from a workflow
step: the trigger hands you `{ id, type, url }`, and both commands accept that directly.

# How

Two commands under a new `testflight` topic:

- `eas testflight:feedback` — screenshot feedback: tester comments, device info, screenshot URLs
- `eas testflight:crashes` — crash reports; with an ID, the full symbolicated crash log

**Credentials.** Same ASC API keys the MCP uses. `src/testflight/app.ts` prefers the key configured
for the project (ASC env vars → `eas.json` submit profile → the key on its iOS submission
credentials, via the existing `getAppStoreAuthAsync`), then falls back to exactly what the MCP does
— try every ASC API key on the owning account until one can see the bundle ID — then, only when
interactive, an Apple ID login.

**Workflow triggers.** `[ID]` accepts either `${{ app_store_connect.beta_feedback.id }}` (with
`--type` when it isn't the command's own kind) or `${{ app_store_connect.beta_feedback.url }}`,
which is the ASC self-link and so encodes both `id` and `type`. The type from a URL wins over the
command name, so either command name works as a workflow entry point. Input is resolved before
authenticating, so a malformed ID or URL fails fast.

**Pagination** uses the repo's standard `EasPaginatedQueryFlags` + `paginatedQueryWithConfirmPromptAsync`,
so interactive runs get the usual `Load more crashes?` prompt and numbering stays absolute across
pages. `apple-utils`' list helpers always walk every ASC page and its low-level `client` isn't
exported, so there is no supported way to expose a `--cursor`; instead one fetch backs all pages
(memoized in `TestFlightQuery`) and `--offset`/`--limit` slice it, which also makes the total count
free. `--limit` defaults to 20 rather than the shared flag's documented 50, because these render as
multi-line blocks.

**`--json`** on every path, with paging metadata, for agent/scripted use.

# Test Plan

25 unit tests across `src/testflight/__tests__/` covering normalization of the ASC `build`/`tester`
relationships, formatting, `{ id, type, url }` resolution, and the paging paths (including the
interactive `Load more?` flow, which can't be driven from a non-TTY shell). `yarn typecheck`,
`yarn lint`, and `yarn fmt:check` are clean.

Also exercised end to end against a real app in TestFlight (`dev.brent.euxy`), non-interactively:

```
$ eas testflight:feedback --limit 2 --offset 2 --non-interactive
Using App Store Connect API Key from EAS credentials service.

3. Feedback AI0AzUc98IV2ObozdgnSJxo
Submitted    2026-07-24, 5:37:38 p.m. (1 day ago)
Build        6
Device       iPhone17_3, iOS 26.5.2
Tester       Brent Vatne <brentvatne@gmail.com>
Locale       en-CA
State        battery 95%, disk 22.0 GB free of 237.9 GB, wifi
Comment      same feedback as velocity for gate
Screenshots  https://tf-feedback.itunes.apple.com/eimg/... (1179x2556)

4. Feedback ALwy1PzJ0qJJgm65uIqU1_4
...

Showing 4 of 5 feedback submissions. Use --offset and --limit to see more.
```

A crash URL passed to `testflight:feedback` still resolves as a crash, with its log:

```
$ eas testflight:feedback https://api.appstoreconnect.apple.com/v1/betaFeedbackCrashSubmissions/AAo2eIIfGzcb1BzuUv3xrh4
Crash AAo2eIIfGzcb1BzuUv3xrh4
Submitted  2026-07-24, 5:42:01 p.m. (1 day ago)
Build      6
Device     iPhone17_3, iOS 26.5.2
Arch       arm64e
Uptime     1.0s
Comment    i moved the numbers of steps down to the minimum and it crashed...

Crash log
Incident Identifier: 043E676D-C2B1-4C4D-825E-70FD5663F026
...357 lines...
```

JSON output, including paging metadata:

```
$ eas testflight:crashes --json
{ "crashes": [ ... ], "total": 1, "offset": 0, "limit": 20, "hasNextPage": false }
```

Error paths: a URL that isn't a beta feedback submission, and a `--type` that contradicts the URL,
both fail with an explanatory message before any Apple request.

Note for reviewers: `printJsonOnlyOutput` strips `null` repo-wide, so a crash with no comment has no
`comment` key rather than `comment: null`. I left that shared helper alone.


brentvatne:
@claude review

PR #4095, opened 2026-07-27
[eas-build-job][build-tools][steps] Accept legacy command/path custom function shape in function.yml; rename references in the consumers

sswrk:
# Why

We want to grandfather the legacy reusable functions from custom builds (https://docs.expo.dev/custom-builds/functions/) into the new reusable functions that can be used in workflows. This is to make the migration from custom builds to workflows easier.

Currently, we only support composite functions. The first step towards extending the support is modifying the `eas-build-job` Zod schema.

# How

- Added the `legacyFunction` module with Zod schemas to match the legacy custom builds functions.
  - Two separate schemas: `LegacyCommandFunctionConfigZ` (functions with `command:` present) and `LegacyPathFunctionConfigZ` (functions with `path:` present).
- Extracted `compositeFunction` and created a new `localFunction` module with schema that is a union of composite function schema and legacy function schemas.
- Renamed references of "composite local functions" schema in `steps` and `build-tools` that should reference "local  functions".

# Test Plan

Added unit tests.

### DIFF
--- a/packages/eas-build-job/src/legacyFunction.ts
+++ b/packages/eas-build-job/src/legacyFunction.ts
@@ -1,0 +1,95 @@
+/**
+ * Legacy single-step local function shape: shell `command` or JS module `path` in `function.yml`.
+ * Branches of `LocalFunctionConfigZ` in `./localFunction`.
+ */
+import { z } from 'zod';
+
+import { CompositeFunctionInputZ } from './compositeFunction';
+
+const LegacyFunctionOutputZ = z.union([
+  z.string().describe('Shorthand for a required output name.'),
+  z
+    .object({
+      name: z.string(),
+      required: z.boolean().optional(),
+    })
+    .strict(),
+]);
+
+const LegacyFunctionPlatformZ = z.enum(['darwin', 'linux']);
+
+const LegacyFunctionBaseZ = z.object({
+  /**
+   * @example
+   * name: Say hi
+   */
+  name: z.string().optional().describe('Display name of the function.'),
+  /**
+   * @example
+   * inputs:
+   *   - greeting
+   *   - name: platform
+   *     type: string
+   *     default_value: ios
+   */
+  inputs: z
+    .array(CompositeFunctionInputZ)
+    .optional()
+    .describe(
+      'Inputs accepted by the function. Each input is required unless it sets `required: false`.'
+    ),
+  /**
+   * @example
+   * outputs:
+   *   - name: version
+   *   - name: sha
+   *     required: false
+   */
+  outputs: z
+    .array(LegacyFunctionOutputZ)
+    .optional()
+    .describe(
+      'Outputs the function sets. Each output is required unless it sets `required: false`.'
+    ),
+  shell: z.string().optional().describe('Shell to run the function with.'),
+  supported_platforms: z
+    .array(LegacyFunctionPlatformZ)
+    .optional()
+    .describe('Runtime platforms the function can run on.'),
+  runs: z.never().optional(),
+});
+
+export const LegacyCommandFunctionConfigZ = LegacyFunctionBaseZ.extend({
+  /**
+   * @example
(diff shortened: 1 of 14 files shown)


packages/eas-build-job/src/legacyFunction.ts
```diff
@@ -0,0 +1,95 @@
...
+    .object({
+      name: z.string(),
+      required: z.boolean().optional(),
+    })
+    .strict(),
+]);
+
+const LegacyFunctionPlatformZ = z.enum(['darwin', 'linux']);
+
+const LegacyFunctionBaseZ = z.object({
```
hSATAC, line 57: Non-blocking, and this may be intentional: should the legacy schema also accept the camelCase spellings?

From what I can tell, those are the native keys in the old Joi schema and snake_case is added as a rename alias, so `supportedRuntimePlatforms`, `defaultValue`, `allowedValues`, and `allowedValueType` work in custom builds today. Here they fail with the generic union error:

```yaml
command: echo hi
supportedRuntimePlatforms: [darwin]
```

Maybe only grandfathering the documented snake_case shape is what we want. If so, could we add a rejection test to make that boundary explicit? Otherwise I think these branches should accept both.

packages/eas-build-job/src/localFunction.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type LocalFunctionConfig = CompositeFunctionConfig | LegacyFunctionConfig;
+
+export function isLegacyFunctionConfig(
+  config: LocalFunctionConfig
+): config is LegacyFunctionConfig {
+  return config.runs === undefined;
+}
+
+export type LocalFunctionCatalog = Record<string, LocalFunctionConfig>;
```
hSATAC, line 63: Minor and non-blocking: do we want to keep a deprecated alias for `CompositeFunctionCatalog`? It has been exported from the published package since 21.3.0, so removing it would be a source break for existing importers.

```ts
/** @deprecated Use `LocalFunctionCatalog`. */
export type CompositeFunctionCatalog = LocalFunctionCatalog;
```

The widened return types of `buildCompositeFunctionCatalogFromStepsAsync` and `buildLocalCompositeFunctionCatalogAsync` look like another possible source break: the loader still only returns `CompositeFunctionConfig`, but downstream code accessing `catalog[path].runs.steps` may stop compiling. Would it make sense to keep those return types composite-only until the loader starts returning legacy functions?

If we'd rather make both changes now, should they go under breaking changes in the changelog so the version bump comes out right?

hSATAC (commented):
I took a pass through this and left two inline questions. I don't have much context on legacy custom build functions, so I'm mostly calling out a couple of compatibility differences I noticed rather than making a call on the intended behavior.

Neither is meant to block from my side, and I'll leave the final approval to Stanley since he has more context on the intended compatibility scope.

packages/eas-build-job/src/legacyFunction.ts
sswrk: 👍 Good point, we should just keep full compatibility

packages/eas-build-job/src/localFunction.ts
sswrk: Restored the old names via aliases to avoid breaking changes. My thinking was that we're fine to change this, thes packages are for internal stuff only, but right, this is public and can be used by some outside consumer.

packages/steps/src/CompositeFunctionExpander.ts
```diff
@@ -66,7 +67,7 @@ type StepOverrides = {
 export class CompositeFunctionExpander {
   constructor(
     private readonly ctx: BuildStepGlobalContext,
-    private readonly compositeFunctionCatalog: CompositeFunctionCatalog,
+    private readonly compositeFunctionCatalog: LocalFunctionCatalog,
```
sjchmiela:
```suggestion
    private readonly localFunctionCatalog: LocalFunctionCatalog,
```
?

PR #4096, opened 2026-07-27
[steps][build-tools][eas-cli] Expand legacy command/path local functions to a single build step; rename references in the consumers

sswrk:
# Why

We want to grandfather the legacy reusable functions from custom builds (https://docs.expo.dev/custom-builds/functions/) into the new reusable functions that can be used in workflows. This is to make the migration from custom builds to workflows easier.

This PR adds support for parsing the legacy custom build functions into workflow steps.

# How

First, across `steps`, and the consumers (`build-tools` and `eas-cli`):

- Rename “composite local function” occurrences to “local function” in places where we support both composite and legacy functions (this is done in the first commit so I recommend taking a look and then reviewing the rest of the commits to avoid the renaming noise).

The actual feature in `steps`:

- Load legacy functions in the local function catalog (`utils/localFunctions.ts`).
  - Functions that specify `path:` have the `path` resolved in this step.
- Map a legacy config to `BuildFunction` (`utils/legacyFunction.ts`).
  -  Inputs and outputs default to required, which match the legacy functions, but is different from composites.
- Expand a legacy call into a single build step (`LocalFunctionExpander`). `BuildStep` already has the machinery to run such function.
  - Note: unlike composite functions, `working_directory` set by the caller is allowed, because there's just one step.

This affects consumers:

- `build-tools`: legacy functions referenced in workflows, hooks or composite functions used in workflows are now loaded.
- `eas-cli`: `eas workflow:validate` accepts legacy functions.

# Test Plan

Added unit tests.

### DIFF
--- a/packages/steps/src/LocalFunctionExpander.ts
+++ b/packages/steps/src/LocalFunctionExpander.ts
@@ -1,16 +1,18 @@
 /**
- * Expands local composite functions (`uses: ./path/to/function`) into a
- * {@link CompositeBuildStep} tree at parse time.
+ * Expands local functions (`uses: ./path/to/function`) into build steps at parse time.
  *
- * Each call becomes a node with prefixed child ids (`caller__inner`); the parser
- * flattens the tree into the workflow. Expanded steps carry a
- * {@link BuildStepCompositeFunctionScope} so `${{ steps.* }}` and `${{ inputs.* }}`
- * resolve against composite-function-local names.
+ * A composite function call becomes a {@link CompositeBuildStep} node with prefixed child ids
+ * (`caller__inner`); the parser flattens the tree into the workflow. Expanded steps carry a
+ * {@link BuildStepCompositeFunctionScope} so `${{ steps.* }}` and `${{ inputs.* }}` resolve
+ * against composite-function-local names. A single-step (`command`/`path`) function call becomes
+ * one ordinary build step keeping the caller's own id.
  */
 import {
   CompositeFunctionConfig,
   FunctionStep,
+  LegacyFunctionConfig,
   LocalFunctionCatalog,
+  LocalFunctionConfig,
   ShellStep,
   Step,
   isLegacyFunctionConfig,
@@ -32,6 +34,7 @@ import {
 import { CompositeBuildStep } from './CompositeBuildStep';
 import { BuildConfigError } from './errors';
 import { duplicates } from './utils/expodash/duplicates';
+import { createBuildFunctionFromLegacyFunctionConfig } from './utils/legacyFunction';
 import {
   getLocalFunctionCallWorkingDirectoryError,
   isLocalFunctionPath,
@@ -48,14 +51,16 @@ export type FunctionMaps = {
 
 type LocalFunctionCall = {
   functionPath: string;
-  /** Caller-assigned id used as prefix for all inner step ids. */
+  /** Caller-assigned id; for a composite function also the prefix of all inner step ids. */
   syntheticStepId: string;
   name?: string;
-  /** Caller-provided input values, consumed when composite function inputs are interpolated. */
+  /** Caller-provided input values, consumed when function inputs are interpolated. */
   callWith?: Record<string, unknown>;
   callIf?: string;
   parentScope?: BuildStepCompositeFunctionScope;
   inheritedEnv?: BuildStepEnv;
+  /** Rejected for composite calls. */
+  workingDirectory?: string;
 };
 
 type StepOverrides = {
@@ -83,9 +88,8 @@ export class LocalFunctionExpander {
     step: FunctionStep,
     functionPath: string,
     syntheticStepId: string
-  ): CompositeBuildStep {
-    this.rejectLocalFunctionCallWorkingDirectory(step);
-    return this.expand(
+  ): BuildStep[] {
+    const expanded = this.expandCall(
       {
         functionPath,
         syntheticStepId,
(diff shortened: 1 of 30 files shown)


sswrk:
Re. AI review:

> Local function load errors use plain Error, not BuildConfigError

Not something specific to this PR, ignoring

packages/steps/src/utils/legacyFunction.ts
```diff
@@ -0,0 +1,56 @@
...
+    return BuildStepInput.createProvider({
+      id: input,
+      required: true,
+      allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+    });
+  }
+  return BuildStepInput.createProvider({
+    id: input.name,
+    required: input.required ?? true,
+    defaultValue: input.default_value,
```
hSATAC: This looks like a compatibility gap for two kinds of `function.yml` that are accepted by custom builds today. Both pass `eas workflow:validate` because it only builds the catalog, then fail when the worker parses the step.

For typed scalars, the Joi schema coerces quoted defaults and allowed values to their declared type. Here the raw values are passed through, so something like `default_value: "42"` for a number input remains a string and fails parse-time validation, even though `getValue()` would coerce it correctly at runtime.

For JSON inputs, an object default can never match an object in `allowed_values` because `isRawValueOneOfAllowedValues()` uses `Array.includes()`, while Joi compares them structurally:

```yaml
inputs:
  - name: payload
    type: json
    default_value: { a: 1 }
    allowed_values: [{ a: 1 }]
command: echo hi
```

Could we preserve the Joi behavior here by coercing typed scalars and comparing JSON values structurally? Otherwise, I think these should at least fail during catalog loading with a field-level error, so `eas workflow:validate` catches them.

packages/steps/src/utils/localFunctions.ts
```diff
@@ -0,0 +1,236 @@
...
+      parsed = YAML.parse(rawContents);
+    } catch (err) {
+      throw new Error(`Failed to parse local function "${functionPath}" YAML at ${absolutePath}`, {
+        cause: err as Error,
+      });
+    }
+
+    const result = LocalFunctionConfigZ.safeParse(parsed);
+    if (!result.success) {
+      throw new Error(`Invalid local function "${functionPath}": ${z.prettifyError(result.error)}`);
```
hSATAC, line 182: Non-blocking, but parsing the union here seems to make the validation errors less useful for both shapes.

A composite `function.yml` with a typo like `shel:` used to report `Unrecognized key: "shel"`. Now it gets the generic three-shapes message. Legacy functions lose their field-level errors too; for example:

```yaml
command: echo hi
supported_platforms: [windows]
```

reports that it must declare one of `runs.steps`, `command`, or `path`, even though `command` is already present.

Could we use the key that is present (`runs`, `command`, or `path`) to re-parse the matching branch and report its errors, keeping the generic message for cases where the shape is actually ambiguous? Since this PR is where the union starts being used in production, I think it may be worth handling here.

packages/steps/src/LocalFunctionExpander.ts
```diff
@@ -79,38 +84,68 @@ export class CompositeFunctionExpander {
...
+    const { functionPath } = call;
+    // One BuildFunction per path. Relative ./../ keys cannot collide with registered ids.
+    const buildFunction = (this.functionMaps.buildFunctionById[functionPath] ??=
+      createBuildFunctionFromLegacyFunctionConfig(functionPath, config));
+    return buildFunction.createBuildStepFromFunctionCall(this.ctx, {
+      id: call.syntheticStepId,
+      name: call.name ?? config.name ?? functionPath,
+      callInputs: call.callWith,
+      workingDirectory: call.workingDirectory,
+      shell: config.shell,
```
hSATAC, line 129: Non-blocking question: is honoring the function-level `shell` here intentional?

From what I can tell, custom builds don't use `BuildFunction.shell`; only the call-site shell is passed through. That means a `shell:` declared by the function is currently ignored and the command runs with the default `/bin/bash -eo pipefail`. After migration, the same function would use its declared shell and lose the default `-eo pipefail` behavior.

Honoring the declaration may be the better behavior, but it does mean migration can change how an existing function runs. If that's intentional, could we adjust the “runs it exactly like a custom build does” docstring in `utils/legacyFunction.ts`?

hSATAC (commented):
I took a pass through this and left three inline comments, mostly around compatibility with existing custom build functions. I don't have much context on the legacy behavior, so I'm calling out differences I noticed rather than making a call on which ones we need to preserve.

I'll leave the final approval to Stanley since he has more context on the intended compatibility scope.

packages/steps/src/utils/legacyFunction.ts
sswrk: I didn't realize it. Added coercion https://github.com/expo/eas-cli/pull/4096/commits/ce7b0c1f76c6b6a238cfb74013e51632b1ed820f

packages/steps/src/utils/localFunctions.ts
sswrk: Addressed it in PR lower in the stack, by implementing a custom error logic: https://github.com/expo/eas-cli/pull/4095

packages/steps/src/LocalFunctionExpander.ts
sswrk: Yes, this was intentional even though it's a real behavior divergence. My thinking is: since we can easily match user's intent, let's do that instead of preserving the confusing behavior.

Clarified the module doc that stated that there are "no differences" in behavior.

PR #4099, opened 2026-07-27
[eas-cli] Support naming simulator sessions

szdziedzic:
# Why

Simulator sessions are only identifiable by a UUID, so juggling several at once means matching ids by hand. Part of ENG-25427, alongside expo/universe#29429 (server) and expo/universe#29430 (dashboard).

Depends on expo/universe#29429 being deployed to production. Until then `verify-graphql-code` fails during codegen document validation with `Cannot query field "name" on type "DeviceRunSession"`, because it introspects `api.expo.dev`. Everything else, including the tests on all three Node versions, passes.

# How

`eas simulator:start --name` forwards the label to `createDeviceRunSession`. `simulator:get` and `simulator:list` both always show a `Name` row, printing `null` when unset to match how those commands already render `Started at`, `Finished at`, and `Metadata`. Both include the name in `--json`, where `printJsonOnlyOutput` drops it when unset.

`generated.ts` and `graphql.schema.json` were produced by running codegen against the committed prod introspection snapshot with the new field injected, so the diff is exactly the `name` additions. They should be a no-op once regenerated against a deployed server.

# Test Plan

CI passes, apart from the GraphQL check noted above, which needs expo/universe#29429 deployed first.

### DIFF
--- a/packages/eas-cli/src/commands/simulator/start.ts
+++ b/packages/eas-cli/src/commands/simulator/start.ts
@@ -115,6 +115,10 @@ export default class SimulatorStart extends EasCommand {
       nonInteractive,
     });
 
+    // The server rejects blank names, so trim here and treat a whitespace-only
+    // --name as if it had been omitted rather than surfacing a validation error.
+    const name = flags.name?.trim() || undefined;
+
     await loadSimulatorEnvAsync(projectDir);
     const existingDeviceRunSessionId = process.env[EAS_SIMULATOR_SESSION_ID];
     if (existingDeviceRunSessionId && !flags.force) {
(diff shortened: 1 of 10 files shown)


gwdp (approved):

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -136,6 +140,7 @@ export default class SimulatorStart extends EasCommand {
     try {
       const session = await DeviceRunSessionMutation.createDeviceRunSessionAsync(graphqlClient, {
         appId: projectId,
+        name: flags.name,
```
gwdp: nit: trim

szdziedzic: fixed

PR #4110, opened 2026-07-29
[build-tools] Install ffmpeg for argent screen recording when missing

szdziedzic:
# Why

Argent screen recording does not work on EAS Simulator sessions. The failure is a missing `ffmpeg` binary on the worker VMs.

Verified live on session `019fadda-2063-7fbd-a6c6-7176b0de2681` (iOS, `--type argent`):

```
POST /tools/screen-recording-start  →
{"error":"[Tool:screen-recording-start] `ffmpeg` was not found on PATH.
 Install it (e.g. `brew install ffmpeg`) to record the screen."}
```

ffmpeg **is** argent's encoder — it pipes simulator frames into ffmpeg's stdin to produce the mp4. Argent resolves it from `PATH`, then `/opt/homebrew/bin`, `/usr/local/bin`, `/usr/bin`. The worker images ship none of them.

Nothing else is wrong. Argent's `screen-recording-start` checks, in order:

1. platform / tvOS
2. simulator-server frame stream → `SCREEN_RECORDING_STREAM_UNAVAILABLE`
3. `startCapture` → `resolveFfmpeg()` → `SCREEN_RECORDING_FFMPEG_NOT_FOUND`

We hit **3**, which proves **2 passed**: the frame stream already works on the VM. ffmpeg is the only blocker.

Also worth noting: with `--type argent` the tool-server runs **on the EAS VM**, and `list-devices` returned a booted `iPhone 17` with no `remote:` prefix. So argent's "remote simulators are unsupported" gate does not apply. This is a missing binary, not an architecture problem.

## Why this lives in eas-cli and not only in the image

There is a companion infra PR that adds ffmpeg to the worker images: https://github.com/expo/turtle-v2/pull/2598. That change applies to **images built from there on**. It deliberately does not update or backfill existing images.

So this PR is not a stopgap for a short rollout window. It is what makes recording work on every worker running a current image, for as long as those images stay in rotation. Once a worker boots an image that already carries ffmpeg, the check finds it and this does nothing.

# How

`ensureFfmpegInstalledAsync` in `remoteDeviceRunSession.ts`, called from `eas/start_argent_remote_session`.

- **Detection matches argent's own resolution** — `PATH` plus the same three Homebrew prefixes — so we never reinstall a binary argent can already find, and the step is a no-op on images that ship it.
- **Both platforms.** Homebrew on macOS (iOS simulators), apt on Linux (Android emulators). Android is covered because argent's capability is `android: { emulator: true, device: true, unknown: true }`, simulator-server has an `android` subcommand, and nothing on that path bypasses `resolveFfmpeg()`. The Linux workers have passwordless sudo (`%sudo ALL=(ALL) NOPASSWD: ALL`).
- **apt index refreshed first.** The worker's index can be older than the image it booted from, which makes the install 404 on a moved package. A failed refresh is not fatal, since the existing index may still resolve.
- **Runs in the background.** `brew install ffmpeg` pulls a large dependency tree and takes minutes. Awaiting it would delay session readiness by that much. Fired with `void` instead, so the session comes up at its usual speed and only a recording started in the first moments misses ffmpeg. This is the main design decision worth pushback if you disagree.
- **Best-effort, never throws.** Screen recording is one optional argent tool. A failed install is logged and reported to Sentry, and the session continues without it. Never rejecting is also what makes the `void` safe.

# Test Plan

Unit tests in `remoteDeviceRunSession.test.ts` cover six paths: already on a fallback path, already on `PATH`, installs with Homebrew on darwin, installs with apt on linux, still installs when the apt refresh fails, and warns without throwing when the install fails.

```
Test Suites: 99 passed, 99 total
Tests:       931 passed, 931 total
```

`yarn typecheck` (13 projects), `yarn lint` (0 errors), and `yarn fmt:check` all pass.

Not yet verified end-to-end on a live session — this is a draft. To verify, start an argent session on a current image, confirm the build log shows the install, then call `screen-recording-start` and expect a started recording instead of the error above.

### DIFF
--- a/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
@@ -20,6 +20,7 @@ import { sleepAsync } from '../../utils/retry';
 import { pollArgentArtifactsForUploadAsync } from '../utils/argentArtifacts';
 import { ARGENT_EVENT_LOG_FILENAME, startArgentEventCollectionAsync } from '../utils/argentEvents';
 import {
+  ensureFfmpegInstalledAsync,
   getDeviceRunSessionIdOrThrow,
   getNgrokAuthtokenOrThrow,
   getNgrokTunnelDomainOrThrow,
@@ -88,6 +89,12 @@ export function createStartArgentRemoteSessionBuildFunction(
       if (runtimePlatform === BuildRuntimePlatform.DARWIN) {
         await selectXcodeDeveloperDirectoryAsync({ env, logger });
       }
+
+      // Argent shells out to ffmpeg to record the screen. Installing it pulls a
+      // large Homebrew dependency tree, so do not block session readiness on it:
+      // the session comes up at its usual speed and only a recording started in
+      // the first moments misses ffmpeg. Never rejects, so `void` is safe.
+      void ensureFfmpegInstalledAsync({ runtimePlatform, env, logger });
 
       logger.info('Enabling the Argent artifacts list endpoint flag.');
       await spawn(
--- a/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
+++ b/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
@@ -230,8 +230,13 @@ async function installFfmpegWithAptAsync({
  * on PATH" — on macOS (iOS simulators) and Linux (Android emulators) alike.
  *
  * Best-effort by design: screen recording is one optional argent tool, so a
- * failure here is logged and the session continues without it. Never rejects,
- * which is what lets the caller run it in the background.
+ * failure here is logged and the session continues without it.
+ *
+ * The whole body is wrapped because the caller runs this in the background with
+ * `void`. There is no unhandledRejection handler in the worker, so a rejection
+ * escaping here would crash the process and take the live session with it.
+ * `spawn` is not an async function and can throw synchronously, which
+ * `asyncResult` cannot catch — it only wraps an already-created promise.
  */
 export async function ensureFfmpegInstalledAsync({
   runtimePlatform,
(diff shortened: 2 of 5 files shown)


packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -179,6 +180,96 @@ async function sleepUntilAbortedAsync(
   }
 }
 
+// Argent encodes screen recordings by piping simulator frames into `ffmpeg`,
+// which it resolves from PATH and then from these prefixes. Keep the list in
+// sync with argent so we never reinstall a binary it can already find.
+const FFMPEG_FALLBACK_PATHS = [
```
sjchmiela: I don't see a reason to do it like this. I think our list should only check our expected places for ffmpeg and our expected list is just $PATH?

Does argent really not adhere to $PATH?

packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
```diff
@@ -89,6 +90,12 @@ export function createStartArgentRemoteSessionBuildFunction(
         await selectXcodeDeveloperDirectoryAsync({ env, logger });
       }
 
+      // Argent shells out to ffmpeg to record the screen. Installing it pulls a
+      // large Homebrew dependency tree, so do not block session readiness on it:
+      // the session comes up at its usual speed and only a recording started in
+      // the first moments misses ffmpeg. Never rejects, so `void` is safe.
+      void ensureFfmpegInstalledAsync({ runtimePlatform, env, logger });
```
sjchmiela, line 97: Probably good for now but I wish we had easier time:
- logging stuff in phases at will
- run steps in parallel

Then we wouldn't mix so much in this one function and one log group…

Going to wait for hooks and custom fns to land and see if adding `parallel: true` would be so hard

sjchmiela (approved):
The detection thing is very weird to me

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
szdziedzic: I simplified it

szdziedzic: they also read these paths, but we don't have to

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -179,6 +180,87 @@ async function sleepUntilAbortedAsync(
   }
 }
 
+// Argent encodes screen recordings by piping simulator frames into `ffmpeg`,
+// which it resolves from PATH. It inherits this step's env, so PATH here is the
+// PATH argent will search — no need to probe install prefixes ourselves.
+async function isFfmpegAvailableAsync(env: BuildStepEnv): Promise<boolean> {
+  return (await asyncResult(spawn('sh', ['-c', 'command -v ffmpeg'], { env }))).ok;
```
sjchmiela: honest question -- do we need `sh -c`?

https://github.com/expo/eas-cli/blob/c3c3f087acab6c7b094bde663bf9e9bc45d5506e/packages/worker/src/env.ts#L72

sjchmiela (approved):
tyty!

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
szdziedzic: it should be better now

PR #4113, opened 2026-07-29
[eas-cli] Stop simulator job when start is canceled

sjchmiela:
Fixes [ENG-25545](https://linear.app/expo/issue/ENG-25545/stop-simulator-job-run-when-start-is-canceled-before-session-is-ready).

The SIGINT handler was registered only after the simulator became ready, so canceling earlier left the Turtle job running until it timed out. This registers one handler immediately after session creation and reuses it for the whole session lifecycle.

Added a regression test and tested it live in `test-production`: canceled while it was still waiting for readiness, the session never started (`startedAt: null`), and its final status was `STOPPED`.

Also ran the targeted tests, typecheck, lint, and formatter.

### DIFF
--- a/packages/eas-cli/src/commands/simulator/start.ts
+++ b/packages/eas-cli/src/commands/simulator/start.ts
@@ -141,6 +141,7 @@ export default class SimulatorStart extends EasCommand {
     const createSpinner = ora('🚀 Creating simulator session').start();
     let deviceRunSessionId: string;
     let deviceRunSessionUrl: string;
+    let sessionInterrupt: SessionInterrupt | undefined;
     try {
       const session = await DeviceRunSessionMutation.createDeviceRunSessionAsync(graphqlClient, {
         appId: projectId,
@@ -157,6 +158,7 @@ export default class SimulatorStart extends EasCommand {
         session.app.slug,
         deviceRunSessionId
       );
+      sessionInterrupt = registerSessionInterrupt(deviceRunSessionId);
       const simulatorEnvWritten =
         !jsonFlag && flags['out-config-type'] === OUT_CONFIG_TYPE_VALUES.Dotenv
           ? await writeSimulatorEnvSafelyAsync(projectDir, {
@@ -170,6 +172,7 @@ export default class SimulatorStart extends EasCommand {
       );
     } catch (err) {
       createSpinner.fail('Failed to create simulator session');
+      sessionInterrupt?.dispose();
       throw err;
     }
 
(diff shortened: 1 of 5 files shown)


packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -207,17 +217,30 @@ export default class SimulatorStart extends EasCommand {
...
 
+    if (sessionInterrupt.signal.aborted) {
+      await stopDeviceRunSessionAfterInterruptAsync({
+        graphqlClient,
+        deviceRunSessionId,
+        projectDir,
+        spinner: pollSpinner,
+        sessionInterrupt,
+      });
+      return;
```
szdziedzic, line 237: This makes cancel exit 0. Before the PR, Ctrl+C here hit Node's default SIGINT and exited 130. Now the handler swallows it and we return.

With --json that's a trap: no JSON on stdout, exit 0. SESSION=$(eas simulator:start --json | jq -r .id) gets an empty id and thinks it worked.

Can we process.exit(130) after cleanup? Exiting 0 is fine once the session is live — Ctrl+C is the point there. But here we never delivered anything.

packages/eas-cli/src/commands/simulator/start.ts
```diff
@@ -207,17 +217,30 @@ export default class SimulatorStart extends EasCommand {
           break;
         }
 
-        await sleepAsync(POLL_INTERVAL_MS);
+        await Promise.race([sleepAsync(POLL_INTERVAL_MS), sessionInterrupt.abortPromise]);
```
szdziedzic: Promise.race picks a winner but doesn't cancel the loser. sleepAsync never exposes its timer handle, so nothing clears it.

On Ctrl+C the abort wins instantly, cleanup runs, we return - but the 5s timer is still pending, and Node won't exit while a timer is pending. I timed it: race settles at 1ms, process exits at 5005ms.

So the user sees "✔ Simulator session stopped", then up to 5s of nothing.

Fix: make sleep abortable so the timer gets cleared.

Then no race needed. Line 391 has the same issue, so one fix covers both.

szdziedzic (approved):

packages/eas-cli/src/commands/simulator/start.ts
sjchmiela: Fixed. Cancellation before readiness now stops the simulator session, cleans up the local environment, and exits with status 130. Cancellation after the session is live keeps the existing successful exit behavior.

packages/eas-cli/src/utils/promise.ts
```diff
@@ -2,8 +2,23 @@
...
+    const onAbort = (): void => {
+      clearTimeout(timeoutId);
+      resolve();
+    };
+    const timeoutId = setTimeout(() => {
+      signal?.removeEventListener('abort', onAbort);
+      resolve();
+    }, ms);
+    signal?.addEventListener('abort', onAbort, { once: true });
+  });
```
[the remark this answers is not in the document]
sjchmiela: No code change needed here: the initial aborted check, timeout creation, and listener registration run synchronously on the same JavaScript event-loop turn. This signal is owned locally and aborted by the SIGINT callback, which cannot interleave between those statements, so the described race cannot occur.

packages/eas-cli/src/utils/__tests__/promise.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    const sleepPromise = sleepAsync(5_000, abortController.signal);
+
+    expect(jest.getTimerCount()).toBe(1);
+    abortController.abort();
+    await sleepPromise;
+
+    expect(jest.getTimerCount()).toBe(0);
+    jest.useRealTimers();
+  });
+});
```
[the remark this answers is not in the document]
sjchmiela: Fixed in 15f4210a1. The suite now uses a string description and restores real timers in afterEach, so cleanup still runs when an assertion fails.

PR #4114, opened 2026-07-30
Run Expo serve-sim under EAS supervision

szdziedzic:
# Why

EAS Simulator currently launches the production-only `serve-sim-szdziedzic` fork and depends on its private tunnel flags and `Tunnel:` stdout format. The official `@expo/serve-sim` package is now published, but it is not a safe drop-in until EAS owns the production concerns that do not belong in the OSS server.

Coordinated serve-sim change: https://github.com/expo/serve-sim/pull/36

That PR must merge and be published to npm's `latest` dist-tag before this change is rolled out, because the currently published package must contain the readiness/encoder startup contract.

# How

- always install `@expo/serve-sim@latest` instead of using the personal fork or a hard-coded package version
- translate the VM-safe production streaming policy to the official CLI flags: explicit WebRTC/VP8, 1280px, 60 fps, 3 Mbps, and MJPEG fallback quality 0.55
- allocate a loopback port, launch serve-sim, and poll its stable `/readyz` endpoint instead of scraping stdout
- keep Cloudflare TURN discovery in EAS and pass the resulting ICE arguments to serve-sim
- make EAS own ngrok setup, the auth token, high-entropy tunnel hostname, and lifecycle
- return explicit cleanup handles for ngrok and detached processes, and stop preview resources from every remote-session caller

This is intentionally split across repositories: serve-sim owns generic streaming/readiness primitives; EAS owns infrastructure, credentials, policy, versioning, and supervision.

# Test Plan

- `yarn build` — all 13 monorepo targets passed, including `@expo/build-tools`
- all directly affected suites outside the sandbox — 24 tests passed
- focused final suites plus build-tools typecheck — 15 tests passed
- `yarn fmt:check`
- `yarn lint` — 0 errors (13 unrelated existing warnings)
- the broader build-tools unit run reached 918 passing tests; its 10 failures were local sandbox restrictions around npm access, process inspection, and macOS keychain creation. The affected process-inspection suite passes outside the sandbox.
- end-to-end verified the coordinated serve-sim build negotiated VP8 and delivered live 589×1280 video while applying the requested 60 fps / 3 Mbps sender policy

## Rollout order

1. Merge https://github.com/expo/serve-sim/pull/36.
2. Publish and verify that npm's `latest` dist-tag contains the coordinated serve-sim changes.
3. Exercise this draft in an EAS Simulator staging session.
4. Merge this PR to switch production off `serve-sim-szdziedzic`.


packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -327,9 +376,108 @@ export function spawnDetached({
...
+    SERVE_SIM_MJPEG_QUALITY,
+    '--video-bitrate',
+    SERVE_SIM_VIDEO_BITRATE,
+    '--video-fps',
+    SERVE_SIM_VIDEO_FPS,
+    ...turnArgs,
+  ];
+}
+
+async function findAvailablePortAsync(): Promise<number> {
```
sjchmiela, line 417: This is weird to me. Shouldn't serve-sim pick a port itself? Or, couldn't we just assign a well known port to serve-sim?

sjchmiela (approved):

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
szdziedzic: serve-sim can choose/scan for a port, but EAS needs to know the selected port to poll /readyz and configure ngrok. I allocated one in EAS to avoid parsing serve-sim output. That said, since the runner is single-tenant and only starts one serve-sim instance, using a well-known port like 3200 is simpler. I can switch this to a constant.

PR #4120, opened 2026-07-30
[build-tools] Publish device run session events in real time

sjchmiela:
## Why

We want to deliver logs faster than every 5 seconds.

## How

Same way we publish logs in worker to `HttpLogStream`, publish them when collecting device run session logs. Whenever we see an event, we write to both loggers.

## Test Plan

CI should pass. Only applies in development and staging.


szdziedzic (approved):
awesome! I love it!

PR #4129, opened 2026-08-01
[eas-cli] Fix `eas login` redirect URI when running in GitHub Codespaces

Ignigena:
# Why

Our team relies heavily on GitHub Codespaces for development, especially for smaller changes that don't require a full local Native development environment. Codespaces lets contributors jump into the codebase from a browser or a lightweight setup without provisioning all of that tooling locally.

However, `eas login` doesn't work from within GitHub Codespaces. The CLI starts a local callback server and sends `expo.dev` a hardcoded `redirect_uri` of `http://localhost:<port>/auth/callback`, but inside a Codespace the browser completing the login can only reach the forwarded port's public URL (e.g. `https://<codespace-name>-<port>.app.github.dev`), not `localhost`. This leaves the login flow unable to complete when developing inside a Codespace without awkardly copying the correct hostname into the resulting 404 redirect in the browser.

Everything else about the flow already works correctly in Codespaces: the callback HTTP server starts, listens, and Codespaces automatically forwards/exposes the port, so a redirect to that port does reach the running server. The only problem is that the URL handed to `expo.dev` as `redirect_uri` needs to use the forwarded hostname instead of `localhost` for the browser to be able to follow it.

> [!NOTE]
> With this change the CLI sends the correct `*.app.github.dev` `redirect_uri`, but the login flow still doesn't complete end-to-end yet, since `expo.dev`/the auth API currently appears to only redirect back to `localhost` redirect URIs and silently drops non-`localhost` ones. To fully fix this for Codespaces users, the server side will also need to allowlist `*.app.github.dev` as a valid redirect host. This should be low risk: even if someone's forwarded port were made public, Codespaces already shows an interstitial warning page to any unauthenticated visitor before they reach the forwarded service.

# How

- Added `getRedirectBaseUrl(port)` in [expoBrowserAuthFlowLauncher.ts](packages/eas-cli/src/user/expoBrowserAuthFlowLauncher.ts), which returns the Codespaces forwarded-port URL (built from the `CODESPACES`, `CODESPACE_NAME`, and `GITHUB_CODESPACES_PORT_FORWARDING_DOMAIN` environment variables) when running in a Codespace, falling back to `http://localhost:<port>` otherwise.
- `buildRedirectUri` now builds the full redirect URI from that base URL using the `URL` constructor instead of hardcoding the `localhost` scheme/host.
- The callback server itself still binds to `localhost`, since Codespaces port forwarding proxies external requests back to `localhost` inside the container, so only the URL handed to `expo.dev` needed to change.

# Test Plan

- Ran `eas login` inside a GitHub Codespace and confirmed the printed login URL and the `redirect_uri` query param now point at the Codespace's forwarded-port URL (`https://<codespace-name>-<port>.app.github.dev/auth/callback`) instead of `localhost`.
- Completed the login in the browser and confirmed the request reaches `expo.dev` with the correct `redirect_uri`, but `expo.dev` does not redirect back to the `*.app.github.dev` URL, so the callback server never receives the code (see note above — needs a server-side allowlist change to close the loop).
- Ran `eas login` outside of Codespaces (`CODESPACES` unset) and confirmed the redirect URI is unchanged (`http://localhost:<port>/auth/callback`) and login still completes as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ This is the log of notable changes to EAS CLI and related packages.
 - [eas-cli] clean up error handling in local builds. ([#4105](https://github.com/expo/eas-cli/pull/4105) by [@douglowder](https://github.com/douglowder))
 - [build-tools] Install ffmpeg when it is missing so Argent screen recording works in EAS Simulator sessions. ([#4110](https://github.com/expo/eas-cli/pull/4110) by [@szdziedzic](https://github.com/szdziedzic))
 - [eas-cli] Stop simulator job runs when `eas simulator:start` is canceled before the session is ready. ([#4113](https://github.com/expo/eas-cli/pull/4113) by [@sjchmiela](https://github.com/sjchmiela))
-- [eas-cli] Fix `eas login` redirect URI to use the forwarded port URL when running in GitHub Codespaces. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@Ignigena](https://github.com/Ignigena))
+- [eas-cli] Fix `eas login` redirect URI to use the forwarded port URL when running in GitHub Codespaces. ([#4129](https://github.com/expo/eas-cli/pull/4129) by [@Ignigena](https://github.com/Ignigena))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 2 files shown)


douglowder:
@claude review once

douglowder:
Adding @ide and @wschurman for a security review on this one.

brentvatne:
@copilot resolve the merge conflicts in this pull request

ide:
1. I don't think we should hardcode GitHub codespaces. Presumably other web-based sandboxes have the same issue.
2. An area to deeply investigate is the redirect URI and whether non-localhost domains should be allowed for a CLI.

ide (changes requested):
Looking into this, including [RFC 8252](https://datatracker.ietf.org/doc/html/rfc8252), I don't think we should add non-localhost or non-loopback addresses for the redirect URL. We enforce this server-side as well as noted in the test plan and to my understanding it is important we do not relax this even if GitHub happens to have some protections today.

The workaround for now is to provision a [personal access token](https://expo.dev/accounts/[account]/settings/access-tokens) and provide it as `EXPO_TOKEN`.

Longer-term, I believe the solution is [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628), which uses device codes.

PR #4106, opened 2026-07-29
[eas-cli] Add integrations:supabase:connect

gwdp:
## Why

Adding `eas integrations:supabase:connect`: authorize Supabase on an Expo account, then either link an existing project or provision a new one. Writes the public URL/key into EAS env and installs the SDK.

## How

- Main path: browser OAuth, pick org/region when needed, then provision a new primary project (async job; CLI polls the receipt) or attach one you already have
- `--link` takes a project ref, dashboard URL, or `*.supabase.co` URL and sets that as the app primary without creating a new Supabase project
- `--environment` is a follow-up after primary connect: provisions a separate hosted project and writes URL/key only into those EAS envs (e.g. preview), leaving the primary alone. Incompatible with `--link` / `--reauth`
- `--reauth` resets the OAuth connection (and cascade-clears the EAS project link), then prompts link-existing (default) vs provision-new; pass `--link` to skip the prompt. `--overwrite` skips prompts when replacing existing env vars

## Test Plan

easd against staging (provision, `--link`, `--environment`)

carbon (1)

PR #4130, opened 2026-08-01
[eas-cli] Supabase integration base 

gwdp:
## Why

Adding the shared layer for the Supabase integration commands (connect / disconnect / dashboard): GraphQL, provisioning, and writing credentials into EAS env.

While extracting it, environment defaults/resolution moved out of integrations/. They aren't integration-specific, and eas env already uses the same list.

## How

- Supabase GraphQL client pieces (mutations/queries/types)
- Helpers under integrations/supabase/ (and a thin shared/ for SDK install + .env.local)
- Top-level environments/ so default env resolution isn't buried in integrations
- Hardened background-job receipt polling for async provision

## Test Plan

CI


packages/eas-cli/src/environments/resolve.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (environment.length < 3 || environment.length > 100 || !/^[a-z0-9_-]+$/.test(environment)) {
+      throw new Error(
+        `Invalid EAS environment "${environment}". Use 3–100 lowercase letters, numbers, dashes, or underscores.`
+      );
+    }
+  }
+  return environments;
+}
+
+export async function resolveTargetEnvironmentsAsync(
```
FiberJW, line 42: Is this meant to be used directly? Or always wrapped by its call site? If it's an internal method, we should rename it as such so we don't need to reassign a name every time we import it

packages/eas-cli/src/integrations/supabase/env.ts
```diff
@@ -0,0 +1,119 @@
...
+import { EnvironmentVariableVisibility } from '../../graphql/generated';
+import { confirmAsync } from '../../prompts';
+import {
+  mergeEnvContent,
+  writeEnvLocalAsync as writeEnvLocalWithLabelAsync,
+} from '../shared/envFile';
+
+export type { EnvVar };
+export { mergeEnvContent };
+export { upsertEasEnvVarAsync } from '../../environments/variables';
```
FiberJW: Let's avoid re-exporting these. We can use the shared modules directly and only keep supabase/integration-specific things in their own separate modules

packages/eas-cli/src/integrations/supabase/provision.ts
```diff
@@ -0,0 +1,324 @@
...
+    return null;
+  }
+}
+
+export async function pollForConnectionAsync(
+  graphqlClient: ExpoGraphqlClient,
+  accountId: string
+): Promise<SupabaseConnectionData> {
+  const deadline = Date.now() + CONNECTION_POLL_TIMEOUT_MS;
+  for (;;) {
```
FiberJW, line 188: Is there a reason this doesn't use `setInterval`?

packages/eas-cli/src/integrations/supabase/provision.ts
```diff
@@ -0,0 +1,324 @@
...
+  appId: string,
+  project: SupabaseProjectData
+): Promise<string> {
+  const spinner = ora('Waiting for the Supabase project to finish provisioning').start();
+  const deadline = Date.now() + READINESS_POLL_TIMEOUT_MS;
+  // The server returns a null key while the project is still provisioning but throws for a real
+  // problem (revoked authorization, etc.). Tolerate a transient blip, but stop retrying for the
+  // full timeout once the errors are persistent — that isn't a provisioning delay.
+  let consecutiveErrors = 0;
+  for (;;) {
```
FiberJW, line 232: Same about `setInterval`

packages/eas-cli/src/integrations/supabase/provision.ts
```diff
@@ -0,0 +1,324 @@
...
+    try {
+      key = await SupabaseMutation.fetchSupabasePublishableKeyAsync(graphqlClient, appId);
+      consecutiveErrors = 0;
+    } catch (error) {
+      consecutiveErrors += 1;
+      Log.debug(`Polling for the Supabase project readiness failed, will retry: ${error}`);
+      if (consecutiveErrors >= MAX_CONSECUTIVE_READINESS_ERRORS) {
+        spinner.fail("Couldn't reach the Supabase project");
+        throw error;
+      }
```
FiberJW, line 248: In these polling functions, we should explicitly handle non-transient errors and cancel the loop

packages/eas-cli/src/integrations/supabase/provision.ts
```diff
@@ -0,0 +1,324 @@
...
+  for (;;) {
+    let connection: SupabaseConnectionData | null = null;
+    try {
+      connection = await SupabaseQuery.getSupabaseConnectionByAccountIdAsync(
+        graphqlClient,
+        accountId,
+        { useCache: false }
+      );
+    } catch (error) {
+      Log.debug(`Polling for the Supabase connection failed, will retry: ${error}`);
```
FiberJW: Let's catch non-transient errors and bail early here

FiberJW (changes requested):
Requesting some small changes to improve polling:

packages/eas-cli/src/environments/resolve.ts
gwdp: It's meant to be used directly; inside this PR it has no caller, which is why it reads ambiguous. [#4106](https://github.com/expo/eas-cli/pull/4106) calls it from `connect.ts`

packages/eas-cli/src/integrations/supabase/env.ts
gwdp: Gone, thank you for catching it!

packages/eas-cli/src/integrations/supabase/provision.ts
gwdp: Plain `setInterval` fires on a fixed schedule no matter how long the request takes, so slow calls could overlap, and avoiding that needs the promise wrapper plus `clearInterval` on every exit path. I only got awareness of `setIntervalAsync` version on this PR tbh, have been using  `for (;;) {` which looks to be more common in this codebase. Let me know how you feel about keeping it; otherwise, happy to explore setIntervalAsync

packages/eas-cli/src/integrations/supabase/provision.ts
gwdp: Good catch!  Updating so both loops bail on non-transient errors

packages/eas-cli/src/integrations/supabase/provision.ts
gwdp: Good catch, updating!

gwdp:
> Requesting some small changes to improve polling:

Thank you for the review, got some changes in and replied to the rest; let me know if you disagree on any, happy to discuss more :)

packages/eas-cli/src/integrations/supabase/provision.ts
FiberJW: Got it! I wasn't aware of that pattern, so it looked a bit odd to me. We can leave it

FiberJW (approved):

PR #4138, opened 2026-08-03
[eas-cli] Add --link-project-id flag to eas build

williamgrosset:
# Why

Users/agents should be able to assert which EAS project a build should belong to.

# How
- Added `--link-project-id <id>` to `eas build` to link the directory to an existing project before building, like `eas init --id <id>`. Supports non-interactive: `eas build -p ios --link-project-id <id> --non-interactive`.
- Default is an assertion: if the config already points at a different project (or the owner/slug conflict), it prompts to overwrite. `--non-interactive --force` deliberately relinks.

# Test Plan

Manually verified an EAS project is successfully linked and `eas build` continues it's execution:

https://github.com/user-attachments/assets/8ccfdc45-51ce-41b3-9fe0-900407c4f4ba

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [eas-cli] Add `--link-project-id` flag to `eas build` to link the directory to an existing project before building, like `eas init --id`. ([#4138](https://github.com/expo/eas-cli/pull/4138) by [@williamgrosset](https://github.com/williamgrosset))
+
 ### 🐛 Bug fixes
 
 - [build-tools] Revert "Pin the default `agent-device` version for remote sessions" now that `agent-device` 0.20.5 fixes the broken release. ([#4144](https://github.com/expo/eas-cli/pull/4144) by [@gwdp](https://github.com/gwdp))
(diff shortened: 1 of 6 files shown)


brentvatne:
why not run `eas init` first? i'm afraid if we add this flag here, we basically have to add it everywhere

williamgrosset:
@brentvatne the idea is to streamline the steps needed for a first build (if `eas init` hasn't been ran yet). Both the new `--link-project-id` and `--auto-configure-update` flags are attempts to reduce the amount of steps that a user/agent would need to run. There is an argument that these flags are redundant and the user/agent will figure out which path to follow, curious to know what you think.

brentvatne:
i would assume that an agent would be able to follow those steps, yeah. what if you want to publish an update before running a build? wouldn't you also need to pass in that project id? what about if you're running a workflow first? or a simulator? and so on.

i'm afraid if we follow this path we end up with lots of flag noise. can you find other similar CLI tools and see how they handle this case for agents?

williamgrosset:
@brentvatne Yeah, the flag noise can be a lot. I think it's better for each command to stay composable and scoped. I looked at a few other CLIs and something like a `EXPO_PROJECT_ID` env var could be a better pattern in the future instead of threading a flag everywhere. I'll close both #4138 and #4139 for now and decide later if it's worth returning to.

PR #4139, opened 2026-08-03
[eas-cli] Add --auto-configure-update flag to eas build

williamgrosset:
# Why

Users/agents should be able to configure EAS update before a build command runs, so that a user doesn't have to complete a rebuild after.

# How

- Added `--auto-configure-update` to `eas build`: before the build starts, installs `expo-updates`, sets `updates.url` and `runtimeVersion` in the app config, syncs native config, and adds channels to build profiles
- Equivalent to running `eas update:configure` before the build, scoped to the build's `--platform`
- Repeat runs change nothing. `ensureEASUpdateIsConfiguredInEasJsonAsync` (shared with `eas update:configure`) now skips the `eas.json` write when no profile changed

# Test Plan

Manually verified  EAS update is configured and `eas build` continues it's execution:

https://github.com/user-attachments/assets/42c9df6f-69bf-4c4e-82c7-8bbd45e47b83

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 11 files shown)


brentvatne:
why not just run `eas build:configure` and `eas update:configure` independently?

williamgrosset:
See #4138 for discussion.

PR #4146, opened 2026-08-03
[gha] Publish eas-cli pointer packages instead of moving npm dist-tags

szdziedzic:
# Why

npm [restricted sensitive operations for access tokens](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/) in August 2026. CI can no longer run `npm dist-tag add`, so the release flow fails to move the `latest-eas-build` and `latest-eas-build-staging` tags (see the [v21.5.0 run](https://github.com/expo/eas-cli/actions/runs/30844588940)). The tags are currently moved by hand after a Slack reminder (9276e22f), and `move-eas-build-tag.yml` only works in dry-run mode.

CI can still run `npm publish` with OIDC trusted publishing (in use since #3342), and every publish moves the published package's `latest` dist-tag. This PR rebuilds the pointer mechanism on top of that one allowed operation.

# How

**New pointer packages** in `pointer-packages/`:

- `eas-cli-for-eas-build` — pins the `eas-cli` version for production EAS Build infrastructure.
- `eas-cli-for-eas-build-staging` — same for staging.

Each package pins an exact `eas-cli` version in `dependencies` and ships a small bin that forwards to the pinned CLI. The `latest` version of each package is the pointer. `pointer-packages/publish.mjs` (dependency-free Node script) stamps the version (`1.0.<epoch seconds>`) and the `eas-cli` pin, verifies the pinned version exists on the registry (with retries), and publishes staging before production.

**New workflow** `publish-eas-cli-pointer.yml` (reusable + manually dispatchable): publishes the pointers for any existing `eas-cli` version with `target` (`staging`/`production`/`both`) and `dry_run` inputs. It replaces `move-eas-build-tag.yml` (deleted) for promotion and rollback.

**Release flow** (`release.yml`): after `lerna publish`, CI now publishes the staging pointer → runs the worker system tests → publishes the production pointer, mirroring the staged rollout in `worker.yml`. A system-test failure leaves production pinned to the previous version; recovery is a manual dispatch of `publish-eas-cli-pointer.yml`.

**Version-source fix**: the old tag steps (and the current Slack reminder) read the version from `lerna.json`. Lerna bumps only changed packages, so that version may not exist for `eas-cli` — the [v21.5.1 run](https://github.com/expo/eas-cli/actions/runs/30848749138) was green but shipped no `eas-cli@21.5.1` (`latest` is still 21.5.0), so the reminder asked for tags npm rejects with 404. The new flow reads `packages/eas-cli/package.json` instead, and the existence check in `publish.mjs` fails loudly instead of pointing at a missing version.

## Not in this PR (follow-ups)

1. The build fleet still resolves the CLI through the legacy dist-tags (`resolveEasCommandPrefixAndEnvAsync()` in `packages/build-tools/src/utils/easCli.ts`). A follow-up PR switches it to the pointer packages; until that ships and the worker is redeployed, the release flow keeps posting the Slack reminder to move the legacy tags manually (now with the corrected version).
2. One-time ops setup, tracked in `pointer-packages/README.md`: confirm the package names are free on npm, publish both packages once manually (first publish of a new package may not work via trusted publishing), and configure trusted publishers for both packages (repo `expo/eas-cli`, workflow `publish-eas-cli-pointer.yml`; if npm validates the caller workflow for reusable workflows, also add `release.yml`).
3. Audit www/turtle for direct reads of the legacy tag names before freezing them, and deprecate `EasCliNpmTags` in `@expo/eas-build-job`.

# Test Plan

- `node pointer-packages/publish.mjs --eas-cli-version 21.5.0 --target both --dry-run` — resolves the version, stamps both packages, and `npm publish --dry-run` packs them (staging first):
  ```
  Resolved eas-cli@21.5.0 on the registry.
  Publishing eas-cli-for-eas-build-staging@1.0.1785794190 (pins eas-cli@21.5.0) [dry run]...
  + eas-cli-for-eas-build-staging@1.0.1785794190
  Publishing eas-cli-for-eas-build@1.0.1785794191 (pins eas-cli@21.5.0) [dry run]...
  + eas-cli-for-eas-build@1.0.1785794191
  Done.
  ```
- The same command with `--eas-cli-version 21.5.1` (not on the registry) retries 5× and exits non-zero — the guard for the lerna.json/eas-cli version mismatch above.
- Both changed workflow files parse as valid YAML; `yarn lint` and `yarn fmt:check` pass.
- After merge, first real verification: manual `dry_run: true` dispatch of `publish-eas-cli-pointer.yml`, then a real `staging`-only publish once the one-time npm setup is done.


douglowder:
Thanks @szdziedzic ! I came up with an alternative way of fixing this without any NPM dependency, see ENG-25633 . Let's sync up tomorrow and discuss.

PR #4147, opened 2026-08-04
[eas-build-job][build-tools] Replace eas-cli npm dist-tags with cli-versions.json

douglowder:
# Why

Add `cli-versions.json` as the source of truth for the staging/production eas-cli versions, to remove the dependency on NPM tags to determine the correct version used in workflow jobs.

# How

Adding new function `fetchEasCliVersionsAsync` in the eas-build-job package. This reads the file from main on GitHub; build-tools uses it to build the eas-cli@<version> npx argument and falls back to the `latest-eas-build*` dist-tags on any failure.  The release workflow now bumps cli-versions.json in the release commit and no longer posts the manual npm dist-tag reminder.

# Test Plan

New unit tests added.


douglowder:
* **#4147** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4147?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4147?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/build-tools/src/utils/easCli.ts
```diff
@@ -1,8 +1,21 @@
...
+ * Resolves the `eas-cli` versions to install for staging/production. Prefers the
+ * versions committed to `cli-versions.json` (fetched from expo/eas-cli); on any
+ * failure falls back to the `latest-eas-build*` npm dist-tags.
+ */
+async function resolveEasCliVersionsAsync(): Promise<EasCliVersions> {
+  try {
+    return await fetchEasCliVersionsAsync();
+  } catch {
+    return { STAGING: EasCliNpmTags.STAGING, PRODUCTION: EasCliNpmTags.PRODUCTION };
+  }
```
szdziedzic: should we report these failures to Sentry so we know when it stops to work for some reason?

packages/eas-build-job/src/common.ts
```diff
@@ -399,3 +399,34 @@ export enum EasCliNpmTags {
...
+ */
+export interface EasCliVersions {
+  STAGING: string;
+  PRODUCTION: string;
+}
+
+const EasCliVersionsZ = z.object({
+  STAGING: z.string(),
+  PRODUCTION: z.string(),
+});
```
szdziedzic, line 422: should we make this validation tighter and make these values match semver regex pattern? I would like to avoid situation where this can become attack vector and is injected into command on user VM.

cli-versions.json
```diff
@@ -0,0 +1,4 @@
+{
+  "STAGING": "21.5.1",
+  "PRODUCTION": "21.5.1"
+}
```
szdziedzic, line 4: As far as I can tell 21.5.1 doesn't exist on npm. Is it intentional?

.github/workflows/release.yml
```diff
@@ -87,6 +53,11 @@ jobs:
         run: yarn install --immutable
       - name: Resolve version
         run: echo "EAS_CLI_VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
+      - name: Update cli-versions.json
+        run: |
+          jq --arg version "$EAS_CLI_VERSION" '.STAGING = $version | .PRODUCTION = $version' \
+            cli-versions.json > cli-versions.json.tmp
```
szdziedzic: Update cli-versions.json uses `$EAS_CLI_VERSION` from the git tag, which is the `lerna.json` version. Any release that doesn't change `eas-cli` (exactly the v21.5.1 case) writes an unpublished version into the file. And unlike the old `npm dist-tag add`, which failed loudly with a 404, `jq` succeeds silently and the fleet breaks ~minutes later. Fix: read `jq -r .version packages/eas-cli/package.json` instead - the step runs in the tag checkout, and at `v21.5.1` that file correctly says `21.5.0`. An optional `npm view eas-cli@$VERSION` version guard before writing would make it airtight.

packages/eas-build-job/src/common.ts
```diff
@@ -399,3 +399,34 @@ export enum EasCliNpmTags {
...
+
+const CLI_VERSIONS_URL = 'https://raw.githubusercontent.com/expo/eas-cli/main/cli-versions.json';
+
+/**
+ * Fetches and parses `cli-versions.json` from the `main` branch of
+ * expo/eas-cli. Throws if the file cannot be fetched or does not match the
+ * expected shape; callers should fall back to {@link EasCliNpmTags}.
+ */
+export async function fetchEasCliVersionsAsync(): Promise<EasCliVersions> {
+  const response = await fetch(CLI_VERSIONS_URL);
```
szdziedzic: I think this fetch should have a timeout so we don't hang here if GH has outage.

douglowder:
@claude review once

packages/build-tools/src/utils/easCli.ts
```diff
@@ -20,25 +33,26 @@ export async function resolveEasCommandPrefixAndEnvAsync(): Promise<{
   extraEnv: Env;
 }> {
   const npxArgsPrefix = (await isAtLeastNpm7Async()) ? ['-y'] : [];
+  const versions = await resolveEasCliVersionsAsync();
```
szdziedzic: should we cache it? every call does separate GH request to fetch versions

szdziedzic (changes requested):

cli-versions.json
douglowder: Good catch, it may be that the release flow failed before NPM publish

packages/build-tools/src/utils/easCli.ts
douglowder: Yes, that would catch situations where GitHub API was unavailable

packages/build-tools/src/utils/easCli.ts
douglowder: I'll have a look to see if this is needed

douglowder: Claude decided not to cache, because "the benefit is marginal (a few calls per build) and it made per-test isolation hard".

szdziedzic (approved):
LGTM overall! Thanks for crafting this! 🙏 

Should we also update `RELEASING.md`? 

Do we want to automatically kick system tests off as a part of release workflow and make this `promote-eas-cli-production.yml` flow automatically start once they pass as a follow up?

douglowder:
@szdziedzic made more changes in response to Claude review, will do some manual checks and then merge

packages/build-tools/src/utils/easCli.ts
douglowder: @szdziedzic after latest AI review, it looks like my AI agrees that a module-level cache would be a good idea, so I've implemented that.

douglowder:
> Uncached GitHub fetch runs on every build, called more than once per build — packages/build-tools/src/utils/easCli.ts:19 (correctness) · id:dc7e6caf1d20

​Fixed in the latest commit.

PR #4151, opened 2026-08-05
[eas-cli] Point gated accounts at the EAS Simulator waitlist

szdziedzic:
# Why

EAS Simulator is behind the `device-run-sessions` account feature gate. Accounts without access got a dead end:

- `eas simulator:availability` said only "it's coming soon", with no way to ask for access.
- `eas simulator` failed with a generic server permission error.

There is a waitlist at https://expo.dev/services/simulators, but nothing in the CLI pointed at it. This PR surfaces it on both paths.

# How

Added a shared message helper in `packages/eas-cli/src/simulator/utils.ts` (`EAS_SIMULATOR_WAITLIST_URL` + `formatSimulatorUnavailableMessage`) so both commands print the same text:

```
EAS Simulator isn't available on <account> yet — it's coming soon.
Join the waitlist to get access: https://expo.dev/services/simulators
```

- **`simulator:availability`**: the `available: false` branch prints that message. `--json` output gains a `waitlistUrl` field, present only when gated, so scripts and agents can pass the link on.
- **`eas simulator`** (`packages/eas-cli/src/commands/simulator/index.ts`): checks the gate with `DeviceRunSessionAvailabilityQuery` right after context resolution, and throws `EasCommandError` with the same message when it is off.

Two decisions worth a look:

**The gate is checked before the create mutation, not by matching the server's error.** The CLI has no coded error to match on for this gate today, so a pre-flight check is the reliable signal. Failing early also means a gated user is never prompted for a platform, and no session or `.env.eas-simulator` file is created. The cost is one extra GraphQL query per run, negligible next to a command that polls for up to 15 minutes.

**Expo admins skip the check entirely.** `actor.isExpoAdmin` already ships on every actor variant via the `LoggedIn` context, so admins keep working against accounts without the gate and the server stays the authority. Wrapping the whole block (rather than extending the `if`) also skips the extra query for them. Note this only removes the *client-side* block — if the API enforces the account gate with no admin override, an admin still gets the server's error, same as before this PR.

`simulator:availability` deliberately still reports the account gate for admins: that command answers "is this account enabled?", which stays true regardless of who asks.

# Test Plan

Automated — `cd packages/eas-cli && yarn test src/commands/simulator src/simulator`: 9 suites, 61 tests pass, including new cases:

- `availability.test.ts`: gated JSON output carries `waitlistUrl`; enabled JSON output omits it; the human-readable message includes the waitlist line.
- `index.test.ts`: a gated account rejects with the waitlist message and reaches neither `createDeviceRunSessionAsync` nor the platform prompt; an Expo admin on a gated account skips the availability query and creates the session.

Also ran `yarn typecheck`, `yarn lint`, and `yarn fmt:check` from the repo root — all clean.

Note on the full `eas-cli` suite: `go-test`, `update/embedded/delete`, and `update/embedded/view` fail intermittently under full-suite parallelism. They fail the same way on the unmodified base commit and the failing set varies per run, so they are pre-existing flakes unrelated to this change.

Not covered: `simulator:list`, `get`, `stop`, `exec`, and `events` still surface the raw server error for gated accounts. The helper is in place if we want to wire them up in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
+- [eas-cli] Point accounts without EAS Simulator access at the waitlist: `eas simulator:availability` includes the link (and a `waitlistUrl` field in `--json`), and `eas simulator` now fails early with the same message instead of a generic permission error. ([#4151](https://github.com/expo/eas-cli/pull/4151) by [@szdziedzic](https://github.com/szdziedzic))
 - [build-tools] Revert "Pin the default `agent-device` version for remote sessions" now that `agent-device` 0.20.5 fixes the broken release. ([#4144](https://github.com/expo/eas-cli/pull/4144) by [@gwdp](https://github.com/gwdp))
 
 ### 🧹 Chores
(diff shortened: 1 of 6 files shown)


krystofwoldrich (approved):
Looks good, this will be helpful for people who discover the command early.

PR #4153, opened 2026-08-05
[eas-cli] Improve non-interactive error for unconfigured projects

williamgrosset:
# Why

The current error tells users to run `eas init`, but that command needs interaction and fails under the same conditions that produced the error.

# How

Extract one builder, `getUnconfiguredProjectError`, and call it from all three throw sites. Every path now lists both recovery commands with their required flags, plus the accounts the user can create projects in. Also suppress the `EAS project not configured.` log warning in non-interactive mode, since every error message now states it.

# Test Plan

Updated the existing tests to assert the message on each path. Added a test showing that a single-account actor in non-interactive mode reaches the project fetch, which is where its error now comes from.

Manually verified with a local build running `project:info --non-interactive`:

Screenshot 2026-08-05 at 6 02 42 PM


brentvatne:
@williamgrosset - can you add a screenshot of the output to the PR description? it's always good to see what it actually looks like when rendered in the terminal

williamgrosset:
@brentvatne Good call, I updated the indentation and line breaks so it matches our existing styles. I think it looks better now

#### Before
Screenshot 2026-08-05 at 4 02 54 PM

#### After
Screenshot 2026-08-05 at 4 07 53 PM

(similar to `init` error):
Screenshot 2026-08-05 at 4 21 29 PM

brentvatne:
can we fix the duplicated "EAS project not configured" message?

williamgrosset:
Updated PR description:

Screenshot 2026-08-05 at 6 02 42 PM

brentvatne (approved):

PR #4154, opened 2026-08-05
[build-tools] Classify Argent session artifacts by MIME type

szdziedzic:
# Why

Currently all argent screenshots and recordings are classified as `other` type on website

# How

Use mime type of argent artifacts to select correct kind and make it display nicely on website

# Test Plan

Tests


gwdp (approved):
🔥

PR #4155, opened 2026-08-05
[build-tools] Make configure-version gradle script compatible with AGP 9

gabrieldonadel:
# Why

EAS injects an `eas-build.gradle` script into Android projects to set `versionCode` and `versionName` from the build job. That script sets per-output overrides through `applicationVariants.all`. AGP 9 removes the `applicationVariants` API from the new DSL, so the injected script breaks the Gradle configuration for project using react-native 0.87
# How

The `defaultConfig` assignments stay unchanged. The per-output override block moves out of the `android {}` block and now picks an API at runtime:

- If the `android` extension still exposes `applicationVariants` (AGP 8 and older), the script keeps the existing `versionCodeOverride` / `versionNameOverride` path.
- Otherwise it uses the `androidComponents.onVariants` variant API and sets the `versionCode` / `versionName` properties on each output.

# Test Plan

CI should be green


sjchmiela (approved):
tested the change across every agp minor from 7.2 to 8.13, all worked

PR #4157, opened 2026-08-05
[build-tools] Stop remote simulator sessions after max_idle_time_minutes without activity

szdziedzic:
# Why

Now we have event logs for agent-device and argent and we can finally add `--max-idel-time-minutes` setting to simulator session.

# How

- `eas/start_argent_remote_session` and `eas/start_agent_device_remote_session` get a new optional `max_idle_time_minutes` input. A missing value disables the timeout.
- we record timestamp of latest event
- we pass it to `waitForDeviceRunSessionStoppedAsync` and use it to stop a session when we go over max idle time

For now only argent/agent-device session events count as activity. Interaction through the serve-sim web preview does not produce these events yet.

# Deployment order

1. **This PR (build-tools)
2. https://github.com/expo/universe/pull/29784 (WWW)
3. https://github.com/expo/eas-cli/pull/4156 (CLI flag)

# Test Plan

CI passes


krystofwoldrich (approved):
Looks good.

PR #4161, opened 2026-08-06
[eas-cli] Fail when --private-key-path is set without codeSigningCertificate

gwdp:
## Why
When `--private-key-path` is passed without `updates.codeSigningCertificate` in the resolved app config: that **implies an intent to sign**, but the CLI previously ignored the flag and published unsigned.

## How
- In `getCodeSigningInfoAsync`, throw if a private key path is provided without a certificate path
- Leave the unsigned path alone when neither is set

## Test Plan
- New unit tests for no-cert/no-key → undefined, and no-cert/with-key → throw
- CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🐛 Bug fixes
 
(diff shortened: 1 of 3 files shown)


wschurman (approved):
Will need to make the equivalent change in @expo/cli I believe.

gwdp:
> Will need to make the equivalent change in @expo/cli I believe.

Thank you, addressing it over https://github.com/expo/expo/pull/48631

PR #4164, opened 2026-08-07
[build-tools] Copy artifacts before upload

sjchmiela:
# Why

Files such as emulator logs can still change while `eas/upload_artifact` archives them. This can make archive creation fail or produce an unclear cutoff.

# How

Add an optional `copy_before_upload` input. When enabled, the function copies all matched artifact paths to a unique temporary directory, uploads the stable copies, and removes the temporary directory after the upload. The default behavior does not change.

# Test Plan

- `yarn workspace @expo/build-tools test src/steps/functions/__tests__/uploadArtifact.test.ts --runInBand`
- `yarn fmt:check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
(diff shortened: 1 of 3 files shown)


sjchmiela:
Superseded by the simplified cross-repository design: #3627 now only exposes native emulator logs through the logcat_directory step output, and WorkflowMaestroTestJob copies that output into the existing Maestro test-results artifact. The generic copy_before_upload option is no longer needed.

PR #4171, opened 2026-08-11
[eas-cli] Print the Supabase authorize URL before opening the browser

gwdp:


gwdp:
Folded into #4106 as a commit on the same branch.

PR #4187, opened 2026-08-13
[build-tools] Allow running Maestro tests with `maestro-runner`

sjchmiela:
# Why

`maestro-runner` promises to be faster, more reliable alternative to the official Maestro CLI.

# How

Added support for `EAS_MAESTRO_BACKEND` and/or input to `eas/maestro_tests` which opts in to using `maestro-runner` over `maestro` command.

`maestro-runner` doesn't support DADB, but we don't error if both are enabled.

`maestro-runner` has slightly different screenshots and JUnit format -- both are handled. It also exposes a nicer JSON file that lets us parse failed flows without having to parse JUnit file.

This adds relatively big chunk of code, but it's all guarded by `backend === maestro-runner` so maybe it's not so bad? I considered splitting logic into `MaestroCliUtils` and `MaestroRunnerUtils` and expose similar interfaces from them, but I feel it was too complicated.

# Test Plan

- Xcode 16 mixed-flow proof: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019fffe8-9bd8-74f9-87b6-d7963ee6d941
  - Automatically selected maestro-runner 1.1.15 without a workflow version input.
  - Attempt 1 ran both mixed flows.
  - `report.json` selected only `flows/mixed/fail.yml` for attempt 2.
  - Uploaded the expected failure screenshot, merged JUnit report, and complete results archive.
- Android runner success: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffcff-4dcb-724f-8671-51c5644f49e1
- Android official Maestro success: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffcf8-ffb2-7c84-bd60-682ffc4a3e7e
- Android mixed expected failure: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffcfa-6123-7f50-bbbf-42ab03560617
- iOS runner success on Xcode 26: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffd22-da8f-79dc-b9c9-3802c5694f7e
- iOS mixed expected failure on Xcode 26: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffd1e-5923-7b3b-b5cb-c8329a357e97
- iOS official Maestro success on Xcode 26: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffd2b-4b37-7601-b665-a675d39f1cdc
- Reusable proof workflows: https://github.com/expo/workflows-testing/pull/11

### DIFF
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -2,6 +2,7 @@ import { asyncResult } from '@expo/results';
 import { XMLBuilder, XMLParser, XMLValidator } from 'fast-xml-parser';
 import fs from 'fs/promises';
 import path from 'path';
+import { z } from 'zod';
 
 export interface MaestroFlowResult {
   name: string;
@@ -190,6 +191,69 @@ export async function junitFileHasFileAttrs(junitFile: string): Promise<boolean>
 
 async function fileExists(absPath: string): Promise<boolean> {
   return (await asyncResult(fs.stat(absPath))).ok;
+}
+
+// maestro-runner writes report.json as the source of truth for a run. Use it for runner control
+// flow because sourceFile preserves the exact flow path, while older runner JUnit reports flatten
+// it to a basename. Keep only passed and failed flows so the result matches Maestro JUnit reports,
+// which do not include skipped flows.
+const MaestroRunnerRecordedFlowSchema = z.object({
+  name: z.string().min(1),
+  sourceFile: z.string().min(1),
+  status: z.enum(['passed', 'failed']),
+});
+const MaestroRunnerReportSchema = z.object({
+  flows: z
+    .array(
+      z.discriminatedUnion('status', [
+        MaestroRunnerRecordedFlowSchema,
+        z.object({ status: z.literal('skipped') }),
+      ])
+    )
+    .transform(flows =>
+      flows.filter(
+        (flow): flow is z.infer<typeof MaestroRunnerRecordedFlowSchema> => flow.status !== 'skipped'
+      )
+    ),
+});
+
+type MaestroRunnerReport = z.infer<typeof MaestroRunnerReportSchema>;
+
+export async function parseFailedFlowsFromMaestroRunnerReport(args: {
+  reportDirectory: string;
+  workingDirectory: string;
+}): Promise<string[] | null> {
+  const report = await parseMaestroRunnerReport(args.reportDirectory);
+  if (report === null) {
+    return null;
+  }
+
+  const failedPaths = [
+    ...new Set(report.flows.filter(flow => flow.status === 'failed').map(flow => flow.sourceFile)),
+  ];
+  if (failedPaths.length === 0) {
+    return null;
+  }
+
+  for (const flowPath of failedPaths) {
+    if (!(await fileExists(path.resolve(args.workingDirectory, flowPath)))) {
+      return null;
+    }
+  }
+  return failedPaths;
+}
+
+export async function parseMaestroRunnerReport(
+  reportDirectory: string
+): Promise<MaestroRunnerReport | null> {
+  let report: unknown;
+  try {
+    report = JSON.parse(await fs.readFile(path.join(reportDirectory, 'report.json'), 'utf8'));
+  } catch {
+    return null;
+  }
+  const result = MaestroRunnerReportSchema.safeParse(report);
+  return result.success ? result.data : null;
 }
 
 // Group by `file=` so two same-named flows in different files stay separate.
--- a/packages/build-tools/src/steps/functions/maestroResultParser.ts
+++ b/packages/build-tools/src/steps/functions/maestroResultParser.ts
@@ -34,10 +34,16 @@ const xmlParser = new XMLParser({
   isArray: name => ['testsuite', 'testcase', 'property'].includes(name),
 });
 
-// A `file=` attribute counts as present only when it is a non-empty string.
+// Official Maestro writes the flow path as a `file=` testcase attribute. maestro-runner writes
+// the same value as a `<property name="file" value="..."/>` child.
 function fileAttrOf(tc: any): string | undefined {
   const f = tc?.['@_file'];
-  return typeof f === 'string' && f.length > 0 ? f : undefined;
+  if (typeof f === 'string' && f.length > 0) {
+    return f;
+  }
+  const properties: { '@_name'?: unknown; '@_value'?: unknown }[] = tc?.properties?.property ?? [];
+  const fileProperty = properties.find(property => property['@_name'] === 'file')?.['@_value'];
+  return typeof fileProperty === 'string' && fileProperty.length > 0 ? fileProperty : undefined;
 }
 
 function parseJUnitContent(content: string): JUnitTestCaseResult[] {
@@ -68,7 +74,6 @@ function parseJUnitContent(content: string): JUnitTestCaseResult[] {
         const timeSeconds = timeStr ? parseFloat(timeStr) : 0;
         const duration = Number.isFinite(timeSeconds) ? Math.round(timeSeconds * 1000) : 0;
 
-        const status: 'passed' | 'failed' = tc['@_status'] === 'SUCCESS' ? 'passed' : 'failed';
         const failureText =
           tc.failure != null
             ? typeof tc.failure === 'string'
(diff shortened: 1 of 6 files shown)


packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -161,6 +193,10 @@ export function createMaestroTestsBuildFunction(ctx: CustomBuildContext): BuildF
       const outputFormat = (inputs.output_format.value as string | undefined)?.toLowerCase();
       const includeTags = inputs.include_tags.value as string | undefined;
       const excludeTags = inputs.exclude_tags.value as string | undefined;
+      const backend = resolveMaestroBackend({
+        input: inputs.backend.value,
+        env,
+      });
```
hSATAC: `resolveMaestroBackend` can throw before the required outputs are published. If the input or `EAS_MAESTRO_BACKEND` is invalid, the downstream `if: always()` upload step may fail to interpolate those outputs and obscure the original error.

This seems to be exactly what the “Outputs are published BEFORE any throw below” comment is meant to prevent. Nothing before the output assignments uses `backend`, so could we move this call below them?

Non-blocking—just moving the call should be enough.

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -449,9 +565,28 @@ async function uploadFailureScreenshotsAsync({
...
+        );
+        flowResults = results.some(result => result === null)
+          ? null
+          : results.flatMap(result => result?.flows ?? []);
+        break;
+      }
+    }
+    if (flowResults === null) {
+      logger.warn('Failed to classify failure screenshots; skipping upload.');
+      return;
```
hSATAC: Runner screenshot classification is currently all-or-nothing: if any attempt’s `report.json` cannot be parsed, all previously harvested screenshots are skipped.

For example, an earlier attempt could fail normally and collect screenshots, then the final retry crashes before writing `report.json`, causing all of them to be discarded. The Maestro path degrades and still uploads, so the comment above about fully failed runs still uploading screenshots does not hold for the runner path.

Could we use the reports that can be parsed, or fall back to uploading everything? The 30-screenshot cap should keep that bounded.

Non-blocking, but I think it’s worth addressing in this PR.

packages/build-tools/src/steps/functions/maestroScreenshots.ts
```diff
@@ -133,6 +133,118 @@ export async function harvestFailureScreenshotsAsync(args: {
...
+      `Skipping maestro-runner screenshot harvest: cannot read ${args.reportDirectory}.`
+    );
+    return [];
+  }
+
+  const shots: HarvestedScreenshot[] = [];
+  for (const flow of report.flows ?? []) {
+    if (flow.status !== 'failed' || !flow.name || !flow.dataFile) {
+      continue;
+    }
```
hSATAC: I don’t think the “Never throws” guarantee currently holds because `report.flows` is accessed outside the `try`.

Valid JSON with an unexpected shape, such as `null`, `{ "flows": 5 }`, or `{ "flows": [null] }`, can throw here. Since the caller does not catch it, a screenshot issue could fail the step and obscure the test result.

Could we move the loop into the `try` or validate the shape before iterating?

Non-blocking, small fix.

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -62,35 +66,58 @@ function isFilesystemError(err: any): boolean {
...
+        args.push(`--include-tags=${includeTags}`);
+      }
+      if (excludeTags) {
+        args.push(`--exclude-tags=${excludeTags}`);
+      }
+      args.push(...flowPaths);
+      return { executable: 'maestro', args };
+    }
+    case 'maestro-runner': {
+      const args = [`--platform=${platform}`, 'test', `--output=${output}`, '--flatten'];
```
hSATAC: Does `maestro-runner` support output formats other than JUnit?

The runner path does not pass a format flag, so `output_format: html` appears to produce no report or warning while the step still succeeds. If there is an equivalent flag, should we pass it through? Otherwise, I think we should log that it is unsupported like DADB, or reject it with a `UserError`.

hSATAC (approved):

hSATAC:
On the “maybe it’s not so bad?” point: I think the current structure is fine for now. `maestro-runner` is still being validated, so adding it behind backend guards while keeping the existing Maestro path unchanged seems like the lowest-cost approach. I wouldn’t ask to restructure this PR.

If we decide to support both backends long-term, I’d prefer to keep the overall flow shared and resolve the backend-specific runtime once before the loop, instead of adding backend checks throughout. The divergence is still small, but there are already some signs of it: the runner path drops all screenshots if any report is unusable while the Maestro path degrades, and the retry logic has an unreachable branch just for TypeScript narrowing.

Not something we need to do in this PR, but I’d prefer to refactor before the two paths diverge further.

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -82,6 +88,17 @@ function parseJUnitContent(content: string): JUnitTestCaseResult[] {
               : (tc.error?.['#text'] ?? null)
             : null;
         const errorMessage: string | null = failureText ?? errorText ?? null;
```
hSATAC, line 100: From [`junit.go`’s failure element construction](https://github.com/devicelab-dev/maestro-runner/blob/v1.1.23/pkg/report/junit.go#L152-L163), the `<failure>` message attribute contains the actual error from `entry.Error`, while [`resolveFailure`](https://github.com/devicelab-dev/maestro-runner/blob/v1.1.23/pkg/report/junit.go#L180-L198) puts the command label/type in the body.

This currently reads `#text`, so the runner’s `errorMessage` may be something like `tapOn` instead of the actual `element not found: ...` error, which will also be exposed through the API and website.

Could we prefer `@_message` and fall back to `#text`? Official Maestro only uses the body, so its behavior should remain unchanged.

Non-blocking.

hSATAC (commented):
One more non-blocking comment.

packages/build-tools/src/steps/functions/maestroScreenshots.ts
```diff
@@ -133,6 +136,156 @@ export async function harvestFailureScreenshotsAsync(args: {
...
+  }
+  return undefined;
+}
+
+function resolvePathInsideDirectory(directory: string, relativePath: string): string | null {
+  const candidate = path.resolve(directory, relativePath);
+  const relative = path.relative(directory, candidate);
+  return relative === '' || (!relative.startsWith(`..${path.sep}`) && !path.isAbsolute(relative))
+    ? candidate
```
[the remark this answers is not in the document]
sjchmiela: Good catch — fixed in 9064bbbb6. `resolvePathInsideDirectory` now also rejects an exact `..` (`relative !== '..'`), so a report-provided path that resolves to the parent directory is treated as outside. Added a regression test feeding a screenshot artifact of `..`.

packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -280,29 +334,35 @@ export function createMaestroTestsBuildFunction(ctx: CustomBuildContext): BuildF
       }
 
       for (let attempt = 0; attempt <= retries; attempt++) {
+        // maestro-runner writes its JUnit report and screenshot metadata to this directory.
+        const runnerOutputDirectory = path.join(
+          testsDirectory,
+          `${platform}-maestro-runner-attempt-${attempt}`
+        );
```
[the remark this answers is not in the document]
sjchmiela: Fixed in 9064bbbb6 — the runner output directory is now cleared best-effort (`fs.rm(dir, { recursive: true, force: true })`) before each attempt's spawn, so a crash before it writes fresh output can't resurrect stale results. Added a test asserting the clear happens before the spawn.

packages/build-tools/src/steps/functions/maestroResultParser.ts
```diff
@@ -68,20 +75,33 @@ function parseJUnitContent(content: string): JUnitTestCaseResult[] {
...
               ? tc.error
-              : (tc.error?.['#text'] ?? null)
+              : (tc.error?.['@_message'] ?? tc.error?.['#text'] ?? null)
             : null;
         const errorMessage: string | null = failureText ?? errorText ?? null;
+        // Official Maestro uses status="SUCCESS". maestro-runner uses standard JUnit semantics:
+        // a testcase passes when it has no failure or error child.
+        const statusAttribute = tc['@_status'];
+        const status: 'passed' | 'failed' =
+          typeof statusAttribute === 'string'
```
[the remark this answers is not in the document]
sjchmiela: Fixed in 9064bbbb6 — `parseJUnitContent` now excludes `<skipped/>` testcases instead of counting them as passed, matching the report.json path which already drops skipped flows. Added a test.

Note this mainly hardens the official-Maestro JUnit path; the maestro-runner path classifies off `report.json`, whose schema already filters `status: 'skipped'`. Excluding skipped is the correct, consistent behavior either way.

PR #4189, opened 2026-08-13
[build-tools] Add Appium remote simulator sessions

sjchmiela:
# Why

Appium is one of the industry standards when it comes to interacting with local and remote mobile devices. I think it may be nice to add support for it to EAS Simulator.

# How

Added a new EAS function that installs Appium v3, necessary plugin, starts it and exposes its endpoint through ngrok.

The event collection is a bit wonky since we're repeatedly calling Appium endpoint for events, but maybe it's ok. Alternatively we could extend session event collection mechanism to support `getEventsAsync`, but decided not to do that (see [diff](https://github.com/expo/eas-cli/compare/8c1e165899c495981ba6bfd506bcd1444cf0c5e8..25c685b5709f66f28341d7f09cc2999d7b0885be)).

# Test Plan

Tested manually locally. A remote test is going to be better!

### DIFF
--- a/packages/build-tools/src/steps/utils/appiumEvents.ts
+++ b/packages/build-tools/src/steps/utils/appiumEvents.ts
@@ -1,0 +1,216 @@
+import { type bunyan } from '@expo/logger';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { setTimeout as setTimeoutAsync } from 'node:timers/promises';
+import { z } from 'zod';
+
+import { type CustomBuildContext } from '../../customBuildContext';
+import { turtleFetch } from '../../utils/turtleFetch';
+import {
+  type DeviceRunSessionEventCollection,
+  type DeviceRunSessionEventSource,
+  startDeviceRunSessionEventCollectionAsync,
+} from './deviceRunSessionEvents';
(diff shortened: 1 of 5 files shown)


packages/build-tools/src/steps/utils/appiumEvents.ts
```diff
@@ -0,0 +1,216 @@
...
+  eventFile,
+  observedCommandKeys,
+  logger,
+}: {
+  appiumUrl: string;
+  eventFile: string;
+  observedCommandKeys: Set<string>;
+  logger: bunyan;
+}): Promise<void> {
+  const response = await turtleFetch(new URL('sessions', appiumUrl).toString(), 'GET', {
```
szdziedzic: Appium 3 removed the `GET /sessions` endpoint. Session discovery now requires `GET /appium/sessions`, with the server launched using `--allow-insecure=*:session_discovery`.

As written, every event-collection poll will fail, so no Appium events are uploaded. When `max_idle_time_minutes` is configured, an actively used session may also be treated as idle and stopped.

Could we update both the endpoint and the Appium launch arguments, and add test assertions for them?

szdziedzic (approved):

PR #4190, opened 2026-08-13
[eas-cli] Add Appium simulator session support

sjchmiela:
# Why

Now that we support Appium sessions, eas-cli needs to support it too.

# How

Added spread on GraphQL fetch, special instructions for printing and env file.

# Test Plan

CI should pass. Also ran a local e2e test.

### DIFF
--- a/packages/eas-cli/src/simulator/env.ts
+++ b/packages/eas-cli/src/simulator/env.ts
@@ -29,7 +29,7 @@ export async function writeSimulatorEnvAsync(
   const simulatorDotenvContent =
     SIMULATOR_DOTENV_FILE_HEADER +
     Object.entries(environmentVariables)
-      .map(([key, value]) => `${key}=${JSON.stringify(value)}`)
+      .map(([key, value]) => `${key}='${value}'`)
       .join('\n') +
     '\n';
 
--- a/packages/eas-cli/src/simulator/utils.ts
+++ b/packages/eas-cli/src/simulator/utils.ts
@@ -22,6 +22,7 @@ export function formatSimulatorUnavailableMessage(accountName: string): string {
 // so adding a new enum value in codegen fails the build until it is wired up here.
 export const DEVICE_RUN_SESSION_TYPE_FLAG_VALUES: Record<DeviceRunSessionType, string> = {
   [DeviceRunSessionType.AgentDevice]: 'agent-device',
+  [DeviceRunSessionType.Appium]: 'appium',
   [DeviceRunSessionType.Argent]: 'argent',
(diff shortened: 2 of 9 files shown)


packages/eas-cli/src/simulator/env.ts
```diff
@@ -29,7 +29,7 @@ export async function writeSimulatorEnvAsync(
   const simulatorDotenvContent =
     SIMULATOR_DOTENV_FILE_HEADER +
     Object.entries(environmentVariables)
-      .map(([key, value]) => `${key}=${JSON.stringify(value)}`)
+      .map(([key, value]) => `${key}='${value}'`)
```
szdziedzic, line 32: not sure if we should be concerned but:

>Raw single-quote wrapping does not round-trip every valid capability value. For example, capabilities containing `It's #1` are truncated at `#` by dotenv 16.3.1 because the inner apostrophe defeats quoted parsing. The new shell-export instructions similarly fail on apostrophes. Could we use format-specific escaping and add an apostrophe-plus-hash regression test?

packages/eas-cli/src/simulator/utils.ts
```diff
@@ -50,6 +51,11 @@ export function getRemoteSessionEnvironmentVariables(
         ARGENT_TOOLS_URL: remoteConfig.toolsUrl,
         ...(remoteConfig.toolsAuthToken ? { ARGENT_AUTH_TOKEN: remoteConfig.toolsAuthToken } : {}),
       };
+    case 'AppiumRunSessionRemoteConfig':
+      return {
+        APPIUM_URL: remoteConfig.appiumUrl,
+        APPIUM_CAPS: JSON.stringify(remoteConfig.capabilities),
```
szdziedzic, line 57: follow-up request:  can we protect it via token as well?

szdziedzic (approved):

packages/eas-cli/src/simulator/utils.ts
sjchmiela: Probably!

packages/eas-cli/src/simulator/env.ts
sjchmiela: yyyyea, that's true. we don't expect `#` or `'` in appium caps i think so let's ship it like this

added https://linear.app/expo/issue/ENG-26066/better-escape-for-appium-caps

packages/eas-cli/src/simulator/utils.ts
sjchmiela: Added https://linear.app/expo/issue/ENG-26067/protect-appium-endpoint-with-a-token

PR #4192, opened 2026-08-14
[eas-cli] Disallow 0-length uploads

wschurman:
# Why

Some fingerprint source uploads are ending up 0-sized. We're not sure why. The goal of this PR is to help isolate the issue by preventing 0-length uploads to begin with so we can trace the "why".

Ref ENG-25965.

# How

Disallow 0-length uploads by checking length ahead of upload, validating uploaded length using `content-length` header, and checking length post-upload.

# Test Plan

1. `EXPO_STAGING=1 neas fingerprint:generate --platform android`
2. Navigate to fingerprint permalink, see sources are uploaded.

### DIFF
--- a/packages/eas-cli/src/uploads.ts
+++ b/packages/eas-cli/src/uploads.ts
@@ -148,6 +148,11 @@ async function uploadWithSignedUrlWithProgressAsync(
 ): Promise<Response> {
   const fileStat = await fs.stat(file);
   const fileSize = fileStat.size;
+  if (fileSize === 0) {
+    throw new Error(
+      `Failed to upload ${file}: the file is empty. It may not have been written correctly. Retry the command, and report this issue if it persists.`
+    );
+  }
 
   const readStream = fs.createReadStream(file);
   const uploadPromise = fetch(signedUrl.url, {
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/uploads.ts
```diff
@@ -148,13 +148,19 @@ async function uploadWithSignedUrlWithProgressAsync(
...
+    );
+  }
 
   const readStream = fs.createReadStream(file);
   const uploadPromise = fetch(signedUrl.url, {
     method: 'PUT',
     body: readStream,
     headers: {
       ...signedUrl.headers,
+      'Content-Length': String(fileSize), // make GCS reject the request if the file size changes during the upload
```
gwdp, line 163: Went digging on this comment; I believe it only covers file shrink (all good for the case you are addressing imo). However, it made me wonder if we should be sending checksums so GCS can validate for real instead of us counting bytes after the fact. File integrity in the whole fingerprint process is not fully there (what we fingerprinted is what we upload) , unless I'm missing something 🤔

gwdp (approved):

packages/eas-cli/src/uploads.ts
wschurman: Good call. Will look into it in a follow-up PR.

PR #4193, opened 2026-08-14
[build-tools] Select compatible `maestro-runner` version for Xcode

sjchmiela:
# Why

`maestro-runner` 1.1.16+ is not compatible with Xcodes older than 26 because when selecing build target it adds an `arch` filter which older Xcodes don't recognize.

# How

Made it so we install 1.1.15 for older Xcodes and 1.1.16+ for new Xcodes.

# Test Plan

- Verified automatic version selection on the Xcode 16.2 staging image without a `maestro_version` input. The worker detected Xcode 16 and installed maestro-runner 1.1.15: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019fffe8-9bd8-74f9-87b6-d7963ee6d941
- Reproduced the unsupported `arch` failure with maestro-runner 1.1.23 on the same image: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019ffd14-cb98-7351-9113-872d01c557ae
- Successful run: https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/019fff26-4055-7ae5-8f64-8c3b9190c380

### DIFF
--- a/packages/build-tools/src/steps/functions/installMaestro.ts
+++ b/packages/build-tools/src/steps/functions/installMaestro.ts
@@ -1,3 +1,4 @@
+import { SystemError, UserError } from '@expo/eas-build-job';
 import { bunyan } from '@expo/logger';
 import { asyncResult } from '@expo/results';
 import {
@@ -14,6 +15,7 @@ import assert from 'assert';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
+import semver from 'semver';
 
 import { MaestroBackend, resolveMaestroBackend } from './maestroBackend';
 import { Datadog } from '../../datadog';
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -154,6 +200,15 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
   });
 }
 
+async function getXcodeVersion({ env }: { env: BuildStepEnv }): Promise<string> {
+  const { stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env });
```
hSATAC: Only the parse failure is wrapped in a `SystemError` here. If `xcodebuild -version` itself fails, the raw spawn error will bubble up and be classified as an unknown error instead of an infra `SERVER_ERROR`, which could affect error attribution and billing.

Could we wrap the `spawn` call and rethrow it with the original cause?

```ts
let stdout: string;
try {
  ({ stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env }));
} catch (err) {
  throw new SystemError('Failed to run xcodebuild -version', { cause: err });
}
```

Non-blocking, but worth fixing in this PR since this is a new failure path for unpinned Darwin installs.

packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -120,19 +122,63 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
...
+            (maestroRunnerVersionToInstall &&
+              maestroRunnerVersionToInstall !== currentMaestroVersion)
+          ) {
+            await installMaestroRunner({
+              version: maestroRunnerVersionToInstall,
+              global,
+              logger,
+              env,
+            });
+          }
```
hSATAC, line 180: After deciding to install `1.1.15`, we don’t verify that it is the version actually resolved. `installMaestroRunner` appends its bin directory to the end of `PATH`, so a newer preinstalled runner earlier in `PATH` would still win. The install step would stay green and only fail later in `maestro_tests` with the architecture issue this PR is trying to prevent.

Could we assert that `getMaestroVersion` matches the target after installation when the target is known and not `latest`?

The downgrade test would not catch this because its second `--version` call always returns `1.1.15` without going through real `PATH` resolution.

Non-blocking and mostly future-proofing—the current staging image does not have a preinstalled runner.

hSATAC (approved):

packages/build-tools/src/steps/functions/installMaestro.ts
sjchmiela: Moved new path element to the beginning of `PATH`

PR #4195, opened 2026-08-14
[build-tools] Restore maestro-runner WDA cache

sjchmiela:
## Why

`maestro-runner` needs WDA driver to run tests. By default it compiles it on first run which takes a minute.

## How

Compiled WDA on all images, uploaded to `turtle-v2` GCS and added optimistic download to `eas/install_maestro`. See https://github.com/expo/turtle-v2/pull/2603.

## Test plan

https://staging.expo.dev/accounts/expo-services/projects/workflows-testing/workflows/01a0009e-b171-7497-9335-af1fc470a8d5

### DIFF
--- a/packages/build-tools/src/ios/xcode.ts
+++ b/packages/build-tools/src/ios/xcode.ts
@@ -1,0 +1,11 @@
+import { SystemError } from '@expo/eas-build-job';
+import spawn from '@expo/turtle-spawn';
+
+export async function getXcodeVersionAsync({ env }: { env: NodeJS.ProcessEnv }): Promise<string> {
+  const { stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env });
+  const version = /^Xcode\s+(\d+(?:\.\d+)*)$/m.exec(stdout)?.[1];
+  if (!version) {
+    throw new SystemError(`Failed to determine Xcode version from: ${stdout.trim()}`);
+  }
+  return version;
+}
--- a/packages/build-tools/src/ios/xcode.ts
+++ b/packages/build-tools/src/ios/xcode.ts
@@ -2,7 +2,12 @@ import { SystemError } from '@expo/eas-build-job';
 import spawn from '@expo/turtle-spawn';
 
 export async function getXcodeVersionAsync({ env }: { env: NodeJS.ProcessEnv }): Promise<string> {
-  const { stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env });
+  let stdout: string;
+  try {
+    ({ stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env }));
+  } catch (error) {
+    throw new SystemError('Failed to get Xcode version', { cause: error });
+  }
   const version = /^Xcode\s+(\d+(?:\.\d+)*)$/m.exec(stdout)?.[1];
   if (!version) {
     throw new SystemError(`Failed to determine Xcode version from: ${stdout.trim()}`);
--- a/packages/build-tools/src/steps/functions/installMaestro.ts
+++ b/packages/build-tools/src/steps/functions/installMaestro.ts
@@ -24,6 +24,8 @@ import { getXcodeVersionAsync } from '../../ios/xcode';
 import { IosSimulatorUtils } from '../../utils/IosSimulatorUtils';
 import { getProxiedDownloadUrl } from '../../utils/download';
 
+const MAESTRO_RUNNER_WDA_CACHE_DOWNLOAD_TIMEOUT_MS = 20_000;
+
 export function createInstallMaestroBuildFunction(): BuildFunction {
   return new BuildFunction({
     namespace: 'eas',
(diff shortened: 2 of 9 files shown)


packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -201,19 +214,120 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
...
+        'DerivedData',
+        'Build',
+        'Products'
+      );
+      for (const runtimeVersion of iosRuntimeVersions) {
+        const productsDirectory = path.join(
+          maestroRunnerHome,
+          'cache',
+          'wda-builds',
+          `sim-ios${runtimeVersion}-iphone`,
```
hSATAC, line 315: I think this may miss the cache on patched runtimes.

`simctl list runtimes --json` gives us the full `version` (e.g. `26.3.1`), but maestro-runner derives the cache key from the runtime identifier (`iOS-26-3` → `26.3`). So we'd write `sim-ios26.3.1-iphone` here while runner looks for `sim-ios26.3-iphone`.

I haven't checked what versions we currently have on the images, so maybe I'm missing something there. Should we derive this from `identifier` too, to match runner?

packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -201,19 +214,120 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
...
+    }
+
+    const tempDirectory = await fs.promises.mkdtemp(
+      path.join(os.tmpdir(), 'install_maestro_runner_wda_cache')
+    );
+    try {
+      const archivePath = path.join(tempDirectory, 'wda-cache.tar.gz');
+      const directArchiveUrl =
+        `https://storage.googleapis.com/turtle-v2/maestro-runner-wda-cache/` +
+        `xcode-${encodeURIComponent(xcodeVersion)}-wda-${encodeURIComponent(wdaVersion)}.tar.gz`;
```
hSATAC, line 269: I'm not sure I understand how this lines up with the turtle-v2 upload side.

This looks for `xcode-<xcodeVersion>-wda-<wdaVersion>.tar.gz`, but turtle-v2#2603 seems to upload image-keyed artifacts like `macos-sequoia-15.6-xcode-26.0-wda-11.1.3.tar.gz`.

The existing `xcode-*` objects are there, so I assume there's another step I'm missing. Are we planning to change the turtle-v2 naming to match this?

If so, I think the two Xcode 16.4 images would also collapse to the same object name, so the artifact count/check there may need an update too.

packages/build-tools/src/steps/functions/installMaestro.ts
```diff
@@ -201,19 +214,120 @@ export function createInstallMaestroBuildFunction(): BuildFunction {
...
+      const proxiedArchiveUrl = getProxiedDownloadUrl({
+        directUrl: directArchiveUrl,
+        proxyBaseUrl: env.EAS_BUILD_COCOAPODS_CACHE_URL,
+      });
+
+      logger.info(`Downloading the prebuilt WebDriverAgent cache for Xcode ${xcodeVersion}`);
+      if (proxiedArchiveUrl) {
+        try {
+          await downloadFile(proxiedArchiveUrl, archivePath, { retry: 3 });
```
hSATAC: Should we add a timeout to these downloads?

AFAICT `@expo/downloader` passes this through to got, and got doesn't have a default timeout. If the connection stalls after being established, this best-effort cache download could hang until the build timeout instead of falling back.

`xcactivitylog.ts` uses `timeout: 20_000` for the same proxied-download pattern.

packages/build-tools/src/ios/xcode.ts
```diff
@@ -0,0 +1,11 @@
+import { SystemError } from '@expo/eas-build-job';
+import spawn from '@expo/turtle-spawn';
+
+export async function getXcodeVersionAsync({ env }: { env: NodeJS.ProcessEnv }): Promise<string> {
+  const { stdout } = await spawn('xcodebuild', ['-version'], { stdio: 'pipe', env });
```
hSATAC: Was dropping the `SystemError` wrapping around the `xcodebuild -version` spawn intentional?

Previously a spawn failure was wrapped as `SystemError('Failed to get Xcode version', { cause })`. Now only the parse failure is a `SystemError`, so things like a bad `DEVELOPER_DIR` would propagate as the raw spawn error at the maestro-runner version-selection call site.

The WDA cache path catches this anyway, so I think this only affects the other caller. Maybe intentional, just wanted to check.

hSATAC (approved):
Looks good overall, approving. Left a few questions inline — I may be missing some context around the WDA/turtle-v2 side.

packages/build-tools/src/steps/functions/installMaestro.ts
sjchmiela: Good catch. I changed the runtime cache version to derive from the CoreSimulator runtime identifier, exactly as maestro-runner does. The regression tests now cover an `iOS-18-3` identifier that reports version `18.3.1` and verify that we create `sim-ios18.3-iphone`.

packages/build-tools/src/steps/functions/installMaestro.ts
sjchmiela: Yes, the deployed objects were Xcode-keyed but the turtle-v2 PR did not reproduce that state. I updated expo/turtle-v2#2603 in d0c38735 to emit `xcode-<version>-wda-<version>.tar.gz`, removed the duplicate Xcode 16.4 job, changed the expected artifact count to 12, and updated the documentation and deployment validation.

packages/build-tools/src/steps/functions/installMaestro.ts
sjchmiela: Agreed. Both proxied and direct downloads now use a 20-second timeout, including the direct fallback path. The installMaestro test verifies the timeout option.

packages/build-tools/src/ios/xcode.ts
sjchmiela: This was not intentional. I restored `SystemError('Failed to get Xcode version', { cause })` wrapping around the spawn call and added a regression test that verifies the error type, message, and cause.

PR #4200, opened 2026-08-14
[eas-cli] Protect simulator environment cleanup

sjchmiela:
# Why

Codex insists we should clean the environment up when stopping the simulator. I think we should also guard the cleanup with ID check.

# How

Added reset to the stop command and an ID guard.

# Test Plan

CI should pass.

### DIFF
--- a/packages/eas-cli/src/simulator/env.ts
+++ b/packages/eas-cli/src/simulator/env.ts
@@ -1,4 +1,5 @@
 import { loadEnvFiles, loadProjectEnv } from '@expo/env';
+import { parse as parseDotenv } from 'dotenv';
 import * as fs from 'fs-extra';
 import path from 'path';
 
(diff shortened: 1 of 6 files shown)


packages/eas-cli/src/simulator/env.ts
```diff
@@ -33,10 +34,17 @@ export async function writeSimulatorEnvAsync(
...
+export async function resetSimulatorEnvAsync(
+  projectDir: string,
+  expectedDeviceRunSessionId: string
+): Promise<void> {
   const simulatorDotenvFilePath = getSimulatorEnvFilePath(projectDir);
 
   try {
+    const currentEnv = parseDotenv(await fs.readFile(simulatorDotenvFilePath, 'utf8'));
+    if (currentEnv[EAS_SIMULATOR_SESSION_ID] !== expectedDeviceRunSessionId) {
+      return;
```
szdziedzic, line 56: should we show warn here?

szdziedzic (approved):

packages/eas-cli/src/simulator/env.ts
sjchmiela: Good question — I went with a debug-level log rather than a warning. This early-return is the guard doing its job: when `.env.eas-simulator` belongs to a newer overlapping session, the older session must not reset it. That's expected, benign behavior that'll happen routinely with overlapping sessions, so a `warn` would read as "something went wrong" and add noise (it'd also be inconsistent with the sibling ENOENT branch, which returns silently). Added a `Log.debug` in 4a07b735a so it's still traceable via `EXPO_DEBUG` without surfacing during normal use.

sjchmiela: Zrzut ekranu 2026-08-19 o 13 06 56

PR #4207, opened 2026-08-14
[steps] Composite functions: skip the outputs step when the call never ran

sswrk:
# Why

In this comment https://github.com/expo/eas-cli/pull/4062#discussion_r3664228189 it was brought up that when no steps of a composite function run (because an earlier step has failed) and the caller sets `if: ${{ always() }}`, the synthetic "outputs" step runs anyway.

The expected behavior is that the synthetic "outputs" step is skipped too.

# How

In CompositeBuildStep, `shouldExecuteStep` is overridden so the outputs step runs if `runByDefault` is true or any authored child has run.

# Test Plan

Added unit tests.

### DIFF
--- a/packages/steps/src/__tests__/StepsConfigParser-composite-functions-outputs-test.ts
+++ b/packages/steps/src/__tests__/StepsConfigParser-composite-functions-outputs-test.ts
@@ -319,6 +319,140 @@ describe('StepsConfigParser local composite functions', () => {
       expect(outputsStep.getOutputValueByName('version_expr')).toBe('');
     });
 
+    it('skips the outputs step when an earlier failure suppressed every inner step', async () => {
+      const workflow = await parseCompositeFunctions({
+        catalog: {
+          [SETUP]: {
+            outputs: { version: { value: '${{ steps.read.outputs.version }}' } },
+            runs: {
+              steps: [{ id: 'read', uses: 'test/set-version' }],
+            },
+          },
+        },
+        steps: [
+          { id: 'boom', uses: 'test/fail' },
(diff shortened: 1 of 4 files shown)


packages/steps/src/__tests__/StepsConfigParser-composite-functions-outputs-test.ts
```diff
@@ -319,6 +319,140 @@ describe('StepsConfigParser local composite functions', () => {
...
+      const error = await getErrorAsync<Error>(() => workflow.executeAsync());
+      expect(error.message).toBe('inner failed');
+
+      expect(workflow.buildSteps.find(s => s.id === 'top__mid__read')?.status).toBe(
+        BuildStepStatus.SKIPPED
+      );
+      expect(workflow.buildSteps.find(s => s.id === 'top')?.status).toBe(BuildStepStatus.SKIPPED);
+    });
+
+    it('runs the outer outputs step when a grandchild of a nested call executed', async () => {
```
hSATAC, line 424: Non-blocking — I think we’re missing a test that actually covers the recursion in `anyAuthoredChildExecuted`. Replacing it with a flat `this.children.some(child => child.wasExecuted)` still passes the whole steps test suite.

This test doesn’t catch it because `mid` declares outputs, so its synthetic outputs step enters `buildSteps` and executes. The no-outputs test above doesn’t catch it either because all grandchildren are skipped.

If you want to make the test coverage more complete, you could add a case where the inner call has no outputs, but one of its grandchildren still runs via `always()`. The inner call itself won’t enter `buildSteps`, so that would verify that the outer outputs step relies on the recursive check.

hSATAC (approved):

PR #4210, opened 2026-08-14
[eas-cli] Handle missing fingerprint source in fingerprint:compare command

wschurman:
# Why

We're seeing two issues:
1. Fingerprint source files in GCS expire after 1 year, and therefore may not exist.
2. Some users fingerprint sources have file size 0.

# How

Either way, we need to make fingerprints usable even if their sources are invalid for a reason above. This PR makes the CLI compare command usable by showing a more user-friendly error message when the source fetch fails.

# Test Plan

`EXPO_STAGING=1 neas fingerprint:compare e6e28b42935c6c53528e647f1f2afe16314b2ed0 eaf33e1d6b16e0ccab0ee569853fdd83f358400e`

### DIFF
--- a/packages/eas-cli/src/commands/fingerprint/compare.ts
+++ b/packages/eas-cli/src/commands/fingerprint/compare.ts
@@ -54,6 +54,18 @@ type FingerprintOrigin = {
   update?: UpdateFragment;
 };
 
+type FingerprintInfo = {
+  fingerprint: Fingerprint;
+  platforms?: AppPlatform[];
+  origin: FingerprintOrigin;
+  /**
+   * Whether the uploaded fingerprint source file could not be downloaded or parsed
+   * (e.g. it expired in GCS or is zero-length). The hash is still valid in that case,
+   * but `fingerprint.sources` is empty and a detailed diff cannot be computed.
+   */
+  isFingerprintSourceUnavailable: boolean;
+};
+
 export default class FingerprintCompare extends EasCommand {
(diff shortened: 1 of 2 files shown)


packages/eas-cli/src/commands/fingerprint/compare.ts
```diff
@@ -190,55 +210,62 @@ export default class FingerprintCompare extends EasCommand {
...
+    if (unavailableSourceHashes.length > 0) {
       Log.newLine();
-      Log.log('📁 Paths with native dependencies:');
-    }
-    const fields = [];
-    for (const diff of filePathDiffs) {
-      const field = getDiffFilePathFields(diff);
-      if (!field) {
-        throw new Error(`Unsupported diff: ${JSON.stringify(diff)}`);
+      Log.warn(
```
gwdp, line 215: Worth noting (unsure if a blocker), but this is a change of behaviour; `unavailableSourceHashes.length > 0` case, unfortunately, would throw on `.json()` before, so if anyone expecting exit != 0 this change things.

`--json` mode does look to exit with an error (on line 191)

gwdp (approved):

packages/eas-cli/src/commands/fingerprint/compare.ts
wschurman: Yep! That's intended. `--json` should still fail. Interactive should hopefully give users all the info, and the previous crash and nonzero exit was a bug.

PR #4216, opened 2026-08-17
[steps] Report a broken composite `if:` on the call, not an inner step

sswrk:
# Why

When a composite function a call errors on a callsite `if:` condition, the runner logs the error as an error coming from inside of the composite function. This is an unexpected behavior that points users to a wrong `if:`. This PR fixes it.

# How

- Added a new `BuildStepConditionEvaluationError` error type.
- In `BuildStepCompositeFunctionScope`, when `evaluateCallIfCondition` throws, set `cachedIsActive` as `false` first, so later steps in the same call skip without re-evaluating. Then throw a `BuildStepConditionEvaluationError` with the caller's details.
- In `BuildWorkflow.logConditionEvaluationError`, handle the `BuildStepConditionEvaluationError` error by logging its details.

# Test Plan

Added and updated unit tests.

### DIFF
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -288,11 +288,8 @@ function logConditionEvaluationError(
   ifCondition: string | undefined
 ): void {
   if (err instanceof BuildStepConditionEvaluationError) {
-    logger.error({ err: err.cause ?? err });
-    logger.error(
-      `Runner failed to evaluate if it should execute ${err.subject}, using its if condition "${err.ifCondition}". This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
-    );
-    return;
+    ({ subject, ifCondition } = err);
+    err = err.cause ?? err;
   }
   logger.error({ err });
   logger.error(
--- a/packages/steps/src/BuildWorkflow.ts
+++ b/packages/steps/src/BuildWorkflow.ts
@@ -7,6 +7,7 @@ import { BuildStep } from './BuildStep';
 import { BuildStepGlobalContext } from './BuildStepContext';
 import { CompositeBuildStep } from './CompositeBuildStep';
 import { StepMetricResult } from './StepMetrics';
+import { BuildStepConditionEvaluationError } from './errors';
 import { AnchorHooks, HookEntry } from './hooks';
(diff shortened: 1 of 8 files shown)


packages/steps/src/BuildWorkflow.ts
```diff
@@ -279,15 +280,20 @@ export async function executeHookStepsAsync(
...
-// `if:`, but the throw may come from a composite call-site `if:` evaluated via
-// the step's scope. In that case the quoted condition is not the one that failed.
+// Shared wording for unevaluable `if:` gates.
 function logConditionEvaluationError(
   logger: bunyan,
   err: unknown,
   subject: string,
   ifCondition: string | undefined
 ): void {
+  if (err instanceof BuildStepConditionEvaluationError) {
```
hSATAC, line 290: Non-blocking: I think this can be simplified a bit. This branch duplicates the user-facing message from the generic path below, which makes it easy for the two messages to drift later. Since `BuildStepConditionEvaluationError` always carries its own `subject` and a non-empty `ifCondition`, we could override those values and fall through to the existing logger:

```ts
function logConditionEvaluationError(
  logger: bunyan,
  err: unknown,
  subject: string,
  ifCondition: string | undefined
): void {
  if (err instanceof BuildStepConditionEvaluationError) {
    ({ subject, ifCondition } = err);
    err = err.cause ?? err;
  }
  logger.error({ err });
  logger.error(
    `Runner failed to evaluate if it should execute ${subject}${
      ifCondition ? `, using its if condition "${ifCondition}"` : ''
    }. This can be caused by trying to access non-existing object property. If you think this is a bug report it here: https://github.com/expo/eas-cli/issues.`
  );
}
```

Similarly, the error message is fully derived from `subject` and `ifCondition`, so I think it would be cleaner for the constructor to build it directly:

```ts
export class BuildStepConditionEvaluationError extends UserError {
  constructor(
    public readonly subject: string,
    public readonly ifCondition: string,
    extra?: { cause?: Error }
  ) {
    super(`Failed to evaluate the if condition "${ifCondition}" of ${subject}.`, extra);
  }
}
```

That also keeps the fields and message from getting out of sync, and makes the throw site smaller.

hSATAC (approved):

packages/steps/src/BuildWorkflow.ts
sswrk: I like this idea, suggestion applied.

PR #4221, opened 2026-08-18
[build-tools] Cache Gradle file access time journal with the build cache

sjchmiela:
# Why

It seems the cache is not getting pruned properly. It looks like Gradle uses journal to keep track of accessed files. Since we're not caching it alongside build cache, Gradle defaults all files to "they have been used now".

# How

When caching Gradle build cache, cache journal too.

# Test Plan

Verified locally.

With `build-cache-1` only kept, everything was retained after cleanup because journal inception was "now".

With both `build-cache-1` and `journal-1` kept AND inception timestamp backdated, cache was pruned.

Didn't test pruning across multiple days.


1337mus (approved):

AbbanMustafa (approved):
thank you for this 📓

PR #4222, opened 2026-08-18
[build-tools] Install builds in simulator sessions

szdziedzic:
# Why

Follow up to https://github.com/expo/universe/pull/30191

Enable EAS Sim users to preinstall and start builds when starting sim session

# How

Add/use `eas/download_build`, `eas/install_build`, and `eas/launch_application` steps to accomplish it.

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/steps/functions/downloadBuild.ts
+++ b/packages/build-tools/src/steps/functions/downloadBuild.ts
@@ -26,6 +26,17 @@ import { pluralize } from '../../utils/strings';
 
 const streamPipeline = promisify(stream.pipeline);
 
+type DownloadBuildSource =
+  | { buildId: string; applicationArchiveUrl?: never }
+  | { buildId?: never; applicationArchiveUrl: string };
+
+type DownloadBuildParams = DownloadBuildSource & {
+  logger: bunyan;
+  graphqlClient: Client;
+  robotAccessToken: string | null;
+  extensions: string[];
+};
+
 const BUILD_BY_ID_QUERY = graphql(`
   query DownloadBuildByIdQuery($buildId: ID!) {
     builds {
@@ -82,17 +93,23 @@ export function createDownloadBuildFunction(ctx: CustomBuildContext): BuildFunct
         ? parseHttpApplicationArchiveUrl(inputs.application_archive_url.value)
         : undefined;
 
-      if (!buildId && !applicationArchiveUrl) {
-        throw new UserError(
-          'EAS_DOWNLOAD_BUILD_INVALID_SOURCE',
-          'Pass build_id or application_archive_url.'
-        );
-      }
-      if (buildId && applicationArchiveUrl) {
-        throw new UserError(
-          'EAS_DOWNLOAD_BUILD_INVALID_SOURCE',
-          'Pass only one of build_id or application_archive_url.'
-        );
+      let source: DownloadBuildSource;
+      if (buildId) {
+        if (applicationArchiveUrl) {
+          throw new UserError(
+            'EAS_DOWNLOAD_BUILD_INVALID_SOURCE',
+            'Pass only one of build_id or application_archive_url.'
+          );
+        }
+        source = { buildId };
+      } else {
+        if (!applicationArchiveUrl) {
+          throw new UserError(
+            'EAS_DOWNLOAD_BUILD_INVALID_SOURCE',
+            'Pass build_id or application_archive_url.'
+          );
+        }
+        source = { applicationArchiveUrl };
       }
 
       logger.info(
@@ -101,8 +118,7 @@ export function createDownloadBuildFunction(ctx: CustomBuildContext): BuildFunct
 
       const { artifactPath } = await downloadBuildAsync({
         logger,
-        buildId,
-        applicationArchiveUrl,
+        ...source,
         graphqlClient: ctx.graphqlClient,
         robotAccessToken: stepsCtx.global.staticContext.job.secrets?.robotAccessToken ?? null,
         extensions,
@@ -149,14 +165,7 @@ export async function downloadBuildAsync({
   graphqlClient,
   robotAccessToken,
   extensions,
-}: {
-  logger: bunyan;
-  buildId?: string;
-  applicationArchiveUrl?: string;
-  graphqlClient: Client;
-  robotAccessToken: string | null;
-  extensions: string[];
-}): Promise<{ artifactPath: string }> {
+}: DownloadBuildParams): Promise<{ artifactPath: string }> {
   const validatedApplicationArchiveUrl = applicationArchiveUrl
     ? parseHttpApplicationArchiveUrl(applicationArchiveUrl)
     : undefined;
--- a/packages/build-tools/src/steps/functions/downloadBuild.ts
+++ b/packages/build-tools/src/steps/functions/downloadBuild.ts
@@ -53,7 +53,7 @@ export function createDownloadBuildFunction(ctx: CustomBuildContext): BuildFunct
         allowedValueTypeName: BuildStepInputValueTypeName.STRING,
       }),
       BuildStepInput.createProvider({
-        id: 'artifact_url',
+        id: 'application_archive_url',
         required: false,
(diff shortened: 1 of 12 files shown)


packages/build-tools/src/steps/functions/downloadBuild.ts
```diff
@@ -26,6 +26,17 @@ import { pluralize } from '../../utils/strings';
...
+type DownloadBuildSource =
+  | { buildId: string; applicationArchiveUrl?: never }
+  | { buildId?: never; applicationArchiveUrl: string };
+
+type DownloadBuildParams = DownloadBuildSource & {
+  logger: bunyan;
+  graphqlClient: Client;
+  robotAccessToken: string | null;
+  extensions: string[];
+};
```
sjchmiela: inline, no reason to extract a single function's params type far from the function itself

packages/build-tools/src/steps/functions/downloadBuild.ts
```diff
@@ -194,3 +258,23 @@ export async function downloadBuildAsync({
...
+      'EAS_DOWNLOAD_BUILD_INVALID_APPLICATION_ARCHIVE_URL',
+      'application_archive_url must be a valid HTTP or HTTPS URL.'
+    );
+  }
+  if (parsedUrl.protocol !== 'http:' && parsedUrl.protocol !== 'https:') {
+    throw new UserError(
+      'EAS_DOWNLOAD_BUILD_INVALID_APPLICATION_ARCHIVE_URL',
+      'application_archive_url must be a valid HTTP or HTTPS URL.'
+    );
+  }
```
sjchmiela: you could move the check as assertion to inside `try`

packages/build-tools/src/steps/functions/installBuild.ts
```diff
@@ -0,0 +1,124 @@
...
+        'iOS Simulator sessions require a .app build artifact.'
+      );
+    }
+
+    const infoPlistPath = path.join(artifactPath, 'Info.plist');
+    const { stdout } = await spawn(
+      'plutil',
+      ['-extract', 'CFBundleIdentifier', 'raw', '-o', '-', infoPlistPath],
+      { stdio: 'pipe', env }
+    );
```
sjchmiela, line 87: i wonder if we should unify this with https://github.com/expo/eas-cli/blob/f3fd05d2dc25d6da75c62d6cd6ab537cb817e22e/packages/build-tools/src/steps/functions/readIpaInfo.ts#L65 somehow (or at least add `readIosAppInfo` or sth

packages/build-tools/src/steps/functions/installBuild.ts
```diff
@@ -0,0 +1,124 @@
...
+    const applicationIdentifier = stdout.trim();
+    if (!applicationIdentifier) {
+      throw new UserError(
+        'EAS_INSTALL_BUILD_MISSING_IDENTIFIER',
+        `Could not read CFBundleIdentifier from ${infoPlistPath}.`
+      );
+    }
+
+    logger.info(`Installing ${artifactPath} on the iOS Simulator.`);
+    await spawn('xcrun', ['simctl', 'install', 'booted', artifactPath], { env, logger });
```
sjchmiela, line 97: technically it would be nice to add support for:
- device identifier to install the app on
- multiple booted devices

packages/build-tools/src/steps/functions/installBuild.ts
```diff
@@ -0,0 +1,124 @@
...
+    return { applicationIdentifier };
+  }
+
+  if (path.extname(artifactPath) !== '.apk' || !artifactStat.isFile()) {
+    throw new UserError(
+      'EAS_INSTALL_BUILD_INVALID_ARTIFACT',
+      'Android Emulator sessions require an .apk build artifact.'
+    );
```
sjchmiela, line 105: would be nice to add support for converting aab to apk on the fly

packages/build-tools/src/steps/functions/installBuild.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  const { stdout } = await spawn('aapt', ['dump', 'badging', artifactPath], {
```
sjchmiela: in submit jobs and submission worker we use aapt2 https://github.com/expo/universe/blob/2b774ea14d4fa6f773e2a125555eaa232641fc92/server/www/src/data/entities/workflow/job/WorkflowSubmitJob.ts#L729-L731

sjchmiela: ah, i see, i presume you need launchable-activity

packages/build-tools/src/steps/functions/installBuild.ts
```diff
@@ -0,0 +1,124 @@
...
+  const activityName = stdout.match(/launchable-activity: name='([^']+)'/)?.[1];
+  if (!applicationIdentifier) {
+    throw new UserError(
+      'EAS_INSTALL_BUILD_MISSING_IDENTIFIER',
+      `Could not read an Android application identifier from ${artifactPath}.`
+    );
+  }
+
+  logger.info(`Installing ${artifactPath} on the Android Emulator.`);
+  await spawn('adb', ['install', '-r', artifactPath], { env, logger });
```
sjchmiela, line 122: what does `-r` do?

packages/build-tools/src/steps/functions/launchApplication.ts
```diff
@@ -0,0 +1,81 @@
...
+        id: 'activity_name',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
+    ],
+    fn: async ({ global, logger }, { inputs, env }) => {
+      const applicationIdentifier = z.string().min(1).parse(inputs.application_identifier.value);
+      const activityName = inputs.activity_name.value
+        ? z.string().min(1).parse(inputs.activity_name.value)
+        : undefined;
```
sjchmiela: i think zod errors don't get converted to usererrors nicely, would need to check

packages/build-tools/src/steps/functions/downloadBuild.ts
```diff
@@ -118,24 +161,45 @@ async function fetchApplicationArchiveUrlAsync({
...
+  const downloadUrl =
+    validatedApplicationArchiveUrl ??
+    (await fetchApplicationArchiveUrlAsync({
+      buildId: z.string().uuid().parse(buildId),
+      graphqlClient,
+    }));
 
   const response = await retryOnDNSFailure(fetch)(downloadUrl, {
-    headers: robotAccessToken ? { Authorization: `Bearer ${robotAccessToken}` } : undefined,
+    // A direct application archive URL is user-controlled, so never send the scoped EAS token to it.
```
sjchmiela: good call, i would even go as far as to make a bigger switch -- one for validatedApplicationArchiveUrl, one for buildid

sjchmiela (approved):

packages/build-tools/src/steps/functions/installBuild.ts
szdziedzic: -r reinstalls/replaces an already installed application while preserving its existing app data - just in case

packages/build-tools/src/steps/functions/installBuild.ts
szdziedzic: Agree this would be useful. I think AAB conversion should be an explicit composable function rather than hidden inside install_build because it also requires bundletool and signing decisions. I’ll treat it as follow-up scope.

packages/build-tools/src/steps/functions/installBuild.ts
szdziedzic: Agree that a shared iOS application metadata abstraction would be cleaner. I’d prefer to handle that as a follow-up so this PR stays focused on the install/launch primitives.

packages/build-tools/src/steps/functions/installBuild.ts
szdziedzic: Agree this would be useful. I’d like to add it separately once we define whether the function accepts one device identifier or supports installing on multiple devices. The current remote-session use case has one booted target.

PR #4227, opened 2026-08-19
[build-tools] Humanize Appium command summaries

sjchmiela:
# Why

We currently present Appium session events as HTTP methods which is not too human-readable (`getWindowRect`).

# How

Added a GHA workflow which runs weekly and submits a pull request adding to an `AppiumCommand` type so TSC fails and we need to add descriptions to commands.

# Test Plan

CI should pass.


.github/workflows/update-appium-commands.yml
```diff
@@ -0,0 +1,54 @@
...
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    name: Regenerate Appium command list
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          token: ${{ secrets.EXPO_BOT_PAT }}
```
szdziedzic: actions/checkout@v4 persists the supplied token by default, then the generator installs and executes the latest @appium/base-driver@^10. A compromised package or dependency could read EXPO_BOT_PAT. Use persist-credentials: false and configure authentication only in the final push step.

.github/workflows/update-appium-commands.yml
```diff
@@ -0,0 +1,54 @@
...
+          branch=bot/update-appium-commands
+          title='[build-tools] Update Appium command list'
+          body=$'Automated update of the Appium command list from `@appium/base-driver`.\n\nIf this PR adds new commands, the `APPIUM_COMMAND_SUMMARIES` coverage test will fail until a curated summary is added for each of them in `packages/build-tools/src/steps/utils/appiumCommandSummary.ts`.'
+
+          git config user.name 'Expo Bot'
+          git config user.email bot@expo.dev
+          git checkout -B "$branch"
+          git add packages/build-tools/src/steps/utils/appiumCommands.generated.ts
+          git commit -m "$title"
+          git push --force-with-lease origin "$branch"
```
szdziedzic, line 83: The default shallow checkout does not fetch origin/bot/update-appium-commands, so --force-with-lease rejects the push with stale info whenever that branch already exists.

packages/build-tools/src/steps/utils/appiumCommandSummary.ts
```diff
@@ -0,0 +1,198 @@
+import { type UpstreamAppiumCommand } from './appiumCommands.generated';
+
+// Driver-level commands that base-driver does not declare (they only show up
+// through drivers such as XCUITest / UiAutomator2).
+type DriverLevelAppiumCommand = 'background' | 'getClipboard' | 'setClipboard';
```
szdziedzic: Include XCUITest and UiAutomator2 commands in coverage

Appium event timings contain more than the commands declared by `@appium/base-driver`. EAS also installs the latest XCUITest or UiAutomator2 driver, and those drivers contribute additional command names such as `mobileScroll`, `openNotifications`, `lock`, and `keys`.

For example, when a client calls `mobile: scroll`, Appium records the resolved internal command name `mobileScroll`. Since this type lists only three driver-specific commands, `humanizeAppiumCommand('mobileScroll')` falls back to returning `mobileScroll`, leaving that event unhumanized.

.github/workflows/update-appium-commands.yml
sjchmiela: Good catch. I disabled persisted checkout credentials and now configure git authentication only in the final publish step, after dependency installation and command generation are complete.

.github/workflows/update-appium-commands.yml
sjchmiela: Good catch. The workflow now checks whether the bot branch exists and fetches its current remote ref before pushing with `--force-with-lease`.

packages/build-tools/src/steps/utils/appiumCommandSummary.ts
sjchmiela: You are right. Appium 3 rewrites known `execute` extensions before recording event timings, so `mobile: scroll` is stored as `mobileScroll`. The generator now imports the BaseDriver, XCUITest, and UiAutomator2 command maps and generates separate command unions. `Record<AppiumCommand, string>` enforces complete coverage, including `mobileScroll`, `openNotifications`, `lock`, and `keys`.

.github/workflows/update-appium-commands.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: ./.github/actions/setup-mise
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Regenerate the Appium command list
+        # The generator installs Appium, XCUITest, and UiAutomator2 on demand,
+        # so it always picks up the latest supported command sets.
+        run: yarn workspace @expo/build-tools generate-appium-commands
+      - name: Open a pull request when the command list changed
+        env:
+          GH_TOKEN: ${{ secrets.EXPO_BOT_PAT }}
```
szdziedzic: Isolate dependency execution from EXPO_BOT_PAT

The generator installs the latest Appium and driver packages and then imports them, which executes npm lifecycle scripts and package module code in this job. Limiting `GH_TOKEN` to this step does not provide isolation: code from an earlier step can modify the same runner—for example, by installing a Git hook that executes during `git commit` and inherits `GH_TOKEN`.

Please split this into two jobs: a secretless `generate` job that uploads only `appiumCommands.generated.ts`, and a `publish` job on a fresh runner that downloads and commits that artifact. The publish job should not execute any code from the artifact.

packages/build-tools/scripts/generate-appium-commands.js
```diff
@@ -0,0 +1,125 @@
...
+  },
+  {
+    packageName: 'appium-uiautomator2-driver',
+    exportName: 'AndroidUiautomator2Driver',
+    typeName: 'UiAutomator2AppiumCommand',
+  },
+];
+
+function renderCommandType(typeName, commands) {
+  return `export type ${typeName} =\n${commands.map(command => `  | '${command}'`).join('\n')};`;
```
szdziedzic: Escape command names when generating TypeScript

`command` comes from the installed Appium packages and is interpolated directly into a single-quoted TypeScript literal. A command containing a quote, backslash, or newline would produce invalid or injected source.

Please serialize the value instead, for example:

commands.map(command => `  | ${JSON.stringify(command)}`)

packages/build-tools/src/steps/utils/appiumCommandSummary.ts
```diff
@@ -0,0 +1,413 @@
...
+};
+
+/**
+ * Translate a raw Appium command name into a short, human-readable summary.
+ *
+ * If a command has no curated summary, the raw command name is returned
+ * unchanged — we intentionally do not guess a phrasing.
+ */
+export function humanizeAppiumCommand(command: string): string {
+  return (APPIUM_COMMAND_SUMMARIES as Record<string, string>)[command] ?? command;
```
szdziedzic: Only return own properties from the summary map

This indexed access also searches `Object.prototype`. Because `command` is an unconstrained string from the Appium response, values such as `toString`, `constructor`, or `__proto__` return inherited functions or objects instead of falling back to the raw command. That also violates this function's declared `string` return type at runtime.

Please check `Object.hasOwn(APPIUM_COMMAND_SUMMARIES, command)` before reading the value, and add regression tests for at least `toString` and `constructor`.

.github/workflows/update-appium-commands.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          if git diff --quiet; then
```
szdziedzic: Scope change detection to the generated file

This checks the entire tracked working tree, but only `appiumCommands.generated.ts` is staged below. If dependency installation modifies another tracked file while the generated file is unchanged, this condition proceeds and `git commit` fails because nothing was staged.

Please scope the check to the generated file:

git diff --quiet -- packages/build-tools/src/steps/utils/appiumCommands.generated.ts

szdziedzic (approved):

PR #4229, opened 2026-08-19
[eas-cli] Use production mode for Expo Updates commands

ramonclaudio:
Stacked on [#4180](https://github.com/expo/eas-cli/pull/4180).

# Why

EAS CLI can use an existing `NODE_ENV` when it resolves the runtime version or syncs Expo Updates config, which can make it load different env files for the same build. EAS CLI can also pass dotenv values loaded by a parent process to Expo Updates, which can make Expo Updates reuse values from a different mode.

# How

I updated the Expo Updates commands to use `production` mode for runtime version resolution and Android and iOS config sync, and kept the original EAS CLI env unchanged. EAS CLI still supports Node 20.0.0, but `@expo/env` v2.5.0 requires Node 20.12.0, so I added the inherited dotenv cleanup locally for now.

# Test Plan

Tests and package checks pass.


ramonclaudio:
@expo-bot verify

PR #4239, opened 2026-08-21
[build-tools] Add ios_signing_backend option to the repack step

gabrieldonadel:
# Why

We recently added experimental support for running repack for iOS apps on Linux machines starting from the package version 0.9.0. The idea for using Linux machines is that these are cheaper and more available than macOS ones. 

This PR exposes that backend as an opt-in input on the `eas/repack` step so we can test it end-to-end on EAS. A follow-up PR in universe adds the matching `ios_signing_backend` param to workflow repack jobs.

# How

- Adds optional `ios_signing_backend` input to `eas/repack` (`fastlane` | `zsign`). Default stays fastlane 
- Bumped the `@expo/repack-app` devDependency to `~0.9.0` for the backend-discriminated `IosSigningOptions` union. Even though the runtime package is still installed dynamically per job via `repack_version`.

# Test Plan

- Added a new unit test: resolving zsign options writes the `.p12` and profiles to the temp dir, returns `backend: 'zsign'`, and never touches `IosCredentialsManager`.  
- After this lands: run a workflow repack job with `ios_signing_backend: zsign` and `runs_on: linux-medium` against a real build.

Note: v0.9.0 is published as `next` on npm, so jobs opting this require setting `repack_version: next` for now

### DIFF
--- a/packages/build-tools/src/steps/functions/repack.ts
+++ b/packages/build-tools/src/steps/functions/repack.ts
@@ -305,7 +305,8 @@ export async function resolveAndroidSigningOptionsAsync({
 }
 
 /**
- * Resolves iOS signing options from the job secrets.
+ * Resolves iOS signing options from the job secrets, dispatching on the
+ * requested signing backend.
  */
 export async function resolveIosSigningOptionsAsync({
   job,
@@ -327,17 +328,28 @@ export async function resolveIosSigningOptionsAsync({
   if (iosJob.simulator || buildCredentials == null) {
     return undefined;
   }
+  const commonOptions = { buildCredentials, logger, useAppEntitlements, entitlementsPath };
+  return backend === 'zsign'
+    ? await createIosZsignOptionsAsync({ ...commonOptions, tmpDir })
+    : await createIosFastlaneOptionsAsync(commonOptions);
+}
 
-  if (backend === 'zsign') {
-    return await resolveZsignSigningOptionsAsync({
-      buildCredentials,
-      logger,
-      useAppEntitlements,
-      entitlementsPath,
-      tmpDir,
-    });
-  }
-
+/**
+ * Creates signing options for the fastlane backend: certificates are imported
+ * into a temporary keychain and provisioning profiles are parsed with the
+ * macOS `security` tool.
+ */
+async function createIosFastlaneOptionsAsync({
+  buildCredentials,
+  logger,
+  useAppEntitlements,
+  entitlementsPath,
+}: {
+  buildCredentials: Ios.BuildCredentials;
+  logger: bunyan;
+  useAppEntitlements?: boolean;
+  entitlementsPath?: string;
+}): Promise<IosSigningOptions> {
   const credentialsManager = new IosCredentialsManager(buildCredentials);
   const credentials = await credentialsManager.prepare(logger);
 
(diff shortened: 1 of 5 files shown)


packages/build-tools/src/steps/functions/repack.ts
```diff
@@ -330,3 +353,51 @@ export async function resolveIosSigningOptionsAsync({
...
+  tmpDir,
+}: {
+  buildCredentials: Ios.BuildCredentials;
+  logger: bunyan;
+  useAppEntitlements?: boolean;
+  entitlementsPath?: string;
+  tmpDir: string;
+}): Promise<IosSigningOptions> {
+  const targets = Object.entries(buildCredentials);
+  const [targetName, targetCredentials] = targets[0];
```
Kudo, line 388: not sure if `targets[0]` here is correct. could we reuse the fastlane flow for provisioningProfile?

https://github.com/expo/eas-cli/blob/3f8ffac11cbb2d1b6765677fba699f2913b252cb/packages/build-tools/src/steps/functions/repack.ts#L318-L324

packages/build-tools/src/steps/functions/repack.ts
```diff
@@ -302,19 +310,34 @@ export async function resolveAndroidSigningOptionsAsync({
...
+  tmpDir: string;
 }): Promise<IosSigningOptions | undefined> {
   const iosJob = job as Ios.Job;
   const buildCredentials = iosJob.secrets?.buildCredentials;
   if (iosJob.simulator || buildCredentials == null) {
     return undefined;
   }
+
+  if (backend === 'zsign') {
+    return await resolveZsignSigningOptionsAsync({
```
Kudo: not a fan of
`resolveIosSigningOptionsAsync -> resolveZsignSigningOptionsAsync`

maybe split out a dedicated resolver for fastlane. i.e.

`resolveIosSigningOptionsAsync` (a factory) -> `createIosZsignOptionsAsync` and `createIosFastlaneOptionsAsync`


or if we can reuse provisioningProfile code, only certificatePath is different, we even not necessarily to split fastlane/zsign function.

Kudo (changes requested):

packages/build-tools/src/steps/functions/repack.ts
gabrieldonadel: From my understanding, we can't really reuse the fastlane profile flow because `ProvisioningProfile.load()` parses the profile with `security cms -D -k <keychain>` which is macOS-only.

About the `targets[0]` here: zsign signs every target with a single identity, and I believe EAS provisions the same distribution certificate for all targets of a build, but I guess we could add a check for mismatches?

packages/build-tools/src/steps/functions/repack.ts
gabrieldonadel: Updated to use `resolveIosSigningOptionsAsync` (a factory) -> `createIosZsignOptionsAsync` and `createIosFastlaneOptionsAsync`, given that we can't use provisioningProfile code

packages/build-tools/src/steps/functions/repack.ts
Kudo: oh good to know. thanks for sharing the context. in most case we would have single provisiong profile except extension like expo-go. if expo-go works then it should be fine.

Kudo (approved):

PR #4244, opened 2026-08-21
[eas-cli] Use an isolated env for local builds

ramonclaudio:
Stacked on [#4229](https://github.com/expo/eas-cli/pull/4229).

# Why

`eas build --local` passes every var from the shell to the local-build plugin, which can make local builds behave differently from EAS Build.

# How

I updated the local-build plugin process to start with an isolated env, then added the EAS env, local runtime vars, and the local build settings it needs.

# Test Plan

Tests, package checks, and local Android and iOS builds pass.

### DIFF
--- a/packages/eas-cli/src/build/local.ts
+++ b/packages/eas-cli/src/build/local.ts
@@ -12,6 +12,17 @@ import { getEnvWithoutInheritedDotenvValues } from '../utils/originalEnv';
 const PLUGIN_PACKAGE_NAME = 'eas-cli-local-build-plugin';
 const PLUGIN_PACKAGE_VERSION = version; // should match version of @expo/eas-build-job
 
+// The plugin starts with an isolated env, so keep the runtime vars it needs from the user's
+// machine.
+const LOCAL_BUILD_RUNTIME_ENV_NAMES = [
+  'ANDROID_HOME',
+  'ANDROID_NDK_HOME',
+  'ANDROID_SDK_ROOT',
+  'HOME',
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/build/local.ts
```diff
@@ -63,11 +64,12 @@ export async function runLocalBuildAsync(
   };
   process.on('SIGINT', interruptHandler);
   try {
+    const processEnv = getEnvWithoutInheritedDotenvValues(process.env);
```
kitten, line 88: `EXPO_UNSAFE_DOTENV_KEYS` and this utility is something I'm not familiar with 🤔 maybe @sjchmiela or @szdziedzic can evaluate this better. I believe, this may be similar to a utility for original-env we have in `@expo/env`, so not sure if we need to reconsolidate?

kitten: Revisiting this; assuming we're keeping parity here with EAS Build, is there any way we could just not inherit any env vars and issue a major bump? It might be easier to build up parity than it is to try to pass through filtered `process.env` vars, and treat the spawned sub-process as "isolated"

ramonclaudio: Yea, that makes sense! I tested local ios and android builds with the runtime var list I added, but can we confirm these are all the vars we want to keep in the isolated env?

https://github.com/expo/eas-cli/pull/4244/changes#diff-a54f38cd6a574317f91ba14ef751c1868db0d9469f924f18b018222c1c956a6aR18-R33

PR #4252, opened 2026-08-21
[eas-cli] Add observe:event command to fetch a single event by ID

douglowder:
# Why

Provide a new `observe:event` command for retrieving either metric or custom events by ID.

# How

Add the new command, and associated queries that call the new event(id:) or customEvent(id:) GraphQL . The command returns whichever query finds the event, or throws if no event is found, similarly to `observe:session`.

# Test Plan

- New unit tests added
- Testing on a real app in production

---


packages/eas-cli/src/graphql/queries/ObserveQuery.ts
```diff
@@ -116,6 +116,23 @@ type AppObserveCustomEventListQueryVariables = {
   orderBy?: AppObserveCustomEventListOrderBy;
 };
 
+type AppObserveEventByIdQuery = {
+  app: {
+    byId: {
+      id: string;
+      observe: {
+        event: AppObserveEvent | null;
+        customEvent: AppObserveCustomEvent | null;
```
kadikraman: The formats don't overlap so we shouldn't try to request both. If the ID matches a UUID, query `customEvent` only; otherwise query `event` only.

This has a real consequence that if you do `eas observe:event` with a valid metric event, e.g. `eyJldmVudEJhdGNoSWQiOiI0ZTFkZmM5NC04MzMyLTQ3M2MtYmY3Zi00ZjllODAyYmE1ODciLCJtZXRyaWNOYW1lIjoiZXhwby5hcHBfc3RhcnR1cC50dHIiLCJtZXRyaWNUaW1lc3RhbXAiOiIyMDI2LTA4LTE0VDEyOjA0OjM5LjAwMFoifQ` it will always fail because the `customEvent` resolver tries to parse it as an UUID (see [Sentry](https://expoio.sentry.io/issues/7221561450/?project=130569&query=is%3Aunresolved%20requestId%3A%EF%80%8DContains%EF%80%8D18633202-3879-4c50-97aa-b1d2d66f31b4&referrer=issue-stream&sort=recommended) error).

douglowder: Fixed, see new code in `event.ts`

packages/eas-cli/src/commands/observe/event.ts
```diff
@@ -0,0 +1,165 @@
...
+import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+export default class ObserveEvent extends EasCommand {
+  static override description = 'display a single Observe event (metric or log) by its ID';
+
+  static override args = {
+    id: Args.string({
+      description:
+        'ID of the event to display (from `eas observe:events` or `eas observe:session`)',
```
kadikraman: `eas observe:session` does not print any IDs so this comment is not accurate

douglowder: It does if you add `--json`. I'll check the results and make this description more precise.

douglowder: Fixed the description to be more complete and accurate.

kadikraman:
@douglowder let's pause this until we've updated the API - the name of the resolver will change. I'll let you know once the API changes are available.

packages/eas-cli/src/commands/observe/event.ts
```diff
@@ -0,0 +1,174 @@
...
+import { buildObserveErrorDetail, buildObserveErrorJson } from '../../observe/formatErrors';
+import { buildObserveEventDetail, buildObserveEventJson } from '../../observe/formatEvents';
+import { ObserveProjectIdFlag } from '../../observe/flags';
+import { withObservePlanGateHandlingAsync } from '../../observe/planGating';
+import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+export default class ObserveEvent extends EasCommand {
+  static override description =
+    'display a single Observe event (metric, log, or error) by its ID. IDs are included in event data when the `--json` flag is passed to `eas observe:session`, `eas observe:metrics`, or `eas observe:events`.';
```
kadikraman, line 26: `eas observe:session --json` does not show event IDs atm, it's already getting fetched, just needs to be passed through.

kadikraman (approved):

packages/eas-cli/src/commands/observe/event.ts
douglowder: Done

PR #4258, opened 2026-08-24
[eas-cli] recognize queued workflow runs and report finishedAt only for final runs

hSATAC:
# Why

Queue semantics for workflow concurrency added a `WAITING` run status: a run queued behind its concurrency group until the run holding the group finishes. eas-cli currently treats it as just another transient status — polling keeps working, but nothing explains why a run isn't starting, queued runs can't be selected in `workflow:cancel`, and run listings print a bogus finish time for runs that haven't finished.

# How

- `workflow:run --wait` and `workflow:status` show a queued run as `Workflow run is queued: Waiting for <run name>`, naming the run that holds the group via the new `blockingWorkflowRun` field. When that run is itself waiting on user action (an approval gate or a device registration), the spinner appends `— it needs your action first: <link to the run>`, matching the CLI's link-out convention for actions that happen on the website. The wording leads with "queued" to match the website's label for this state.
- `workflow:runs` reports `finishedAt` only for final statuses, reading the server's `finalizedAt` instead of `updatedAt` (which any later write to the row refreshes). Final runs predating the `finalizedAt` column report null rather than an approximated time.
- `workflow:cancel` includes queued runs in the interactive picker. While in there, this also fixes `--non-interactive`, which was parsed but never applied (the command scanned argv for the flag, but oclif consumes declared flags out of argv), so automation could hang on the interactive picker.
- The committed GraphQL snapshot is regenerated from production plus a hand-added `blockingWorkflowRun` field, which ships with the www concurrency stack. Once that stack is live, `yarn generate-graphql-code` should reproduce the snapshot with no diff.

Merge gate: do not merge until [expo/universe#30099](https://github.com/expo/universe/pull/30099) is live in production — the polled query selects `blockingWorkflowRun` at runtime, so a release cut before that deploy would break `workflow:run --wait` and `workflow:status`. Gating the merge instead of the release keeps main releasable at all times.

# Test Plan

Jest covers `finishedAt` for every status (including the `finalizedAt` preference and the legacy fallback), the cancel picker listing both in-progress and queued runs, and two `--non-interactive` regressions. Lint, format, typecheck, and the changelog check are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,12 @@ This is the log of notable changes to EAS CLI and related packages.
 
 - [build-tools] Add `ios_signing_backend` option to the repack step. ([#4239](https://github.com/expo/eas-cli/pull/4239) by [@gabrieldonadel](https://github.com/gabrieldonadel))
 - [build-tools] Support an optional `package_version` input on `eas/start_serve_sim_remote_session`, so a simulator session can pin the `@expo/serve-sim` version instead of always running `latest`. ([#4253](https://github.com/expo/eas-cli/pull/4253) by [@gwdp](https://github.com/gwdp))
+- [eas-cli] Display queued workflow runs: `workflow:run --wait` and `workflow:status` show a run waiting for its concurrency group and the run blocking it, and `workflow:cancel` includes queued runs in its selection list. ([#4258](https://github.com/expo/eas-cli/pull/4258) by [@hSATAC](https://github.com/hSATAC))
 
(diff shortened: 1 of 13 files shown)


hSATAC:
Heads up: the failing GraphQL schema check is expected for now. This PR uses `blockingWorkflowRun` from expo/universe#30099, which isn't in production yet, so the live-schema validation fails.

Review can proceed in the meantime — the check should go green once #30099 is deployed.

sjchmiela (approved):

PR #4263, opened 2026-08-24
[build-tools] Support simulator application launch options

szdziedzic:
# Why

Companion to https://github.com/expo/universe/pull/30416

# How

Use launch args and tunnel url to actually do things.

Use launch args when launching the app.

Open tunnel url in launched build (it will work only if this is dev client or expo go)

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/steps/functions/launchApplication.ts
+++ b/packages/build-tools/src/steps/functions/launchApplication.ts
@@ -26,6 +26,16 @@ export function createLaunchApplicationFunction(): BuildFunction {
         required: false,
         allowedValueTypeName: BuildStepInputValueTypeName.STRING,
       }),
+      BuildStepInput.createProvider({
+        id: 'launch_args',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.JSON,
+      }),
+      BuildStepInput.createProvider({
+        id: 'tunnel_url',
+        required: false,
+        allowedValueTypeName: BuildStepInputValueTypeName.STRING,
+      }),
     ],
     fn: async ({ global, logger }, { inputs, env }) => {
       const applicationIdentifier = parseNonEmptyStringInput(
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/functions/launchApplication.ts
```diff
@@ -77,10 +100,32 @@ export async function launchApplicationAsync({
   }
 
   logger.info(`Launching ${applicationIdentifier}.`);
-  await spawn('adb', ['shell', 'am', 'start', '-n', `${applicationIdentifier}/${activityName}`], {
-    env,
-    logger,
-  });
+  await spawn(
+    'adb',
+    ['shell', 'am', 'start', ...launchArgs, '-n', `${applicationIdentifier}/${activityName}`],
```
sjchmiela, line 107: is this how we should do launch args? or https://stackoverflow.com/a/59445074?

packages/build-tools/src/steps/functions/launchApplication.ts
```diff
@@ -77,10 +100,32 @@ export async function launchApplicationAsync({
   }
 
   logger.info(`Launching ${applicationIdentifier}.`);
```
sjchmiela: should we note launch args here?

sjchmiela (approved):

packages/build-tools/src/steps/functions/launchApplication.ts
szdziedzic: This is intentionally passing raw am start intent arguments through, so callers can use --es, --ez, etc. as described in the linked answer. Android doesn’t have an equivalent of iOS process arguments, and automatically converting string pairs would be ambiguous because Intent extras can have different types. I’ll clarify that launch_args means process arguments on iOS and am start intent arguments/options on Android.

packages/build-tools/src/steps/functions/launchApplication.ts
szdziedzic: Good point. turtle-spawn doesn’t log the command itself, so these arguments aren’t currently visible. I’ll include the launch arguments in the log when they’re present.

packages/build-tools/src/steps/functions/launchApplication.ts
szdziedzic: Yes — that is the intended Android model. launch_args are passed through as raw am start Intent arguments, so callers can use --es, --ez, and similar options as in the linked answer; Android does not expose iOS-style process arguments. I added an inline comment clarifying those platform-specific semantics.

packages/build-tools/src/steps/functions/launchApplication.ts
szdziedzic: Yep. turtle-spawn does not log the command, so the launch log now includes the JSON-serialized argument list when arguments are present. I also added a regression assertion for the logged arguments.

PR #4266, opened 2026-08-24
[build-tools] support remote CocoaPods caching

AbbanMustafa:
## Why

CocoaPods dependencies are installed from scratch on every iOS build. For projects with large dependency graphs, `pod install` can add several minutes even when `Podfile.lock` has not changed.

Local validation with 107 pods showed that remote Pods caching reduced the relevant phases from 111 seconds to 42 seconds.

The cache-hit validation started with no `Podfile.lock` present. It requested the `ios-pods-1.16.2-` prefix, restored the hash-qualified `ios-pods-1.16.2-5c51d462...` entry, and completed the iOS build successfully.

## How

- Add opt-in CocoaPods caching with `EAS_PODS_CACHE=1`.
- Cache `ios/Pods` under the exact key `ios-pods-<cocoapods-version>-<sha256-of-Podfile.lock>`.
- When a managed project has no lockfile before prebuild, request the `ios-pods-<cocoapods-version>-` prefix and restore its newest matching entry.
- Always run `pod install` to reconcile restored Pods with the generated lockfile.
- Save the resulting Pods directory under the exact lockfile key.
- Support traditional, custom, and simulator iOS builds.
- Treat cache failures as non-fatal.

`~/Library/Caches/CocoaPods` is intentionally excluded. In local testing, its 435 MB archive added approximately 33 seconds of restore time and 42 seconds of recompression on a cache hit.

## Test Plan

- [x] Run focused build-tools tests: 16 passed.
- [x] Run build-tools typecheck.
- [x] Run repository lint.
- [x] Run changelog lint.
- [x] Populate the cache from a managed iOS build.
- [x] Restore the cache on a subsequent managed iOS build.
- [x] Validate prefix fallback when no `Podfile.lock` exists before prebuild.
- [x] Run the cache-hit build with `~/Library/Caches/CocoaPods` removed to simulate a fresh worker.
- [x] Confirm the cache-hit build completes without CocoaPods downloads.
- [x] Measure a 242 MB Pods archive.
- [x] Measure install phases:

| Scenario | Restore | Install pods | Save | Total |
| --- | ---: | ---: | ---: | ---: |
| Cold, cache disabled | 1s | 110s | — | 111s |
| Cold worker, cache hit | 15s | 17s | 10s | 42s |


packages/build-tools/src/builders/ios.ts
```diff
@@ -107,6 +111,12 @@ async function buildInnerAsync(
         env: ctx.env,
         secrets: ctx.job.secrets,
       });
+      await restoreCocoapodsCacheAsync({
+        logger: ctx.logger,
+        workingDirectory,
+        env: ctx.env,
+        secrets: ctx.job.secrets,
+      });
```
sjchmiela, line 119: i wonder if it would make sense to parallelize them some time

maybe not in this PR

this is going to create weird log output

packages/build-tools/src/builders/ios.ts
```diff
@@ -250,6 +260,12 @@ async function buildInnerAsync(
       env: ctx.env,
       secrets: ctx.job.secrets,
     });
+    await saveCocoapodsCacheAsync({
+      logger: ctx.logger,
+      workingDirectory,
+      env: ctx.env,
+      secrets: ctx.job.secrets,
+    });
```
sjchmiela, line 268: same

packages/build-tools/src/utils/cocoapodsCache.ts
```diff
@@ -0,0 +1,106 @@
...
+    return { key: keyPrefix, keyPrefix };
+  }
+
+  return {
+    key: `${keyPrefix}${hashFiles([podfileLockPath])}`,
+    keyPrefix,
+  };
+}
+
+export async function compressCocoapodsCacheAsync({
```
sjchmiela, line 49: why do we prefer this over `compressCacheAsync`?

packages/build-tools/src/steps/functions/saveBuildCache.ts
```diff
@@ -78,11 +83,77 @@ export function createSaveBuildCacheFunction(evictUsedBefore: Date): BuildFuncti
...
+          workingDirectory,
+          env,
+          secrets: stepCtx.global.staticContext.job.secrets,
+        });
       }
     },
   });
 }
 
+export async function saveCocoapodsCacheAsync({
```
sjchmiela, line 98: we need to prune the cache somehow too? what's the plan here?

sjchmiela (commented):
I like it!

packages/build-tools/src/steps/functions/saveBuildCache.ts
AbbanMustafa: Pods does not accumulate like gradle/ccache. pod install is responsible for reconciling the additions and removals https://guides.cocoapods.org/using/pod-install-vs-update.html, and CocoaPods cleanup removes pods marked for deletion https://github.com/CocoaPods/CocoaPods/blob/b80e113e28a7e23cd6dbc169f0c923f6b48bf7a2/lib/cocoapods/installer.rb#L452-L476

packages/build-tools/src/utils/cocoapodsCache.ts
AbbanMustafa: Its because compressCacheAsync does not preserve symlink entries as we need them to be, and Pods commonly contains symlinks. I can fix the helper in a follow up!

packages/build-tools/src/builders/ios.ts
AbbanMustafa: yes we probably could, I will follow up on this

sjchmiela (approved):

PR #4274, opened 2026-08-24
[build-tools] Bypass simulator URL confirmation

szdziedzic:
# Why

Opening a custom URL scheme with `eas/launch_application` on a fresh iOS Simulator shows the system “Open in <application>?” confirmation. `simctl openurl` delivers the URL, but the simulator session remains blocked on that prompt instead of opening the URL automatically.

This is a follow-up to #4263 and #4264.

# How

Before calling `simctl openurl` for a custom scheme, write the corresponding application approval to the booted Simulator's `com.apple.launchservices.schemeapproval` preferences. The approval maps the scheme opened by CoreSimulatorBridge to the installed application's bundle identifier.

HTTP and HTTPS URLs keep their existing behavior. Android launch behavior is unchanged.

# Test Plan

Tests

### DIFF
--- a/packages/build-tools/src/steps/functions/launchApplication.ts
+++ b/packages/build-tools/src/steps/functions/launchApplication.ts
@@ -145,26 +145,33 @@ async function preapproveIosUrlSchemeAsync({
   env: BuildStepEnv;
   logger: bunyan;
 }): Promise<void> {
-  const urlScheme = new URL(openUrl).protocol;
-  if (urlScheme === 'http:' || urlScheme === 'https:') {
+  const urlScheme = new URL(openUrl).protocol.slice(0, -1);
+  if (urlScheme === 'http' || urlScheme === 'https') {
     return;
   }
 
-  await spawn(
-    'xcrun',
-    [
-      'simctl',
-      'spawn',
-      'booted',
-      'defaults',
-      'write',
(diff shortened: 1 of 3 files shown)


packages/build-tools/src/steps/functions/launchApplication.ts
```diff
@@ -130,6 +134,39 @@ export async function launchApplicationAsync({
...
+  openUrl: string;
+  env: BuildStepEnv;
+  logger: bunyan;
+}): Promise<void> {
+  const urlScheme = new URL(openUrl).protocol;
+  if (urlScheme === 'http:' || urlScheme === 'https:') {
+    return;
+  }
+
+  await spawn(
```
gwdp: Worth enveloping in a try/catch (keep it best effort since we use unstable `com.apple.CoreSimulator.CoreSimulatorBridge-->`)?

gwdp (approved):

packages/build-tools/src/steps/functions/launchApplication.ts
szdziedzic: Good call. The scheme-preapproval write is now wrapped in a try/catch. On failure it logs the original error plus a warning that manual confirmation may be required, then continues to simctl openurl. I also added a regression test that rejects the preference write and verifies the URL is still opened.

gwdp (approved):

PR #4275, opened 2026-08-25
[eas-cli] add environment filter to observe commands

kadikraman:
# Why

Part of ENG-26159

We want to match the filters on the dashboard to we could correctly pre-fill them in the prompts.

# How

Add the `--environment` filter to `metrics-summary`, `routes`, `events`, `versions` and `session`

# Test Plan

Tested locally against staging (I used `--environment playground` in my personal app just fyi, but I'll past the examples with the more popular `--environment production`):

## Metrics summary

```sh
eas observe:metrics-summary --environment production
eas observe:metrics-summary --metric tti --stat median --days 7 --environment production
```

## Metrics

```sh
eas observe:metrics tti --environment production
eas observe:metrics cold_launch --environment development --sort slowest --limit 20
eas observe:metrics tti --environment production --app-version 2.1.0 --platform ios
```

## Routes

```sh
eas observe:routes --environment production
eas observe:routes --metric cold_ttr --route-name /home --environment preview
```

## Events

```sh
eas observe:events --environment production            # summary of event names + counts
eas observe:events login_pressed --environment production
eas observe:events --all-events --environment development --days 3
```

## Versions

```sh
eas observe:versions --environment production
eas observe:versions --platform android --days 30 --environment production
```

## Session

```sh
eas observe:session --event-name tti --sort slowest --environment production
```

### DIFF
--- a/packages/eas-cli/src/observe/flags.ts
+++ b/packages/eas-cli/src/observe/flags.ts
@@ -15,6 +15,12 @@ export const ObservePlatformFlag = {
   })(),
 };
 
+export const ObserveEnvironmentFlag = {
+  environment: Flags.string({
+    description: 'Filter by environment (e.g. production, development)',
+  }),
+};
+
 export const ObserveTimeRangeFlags = {
   start: Flags.string({
     description: 'Start of time range (ISO date)',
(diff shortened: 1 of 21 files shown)


packages/eas-cli/src/observe/flags.ts
```diff
@@ -15,6 +15,12 @@ export const ObservePlatformFlag = {
   })(),
 };
 
+export const ObserveEnvironmentFlag = {
```
douglowder, line 18: Can we not use `EASEnvironmentFlag`?

douglowder (changes requested):
See inline

packages/eas-cli/src/observe/flags.ts
kadikraman: Type-wise, we could, but logic-wise, we shouldn't because it's not referring to the same environment value. We are not currently adding the EAS environment to events, but if we were, these two could be different.

douglowder (approved):

PR #4290, opened 2026-08-26
[ci] Bump both STAGING and PRODUCTION cli-versions in release workflow

douglowder:
# Why

Since updating the release workflow to use `cli-versions.json` instead of NPM tags, there have been issues where the version of the CLI used in www production has gone out of date.

# How

Align the new workflow with the old workflow, which updated both staging and production NPM tags on release. After this change, both STAGING and PRODUCTION values in `cli-versions.json` will be bumped for a new release.

# Test Plan

CI should pass.

---


szdziedzic (approved):
awesome

sjchmiela (approved):

PR #4299, opened 2026-08-27
[eas-cli] Add dynamic retries for deploy command

kitten:
# Why

When we encounter transient errors, we don't account for network errors that could be an ongoing internal failure or outage on the machine that's deploying. This is especially punishing when the deployment is larger and has more missing assets to upload, since it has to fully retry.

# How

- Expand maximum retry attempts and timeouts to be much longer
- Dynamically scale retry attempts and timeouts based on amount of asset requests we'll have
- Scale up retries for network errors (ENOTFOUND and ECONNRESET, and others)
- Issue a visible warning to the user when we're still retrying but have waited for longer than 30s for a successful upload

# Test Plan

- LLM-derived tests added to capture this behaviour

### DIFF
--- a/packages/eas-cli/src/worker/upload.ts
+++ b/packages/eas-cli/src/worker/upload.ts
@@ -16,7 +16,9 @@ import {
 } from './utils/multipart';
 
 const MAX_CONCURRENCY = Math.min(10, Math.max(os.availableParallelism() * 2, 20));
-const RETRY_WARNING_DELAY_MS = 30_000;
+const MAX_RETRY_WARNING_DELAY_MS = 30_000;
+const UPLOAD_RETRY_LIMITS: RetryLimits = { retries: 4, maxTimeout: 5_000 };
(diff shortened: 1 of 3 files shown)


packages/eas-cli/src/worker/upload.ts
```diff
@@ -216,7 +354,10 @@ export async function* batchUploadAsync(
             progressTracker[currentIndex] = progress;
             sendProgressUpdate();
           });
-        const uploadPromise = uploadAsync(initWithSignal, payload, onChildProgressUpdate).then(
+        const uploadPromise = uploadAsync(initWithSignal, payload, onChildProgressUpdate, {
+          totalRequests: payloads.length,
+          state: retryState,
```
kadikraman: The retry state is shared, so for parallel uploads, a sibling upload can reset the `firstRetryAt` which suppresses the warning.

kadikraman (approved):

PR #4303, opened 2026-08-28
[eas-cli] Display build logs when --wait flag is set

AHGIJMKLKKZNPJKQR:
# Why

Tracking progress of builds created with `eas build` requires going to the web UI.
The `--wait` flag only displays a spinner, which is in contrast to workflows, which display logs and steps as they progress.

# How

We use the existing log polling and realtime log centrifugo clients used by workflows to ingest logs as they appear.

- Extracted common logging types and logic from `packages/eas-cli/src/commandUtils/workflow/logs` into `packages/eas-cli/src/commandUtils/logs`.
- Updated `packages/eas-cli/src/build/build` to store the log state and rerender the spinner text when logs change (same as workflows).

# Test Plan

CI passes, manual testing with local build of eas cli

---


packages/eas-cli/src/build/logs.ts
```diff
@@ -0,0 +1,111 @@
...
+
+const SINGLE_BUILD_MAX_LOG_LINES = 5;
+const MULTIPLE_BUILDS_MAX_LOG_LINES = 3;
+
+async function fetchAndParseLogsForBuildAsync(build: BuildFragment): Promise<RawLogLine[] | null> {
+  const logFileUrl = build.logFiles[0];
+  if (!logFileUrl) {
+    return null;
+  }
+  const response = await fetch(logFileUrl);
```
sjchmiela, line 20: use `fetch` from teh package

packages/eas-cli/src/build/build.ts
```diff
@@ -538,17 +547,62 @@ export async function waitForBuildEndAsync(
...
+          render();
+        }
+      )
+    : null;
+
+  try {
+    while (true) {
+      render = (): void => {};
+      const builds = await getBuildsSafelyAsync(graphqlClient, buildIds);
+      const logsStates = await syncBuildLogsAsync(watcher, builds);
```
sjchmiela, line 564: should we have a timeout for this in case something breaks or is super slow??

sjchmiela: and/or parallelize with getbuidlssafelyasync

packages/eas-cli/src/build/logs.ts
sjchmiela, line 20: we might want to check the status code on the response

sjchmiela (approved):

packages/eas-cli/src/build/logs.ts
AHGIJMKLKKZNPJKQR: If the status code is `>= 400` then the fetch throws, which is what we want, only success will continue down

packages/eas-cli/src/build/build.ts
AHGIJMKLKKZNPJKQR: parallelizing with `getBuildsSafelyAsync` isn't really feasible, since we want to fetch the logs of new builds, but I've added timeouts to the fetches and websocket connection

PR #4306, opened 2026-08-28
[build-tools] gate device-run-session serve-sim with the session token

gwdp:
# Why

serve-sim can now require a session token before it serves anything (assuming expo/serve-sim#49 merge). This turns that on for device run sessions.

# How

Pass `--require-token` when we start serve-sim, then carry the token:

- Read it from serve-sim state file
- Send it to www as its own field, `previewToken` or `webPreviewToken`; Usage on expo/universe#30571
- Send it as a bearer header on the `/metrics` stream we collect, which is gated now

Needs a serve-sim release with the flag first. An older serve-sim rejects the unknown option and will not start, so this should land after that publishes.

# Test Plan

- CI
- Local test with new serve-sim

### DIFF
--- a/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
+++ b/packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
@@ -222,6 +222,7 @@ export function createStartArgentRemoteSessionBuildFunction(
             toolsUrl: publicToolsUrl,
             ...(toolServerToken ? { toolsAuthToken: toolServerToken } : {}),
             webPreviewUrl: webPreview.previewUrl,
+            ...(webPreview.previewToken ? { webPreviewToken: webPreview.previewToken } : {}),
           },
           logger,
         });
--- a/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
+++ b/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
@@ -881,11 +881,13 @@ export async function startServeSimWithTunnelAsync(
     readPreviewTokenAsync: async device => {
       const previewToken = await readServeSimPreviewTokenAsync(device);
       if (!previewToken) {
+        // A serve-sim that does not know --require-token fails earlier, in the readiness check, so
+        // reaching here means it started and left no token in its state file.
         throw new SystemError(
-          `serve-sim started without a session token for device ${device}. It is always launched ` +
-            'with --require-token, so this usually means the pinned @expo/serve-sim predates that ' +
-            'flag and ignored it, leaving the preview ungated on a public tunnel. Pin a serve-sim ' +
-            'version that supports --require-token.'
+          `serve-sim became ready but wrote no session token for device ${device}. The preview is ` +
+            'on a public tunnel and would be reachable without one, so the session cannot continue. ' +
+            'This usually means the state file was not written as expected; retry the session, and ' +
+            'report it if it repeats.'
         );
       }
       return previewToken;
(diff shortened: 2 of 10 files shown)


packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
```diff
@@ -219,6 +221,7 @@ export function createStartArgentRemoteSessionBuildFunction(
             toolsUrl: publicToolsUrl,
             ...(toolServerToken ? { toolsAuthToken: toolServerToken } : {}),
             ...(webPreviewUrl ? { webPreviewUrl } : {}),
+            ...(webPreviewToken ? { webPreviewToken } : {}),
```
szdziedzic: I don't think we support this input in WWW as of today, right? we need to add it first?

szdziedzic: oh ok we have PR to add it great

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -729,7 +740,13 @@ export async function startServeSimWithTunnelAsync(
 
   try {
     logger.info('Waiting for serve-sim to become ready.');
-    await waitForServeSimReadyAsync({ serveSim, port, timeoutMs });
+    const device = await waitForServeSimReadyAsync({ serveSim, port, timeoutMs });
+    const previewToken = await readServeSimPreviewTokenAsync(device);
+    if (!previewToken) {
+      logger.warn(
+        'serve-sim did not report a session token; the preview link will not be pre-authorized.'
+      );
```
szdziedzic: Can we fail startup here instead of continuing? We unconditionally pass `--require-token`, so a successfully started serve-sim without a token is not a valid ungated fallback?

szdziedzic (changes requested):
[P1] EAS CLI prints preview URLs without the token. The CLI query still requests only previewUrl/webPreviewUrl, then prints the bare URL. Gated serve-sim returns 401 for those links. Request previewToken/webPreviewToken in [DeviceRunSessionQuery.ts](https://github.com/expo/eas-cli/blob/d48b193453b90ab88fd50b44a9edb57743145b49/packages/eas-cli/src/graphql/queries/DeviceRunSessionQuery.ts#L85-L102), regenerate types, and append ?token= when formatting preview instructions. This affects every iOS session type; serve-sim-only sessions become unusable from the CLI.

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
gwdp: Good call. Yes, updated to throw now. Thank you!

gwdp:
> [P1] EAS CLI prints preview URLs without the token. The CLI query still requests only previewUrl/webPreviewUrl, then prints the bare URL. Gated serve-sim returns 401 for those links. Request previewToken/webPreviewToken in [DeviceRunSessionQuery.ts](https://github.com/expo/eas-cli/blob/d48b193453b90ab88fd50b44a9edb57743145b49/packages/eas-cli/src/graphql/queries/DeviceRunSessionQuery.ts#L85-L102), regenerate types, and append ?token= when formatting preview instructions. This affects every iOS session type; serve-sim-only sessions become unusable from the CLI.

Yep, I inverted order here; expo/universe#30570 was coming, but I missed that it needs to be before this.

Deploy order: [#30628](https://github.com/expo/universe/pull/30628) → [#30571](https://github.com/expo/universe/pull/30571) → serve-sim#49 publish → [#4312](https://github.com/expo/eas-cli/pull/4312) → [#4306](https://github.com/expo/eas-cli/pull/4306).

packages/build-tools/src/steps/functions/startArgentRemoteSession.ts
gwdp: Sorry, should have done that more explicit in the description, only linked the website which was the last on universe stack

szdziedzic (approved):

PR #4312, opened 2026-08-31
[eas-cli] carry the serve-sim session token in simulator preview links

gwdp:
# Why

`eas simulator` and `eas simulator get` print the preview link. Once the preview is gated by expo/serve-sim#49, a bare link returns 401, so the printed instructions stop working. This affects every iOS session type, not only serve-sim ones.

# How

- Ask for the token in the session query, then append `?token=` when there is one. All four print sites go through the same `formatPreviewUrl` helper.
- Nothing changes for a session without a token, which is every session today.
- Ships before #4306, which is what turns gating on.

# Test Plan

- CI

### DIFF
--- a/packages/eas-cli/src/graphql/queries/DeviceRunSessionQuery.ts
+++ b/packages/eas-cli/src/graphql/queries/DeviceRunSessionQuery.ts
@@ -87,22 +87,27 @@ export const DeviceRunSessionQuery = {
                       agentDeviceRemoteSessionUrl
                       agentDeviceRemoteSessionToken
                       webPreviewUrl
+                      webPreviewToken
                     }
                     ... on ArgentRunSessionRemoteConfig {
                       toolsUrl
                       toolsAuthToken
                       webPreviewUrl
+                      webPreviewToken
                     }
                     ... on AppiumRunSessionRemoteConfig {
                       appiumUrl
                       capabilities
                       webPreviewUrl
+                      webPreviewToken
                     }
                     ... on ServeSimRunSessionRemoteConfig {
                       previewUrl
+                      previewToken
                     }
                     ... on WebPreviewOnlyRunSessionRemoteConfig {
                       previewUrl: webPreviewUrl
+                      previewToken: webPreviewToken
                     }
                   }
                   turtleJobRun {
--- a/packages/eas-cli/src/simulator/utils.ts
+++ b/packages/eas-cli/src/simulator/utils.ts
@@ -99,6 +99,34 @@ export function formatPreviewUrl(url: string, token: string | null | undefined):
   return withToken.toString();
 }
 
+/**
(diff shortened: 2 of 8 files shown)


packages/eas-cli/src/simulator/utils.ts
```diff
@@ -86,6 +86,19 @@ export function getRemoteSessionEnvironmentVariables(
...
+/**
+ * Preview link for a session. A gated serve-sim needs the session token, and a browser cannot send
+ * a header on a page load, so it rides the query. serve-sim swaps it for a cookie on the first load.
+ */
+export function formatPreviewUrl(url: string, token: string | null | undefined): string {
+  if (!token) {
+    return url;
+  }
+  const withToken = new URL(url);
+  withToken.searchParams.set('token', token);
```
szdziedzic, line 98: Follow up idea: What if the URL would carry a distinct, least-privilege credential - ideally a short-lived, one-time bootstrap token exchanged for a secure cookie. Browser controls should authorize only typed simulator operations, never arbitrary commands. We are also planning to move streaming to expo.dev so we should also plan having it in mind.

packages/eas-cli/src/graphql/queries/DeviceRunSessionQuery.ts
```diff
@@ -87,22 +87,27 @@ export const DeviceRunSessionQuery = {
...
                     }
                     ... on AppiumRunSessionRemoteConfig {
                       appiumUrl
                       capabilities
                       webPreviewUrl
+                      webPreviewToken
                     }
                     ... on ServeSimRunSessionRemoteConfig {
                       previewUrl
+                      previewToken
```
szdziedzic, line 106: [P1] Format and sanitize gated preview data in JSON output too

Both `eas simulator --json` and `eas simulator get --json` serialize `remoteConfig` directly, bypassing `formatPreviewUrl`. A gated session therefore returns a bare `previewUrl`/`webPreviewUrl` that responds with 401, alongside the bearer token as a separate field.

Please produce a sanitized JSON remote config whose preview URL has passed through `formatPreviewUrl` and omit the standalone preview token. Alternatively, explicitly document and test JSON output as a separate low-level `{ url, token }` contract.

gwdp: Fixed, both now go through a sanitizer that folds the token into the preview URL and drops the standalone field 🙏 

Kept `toolsAuthToken` as-is since simulator:exec reads it

packages/eas-cli/src/simulator/utils.ts
gwdp: I like the idea! Not sure how applicable that would be with the current serve-sim `/api` provides-the-token approach, but maybe a read-only (no-exec, no /api) token access could be quite useful. 🤔

szdziedzic (approved):

PR #4325, opened 2026-08-31
[build-tools] Limit simulator recording bitrate

szdziedzic:
# Why

iOS Simulator recordings currently inherit `record-sim`'s 30 Mbps default. Long device run sessions can therefore produce multi-gigabyte artifacts that approach or exceed the single-part upload limit.

Companion WWW change: https://github.com/expo/universe/pull/30668

# How

Pass an explicit 5 Mbps bitrate for device run session recordings. A two-hour recording at that target is about 4.5 GB before MP4 container overhead, leaving headroom below the 4.995 GiB R2 limit while retaining a reasonable H.264 screen-recording bitrate. The standalone `record-sim` default is unchanged.

# Test Plan

CI

### DIFF
--- a/packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
+++ b/packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
@@ -16,6 +16,9 @@ const RECORD_SIM_FINISH_TIMEOUT_MS = 70_000;
 const RECORD_SIM_FORCE_STOP_TIMEOUT_MS = 5_000;
 const RECORD_SIM_MAX_ATTEMPTS_PER_BOOT = 3;
 const RECORD_SIM_COMMAND = 'record-sim';
+// A two-hour recording at this target is about 4.5 GB, leaving room below R2's
+// single-part upload limit for MP4 container overhead.
+const RECORD_SIM_BITRATE = '5000000';
 
 type IosSimulatorRecording = {
   id: string;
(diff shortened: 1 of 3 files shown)


gwdp (approved):

packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
```diff
@@ -16,6 +16,9 @@ const RECORD_SIM_FINISH_TIMEOUT_MS = 70_000;
 const RECORD_SIM_FORCE_STOP_TIMEOUT_MS = 5_000;
 const RECORD_SIM_MAX_ATTEMPTS_PER_BOOT = 3;
 const RECORD_SIM_COMMAND = 'record-sim';
+// A two-hour recording at this target is about 4.5 GB, leaving room below R2's
+// single-part upload limit for MP4 container overhead.
```
sjchmiela, line 20: this feels very handwavy cause if the screen does not change, a two hour recording should be under 10 MB

packages/build-tools/src/steps/utils/IosSimulatorRecordingUtils.ts
```diff
@@ -285,6 +288,25 @@ async function startIosSimulatorRecordingAsync(
   });
 }
 
+export function createRecordSimArgs({
```
sjchmiela, line 291: i don't think extracting to a separate function makes sense

sjchmiela (approved):

PR #4326, opened 2026-09-01
[build-tools] Use lower 980px Android preview resolution

krystofwoldrich-agent:
Android expo-device-hub previews use a 1280 px maximum dimension while iOS serve-sim previews use 960 px. Matching the lower cap reduces Android streaming bandwidth and improves the stream fps. (not stable 60 fps yet)


szdziedzic (approved):
One of the problems we had in serve sim is that we did multiple unneeded copies of whole high res pixel matrix. I'm not sure if we have some optimization to do here for it as well.

krystofwoldrich:
Yes, there are similar improvement to be made on the serve-emu side.

https://github.com/expo/serve-emu/pull/16

PR #4330, opened 2026-09-02
[worker] Add access-only agent runtimes

sjchmiela:
## Why

Agent jobs must run Claude Code and Codex without copying canonical provider refresh tokens into workers. This change adds worker-side adapters for the access-only lease contract from expo/universe#30717.

## What changed

- add a typed job-bound auth lease client with strict response validation
- add bounded Claude CLI and Codex exec adapters
- add a Codex app-server adapter with external auth, typed thread and turn handling, proactive and unauthorized refresh, cancellation, timeouts, and isolated temporary state
- pin and probe Claude Code 2.1.246 and Codex CLI 0.149.1 contracts
- add dynamic token redaction to all worker build log sinks

## Verification

- `yarn workspace @expo/build-tools jest-unit src/agent-runtime src/__tests__/context.test.ts --runInBand`
- `EAS_AGENT_RUNTIME_CONTRACT_TESTS=1 yarn workspace @expo/build-tools jest-integration src/agent-runtime/__integration-tests__/AgentRuntimeContracts.test.ts --runInBand`
- `yarn workspace @expo/worker test:unit src/__unit__/logger.test.ts src/__unit__/service.test.ts --runInBand`
- `yarn workspace @expo/worker typecheck`
- `yarn lint`
- `yarn fmt:check`

## Follow-up

The worker image must pin these exact CLI versions before the runtime is used. The Universe job entry point will then select an adapter and request its first lease after queueing.

### DIFF
--- a/packages/build-tools/src/agent-runtime/CodexExecRuntime.ts
+++ b/packages/build-tools/src/agent-runtime/CodexExecRuntime.ts
@@ -1,0 +1,87 @@
+import { errors } from '@expo/eas-build-job';
+import type { bunyan } from '@expo/logger';
+import type { SpawnResult } from '@expo/turtle-spawn';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+
+import { AgentAuthLeaseClient } from './AgentAuthLeaseClient';
+import {
+  assertAgentExecutableVersionAsync,
+  assertLeaseCoversInvocation,
+  runBoundedAgentProcessAsync,
+} from './processUtils';
+
+export const CODEX_VERSION = '0.149.1';
+
+const NON_REFRESHING_SENTINEL = 'eas-access-only-lease';
+
+export class CodexExecRuntime {
+  public constructor(
+    private readonly leaseClient: AgentAuthLeaseClient,
+    private readonly executable = 'codex'
+  ) {}
+
+  public async runAsync({
+    args,
+    cwd,
+    env,
+    logger,
+    maximumInvocationSeconds,
+    signal,
+  }: {
+    args: string[];
+    cwd?: string;
+    env: NodeJS.ProcessEnv;
+    logger: bunyan;
+    maximumInvocationSeconds: number;
+    signal?: AbortSignal;
+  }): Promise<SpawnResult> {
+    await assertAgentExecutableVersionAsync({
+      executable: this.executable,
+      expectedVersion: CODEX_VERSION,
+      displayName: 'Codex CLI',
+    });
+    const lease = await this.leaseClient.getLeaseAsync({ reason: 'startup' }, signal);
+    if (lease.provider !== 'openai') {
+      throw new errors.UserError(
+        'EAS_AGENT_PROVIDER_MISMATCH',
+        'This agent job is bound to a different provider. Select a Codex connection and start a new job.'
+      );
+    }
+    assertLeaseCoversInvocation(lease.expiresAt, maximumInvocationSeconds);
+
+    const codexHome = await fs.mkdtemp(path.join(os.tmpdir(), 'eas-codex-'));
+    try {
+      await fs.writeFile(
+        path.join(codexHome, 'auth.json'),
+        JSON.stringify({
+          auth_mode: 'chatgpt',
+          OPENAI_API_KEY: null,
+          tokens: {
+            id_token: lease.idToken,
+            access_token: lease.accessToken,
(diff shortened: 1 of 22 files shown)


sjchmiela:
## APJ-3 verification receipt

Head: `7518585a78760ab716d882d75ca3192570627db1`

- Unit suite: 22/22 passed.
- Exact-version contract suite: 2/2 passed with Claude Code 2.1.246 and Codex CLI 0.149.1.
- Typecheck passed.
- Real CLI live harness with local provider endpoints passed:
  - Claude completed a prompt with only `CLAUDE_CODE_OAUTH_TOKEN` and resumed the same session in a new process.
  - A too-short Claude lease was rejected before any provider request.
  - `codex exec` completed a prompt, used a mode-0600 access-only `auth.json`, and removed its temporary `CODEX_HOME`.
  - A too-short Codex exec lease was rejected before any provider request.
  - Codex app-server completed a real thread and turn through external ChatGPT auth, wrote no `auth.json`, and removed its temporary home.
  - A real backend 401 produced the sequence token A -> `unauthorized` lease request with the old fingerprint -> token B retry. Callback time was 3 ms.
  - Two real app-server processes received simultaneous 401 responses. Both retried with token B and the shared refresh manager counted exactly one provider refresh.
  - Proactive replacement completed on the same app-server process and the next turn used token B.
  - No access token or ID token appeared in captured runtime logs.
- Performance probe:
  - Claude adapter startup p95: 8.26 ms.
  - Codex exec adapter startup p95: 8.53 ms.
  - Full app-server token-replacement upper bound, including process start and turn, p95: 505.30 ms.
  - Actual 401 callback: 3 ms, below the 8-second worker limit.

Result: PASS for code, tests, live CLI behavior, concurrent refresh, redaction, and performance. The screenshot/video operator review gate remains open because macOS Computer Use blocks both Terminal and Codex app control in this environment.

packages/build-tools/src/agent-runtime/CodexExecRuntime.ts
```diff
@@ -0,0 +1,87 @@
...
+      );
+    }
+    assertLeaseCoversInvocation(lease.expiresAt, maximumInvocationSeconds);
+
+    const codexHome = await fs.mkdtemp(path.join(os.tmpdir(), 'eas-codex-'));
+    try {
+      await fs.writeFile(
+        path.join(codexHome, 'auth.json'),
+        JSON.stringify({
+          auth_mode: 'chatgpt',
```
hSATAC, line 59: I dug into how Codex handles this, and I think we can avoid the sentinel here.

With `auth_mode: 'chatgpt'`, Codex treats this as a normal login it owns and may try to refresh the token, both proactively near expiry and again after a 401. In our case that would end up trying to use `NON_REFRESHING_SENTINEL` as the refresh token.

There’s a `chatgptAuthTokens` mode for externally provided tokens, which skips refresh entirely. Codex’s own `AuthDotJson::from_external_access_token` uses that mode with an empty refresh token.

Would it make sense to use:

```ts
auth_mode: 'chatgptAuthTokens',
...
refresh_token: '',
```

and drop `NON_REFRESHING_SENTINEL`?

packages/build-tools/src/agent-runtime/CodexExecRuntime.ts
```diff
@@ -0,0 +1,87 @@
...
+            access_token: lease.accessToken,
+            refresh_token: NON_REFRESHING_SENTINEL,
+            account_id: lease.accountId,
+          },
+          last_refresh: new Date().toISOString(),
+        }),
+        { mode: 0o600 }
+      );
+      const processEnv: NodeJS.ProcessEnv = { ...env, CODEX_HOME: codexHome };
+      delete processEnv.OPENAI_API_KEY;
```
hSATAC, line 72: I think we should also clear `CODEX_API_KEY` and `CODEX_ACCESS_TOKEN` here.

Codex checks those env vars before `auth.json`, so if either is inherited from the job environment it could silently bypass the leased connection and run under a different identity.

Something like:

```ts
delete processEnv.OPENAI_API_KEY;
delete processEnv.CODEX_API_KEY;
delete processEnv.CODEX_ACCESS_TOKEN;
```

hSATAC (commented):
Not a full review yet — just noticed a couple of things while looking through this.

sjchmiela:
oh yea it was a slop pull request -- i asked codex to build all this independently it opened a bunch of pull requests here and there but i didn't get to polish or review them

PR #4332, opened 2026-09-03
[build-tools] Install FFmpeg before Android device hub

krystofwoldrich-agent:
## Why

`serve-emu` backend used in `expo-device-hub` uses `ffmpeg` for encoding the emulators stream collected using gRPC.

I'm looking into removing the `ffmpeg` dependency, but in the mean time, we need to ensure `ffmpeg` is installed on the host for the better smoother streaming performance we are getting from using gRPC.

## How
 
This PR reused the existing `ffmpeg` install function which was used in `argent` sessions to install `ffmpeg` before starting `expo-device-hub`.

### DIFF
--- a/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
+++ b/packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
@@ -836,12 +836,14 @@ export async function startServeSimWithTunnelAsync(
 export async function startExpoDeviceHubWithTunnelAsync(
   ctx: CustomBuildContext,
   {
+    runtimePlatform,
     baseDomain,
     env,
     logger,
     timeoutMs,
     packageVersion,
   }: {
+    runtimePlatform: BuildRuntimePlatform;
     baseDomain: string;
     env: BuildStepEnv;
     logger: bunyan;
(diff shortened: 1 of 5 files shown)


packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -333,22 +332,23 @@ async function installFfmpegWithAptAsync({
...
+ * without FFmpeg-dependent features.
  *
  * The whole body is wrapped because the caller runs this in the background with
  * `void`. There is no unhandledRejection handler in the worker, so a rejection
  * escaping here would crash the process and take the live session with it.
  * `spawn` is not an async function and can throw synchronously, which
  * `asyncResult` cannot catch — it only wraps an already-created promise.
  */
-export async function ensureFfmpegInstalledAsync({
+async function ensureFfmpegInstalledOnceAsync({
```
sjchmiela: this does not check "once"?

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -377,16 +377,41 @@ export async function ensureFfmpegInstalledAsync({
...
+  }
+
+  const setupPromise = ensureFfmpegInstalledOnceAsync({ runtimePlatform, env, logger });
+  ffmpegSetupPromise = setupPromise;
+  try {
+    await setupPromise;
+  } finally {
+    if (ffmpegSetupPromise === setupPromise) {
+      ffmpegSetupPromise = undefined;
+    }
```
sjchmiela, line 411: and what does this do?

sjchmiela (approved):

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
krystofwoldrich: Just a clean up so we don't hold the promise in memory after it's done installing.

packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
krystofwoldrich: Good point, I'll switch the names. So the *once* does the check and ensureFfmpegInstalledAsync is the internal helper.

PR #4334, opened 2026-09-03
[build-tools] Add custom log phase helper

sjchmiela:
# Why

I want a single EAS function to appear as running as two separate phases.

# How

Added `withLogPhaseAsync` which is going to create a logger child with necessary fields overwritten so a new log phase appears.

# Test Plan

CI should pass. Verified locally.

Zrzut ekranu 2026-09-4 o 12 17 25

### DIFF
--- a/packages/build-tools/src/utils/logPhase.ts
+++ b/packages/build-tools/src/utils/logPhase.ts
@@ -1,0 +1,30 @@
+import { BuildPhase, BuildPhaseResult, LogMarker } from '@expo/eas-build-job';
+import { type bunyan } from '@expo/logger';
+import { BuildStep } from '@expo/steps';
+
+export async function withLogPhaseAsync<T>(
+  logger: bunyan,
+  name: string,
+  fn: (logger: bunyan) => Promise<T>
+): Promise<T> {
+  const phaseLogger = logger.child({
+    phase: BuildPhase.CUSTOM,
+    buildStepId: BuildStep.getNewId(),
+    buildStepDisplayName: name,
+  });
+  phaseLogger.info({ marker: LogMarker.START_PHASE }, `Start phase: ${name}`);
+  try {
+    const result = await fn(phaseLogger);
+    phaseLogger.info(
+      { marker: LogMarker.END_PHASE, result: BuildPhaseResult.SUCCESS },
+      `End phase: ${name}`
+    );
+    return result;
+  } catch (error) {
+    phaseLogger.info(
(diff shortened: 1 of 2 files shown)


packages/build-tools/src/utils/logPhase.ts
```diff
@@ -0,0 +1,30 @@
...
+import { type bunyan } from '@expo/logger';
+import { BuildStep } from '@expo/steps';
+
+export async function withLogPhaseAsync<T>(
+  logger: bunyan,
+  name: string,
+  fn: (logger: bunyan) => Promise<T>
+): Promise<T> {
+  const phaseLogger = logger.child({
+    phase: BuildPhase.CUSTOM,
```
AHGIJMKLKKZNPJKQR, line 11: Why is the build phase `CUSTOM`? Is there something stopping us from making the sandbox phases that later use this predefined?

sjchmiela: the `BuildPhase` enum contains a list of all the "known" phases of builds

https://github.com/expo/eas-cli/blob/b0a4838c0db4edc1e613f830ad2615bcf7c22cc1/packages/eas-build-job/src/logs.ts#L1-L65

when you want to add a new known build phase, not only do you need to add a new `BuildPhase` enum value, but you also need to [set up the website id](https://github.com/expo/eas-cli/blob/b0a4838c0db4edc1e613f830ad2615bcf7c22cc1/packages/eas-build-job/src/logs.ts#L142-L201) (because for some reason we expect each phase only to exist once in a given build; not sure how it works on the workflow page actually), set up a name in [the display name mapping](https://github.com/expo/eas-cli/blob/b0a4838c0db4edc1e613f830ad2615bcf7c22cc1/packages/eas-build-job/src/logs.ts#L74-L133), publish a new release of `@expo/eas-build-job`, go to `website`, bump the package, deploy and ONLY THEN technically you can use the new build phase. until then the log phase is going to appear as the build phase value (`"START_SANDBOX"`)

i don't like this system and i think we should move away from it to phases which get named by the creator, when they're created, not through some arbitrary mapping. `BuildPhase.CUSTOM` is currently a way to achieve that https://github.com/expo/universe/blob/2ff6ee15e9867cee5613b9eb346da42f97252fb0/server/website/common/logs-utils.ts#L399-L400

this never has been a topmost priority though

AHGIJMKLKKZNPJKQR (approved):

PR #4338, opened 2026-09-03
[eas-cli] Align Observe code with new server schema

douglowder:
# Why

We need to adjust the Observe commands and queries to use the new GraphQL schema from the server.

# How

Rework all the commands to use the new schema. Results from commands using the `--json` flag will change slightly, with the `metricName` and `eventName` properties replaced by `name`.

# Test Plan

- CI and unit tests should pass.
- Tested locally against real EAS projects in production.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- [eas-cli] Align the `eas observe:*` commands with the renamed Observe GraphQL schema. `--json` output now uses `name`/`value` instead of `metricName`/`metricValue` (metrics) and `name` instead of `eventName` (log events and event-name summaries), and `eas observe:events` now lists user-defined events only — exceptions moved to the new `eas observe:errors`. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@douglowder](https://github.com/douglowder))
+
 ### 🎉 New features
 
 ### 🐛 Bug fixes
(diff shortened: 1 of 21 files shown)


packages/eas-cli/src/project/android/__tests__/fixtures/build/reports/problems/problems-report.html
kadikraman, line 1: I think this and the other files under `/project/android/` were committed accidentally.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🛠 Breaking changes
 
+- [eas-cli] Align the `eas observe:*` commands with the renamed Observe GraphQL schema. `--json` output now uses `name`/`value` instead of `metricName`/`metricValue` (metrics) and `name` instead of `eventName` (log events and event-name summaries), and `eas observe:events` now lists user-defined events only — exceptions moved to the new `eas observe:errors`. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@douglowder](https://github.com/douglowder))
```
kadikraman: "moved to the new `eas observe:errors`" is incorrect (this change is in the stacked PR)

kadikraman: Also need to fill in the PR number `4338`. Side-note: I think in expo/expo there's a PR check that parses the changelog and if you add an entry without attribution, it code comments the right PR and author to add. Maybe we can add the same thing to this repo so we wouldn't have to look it up separately all the time?

kadikraman (approved):
Tested against real data, LGTM!

CHANGELOG.md
douglowder: Thanks for the catch, and we should at least fail the CHANGELOG check if the PR number is not filled in.

PR #4339, opened 2026-09-03
[eas-cli] Add observe:errors command

douglowder:
# Why

A new command is needed to support the display of errors and exceptions in the CLI.

# How

Add `observe:errors` to display error and exception issue groups grouped by fingerprint. Exceptions are now excluded from `observe:events` and surfaced here instead.  Supports the following flags to filter output: `--platform`, `--severity` (fatal|error), time range flags, `--app-version`, `--build-number`, `--update-id`, and `--environment`.

# Test Plan

- New unit tests added
- Tested against real projects in EAS production

---

### DIFF
--- a/packages/eas-cli/src/commands/observe/errors.ts
+++ b/packages/eas-cli/src/commands/observe/errors.ts
@@ -1,0 +1,105 @@
+import { Flags } from '@oclif/core';
+
+import EasCommand from '../../commandUtils/EasCommand';
+import {
+  EasNonInteractiveAndJsonFlags,
+  resolveNonInteractiveAndJsonFlags,
+} from '../../commandUtils/flags';
+import { AppObserveErrorSeverity } from '../../graphql/generated';
+import Log from '../../log';
+import { fetchObserveErrorGroupsAsync } from '../../observe/fetchErrors';
+import {
+  ObserveAppVersionFlag,
+  ObserveBuildNumberFlag,
+  ObserveEnvironmentFlag,
+  ObservePlatformFlag,
+  ObserveProjectIdFlag,
+  ObserveTimeRangeFlags,
+  ObserveUpdateIdFlag,
+} from '../../observe/flags';
+import {
+  buildObserveErrorGroupsJson,
+  buildObserveErrorGroupsTable,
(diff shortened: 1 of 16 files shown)


packages/eas-cli/src/commands/observe/errors.ts
```diff
@@ -0,0 +1,156 @@
...
+import { resolveObserveCommandContextAsync } from '../../observe/resolveProjectContext';
+import { resolveTimeRange } from '../../observe/startAndEndTime';
+import { enableJsonOutput, printJsonOnlyOutput } from '../../utils/json';
+
+const SEVERITY_BY_FLAG: Record<string, AppObserveErrorSeverity> = {
+  fatal: AppObserveErrorSeverity.Fatal,
+  error: AppObserveErrorSeverity.Error,
+};
+
+const DEFAULT_OCCURRENCES_LIMIT = 5;
```
kadikraman: Should we limit to 10 like other observe commands?

packages/eas-cli/src/commands/observe/errors.ts
```diff
@@ -0,0 +1,156 @@
...
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  private static loggedInOnlyContextDefinition = {
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    const { flags } = await this.parse(ObserveErrors);
+    const { json, nonInteractive } = resolveNonInteractiveAndJsonFlags(flags);
```
kadikraman, line 85: The `--after` and `--limit` flags are only used when `--fingerprint` is supplied, else they are silently ignores. Could we error instead so we're more in line with what we do elsewhere (like how `observe:events` rejects `--all-events` with an event name)

douglowder: Fixed below.

packages/eas-cli/src/commands/observe/errors.ts
douglowder: Agreed

kadikraman (approved):

PR #4346, opened 2026-09-04
[eas-build-job] Add sandbox daemon protocol

sjchmiela:
## Why

"sandboxd" from `build-tools` needs to connect to the MCP server in `universe`. They need to be able to talk to each other. I'm thinking it might be nice if they shared an "API protocol"?

## How

Added Zod schema for input and output of to-be-supported messages to `eas-build-job`.

## Test plan

CI should pass. I think I tested this locally too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [eas-build-job] Add a shared protocol for sandbox daemon commands. ([#4346](https://github.com/expo/eas-cli/pull/4346) by [@sjchmiela](https://github.com/sjchmiela))
+
 ### 🐛 Bug fixes
 
 ### 🧹 Chores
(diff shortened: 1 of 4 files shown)


packages/eas-build-job/src/sandbox.ts
```diff
@@ -0,0 +1,83 @@
...
+  jsonrpc: z.literal('2.0'),
+  id: z.string(),
+  method: z.string(),
+  params: z.unknown().optional(),
+});
+
+export type SandboxDaemonRequest = z.output<typeof SandboxDaemonRequestZ>;
+
+export const SandboxDaemonResponseZ = z.union([
+  z.strictObject({ jsonrpc: z.literal('2.0'), id: z.string(), result: z.unknown() }),
```
hSATAC: Is it intentional to allow a response without either `result` or `error` here? Since `z.unknown()` accepts `undefined`, `{ jsonrpc: '2.0', id: 'x' }` passes this success branch even when `result` is missing. Just checking whether that's expected, or whether we'd want to require the key.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [eas-build-job] Add a shared protocol for sandbox daemon commands. ([#4346](https://github.com/expo/eas-cli/pull/4346) by [@sjchmiela](https://github.com/sjchmiela))
```
hSATAC: Small thing: this entry ended up under the already released `24.0.0` section. Could we move it to `main`?

hSATAC (approved):
Looks good overall! Left a question about the response schema and a small changelog note.

packages/eas-build-job/src/sandbox.ts
sjchmiela: Thanks for noticing — I thought `z.unknown()` requires a value! Added `.nonoptional()`.

PR #4347, opened 2026-09-04
[build-tools] Execute sandbox commands over MCP connection

sjchmiela:
## Why

Clients are telling sandboxes through MCP server what to do. Let's handle the requests!

## How

The endeavour starts from a (relatively, after all) complex task of starting commands and reading/writing to them.

Since commands may live longer than a tool call (request followed by a response), we need some kind of manager of running commands — Shell Session Manager. (Session is the running command.)

This is where most of the complexity lays, the rest is mostly a wrapper.

## Test plan

CI should pass. Tested manually.

### DIFF
--- a/packages/build-tools/src/steps/functions/startSandbox.ts
+++ b/packages/build-tools/src/steps/functions/startSandbox.ts
@@ -48,6 +48,7 @@ export function createStartSandboxBuildFunction(ctx: CustomBuildContext): BuildF
         reconnectDelayMs: RECONNECT_DELAY_MS,
         logger: stepCtx.logger,
         signal,
+        workingDirectory: stepCtx.workingDirectory,
       });
       try {
         await daemon.ready;
--- a/packages/build-tools/src/steps/utils/__tests__/sandboxDaemonCommands.test.ts
+++ b/packages/build-tools/src/steps/utils/__tests__/sandboxDaemonCommands.test.ts
@@ -1,0 +1,297 @@
+jest.unmock('fs');
+jest.unmock('fs/promises');
+jest.unmock('node:fs');
+jest.unmock('node:fs/promises');
+jest.unmock('@expo/logger');
+
+import {
+  type SandboxDaemonCommandResult,
+  type SandboxDaemonMethod,
+  SandboxDaemonResponseZ,
+} from '@expo/eas-build-job';
+import fs from 'node:fs/promises';
+import Log from '@expo/logger';
+import http from 'node:http';
+import { AddressInfo } from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+import { setTimeout as setTimeoutAsync } from 'node:timers/promises';
+import WebSocket, { WebSocketServer } from 'ws';
+
+import { startSandboxDaemonAsync } from '../sandboxDaemon';
+
+describe('sandbox daemon commands', () => {
+  let sendCommandAsync: (
+    method: SandboxDaemonMethod,
+    params: unknown
+  ) => Promise<SandboxDaemonCommandResult<SandboxDaemonMethod>>;
+  let stopAsync: () => Promise<void>;
+  let workingDirectory: string;
+
+  beforeEach(async () => {
+    workingDirectory = await fs.mkdtemp(path.join(os.tmpdir(), 'sandbox-command-executor-'));
+    ({ sendCommandAsync, stopAsync } = await startTestDaemonAsync(workingDirectory));
+  });
+
+  afterEach(async () => {
+    await stopAsync();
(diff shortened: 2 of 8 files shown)


packages/build-tools/src/steps/utils/shellSessionManager.ts
```diff
@@ -0,0 +1,282 @@
...
+    detached: true,
+  });
+  const session = createSession(child);
+  session.write = (chars: string): void => {
+    child.stdin.write(chars);
+  };
+  child.stdout.setEncoding('utf8');
+  child.stderr.setEncoding('utf8');
+  child.stdout.on('data', data => {
+    session.output += data;
```
hSATAC, line 171: Looks like the output keeps growing until it's read, with no limit. Wouldn't a command producing a lot of output eventually take down the worker? This could happen even during the initial call while we're waiting for the yield timeout.

packages/build-tools/src/steps/utils/shellSessionManager.ts
```diff
@@ -0,0 +1,282 @@
...
+function startPipeCommand({
+  command,
+  workingDirectory,
+}: {
+  command: string;
+  workingDirectory: string;
+}): CommandSession {
+  const child = spawn(command, {
+    cwd: workingDirectory,
+    env: process.env,
```
hSATAC: Is using `process.env` here intentional? I was expecting sandbox commands to get the env prepared by the earlier steps. Looks like they'd miss those vars and also get the worker vars that `getBuildEnv()` normally filters out.

packages/build-tools/src/steps/utils/__tests__/sandboxDaemonCommands.test.ts
```diff
@@ -0,0 +1,352 @@
...
+    await fs.chmod(directory, 0o700);
+  });
+
+  it.each([false, true])('stops command descendants when tty is %s', async tty => {
+    const started = await sendCommandAsync('execCommand', {
+      cmd: `node -e "const { spawn } = require('node:child_process'); const child = spawn(process.execPath, ['-e', 'setInterval(() => {}, 1000)'], { stdio: 'ignore' }); console.log(child.pid); setInterval(() => {}, 1000)"`,
+      tty,
+      yieldTimeMs: 100,
+    });
+    const ready = await readUntilAsync(started, output => /\d+\r?\n/.test(output));
```
hSATAC, line 231: I think this is why the two tests are timing out in CI. `console.log(child.pid)` can add ANSI color codes around the number, so there's a reset code before the newline and this regex never matches. The test would keep polling without reaching the shutdown check.

hSATAC (changes requested):
Left a couple of questions about output buffering and the command environment, plus a note on the CI timeouts.

packages/build-tools/src/steps/utils/shellSessionManager.ts
sjchmiela: yes. added an inline comment. i think as a follow up we might redirect data to filesystem or something like that

packages/build-tools/src/steps/utils/shellSessionManager.ts
sjchmiela: it wasn't! thank you for finding it

packages/build-tools/src/steps/utils/__tests__/sandboxDaemonCommands.test.ts
sjchmiela: fixed

packages/build-tools/src/steps/functions/startSandbox.ts
```diff
@@ -48,6 +48,8 @@ export function createStartSandboxBuildFunction(ctx: CustomBuildContext): BuildF
         reconnectDelayMs: RECONNECT_DELAY_MS,
         logger: stepCtx.logger,
         signal,
+        workingDirectory: stepCtx.workingDirectory,
+        env,
```
hSATAC: One question: since www passes `__EAS_SANDBOX_MCP_TOKEN` as an environment secret, looks like sandbox commands will now get the daemon's MCP token too. Is that expected? I guess it's still within the same sandbox's scope, but wanted to check.

hSATAC (approved):
Thanks for the fixes! Just one non-blocking question.

packages/build-tools/src/steps/functions/startSandbox.ts
sjchmiela: Yea, they don't need it. Removed!

(Technically it doesn't prevent reading the memory by commands to find it, but you're right they shouldn't expect to have access to it.)

PR #4349, opened 2026-09-04
[build-tools] Allow configuring Android emulator display

krystofwoldrich-agent:
# Why

On a large Linux runner, Pixel 9 rendered 2.618M pixels (1080×2424) at about 30–35 FPS. Medium Phone renders 1.152M pixels (720×1600), 1.466M fewer pixels (~56%), and reaches 50–60 FPS. Medium Linux runners deliver roughly half these frame rates.

# How

Expose lcd_width, lcd_height, and lcd_density inputs for eas/start_android_emulator. So device sessions can set the resolution and density.

# Test Plan

The videos in the comment below show that the emulator runs fast enough to produce 60 fps on large running in the browser `testufo.com` while streaming using `expo-device-hub`. Ofc depending on the running app workload the produced fps can drop.

Focused tests, typecheck, lint, and format check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [build-tools] Add `lcd_width`, `lcd_height`, and `lcd_density` inputs to configure the Android emulator display. ([#4349](https://github.com/expo/eas-cli/pull/4349) by [@krystofwoldrich-agent](https://github.com/krystofwoldrich-agent))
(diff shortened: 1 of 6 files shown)


krystofwoldrich-agent:
Performance samples:

Medium Linux runner (~40 FPS):

https://github.com/user-attachments/assets/8250d51b-a35f-48d2-b3d3-74d285cdb921

Large Linux runner (~60 FPS):

https://github.com/user-attachments/assets/b8a317f9-14c0-49b7-92fd-22a5550602f0

sjchmiela (approved):
previously i advised people to use `ANDROID_EMULATOR_EXTRA_CONFIG` env var for this but i think it might be nice to have this as first class params

Issue #4273, opened 2026-08-24
integrations:posthog:connect creates an org and project before any confirmation, with no way to use an existing PostHog project


designorant:
### Summary

Two related gaps in `eas integrations:posthog:connect`:

1. It creates a PostHog organization and project **before any confirmation**. The only prompt beforehand is region selection.
2. There is no way to point the connection at a **pre-existing** PostHog project.

Together these mean a developer who already uses PostHog, and whose PostHog account email differs from their Expo account email, gets a brand-new PostHog account, organization, and project created for them — with no opportunity to say "use the project I already have," and no chance to decline before the resources exist.

To be clear about scope: the command **does** prompt before overwriting `EXPO_PUBLIC_POSTHOG_*` in `.env.local` and in EAS, so existing configuration is not silently replaced. The problem is confined to the provisioning step, which happens earlier and unconditionally.

### Why it happens

In `build/commands/integrations/posthog/connect.js` (eas-cli 22.3.0), the project mutation accepts no project selector:

```js
project = await PostHogMutation.setupPostHogProjectAsync(graphqlClient, {
  appId: projectId,
  posthogOrganizationConnectionId: connection.id,
})
```

and the lookup only finds a project EAS previously created for that app:

```js
let project = await PostHogQuery.getPostHogProjectByAppIdAsync(graphqlClient, projectId)
```

There is an org-level reuse path, but it is gated on email identity:

```js
throw new Error(`The email on the ${account.name} account already has a PostHog account,
  which must be connected by approving it in a browser. ...`)
```

If the Expo account's email has no PostHog account, the server takes the create branch. So a developer whose PostHog login differs from their Expo login always gets new resources. Even the browser-approval branch reuses only the **organization** — the project is still newly created.

### Managed or bare?

Bare — native directories are committed. Not relevant to the failure; the command never reaches native code.

### Environment

- eas-cli 22.3.0, darwin-arm64, node 22.22.0
- Expo SDK 54, pnpm workspace monorepo
- PostHog EU Cloud

### Reproducible steps to reproduce from a blank project

1. Create a PostHog project manually and note that its account email differs from your Expo account email.
2. Run `eas integrations:posthog:connect` and select a region.

Observed:

```
✔ Select a PostHog region › European Union (EU)
✔ Created PostHog organization <account>-XXXXXXXX
✔ Created PostHog project <App Name>
? PostHog features to set up ›
```

[...]

PR #4350, opened 2026-09-04
[eas-cli] Confirm before creating a PostHog organization

LizunovSergey:
# Why

Fixes the first half of #4273.

`eas integrations:posthog:connect` creates a PostHog organization and project as soon as a region is picked. Region selection is the only prompt in front of that, so the first opportunity to stop is the feature prompt that comes *after* — by which point the resources already exist.

That matters more than a typical missing confirmation because the CLI cannot undo it. `eas integrations:posthog:disconnect` removes only the EAS-side link record; the organization then has to be deleted in the PostHog dashboard, under an account auto-created for the Expo account's email — which, as the reporter points out, the developer may not realize exists or have credentials for.

Scope note: #4273 raises two things, and this PR is deliberately only the first. The second — pointing the connection at a **pre-existing** PostHog project — is not fixable from the CLI: `setupPostHogProjectAsync` takes `appId` and `posthogOrganizationConnectionId` and no project selector, so choosing an existing project needs a server-side field first. The reporter wrote that either change would resolve the issue; this is the one that can land in this repo. Happy to follow up on the other if the mutation grows a selector.

# How

One confirmation immediately before the create path, in the branch that runs only when there is no organization to reuse:

```ts
} else {
  const region = await this.resolveRegionAsync(regionFlag, nonInteractive);
  if (!nonInteractive) {
    const confirmed = await confirmAsync({
      message: `The ${chalk.bold(account.name)} account has no PostHog organization. Create a new PostHog organization and project in ${region}?`,
    });
    if (!confirmed) {
      Log.log('Nothing was created. …');
      return;
    }
  }
  connection = await this.startConnectionAsync(...);
}
```

Three things this deliberately does **not** change:

- **Non-interactive runs behave exactly as before.** The prompt is skipped when `nonInteractive` is set, so scripted and CI use is untouched. `--json` implies `nonInteractive` via `resolveNonInteractiveAndJsonFlags`, so JSON output is unaffected too.
- **Reusing an existing organization does not prompt.** Nothing is created on that path, so there is nothing to confirm.
- **The declined path stops before the mutation** rather than creating and cleaning up, since cleanup is exactly what is impossible here.

The prompt sits after the region is resolved so the message can name the region, which is the part that cannot be changed later.

# Test Plan

Four tests added to the existing `connect.test.ts`; the file's `beforeEach` already defaults `confirmAsync` to `true` and to reusing an existing connection, so no existing test needed changing.

```
yarn jest src/commands/integrations/posthog/
Test Suites: 3 passed, 3 total
Tests:       62 passed, 62 total
```

That is 40 pre-existing tests in `connect.test.ts` still green, plus the 4 new ones:

| test | pins |
|---|---|
| `confirms before creating a new organization and project` | the prompt fires on the create path, and provisioning still proceeds on yes |
| `creates nothing when the confirmation is declined` | no org mutation, no project mutation, no install, no file writes |
| `does not confirm when reusing an existing organization` | the reuse path is untouched |
| `provisions non-interactively without a confirmation prompt` | `--non-interactive` behaviour is unchanged |

Each new test was checked against a mutant rather than assumed to discriminate:

| mutant | failing test |
|---|---|
| confirmation block deleted | `confirms before creating…` and `creates nothing when…declined` |
| `if (!nonInteractive)` → always prompt | `provisions non-interactively without a confirmation prompt` |

Repo gates on this branch: `yarn typecheck` clean, `yarn lint` — 0 warnings, 0 errors, `yarn fmt:check` — all matched files correctly formatted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This is the log of notable changes to EAS CLI and related packages.
 - [build-tools] Reduce `expo-device-hub` preview resolution from 1280 px to 960 px to match `serve-sim` and lower streaming bandwidth. ([#4326](https://github.com/expo/eas-cli/pull/4326) by [@krystofwoldrich-agent](https://github.com/krystofwoldrich-agent))
 - [build-tools] Install FFmpeg before launching `expo-device-hub` in Android device sessions. ([#4332](https://github.com/expo/eas-cli/pull/4332) by [@krystofwoldrich-agent](https://github.com/krystofwoldrich-agent))
 - [eas-cli] Print the session token in simulator preview links, so they still open once the preview is gated. ([#4312](https://github.com/expo/eas-cli/pull/4312) by [@gwdp](https://github.com/gwdp))
+- [eas-cli] Confirm before `eas integrations:posthog:connect` creates a PostHog organization and project, since the CLI cannot remove them afterwards. ([#4350](https://github.com/expo/eas-cli/pull/4350) by [@LizunovSergey](https://github.com/LizunovSergey))
 
 ### 🧹 Chores
 
--- a/packages/eas-cli/src/commands/integrations/posthog/__tests__/connect.test.ts
+++ b/packages/eas-cli/src/commands/integrations/posthog/__tests__/connect.test.ts
@@ -536,12 +536,49 @@ describe(IntegrationsPostHogConnect, () => {
     expect(fs.writeFile).not.toHaveBeenCalled();
   });
 
-  it('does not confirm when reusing an existing organization', async () => {
+  it('does not ask about an organization when one is reused', async () => {
     jest.mocked(PostHogQuery.getPostHogProjectByAppIdAsync).mockResolvedValue(null);
 
     await createCommand([]).runAsync();
 
     expect(confirmAsync).not.toHaveBeenCalledWith({
+      message: expect.stringContaining('Create a new PostHog organization and project'),
+    });
+    expect(PostHogMutation.setupPostHogProjectAsync).toHaveBeenCalled();
+  });
+
+  it('confirms before creating a project when the organization is reused', async () => {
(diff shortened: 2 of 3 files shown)


LizunovSergey:
Correction to my own description above, and a follow-up commit (`2ea3bd1a`).

I wrote that reusing an existing organization needs no prompt because "nothing is created on that path". **That was wrong.** When an organization is reused but the app has no PostHog project yet, execution falls straight through to `setupPostHogProjectAsync` and creates one — unprompted. It is the case the issue names when it says the browser-approval branch "reuses only the **organization** — the project is still newly created", and my first commit did not cover it.

So the same confirmation now guards project creation:

```ts
if (!nonInteractive && !confirmedProvisioning) {
  const confirmed = await confirmAsync({
    message: `There is no PostHog project for this app in ${connection.posthogOrganizationName}. Create one?`,
  });
  if (!confirmed) {
    Log.log('Nothing was created.');
    return;
  }
}
```

`confirmedProvisioning` is set when the organization prompt has already run, so a run that creates both still asks exactly **once** — the first prompt already says "organization and project".

Non-interactive behaviour is still unchanged on both paths.

Tests are now 47 in `connect.test.ts` (65 across the posthog suites), with three added for this path, each verified against a mutant:

| mutant | failing test |
|---|---|
| `!confirmedProvisioning` → `confirmedProvisioning` (drops the reuse-path prompt) | `confirms before creating a project when the organization is reused`, `creates nothing when the project confirmation is declined`, `asks once, not twice…` |
| `confirmedProvisioning = true` never assigned (would double-prompt) | `asks once, not twice, when it creates both the organization and the project` |

`yarn typecheck`, `yarn lint` (0/0) and `yarn fmt:check` clean.

Thanks for bearing with the extra round — I found this by re-checking the claim in my own description against the code rather than against the tests, which all passed either way.

PR #4351, opened 2026-09-04
[eas-cli] Use large runners for Android device sessions

krystofwoldrich-agent:
# Why

Android FPS is roughly halved on medium Linux runners. This complements #4349, which reduces the emulator pixel load.

# How

Default Android device sessions to the large resource class. Explicit overrides still work and iOS behavior is unchanged.

# Test Plan

Simulator command tests, build, typecheck, lint, and format check.


krystofwoldrich:
We need to change this in the API, not CLI.

PR #4352, opened 2026-09-04
[eas-cli] print the preview page link instead of the tunnel url

gwdp:
## Why

The worker now reports the expo.dev preview page as a session's preview URL and the tunnel separately as previewApiUrl ([expo/eas-cli#4353](https://github.com/expo/eas-cli/issues/4353)). The CLI prints and returns both, so it has to put the token where each one expects it.

## How
- printed preview link puts the token in the fragment
- `--json` returns both URLs with the token applied and drops the standalone token field

Deploy order: expo/universe#30946 + expo/universe#30817 -> expo/serve-sim#143 -> expo/eas-cli#4353 + expo/eas-cli#4352

## Test Plan

CI + local e2e (full stack running locally and testing all the links/outputs and iframe w/ token)


packages/eas-cli/src/simulator/utils.ts
```diff
@@ -8,6 +9,25 @@ import { link } from '../log';
 type DeviceRunSessionByIdResult = DeviceRunSessionByIdQuery['deviceRunSessions']['byId'];
 export type DeviceRunSessionRemoteConfig = NonNullable<DeviceRunSessionByIdResult['remoteConfig']>;
 
+const PREVIEW_HOST = /^https:\/\/web-preview-([^./]+)\./;
```
szdziedzic: What if we ever change it on backend? I don't think that EAS CLI impl should care about URL format we have on backend. I believe we should just save correct webPreviewUrl in our WWW backend. It's also consumed by REST API outputs, so there are more places where we would need to handle it. So I think it's best to just report correct value from worker to backend. If it will break using serve-sim API on website, we should alter our configs to allow for serve sim API URL and web preview URL. WDYT?

szdziedzic (changes requested):

packages/eas-cli/src/simulator/utils.ts
gwdp: Good call, reworked the stack on this line. Worker reports the page as previewUrl, tunnel goes in a new previewApiUrl (universe#30946), CLI just prints what it gets. Let me know what do you think

szdziedzic (approved):

PR #4353, opened 2026-09-05
[build-tools] print the preview page link instead of the tunnel url

gwdp:
## Why

`previewUrl` is the raw ngrok tunnel, so opening a link while the session is still booting or after it ended lands on an ngrok error page. The expo.dev preview page handles both states, which is why we are changing to point there. 

## How
- previewUrl/webPreviewUrl becomes `https://expo.dev/simulator-preview/:tunnel_id`
- Tunnel itself goes in `previewApiUrl`
- Website origin is picked per stage, the same way the realtime logs URL already does. Considered adding a new env for this but havent done.
  - serve-sim gets `--frame-ancestor` with that origin, so the preview page is allowed to frame a gated preview

Deploy order: universe#30946+universe#30817 -> serve-sim#143 -> eas-cli#4353+eas-cli#4352

Test Plan

- CI + local e2e (full stack running locally and testing all the links/outputs and iframe w/ token)


packages/build-tools/src/steps/utils/remoteDeviceRunSession.ts
```diff
@@ -814,6 +825,7 @@ async function startWebPreviewWithTunnelAsync(
       authtoken: getNgrokAuthtokenOrThrow(env),
       logger,
     });
+    logger.info(`Web preview URL: ${simulatorPreviewUrl(tunnel.url, env)}`);
     return {
       previewUrl: tunnel.url,
```
szdziedzic: should we alter what we return here and generally what we report as preview URL to our API? AFAIK this API `webPreviewUrl` value is consumed in many places (EAS CLI, REST API, Webiste for mem/cpu analytics)

gwdp: Yes and thanks for pushing on it! Reworked things a bit to report previewUrl and previewApi, so metrics and the preview page liveness check keep working without any hacks.

szdziedzic (approved):

PR #4367, opened 2026-09-08
[eas-cli] add eas workflow:insights command

hSATAC:
# Why

Workflows Insights is only available on the expo.dev dashboard today. This brings it to EAS CLI so developers and agents can get to the same data more easily: in the terminal, or as JSON for scripts and automation.

Linear: [ENG-26599](https://linear.app/expo/issue/ENG-26599). Maestro insights (`eas workflow:insights:maestro`) follows in a separate PR.

# How

`eas workflow:insights` follows the shape of `eas update:insights` and `eas channel:insights`: the same `--days` / `--start` / `--end` flags with a 7-day default, `--json` for scripts, and human output built with `formatFields` and `renderTextTable`.

- One request to the existing `App.workflowsInsights` fields (overview metrics, runs over time, workflows connection). The server returns current and previous-period counts, so the CLI computes the success rate and the deltas.
- Filters mirror the dashboard: `--workflow` (file name, resolved to an ID because the server filters by ID), `--status` (the three finished statuses), `--trigger`, and `--git-ref` (a bare branch name becomes `refs/heads/<name>`). `--trigger` accepts every trigger a customer's run can carry; the dashboard's filter currently lists only nine of them, which is a gap on the dashboard side. The overview shows the dashboard's four tiles (total runs, success rate, active workflows, failed runs).
- `--project-id` lets it run outside a project directory, for cron jobs with a robot token.
- Plan-gate rejections are rethrown with the server's message plus the plan's day limit when the server reports it.

## The time range is widened to whole buckets before it is sent

The server widens only the `runsOverTime` query to whole buckets and leaves `overviewMetrics` and `workflows` on the exact bounds it was given. A window that is not already on a bucket boundary therefore makes those sections count different runs, and the bucket totals in the table can add up to more than the overview's total. The dashboard avoids this by widening the selected range client-side before it issues its queries, and the server's plan-limit check already grants a day of slack for exactly that.

So the CLI widens the window itself and sends the widened one, which makes the server's own widening a no-op and leaves every section covering the same period.

Widening lengthens the window, which can push it past the threshold of the bucket size it was widened by, so the bucket size is chosen by checking the widened window rather than the requested one. Without that, `--days 4` would ask for hourly buckets over a 97-hour window and the server would reject the request. Because the window no longer matches the `--days N` that was asked for, the `Time range` row and the `--json` timespan now report the bounds actually queried instead of "last N days".

One consequence worth naming: widening the end past now puts unelapsed time in the current period while the previous period it is compared against is complete, so the count metrics span unequal exposure. The dashboard behaves the same way, so this keeps one interpretation of the numbers rather than introducing a second.

## Other changes

- `enableJsonOutput()` now runs before the time range and the project context are resolved. Both can write to stdout — a time-range error goes through `Log.error`, and the logged-in context warns when there is no session — and `enableJsonOutput` works by swapping `process.stdout.write`, so anything logged before it corrupts `--json` output.
- A metric's change is shown in parentheses (`100 (+25.0%)`), matching how the observe formatters pair a statistic with its secondary figure. The previous two-space form read as a column that `formatFields` cannot pad, since it pads only the label.

Three shared helpers also change, all called out in the changelog:

- `resolveTimeRange` (observe) now rejects `--end` without `--start`. Before, `--end` alone was silently ignored and the default window was used.
- `resolveTimeRange` also rejects a range that does not run forwards, which is reachable by swapping the two flags or by passing a `--start` in the future. Widening would otherwise turn a backwards range into a valid one-bucket query rather than letting the server reject it.
- The shared `--limit` / `--offset` parser now rejects non-integers such as `1.5`, which it used to forward to the server where `Int!` arguments fail.

Both `resolveTimeRange` changes apply to the `observe:*` commands and the update/channel insights commands as well.

# Test Plan

- `yarn typecheck`, `yarn lint`, `yarn fmt:check`, `yarn lint-changelog`, and the tests for the new and touched areas pass locally (43 suites, 570 tests).
- New tests cover the command's flags and context resolution, filter resolution and git-ref normalization, the formatter (summary, JSON, table, empty state), granularity thresholds, plan-gate handling, and the integer parser.
- The widening has its own tests for a window that already sits on bucket boundaries, a window that is widened, a window whose widening crosses a granularity threshold and has to drop to a coarser bucket, an unaligned single day, and the fixed-point property that the server widening the result again changes nothing.
- At command level, `--days 4` on a clock that is not on the hour is asserted to send whole-day bounds with `DAY` granularity, which is the case that would otherwise be rejected by the server, and a backwards explicit range is asserted to fail after `enableJsonOutput()` and before the project context is resolved.
- `resolveTimeRange` has tests for both rejections (a range that ends before it starts, and a zero-length range).
- Manual run against production with `--project-id` on a project with runs in the window: the table renders the header, the overview with its deltas, the runs-over-time rows, and the workflows table; `--json --non-interactive --status FAILURE --limit 3` returns `project`, `timespan`, `filters`, `overview`, `runsOverTime`, `workflows`, and `hasMoreWorkflows`.
- `--days 400` returns the server's plan message with the limit appended: "The selected timeframe exceeds the supported range for workflow insights. Please choose a shorter window. Your plan includes the last 365 days."
- `--end 2026-09-01` alone returns "--end requires --start. Pass both, or use --days instead."
- `--limit 1.5` returns "Unable to parse 1.5 as an integer" before any request is made.


hSATAC:
Hey @kadikraman , this follows the same patterns as your `update:insights` / `channel:insights` commands. I’m planning to add `eas workflow:insights:maestro` next.

I also made two changes to shared helpers that affect your commands and wanted to check with you:

* `resolveTimeRange` now rejects `--end` without `--start`, which currently falls back to the default time range. This also affects `update:view --insights` and the `observe:*` commands.
* The shared `--limit` / `--offset` parser now rejects non-integers like `1.5` before they reach the server and fail the GraphQL query.

I put these under bug fixes in the changelog, but I wasn’t sure whether accepting `--end` on its own was intentional. Do you think this should count as a breaking change, or would you prefer keeping the validation in the insights layer? Happy to adjust.

When you have a chance, could you take a look at those two changes? No need to review the whole PR. Thanks!

packages/eas-cli/src/commands/workflow/insights.ts
```diff
@@ -0,0 +1,164 @@
...
+      Log.addNewLineIfNone();
+      Log.log(buildWorkflowsInsightsTable(summary));
+    }
+  }
+
+  /**
+   * With `--project-id` the command runs outside a project directory, so only the
+   * login context is needed.
+   */
+  private async resolveProjectContextAsync(
```
kadikraman: This looks very similar to `packages/eas-cli/src/observe/resolveProjectContext.ts` - could we make it generic and pull to a shared util?

kadikraman (commented):
Nice one, this looks really good! A couple of suggestions based on testing:

### The workflow name from `workflow:insights` doesn't work

It needs the workflow filename including extension.

`easd workflow:insights --project-id 647dc2fd-5a2b-456f-b450-9eb9b119ca51` shows a workflow called `test`

Image

`easd workflow:insights --project-id 647dc2fd-5a2b-456f-b450-9eb9b119ca51 --workflow test` errors.

It looks like these can also differ in that multiple workflows in the list will have the same "name". Maybe the table should show the filenames instead or workflow names?

### Empty rows are all written out

`easd workflow:insights --days 60`

E.g. here I don't have any workflow runs at all so they're all zeros.

Image

Maybe instead when we see that it's all 0 we could skip the identical rows for readability, like print only 2 or 3 and then `  ... xx more identical rows ...`

hSATAC:
@kadikraman Thanks for the review and good catch! Will fix accordingly.

packages/eas-cli/src/workflow/insights/planGating.ts
```diff
@@ -0,0 +1,37 @@
...
+export const EAS_WORKFLOWS_INSIGHTS_TIMESPAN_LIMIT_EXCEEDED_ERROR_CODE =
+  'EAS_WORKFLOWS_INSIGHTS_TIMESPAN_LIMIT_EXCEEDED';
+
+const PLAN_GATE_ERROR_CODES: ReadonlySet<string> = new Set([
+  EAS_WORKFLOWS_INSIGHTS_NOT_AVAILABLE_ERROR_CODE,
+  EAS_WORKFLOWS_INSIGHTS_TIMESPAN_LIMIT_EXCEEDED_ERROR_CODE,
+]);
+
+/** Plan-gate rejections become a plain error message instead of a raw GraphQL error. */
+export async function withWorkflowsInsightsPlanGateHandlingAsync<T>(
```
sjchmiela, line 16: is this helpful (i.e. what does the regular error handling lack that we need to special-handle plan gate error)?

packages/eas-cli/src/workflow/insights/formatInsights.ts
```diff
@@ -0,0 +1,365 @@
...
+  if (metric.previous === 0) {
+    return chalk.dim('n/a');
+  }
+  const pct = ((metric.current - metric.previous) / metric.previous) * 100;
+  return formatSignedChange(pct, '%', lowerIsBetter ? pct < 0 : pct > 0);
+}
+
+function formatSignedChange(value: number, unit: string, isImprovement: boolean): string {
+  if (value === 0) {
+    return chalk.dim(`0.0${unit}`);
```
sjchmiela, line 326:
```suggestion
    return chalk.dim(`0.0 ${unit}`);
```
same in other instances?

sjchmiela: ah i see `' pts'` has the extra space

packages/eas-cli/src/workflow/insights/formatInsights.ts
```diff
@@ -0,0 +1,365 @@
...
+    parts.push(`git ref: ${filters.gitRef}`);
+  }
+  return parts.join('; ');
+}
+
+function formatCount(
+  metric: WorkflowsInsightsMetricSummary,
+  options?: { lowerIsBetter?: boolean }
+): string {
+  return `${metric.current.toLocaleString()}  ${formatTrend(metric, options)}`;
```
sjchmiela: two spaces?

packages/eas-cli/src/commands/workflow/insights.ts
```diff
@@ -0,0 +1,137 @@
...
+      projectId,
+      loggedIn: { graphqlClient },
+    } = await this.getContextAsync(WorkflowInsights, {
+      nonInteractive,
+      projectIdOverride: flags['project-id'],
+    });
+
+    if (json) {
+      enableJsonOutput();
+    }
```
sjchmiela: should we move this earlier?

packages/eas-cli/src/commands/workflow/insights.ts
```diff
@@ -0,0 +1,137 @@
...
+    ...this.ContextOptions.ProjectId,
+    ...this.ContextOptions.LoggedIn,
+  };
+
+  async runAsync(): Promise<void> {
+    const { flags } = await this.parse(WorkflowInsights);
+    const { json, nonInteractive } = resolveNonInteractiveAndJsonFlags(flags);
+
+    const timespan = resolveInsightsTimeRange(flags);
+    const granularity = granularityForTimespan(timespan.startTime, timespan.endTime);
```
sjchmiela: (AI!)

Could we use the same effective time range for the overview and the runs-over-time table?

`resolveInsightsTimeRange` returns an exact rolling window, but the server's `getRunsOverTimeAsync` expands its SQL bounds to full buckets with `getAlignedTimespan`. The overview and workflows queries retain the original bounds. Unlike this command, the dashboard aligns the selected range before sending all three queries.

For example, at September 8, 12:00 UTC, the default window starts September 1, 12:00 UTC, but the daily query starts September 1, 00:00 UTC. With a failed run at September 1, 08:00 and a successful run at September 5, 16:00, the overview reports one run and zero failures, while the daily table reports two runs and one failure. I reproduced this with the server query builder and synthetic rows.

Could we align the range consistently for all sections and report that effective range, or keep the exact requested SQL bounds in the server's chart query? A test with a run just before the requested start would cover this.

sjchmiela (approved):

packages/eas-cli/src/commands/workflow/insights.ts
hSATAC: Good catch, let me think how to address this. I think we should mirror web behavior as much as possible.

PR #4369, opened 2026-09-09
[eas-cli] add --egress-allow to reach local dev servers through local egress

brentvatne:
# Why

The local-egress proxy blocks private/local destinations by default. Developers need explicit access to dev servers such as `localhost:3000`. This PR is stacked on https://github.com/expo/eas-cli/pull/4364.

# How

- Adds repeatable `--egress-allow host:port`, requiring `--egress local`. Entries are normalized and deduplicated; wildcards, ranges, and malformed values are rejected before session creation.
- Exceptions match the exact host and port. Allowing `localhost:3000` does not allow another port or the separately named `127.0.0.1`. Unlisted destinations retain the default policy.
- Persists exceptions in session configuration and displays them when starting or reconnecting. All iOS controllers are supported through the base PR, including Argent and Appium; tests cover exception propagation and stale-value clearing across all five remoteConfig variants.
- Forwards loopback entries like `adb reverse`. iOS never sends `127.0.0.1` requests to the system proxy, and Expo CLI rewrites `localhost` to `127.0.0.1` in manifests, so a proxy-only allow entry could not serve an Expo Go or dev client bundle. For each `localhost:<port>` or `127.0.0.1:<port>` entry, the egress client opens one chisel reverse remote so `127.0.0.1:<port>` on the simulator host reaches the same port on this machine. Privileged ports and the proxy port stay name-only. The base PR’s worker permits these remotes; a rejected remote is surfaced with a hint instead of looping silently. Reverse-remote traffic does not pass through the proxy’s per-request check or logging: the entry itself is the decision, and the forward can only reach that one port.

# Test Plan

- Local validation: 196 CLI tests across the simulator suites pass, including parsing, exact matching, live proxy requests, controller configuration, dotenv/environment isolation, the loopback forward plan, and the chisel remote arguments. CLI typecheck, lint, and format checks pass.
- Real-worker validation on staging (worker built from the base branch, Expo Go session, agent-device controller, Metro on port 8082 on the developer machine):
  - With `--egress-allow localhost:8082` only, `exp://localhost:8082` in Expo Go fetched the manifest through the proxy and loaded the bundle from `127.0.0.1:8082` through the forwarded port. Metro logged the bundle build and the app rendered.
  - Fast Refresh works: with Metro in watch mode, editing `src/app/index.tsx` updated the running app in the simulator within seconds and without a manual reload. The dev server WebSocket rides the forwarded loopback port like the bundle does.
  - Safari in the simulator reached `http://127.0.0.1:8082/status` on the developer machine. `http://127.0.0.1:8081`, a port the developer machine was also serving but had not allowed, stayed unreachable.
  - Before the forward existed, the same `localhost:8082` entry loaded the manifest but the bundle request to `127.0.0.1:8082` never reached the proxy and failed on the worker’s loopback, which is the failure this change fixes.
  - Unlisted private destinations such as `localhost:8081` and a LAN address on an unlisted port were refused with the documented message. Proxied requests failed while the egress client was stopped and recovered when it restarted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [eas-cli] Add `--egress-allow <host:port>` to `eas simulator:start` so local egress sessions can reach dev servers on this machine or its network. ([#4369](https://github.com/expo/eas-cli/pull/4369) by [@brentvatne](https://github.com/brentvatne))
 - [build-tools] Add the `eas/start_local_egress` step for iOS simulator sessions with local egress: the device host's system proxy points at a reverse tunnel to the EAS CLI machine, so HTTP(S) requests that honor it exit from that machine and third parties see its IP. ([#4364](https://github.com/expo/eas-cli/pull/4364) by [@brentvatne](https://github.com/brentvatne))
 - [eas-cli] Add `--egress local` to `eas simulator:start` and the `eas simulator:egress` command, which run the local egress client for the session. ([#4364](https://github.com/expo/eas-cli/pull/4364) by [@brentvatne](https://github.com/brentvatne))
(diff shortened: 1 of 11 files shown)


szdziedzic (approved):
Can we split

PR #4370, opened 2026-09-09
[eas-cli] add eas workflow:insights:maestro command

hSATAC:
# Why

Maestro insights is only available on the expo.dev dashboard today. This brings it to EAS CLI next to `eas workflow:insights`, so developers and agents can read the same data from the terminal or as JSON.

Linear: [ENG-26601](https://linear.app/expo/issue/ENG-26601). Stacked on #4367; review this PR against that branch.

# How

`eas workflow:insights:maestro` is a subcommand of `eas workflow:insights` (same layout as `eas deploy:alias` and `eas deploy:alias:delete`), with the same time flags, `--json`, and `--project-id`.

- The overview shows the Maestro tab's four tiles (Maestro runs, pass rate, flaky flows, average duration, each against the previous period), runs over time, and a per-flow table with the dashboard's columns. `--sort` and `--sort-direction` map to the server's sort fields; `--search`, `--status`, `--tag`, `--workflow`, and `--git-ref` mirror the dashboard filters. `--status` uses the dashboard's words (`PASSED`, `FLAKY`, `FAILED`) and maps `PASSED` to the server's `PASSED_CLEAN`.
- `--flow <path>` switches to one flow's history, shaped like the dashboard's flow detail: its four tiles (runs, pass rate, flaky runs, p90 duration, without a previous-period comparison), runs over time, the top five error patterns, and recent runs with duration, branch, commit, and workflow run. The flags that only apply to the flows table are declared exclusive with `--flow`.
- Pass rate is `(passed + flaky) / runs` and flake rate is `flaky / runs`, the dashboard's definitions; the server returns counts only. Durations follow the dashboard's format. When no run reported a duration the value is unknown, and `--json` leaves that key out, the same way every EAS CLI command's JSON output omits null fields.
- Shared pieces from the first command moved so both use them: the metric formatting helpers now live in `src/insights/metrics.ts`, the time-range and workflow filter flag definitions are shared, the timespan JSON and the runs-over-time presentation table live next to `formatTimespan`, workflow file-name resolution is exported from `filters.ts`, and the plan-gate helper knows the Maestro error codes. The first command moved from `insights.ts` to `insights/index.ts` to make room for the subcommand.
- The Maestro server has the same split as the Workflows one, where only the time-series query widens its bounds to whole buckets, so this command sends a widened window too. `#4367` adds the widening; here it is reused through a helper that maps the resulting bucket size onto the Maestro granularity enum, which keeps the overview, the time series, and the per-flow table covering the same period. Because the window no longer matches the `--days N` that was asked for, the `Time range` row and the `--json` timespan report the bounds actually queried.

# Test Plan

- `yarn typecheck`, `yarn lint`, `yarn fmt:check`, `yarn lint-changelog`, and the tests for the new and touched areas pass locally (47 suites, 610 tests).
- New tests cover the command's flags (filters, sort, search, limit, `--flow`, the exclusive-flag rejection, `--project-id`, JSON output, plan-gate error), the filter resolver and status mapping, the formatter (totals, rates, buckets, flows, history, durations), and the plan-gate codes.
- Both Maestro branches, the overview and `--flow`, are asserted to send whole-day bounds with `DAY` granularity for the default window, so the widening reaches the time series and the totals alike.
- The widening helper has its own tests for the bucket-size mapping onto the Maestro enum and for forwarding the widened bounds unchanged.
- Manual run against production on a project with Maestro runs: the overview renders its four tiles with deltas, the flows table renders the dashboard's columns, and `--flow <path>` renders the flow's tiles, error patterns, and recent runs. `--json` returns the same data with the unknown duration key omitted.


packages/eas-cli/src/commands/workflow/insights/maestro.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  async runAsync(): Promise<void> {
+    const { flags } = await this.parse(WorkflowInsightsMaestro);
+    const { json, nonInteractive } = resolveNonInteractiveAndJsonFlags(flags);
+
+    if (json) {
+      enableJsonOutput();
+    }
+
+    const timespan = resolveInsightsTimeRange(flags);
```
sjchmiela: I think this has the same problem as the downstack PR with time ranges: the server expands the daily query to full UTC days, while the summary and flow history use the exact selected times.

sjchmiela (approved):

PR #4379, opened 2026-09-10
[eas-cli] Regenerate graphql and fix types

wschurman:
# Why

When creating #4380, I noticed that upon regenerating the graphql types there were tsc errors. This fixes those errors.

Likely blame rev: #4339

# How

I ran re-generate, and then had claude fix the missing fields by telling it:

> add the missing fields exposed by tsc to the test

# Test Plan

`yarn tsc`

PR #4380, opened 2026-09-10
[eas-cli] Remove broad primaryAccount querying to support scoped token access

wschurman:
# Why

Expo is adding the ability to scope token access, where a session may be downscoped to just a single account or subset of an actor's permissions.

One constraint this imposes upon the tools that communicate with the API is that a session may no longer be able to read all accounts, including the user's primary account.

This PR updates eas-cli to handle this case gracefully for commands that do require a primary account (shows a better error) and at the same time stops querying for primary accounts for all `Context.LoggedIn` commands.

# How

Remove `primaryAccount` from main user query, expose a different user query for getting just the primary account. Use the new query where primary account is required.

# Test Plan

1. Set `EXPO_STAGING`
1. `neas login`
1. Scope to just a single account and VIEWER role.
1. Use various commands, see things don't throw and function mostly normally.


wschurman:
* **#4380** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4380?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/eas-cli/4380?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4379** <a href="https://app.graphite.com/github/pr/expo/eas-cli/4379?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/eas-cli/src/commands/go.ts
```diff
@@ -287,14 +291,12 @@ export default class Go extends EasCommand {
 
   private async ensureEasProjectAsync(
     graphqlClient: ExpoGraphqlClient,
-    actor: Actor,
     slug: string
   ): Promise<string> {
-    const account = ensureActorHasPrimaryAccount(actor);
-
+    const userPrimaryAccount = await UserQuery.requireCurrentUserPrimaryAccountAsync(graphqlClient);
```
byronkarlen: nit: this is unnecessary query since we already run it above inside `runAsync()`. maybe just hits a cache though so doesn't matter anyway?

packages/eas-cli/src/graphql/queries/UserQuery.ts
```diff
@@ -52,4 +36,52 @@ export const UserQuery = {
...
+              additionalTypenames: ['User', 'SSOUser'],
+            }
+          )
+          .toPromise()
+      );
+    } catch (error) {
+      throw new Error(
+        'An error occurred while fetching the primary account of the current user. Check to ensure your session has sufficient scope on your primary account.'
+      );
+    }
```
byronkarlen, line 74: we should probably throw the original error if it's not auth-related(?)

packages/eas-cli/src/graphql/queries/UserQuery.ts
```diff
@@ -16,31 +21,10 @@ export const UserQuery = {
               meActor {
                 __typename
                 id
```
byronkarlen, line 23: doesn't matter at all but these are redundant with the fragment

packages/eas-cli/src/commands/go.ts
byronkarlen, line 296: The one caller of this function (`runAsync()` defined above) already calls this query so we could just pass the data in as an arg instead. Maybe this second query just hits a cache though so it doesn't matter(?)

packages/eas-cli/src/graphql/queries/UserQuery.ts
wschurman: I tend to over-include them since I don't (try to) understand how fragment caching works and know that sometimes these have weird quirks about being required.

packages/eas-cli/src/graphql/queries/UserQuery.ts
wschurman: We don't have a way to reliably detect auth-related errors :( I know in some places in this codebase we do string or error code introspection, but those are fairly fragile and/or require us to know the precise error ahead of time. In this case, the error that occurs when fetching the `primaryAccount` for a token without permission on it isn't stable enough to hardcode in the client, so I figured the best approach here was to just assume that if the callsite got this far (likely already fetched normal user document via query above) and there was an error, it's most likely a session scope error and it's ok to throw an error indicating it as a possibility.

byronkarlen: Dang ok makes sense

byronkarlen (approved):

PR #4393, opened 2026-09-11
[build-tools] inject a guard into simulator processes that refuses connections bypassing the local egress proxy

brentvatne:
# Why

`--egress local` points the simulator's system proxy at the egress tunnel, and #4390 sets proxy environment variables for clients that read them. Both are conventions: a library with its own socket layer still exits from the worker, silently. A packet filter cannot help, since simulator processes share the worker's uid. This PR makes the guarantee structural: a dylib injected into every simulator process refuses any connection that is not destined for loopback, in the process that made it, and reports which library tried. The proxy is the only way a request ends up on loopback, so this refuses everything that does not use it.

# How

- `resources/egress-guard/`: a small C dylib. `guard.c` interposes `connect`, `connectx`, `sendto`, `sendmsg` and their `$NOCANCEL` variants through dyld's `__DATA,__interpose`, which reaches calls made inside CFNetwork and Network.framework. Remote destinations get `ECONNREFUSED`; each process logs one event per destination with the calling image names. `policy.c` classifies destinations, including a sockaddr left as `AF_UNSPEC`, which the kernel connects as the family its length implies. `check.c` is a self-check run inside the simulator after boot.
- Recording is best-effort and cannot hurt the process: the log is opened per event rather than held as a descriptor, the dedupe lock is a bounded trylock so signal re-entry or a fork child skips the event instead of aborting, and a process that lists 128 destinations writes one `overflow` line and keeps refusing.
- Injection rides the boot: `simctl boot` forwards `SIMCTL_CHILD_` variables to the simulator's launchd, so every process of the boot inherits the guard. `launchctl setenv` cannot change variables given at boot and only reaches later processes, so it is used only for a device that was already booted.
- `localEgressGuard.ts` resolves the boot environment from the local egress handoff, relays refusals into the session log under the session step, runs the self-check once boot completes, and reports coverage from `lsof`. The self-check proves in a fresh process that the guard is loaded and behaves as the mode says on `connect`, `connect$NOCANCEL` and an `AF_UNSPEC` connect.
- Fail closed: a missing library, a failed `launchctl setenv`, or a failed self-check fails the session. An unwritable event log is a warning.
- The `simulator:start` banner and `--egress` help say that bypassing connections are refused and link the session page, where the job log lists them.
- Built into `bin/` by `package.sh` for the iOS worker tarball; the `test-egress-guard` EAS workflow runs the host suites, unit tests and simulator e2e on pull requests touching the guard.

Not covered: DNS names still resolve through the system resolver (metadata only; direct DNS is refused), and a hostile app with the worker's uid could unload the guard or make raw syscalls. The target is accidental bypass by libraries.

# Test Plan

- `yarn test:egress-guard`: `policy_test.c` covers classification, mode parsing, peer and event formatting, the dedupe table. `guard_test.c` compiles `guard.c` in and forces signal re-entry, fork with a held lock, and eight-thread contention. `guard_insert_test.c` loads the built dylib and checks descriptor reuse, closed stdout, `$NOCANCEL` refusal, every `AF_UNSPEC` shape, the UDP disconnect, caller attribution, and the overflow line. No test opens a socket to a non-loopback address. Both suites also pass as x86_64 under Rosetta.
- Unit tests for the boot environment, log parser, relay, tailer, coverage parsing, installer outcomes, and the simulator step ordering. `oxlint` and `tsc` clean.
- Simulator e2e (`EAS_LOCAL_EGRESS_E2E=1`, iOS 26.5): every process of the boot covered; URLSession, Network.framework, BSD TCP and UDP, `AF_UNSPEC` connect and `sendto$NOCANCEL` refused while proxied URLSession, loopback UDP, UDP disconnect and DNS work; one event per destination per process; daemons the simulator launches are guarded; simctl tooling unaffected; log mode observes without refusing; a device booted without the guard fails the self-check and can be covered afterwards through `launchctl`; an unwritable log still refuses. 8 of 8 pass locally and in the EAS workflow.
- Staging: sessions on a worker built from this branch (tarball hash-matched to the promotion) showed the guard installed at boot, the self-check passing, all simulator processes covered, system daemons' direct connections refused with callers attributed, and Safari rendering a page through the egress client. Without the client connected the same page failed closed.

### DIFF
--- a/.github/workflows/test-egress-guard.yml
+++ b/.github/workflows/test-egress-guard.yml
@@ -1,0 +1,31 @@
+name: test-egress-guard
+
+on:
+  workflow_dispatch: {}
+  pull_request:
+    paths:
+      - packages/build-tools/resources/egress-guard/**
+      - packages/build-tools/src/steps/utils/localEgressGuard.ts
+      - packages/build-tools/src/steps/utils/__tests__/localEgressGuard*.ts
+      - packages/build-tools/src/steps/functions/startIosSimulator.ts
+      - packages/worker/.eas/workflows/test-egress-guard.yml
+      - .github/workflows/test-egress-guard.yml
+
+jobs:
+  test-egress-guard:
+    # The simulator tests need macOS with Xcode, which the EAS macOS workers have.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+      - uses: ./.github/actions/setup-mise
+      - name: Install eas-cli
+        run: npm install --global eas-cli@latest
+      - name: Run the test-egress-guard EAS workflow on macOS
+        working-directory: packages/worker
+        env:
+          EXPO_TOKEN: ${{ secrets.EXPO_DEV_EXPO_GITHUB_ROBOT_ACCESS_TOKEN }}
+        run: |
+          eas workflow:run test-egress-guard.yml \
+            --ref ${{ github.event.pull_request.head.sha || github.sha }} \
+            --non-interactive \
+            --wait
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
-- [build-tools] Inject a guard into every iOS Simulator process during `--egress local` sessions that refuses connections which bypass the system proxy and reports them, with the calling frameworks, in the session log. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@brentvatne](https://github.com/brentvatne))
+- [build-tools] Inject a guard into every iOS Simulator process during `--egress local` sessions that refuses connections which bypass the system proxy and reports them, with the calling frameworks, in the session log. ([#4393](https://github.com/expo/eas-cli/pull/4393) by [@brentvatne](https://github.com/brentvatne))
 - [build-tools] Set proxy environment variables inside the iOS Simulator for `--egress local` sessions, so clients that read them (gRPC, libcurl) also exit from the egress client. ([#4390](https://github.com/expo/eas-cli/pull/4390) by [@brentvatne](https://github.com/brentvatne))
 
 ### 🐛 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This is the log of notable changes to EAS CLI and related packages.
 
 ### 🎉 New features
 
+- [build-tools] Set proxy environment variables inside the iOS Simulator for `--egress local` sessions, so clients that read them (gRPC, libcurl) also exit from the egress client. ([#XXXX](https://github.com/expo/eas-cli/pull/XXXX) by [@brentvatne](https://github.com/brentvatne))
+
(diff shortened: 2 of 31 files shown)


brentvatne:
Staging run with the worker built from 340ac1e2 (session `01a092b2-55fe-7536-bab9-6229f1896c95`, `--egress local --egress-allow localhost:8082`, Expo Go loading from local Metro through the tunnel). Worker log, guard lines only:

```
22:59:39 Local egress guard: no simulator process was running before the guard was installed.
22:59:49 Local egress guard installed in the Simulator (mode block): connections that bypass the system proxy are refused in the process that makes them and reported here as they happen.
22:59:51 Local egress: proxy environment variables set in the Simulator (http_proxy, https_proxy, HTTP_PROXY, HTTPS_PROXY, grpc_proxy, no_proxy, NO_PROXY), ...
22:59:52 Local egress guard verified in the Simulator: egress-guard-check: guard loaded; non-loopback connections are refused.
22:59:54 Local egress guard coverage: all 143 simulator process(es) have the guard loaded.
23:00:27 Local egress guard: refused sendto from AssetCacheLocatorService (pid 2881) to 192.168.2.1:53; callers: libresolv.9.dylib, AssetCacheLocatorService
```

For comparison, the earlier revision that set the environment with `launchctl setenv` after `simctl boot` returned covered 26 of 150 processes on the same worker: the single setenv spawn took ~9s there and the boot raced ahead. Passing the variables to the boot itself as `SIMCTL_CHILD_*` is what closes that.

The `test-egress-guard` EAS workflow (macOS, simulator end-to-end) passed for this commit: run 34655997474. Staging worker restored from `main` afterwards.

brentvatne:
Second staging run with the worker built from 340ac1e2 (session `01a092bb-f6dc-7481-8298-ab7016f94608`), exercising the developer workflow end to end:

- `--egress local --egress-allow localhost:8082`, Metro in watch mode on this machine, Expo Go opened with `exp://127.0.0.1:8082`: the app loaded through the forwarded port and rendered.
- Edited `src/app/index.tsx` (title text): Fast Refresh updated the remote simulator within seconds, and reverting the file refreshed it back. Screenshots taken on the remote device before and after each change.
- Opened `https://example.com/` in Safari on the remote simulator: the page rendered through the tunnel, while Safari's SafeBrowsing service, which does not honor the system proxy, was refused by the guard, which is the bypass that started this whole thread.

Worker log, guard lines only:

```
23:10:28 Local egress guard: no simulator process was running before the guard was installed.
23:10:38 Local egress guard installed in the Simulator (mode block): ...
23:10:41 Local egress guard verified in the Simulator: egress-guard-check: guard loaded; non-loopback connections are refused.
23:10:43 Local egress guard coverage: all 144 simulator process(es) have the guard loaded.
23:11:23 Local egress guard: refused sendto from AssetCacheLocatorService (pid 3255) to 192.168.2.1:53; callers: libresolv.9.dylib, AssetCacheLocatorService
23:14:30 Local egress guard: refused connect from com.apple.Safari.SafeBrowsing.Service (pid 17071) to 17.253.27.137:80; callers: Network, libboringssl.dylib, Network, CFNetwork, Network
23:14:30 Local egress guard: refused connect from com.apple.Safari.SafeBrowsing.Service (pid 17071) to 17.253.27.136:80; callers: ...
```

No warnings or errors in the job log. Staging worker restored from `main` afterwards.

PR #4398, opened 2026-09-14
[build-tools] Refresh simulator session preview artifacts

szdziedzic:
# Why

Keep a current simulator screen image available for future session list-page thumbnails.

# How

Capture a WebP up to 320px once the shared preview server and FFmpeg are ready, then refresh it 60 seconds after each attempt completes. Obtain one signed upload URL and reuse it for the entire session; its lifetime already covers the session duration. Stop capture with the session and retry capture or upload failures without failing the session.

The upload-session request preserves worker authentication and combines the attempt and URQL cancellation signals at the fetch boundary. A stalled request is aborted on timeout or shutdown. FFmpeg setup runs in the background on macOS without delaying session readiness; Android retains its existing setup before streaming starts. Stopping a session during setup prevents the thumbnail loop from starting later.

Supports iOS and Android. Because Android readiness currently returns a placeholder ID, resolve the sole connected ADB device and skip capture if selection is ambiguous.

Deploy the [API PR](https://github.com/expo/universe/pull/31061) first; CI GraphQL schema generation also needs that mutation deployed. The [website PR](https://github.com/expo/universe/pull/31065) hides preview artifacts without adding thumbnail UI.

# Test Plan

Tests


packages/build-tools/src/steps/utils/deviceRunSessionPreview.ts
```diff
@@ -0,0 +1,181 @@
...
+        body: image,
+        signal,
+        timeout: 30_000,
+      });
+      if (!response.ok) {
+        throw new Error(`Session preview upload failed: ${response.status} ${response.statusText}`);
+      }
+    } catch (err) {
+      if (!controller.signal.aborted) {
+        logger.warn({ err }, 'Could not refresh the session preview; retrying in 60 seconds.');
```
gwdp, line 97: `err` could expose a signed url (recall getting this flagged in previous reviews). Maybe we want to log only `err.code`?

packages/build-tools/src/steps/utils/deviceRunSessionPreview.ts
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      let serial = device;
+      // expo-device-hub's readiness endpoint currently reports a placeholder instead of a serial.
+      if (serial === 'no-device-id') {
+        const result = await spawn('adb', ['devices'], options);
+        const devices = result.stdout
+          .split('\n')
+          .map(line => line.trim().split(/\s+/))
+          .filter(([, status]) => status === 'device');
+        if (devices.length !== 1) {
```
gwdp, line 146: I could be off (if the case, disregard), but I believe iOS is mostly always trying to get the first available device, not necessarily stopping/throwing if multiple devices. Could/should we do the same here and allow multiple devices, but only consider the first on the list for now?

packages/build-tools/src/steps/utils/deviceRunSessionPreview.ts
hirbod: Good catch! Let’s log the type/code and status instead. Just `err.code` would miss some errors, like timeouts.

packages/build-tools/src/steps/utils/deviceRunSessionPreview.ts
hirbod: iOS gives us the device ID from the preview server, but Android only returns a placeholder. I’d keep this check so we don’t grab the wrong screen. It only skips the thumbnail, the session keeps running.

PR #4405, opened 2026-09-15
[eas-cli] Support Xcode 27 Device Hub when running iOS simulator builds

brentvatne:
# Why

Xcode 27 removes the standalone `Simulator.app` and replaces it with `DeviceHub.app` (bundle id `com.apple.dt.Devices`), and moves the developer apps from `Xcode.app/Contents/Developer/Applications/` to `Xcode.app/Contents/Applications/`. The iOS run path in eas-cli hardcoded `Simulator` in three places, so `eas build:dev`, `eas build:run`, and `eas run` fail on Xcode 27 before ever touching the build:

```
Can't determine id of Simulator app; the Simulator is most likely not installed on this machine. Run 'sudo xcode-select -s /Applications/Xcode.app/Contents/Developer'
    Error: build:dev command failed.
```

This ports the handling that `@expo/cli` already shipped in expo/expo#46757 and expo/expo#46809.

# How

- `getSimulatorAppIdAsync` asks LaunchServices for `Simulator`, then `DeviceHub`, then falls back to reading `CFBundleIdentifier` from the Info.plist inside the `xcode-select` path at both the old and new locations. Uses `safeIdOfAppAsync` from the `@expo/osascript` version already pinned.
- `assertSimulatorAppInstalledAsync` accepts `com.apple.dt.Devices` as a valid host app id.
- `openSimulatorAppAsync` still tries `open -a Simulator` first. When that fails it opens `devices://device/open?id=<udid>` so Device Hub focuses the requested device, or `open -a DeviceHub` without a UDID.
- The "is the simulator app running" check counts a `DeviceHub` process as well as `Simulator`.

Every Simulator.app call stays first in each sequence, so on Xcode 26 and older the commands issued are unchanged.

`@expo/build-tools` is not affected. It only uses `xcrun simctl`, which is unchanged on Xcode 27, and never opens the GUI.

# Test Plan

Added unit tests covering both generations: the Simulator.app paths resolve without ever touching Device Hub, and the Device Hub paths only engage after the Simulator lookups fail. Also covers the Info.plist fallbacks and the accepted app ids.

Manually verified on Xcode 27.0 (27A266a), where `osascript -e 'id of app "Simulator"'` fails and `id of app "DeviceHub"` returns `com.apple.dt.Devices`, by running the built CLI against a project with a cached dev build:

```
eas build:dev -p ios --simulator "iPhone 17 Pro" --skip-bundler
```

With Device Hub already running, the app installed and launched. With no Simulator or DeviceHub process running, the CLI printed "Waiting for Simulator app to start...", launched `Xcode.app/Contents/Applications/DeviceHub.app`, then installed and launched the app on the requested simulator.

Also verified the Simulator.app branches on the same machine with Xcode 26.6 installed alongside Xcode 27. LaunchServices then resolves `id of app "Simulator"` to `com.apple.iphonesimulator` from `Xcode-26.6.0.app`, and with no host app running the CLI opened that Simulator.app via `open -a Simulator --args -CurrentDeviceUDID <udid>`, waited for it, then installed and launched the app. Device Hub was never opened in that run, matching the pre-Xcode 27 command sequence.


packages/eas-cli/src/run/ios/simulator.ts
```diff
@@ -205,9 +219,41 @@ export async function installAppAsync(deviceId: string, filePath: string): Promi
   Log.succeed('Successfully installed your app on the simulator!');
 }
 
+const XCODE_SIMULATOR_INFO_PLIST_PATH = './Applications/Simulator.app/Contents/Info.plist';
+const XCODE_DEVICE_HUB_INFO_PLIST_PATH = '../Applications/DeviceHub.app/Contents/Info.plist';
+
 export async function getSimulatorAppIdAsync(): Promise<string | undefined> {
+  // TODO: resolve DeviceHub before Simulator once Xcode 27 is stable, see openSimulatorAppAsync
```
brentvatne, line 226: i left this comment in from expo/cli so that we can update them both at the same time cc @krystofwoldrich

PR #4406, opened 2026-09-15
[build-tools] Support `maestro-runner` HTML and Allure reports

sjchmiela:
We're seeing customers have good experience with `maestro-runner`. Let's add some support for exporting HTML and Allure reports from it.

Runner report generation: https://github.com/devicelab-dev/maestro-runner/blob/main/pkg/cli/test.go


packages/build-tools/src/steps/functions/maestroTests.ts
```diff
@@ -521,23 +530,22 @@ export function createMaestroTestsBuildFunction(ctx: CustomBuildContext): BuildF
...
             );
           }
         }
       }
 
-      // Upload before the ERR_MAESTRO_TESTS_FAILED throw below so fully-failed runs (which need
-      // screenshots most) still upload. Harvest only ran for junit, so guard the same way.
-      if (outputFormat === 'junit') {
+      // Upload before the failure verdict so fully-failed runs still get screenshots.
+      if (mayHaveJUnitReports) {
```
hSATAC, line 548: Do we want to upload the individual failure screenshots for HTML/Allure too? The pre-packaged job still only reports test results for `junit`, and the website hides these screenshot artifacts from the downloads list, so I think they'd get uploaded without being visible anywhere. The HTML artifact already includes its screenshots in the attempt directory, so just wondering if there's another use case I'm missing here.

hSATAC (approved):

packages/build-tools/src/steps/functions/maestroTests.ts
hSATAC: Just to clarify my question above: I realize the runner generates JUnit regardless of `output_format`, so collecting it here makes sense. I was wondering about the pre-packaged job, which still only runs `report_maestro_test_results` when `output_format` is `junit`. Are we planning to update that too so the screenshots can show up for HTML/Allure runs?

PR #4409, opened 2026-09-15
[build-tools] Clear old Maestro HTML reports before each attempt

sjchmiela:
It seems when auto-retrying an old HTML report may be uploaded instead of a more recent one. This clears reports between attempts.

(I think behavior of auto-retry + relying on HTML reports is a bit wonky so I guess users should then use the "retry-all" retry behavior.)


hSATAC (approved):

hSATAC:
Users choose `html` format missed so much from our Maestro insights & screenshots & recordings & etc...

PR #4419, opened 2026-09-16
[build-tools] Separate recording ownership from preview lifetime

vonovak:
# Why

The EAS session must own recording and the host process independently of preview tunnels. Closing or replacing a preview must not stop recording, and artifact upload must not postpone automation cleanup.

# How

- Introduce `DeviceSessionHost` with terminal `finishAsync()` and separately owned preview handles.
- Migrate Agent Device, Appium, Argent and web-preview callers; remove the old combined preview/host API.
- Move shared recording validation/upload into a utility; retain the iOS build function as an adapter.
- Run automation-resource cleanup alongside host finalization and await both.

## Design alternatives and shutdown constraints

A combined preview-and-recording handle makes closing a preview stop the shared host. Separate recording, process, and tunnel teardown handles would make each caller responsible for cleanup order. The session host instead owns finalization, while preview handles own only their tunnels. Concurrent preview opens share a handle, and closing a stale handle cannot close its replacement.

EAS requests recording finalization before stopping the process group. Group shutdown also signals the capture encoder and could interrupt the MP4 before it is complete. The finalization endpoint shares the preview server, so the runner supplies a random control token rather than exposing an unauthenticated stop operation.

This design isolates recording from preview connections, not from a Hub process crash. A separate recorder process, rotation recovery, segment stitching, and multi-device recording remain out of scope. The package README contains the stable ownership contract; shutdown constraints are documented beside the code.

## Stack

EAS **2/3**, based on #4418, followed by #4420. The companion Hub lifecycle work is folded into https://github.com/expo/expo-device-hub/pull/112; #113 is closed.

Android recording is enabled by default in #4418 and remains enabled when ownership moves into DeviceSessionHost. Publish the compatible Hub release before deploying the worker. iOS keeps its existing independent native-recording steps and authenticated preview token behavior.

# Test plan

Added host ownership and orchestration coverage for all four callers, no-preview recording, concurrent opens/finish, close/reopen, stale handles, late opens and failure cleanup. Existing artifact tests cover both iOS and Android recordings. Startup and preview/config publication failures must still finish the host.

At the **cumulative stack heads**, 128 focused build-tools tests and its typecheck passed; 1,163 Hub/serve-emu tests and its typecheck passed. The local capture/events/upload/website replay flow also passed with a local worker adapter. These are aggregate results, not separate validation of this intermediate commit. The production Linux launcher and a new live iOS session have not been exercised.

After rebasing onto default-on recording, this intermediate branch passed 60 host, orchestration, and remote-session tests plus the build-tools typecheck.


vonovak:
Folded into #4418. Fast-forwarding the #4418 branch to this branch's head made GitHub mark this PR as merged; the session-host commits now live in #4418, whose title and description cover both the default-on recording and the ownership split. #4420 is based on #4418.

PR #4424, opened 2026-09-16
[eas-cli] Skip managed-workflow checks for custom builds

douglowder:
# Why

Testing a custom build YAML led to problems with the initial checks in EAS build command, which ran checks that should only apply to normal managed or generic mobile app builds.

# How

Suppress the checks when build type = CUSTOM

# Test Plan

New unit tests added

---


sjchmiela (approved):
well, technically custom builds were, and i think still are, advertized as good solution for running stuff before / after build or customize the build process with steps in between the regular steps. that is to say i think if someone is using custom steps they _might_ benefit from the checks (maybe just informative, non-erroring?)

douglowder:
> well, technically custom builds were, and i think still are, advertized as good solution for running stuff before / after build or customize the build process with steps in between the regular steps. that is to say i think if someone is using custom steps they _might_ benefit from the checks (maybe just informative, non-erroring?)

Good point! Maybe we can add an expo-doctor build/workflow step that customers could optionally insert in a custom build or workflow.

PR #4429, opened 2026-09-17
[build-tools] Preserve captured simulator app logs

sjkim-expo:
# Why

Agent Device can capture app logs on the worker, but those files are not automatically registered in its downloadable artifact inventory. I added a first slice of [ENG-26642](https://linear.app/expo/issue/ENG-26642/upload-native-logs-as-artifacts-in-dashboard) to preserve already-captured logs after the simulator session ends.

# How

I recorded the existing daemon-session directories before startup, then collected `app.log` from new directories after the daemon stopped. This uses the existing simulator artifact upload and download path, so it does not need a new dashboard viewer or a change to the daemon API.

Collection is best-effort, with a 10 MiB total attempted-upload limit and a 30-second final collection/upload budget. Oversized logs are skipped with a warning rather than truncated. Upload failures do not fail the session, and there are no automatic retries because allocating another upload session could create duplicate artifact rows. Cancellation applies to both the GraphQL request and the signed-URL upload without replacing authorization headers.

This does not turn on log capture, collect rotated logs, include preexisting daemon-session directories, or guarantee recovery after a worker is killed. The size and time limits are conservative starting points for review.

# Test Plan

- The focused suite passed 33 tests across five suites, covering final contents, shutdown ordering, session isolation, symlink rejection, missing/oversized logs, failure handling, cancellation and authorization forwarding.
- Build-tools typechecking, changed-file type-aware lint, formatting and diff checks passed. Root lint reports an existing unnecessary cast in `packages/eas-cli/src/build/utils/resourceClass.ts:58`.
- The full build-tools unit run passed 120 suites and 1,222 tests. Five tests in two unchanged iOS provisioning-profile suites could not run successfully because `fastlane` is missing locally (`spawn fastlane ENOENT`).
- A bounded independent read-only review found no concrete introduced issues.
- Live worker-to-dashboard verification is still pending. Before marking this ready, run with a known agent-device version, capture initial and final log markers, stop the session, and confirm the downloaded artifact contains both.


sjkim-expo:
Closing in favor of https://github.com/expo/eas-cli/pull/4463, which uses shared workflow collection across iOS session types rather than collecting files only from Agent Device. Filter support is in https://github.com/expo/serve-sim/pull/196 and WWW wiring is in https://github.com/expo/universe/pull/31439.

PR #4436, opened 2026-09-17
[eas-cli] Add --upload-source-maps to eas update

douglowder:
Upload the source map that expo export emits for each native platform, so EAS Observe can symbolicate stack traces from an update.
    
The maps are not listed in metadata.json, so each one is located from the bundle that metadata.json does name, falling back to a single sibling .map file. sourcesContent is always stripped, which keeps application source text out of EAS storage and gives the map a chance of fitting under the 5 MiB upload limit. A map that is still too large is skipped with a warning; publishing never fails because of a source map.
    
The GraphQL schema for this is not deployed yet, so the new types live in graphql/sourceMapShim.ts. Once expo/universe#30154 ships, regenerate the GraphQL types, delete that file, and repoint its four importers. Grep for ENG-26889.

---


packages/eas-cli/src/project/maybeUploadSourceMapsAsync.ts
```diff
@@ -0,0 +1,203 @@
...
+  sourceMapPath: string,
+  platform: SourceMapPlatform
+): Promise<SourceMapSourceInput | null> {
+  try {
+    // Always strip, so application source text never reaches EAS storage and the map has a chance
+    // of fitting under the upload limit.
+    const strippedPath = await stripSourcesContentAsync(sourceMapPath, platform);
+    const { size } = await fs.promises.stat(strippedPath);
+
+    if (size > MAX_SOURCE_MAP_SIZE_BYTES) {
```
AbbanMustafa: Is the 5 MiB limit too conservative now? Not sure what was determined as far as from the Update side

PR #4452, opened 2026-09-21
[eas-cli] add support for build id in workflow:ssh

AHGIJMKLKKZNPJKQR:
# Why

It's not possible to connect with build id, causing the `eas workflow:ssh <id>` command to fail when copied verbatim.

# How

Use the job id from context instead of the build id when printing .

# Test Plan

Updated test, tested `eas workflow:ssh` with a job id manually


packages/worker/src/sshSession.ts
```diff
@@ -31,7 +31,9 @@ export async function startSshSessionPhaseAsync({
           relayServerUrl: TurtleSshSession.getSshRelayServerUrl(ctx.job),
           idleTimeoutSeconds: TurtleSshSession.getSshIdleTimeoutSeconds(ctx.job),
         });
-        ctx.logger.info(`SSH session ready. Connect with: eas workflow:ssh ${buildId}`);
+        ctx.logger.info(
+          `SSH session ready. Connect with: eas workflow:ssh ${ctx.env.__WORKFLOW_JOB_ID}`
```
gwdp: I remember using this in my earlier versions, but I didn’t include it in the final one so we could design the feature to eventually support SSH on `eas build`, where `__WORKFLOW_JOB_ID` would not be set. 

Do you think we could patch the `eas workflow:ssh` lookup instead? It already supports a few entity IDs, but I overlooked `TurtleBuild`. 🙇‍♂️

AHGIJMKLKKZNPJKQR: Good point, I've done that instead. I kind of defaulted to the job run id, since it's always present but it definitely makes more sense to just accept the build id

gwdp (approved):
LGTM, thanks for the update 🔥

